PR #2, opened 2021-04-30
Subscribe/Unsubscribe RPC handlers

### LINKED ISSUE #7: Implement subscriptions
Rippled has a subscribe RPC, where clients can subscribe to various events. We need to implement the infrastructure, and support "transactions", "transactions_proposed","accounts", "accounts_proposed" and "ledgers" at a minimum

natenichols:
Adds support for subscriptions to `ledger` and `transactions` streams as well as `account` subscriptions.

fixes #7 

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/xrpl-reporting/2)

### DIFF
--- a/handlers/RPCHelpers.cpp
+++ b/handlers/RPCHelpers.cpp
@@ -41,6 +41,23 @@ deserializeTxPlusMeta(Backend::TransactionAndMetadata const& blobs)
     return result;
 }
 
+
+std::pair<
+    std::shared_ptr<ripple::STTx const>,
+    std::shared_ptr<ripple::TxMeta const>>
+deserializeTxPlusMeta(Backend::TransactionAndMetadata const& blobs, std::uint32_t seq)
+{
+    auto [tx, meta] = deserializeTxPlusMeta(blobs);
+
+    std::shared_ptr<ripple::TxMeta> m = 
+        std::make_shared<ripple::TxMeta>(
+            tx->getTransactionID(),
+            seq,
+            *meta);
+
+    return {tx, m};
+}
+
 boost::json::object
 getJson(ripple::STBase const& obj)
 {
--- a/handlers/RPCHelpers.cpp
+++ b/handlers/RPCHelpers.cpp
@@ -4,11 +4,21 @@
 std::optional<ripple::AccountID>
 accountFromStringStrict(std::string const& account)
 {
+    auto blob = ripple::strUnHex(account);
+
+    boost::optional<ripple::PublicKey> publicKey = {};
+    if (blob && ripple::publicKeyType(ripple::makeSlice(*blob)))
+    {
+        publicKey = ripple::PublicKey(
+            ripple::Slice{blob->data(), blob->size()});
+    }
+    else 
+    {
+        publicKey = ripple::parseBase58<ripple::PublicKey>(
+            ripple::TokenType::AccountPublic, account);
+    }
+
     boost::optional<ripple::AccountID> result;
-
-    auto const publicKey = ripple::parseBase58<ripple::PublicKey>(
-        ripple::TokenType::AccountPublic, account);
-
     if (publicKey)
         result = ripple::calcAccountID(*publicKey);
     else
--- a/handlers/Subscribe.cpp
+++ b/handlers/Subscribe.cpp
@@ -1,0 +1,234 @@
+#include <boost/json.hpp>
+#include <reporting/server/session.h>
+#include <handlers/RPCHelpers.h>
+
+static std::unordered_set<std::string> validStreams { 
+    "ledger",
+    "transactions" };
+
+boost::json::value
+validateStreams(boost::json::object const& request)
+{
+    if (!request.at("streams").is_array())
+    {
+        return "missing or invalid streams";
+    }
+
+    boost::json::array const& streams = request.at("streams").as_array();
+
+    for (auto const& stream : streams)
+    {
+        if (!stream.is_string())
+        {
+            return "streams must be strings";
+        }
+
+        std::string s = stream.as_string().c_str();
+
+        if (validStreams.find(s) == validStreams.end())
+        {
+            return boost::json::string("invalid stream " + s);
+        }
+    }
+
+    return nullptr;
+}
+
+void
+subscribeToStreams(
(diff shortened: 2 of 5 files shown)


natenichols:
This should be reviewable now. Changed instances of `finishSequence_` to first check if `{}`

natenichols:
This does not make changes to databases, so nothing has to be changed w/ postgres.

reporting/ReportingETL.cpp
```diff
@@ -407,7 +461,7 @@ ReportingETL::runETLPipeline(uint32_t startSequence, int numExtractors)
             // ETL mechanism should stop. The other stopping condition is if
             // the entire server is shutting down. This can be detected in a
             // variety of ways. See the comment at the top of the function
-            while (currentSequence <= finishSequence_ &&
+            while ((!finishSequence_ || currentSequence <= *finishSequence_) &&
```
natenichols, line 464: changes to `finishSequence_` optional.

reporting/ReportingETL.cpp
```diff
@@ -331,8 +384,9 @@ ReportingETL::buildNextLedger(org::xrpl::rpc::v1::GetLedgerResponse& rawData)
 std::optional<uint32_t>
 ReportingETL::runETLPipeline(uint32_t startSequence, int numExtractors)
 {
-    if (startSequence > finishSequence_)
+    if (finishSequence_ && startSequence > *finishSequence_)
```
natenichols, line 387: As discussed, `finishSequence_` is a `std::optional`, and as such we should check if it is `std::nullopt` before doing comparision.

reporting/ReportingETL.cpp
```diff
@@ -445,7 +499,7 @@ ReportingETL::runETLPipeline(uint32_t startSequence, int numExtractors)
 
                 transformQueue->push(std::move(fetchResponse));
                 currentSequence += numExtractors;
-                if (currentSequence > finishSequence_)
+                if (finishSequence_ && currentSequence > *finishSequence_)
```
natenichols, line 502: same changes to `finishSequence` `std::optional`.

reporting/ReportingETL.cpp
```diff
@@ -252,7 +252,7 @@ ReportingETL::publishLedger(uint32_t ledgerSequence, uint32_t maxAttempts)
         }
 
         publishStrand_.post([this, &ledger]() {
-            publishLedger(*ledger);
+            this->publishLedger(*ledger);
```
cjcobb23, line 255: you don't need `this->` here.

handlers/Subscribe.cpp
```diff
@@ -0,0 +1,233 @@
...
+            return "account must be strings";
+        }
+
+        std::string s = account.as_string().c_str();
+        auto id = ripple::parseBase58<ripple::AccountID>(s);
+
+        if (!id)
+        {
+            return boost::json::string("invalid account " + s);
+        }
```
cjcobb23, line 101: Can we add the ability to accept accounts as 128 bit hex blobs too, instead of just base58?

handlers/Subscribe.cpp
```diff
@@ -0,0 +1,233 @@
...
+    if (request.contains("streams"))
+    {
+        boost::json::value error = validateStreams(request);
+
+        if(!error.is_null())
+        {
+            response["error"] = error;
+            return response;
+        }   
+    }
```
cjcobb23, line 172: Let's combine this `if` body with the one starting on line 185. Same for the `if` body on line `174` and the one on line `188`

handlers/Subscribe.cpp
```diff
@@ -0,0 +1,233 @@
...
+            response["error"] = error;
+            return response;
+        }  
+    }
+
+    if (request.contains("streams"))
+        unsubscribeToStreams(request, session, manager);
+
+    if (request.contains("accounts"))
+        unsubscribeToAccounts(request, session, manager);
```
cjcobb23, line 229: Let's combine the `if`s here, so there are two instead of four.

reporting/server/session.cpp
```diff
@@ -0,0 +1,47 @@
...
+            return doTx(request, backend);
+            break;
+        case account_tx:
+            return doAccountTx(request, backend);
+            break;
+        case book_offers:
+            return doBookOffers(request, backend);
+            break;
+        case ledger:
+            break;
```
cjcobb23, line 30: missing a `return doLedger(request, backend);`

handlers/RPCHelpers.cpp
```diff
@@ -41,6 +41,33 @@ deserializeTxPlusMeta(Backend::TransactionAndMetadata const& blobs)
...
+        // ripple::Blob{blobs.metadata.data(), blobs.metadata.size()};
+
+        result.second =
+            std::make_shared<ripple::TxMeta const>(
+                result.first->getTransactionID(),
+                seq,
+                blobs.metadata);
+    }
+    return result;
+}
```
cjcobb23, line 69: Let's use the first `deserializeTxPlusMeta` function in the body of this one. `TxMeta` has a constructor that takes in an `STObject`, which is how the first function returns the metadata.

cjcobb23 (changes requested):
Looks good. I requested some small changes

handlers/Subscribe.cpp
natenichols: So here, if we combine the `if`s,  then there's a chance you get a `subscribe` request w/ an error in accounts, that successfully subscribes to streams, but then returns a failure code because accounts failed. here we are validating both streams and accounts, and then subscribing.

handlers/Subscribe.cpp
natenichols: See comment below

handlers/Subscribe.cpp
natenichols: Yeah, I'm gonna change all instance of `parseBase58<AccountID>()` to use `accountFromStringStrict()` in RPCHelpers, and then adjust `accountFromStringStrict()` to be able to parse hex strings

cjcobb23 (approved):
LGTM. I'll squash these commits.

PR #3, opened 2021-05-03
Book Offers RPC handler

### LINKED ISSUE #8: Sort the offers returned by book_offers by quality
Right now, they are sorted by the index of the offer object.

natenichols:
Book offers with order enforced by using the quality of offer encoded in `BookDirectory`. 

Modified to use the most recent xrpl-reporting flat map data model.

fixes #8 

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/xrpl-reporting/3)


reporting/CassandraBackend.cpp
```diff
@@ -573,49 +573,87 @@ CassandraBackend::fetchBookOffers(
...
+
+            if(!object)
+                return {{}, {}};
+
+            ripple::SerialIter it{object->data(), object->size()};
+            ripple::SLE offer{it, *cursor};
+            ripple::uint256 bookDir = offer.getFieldH256(ripple::sfBookDirectory);
+
+            statement.bindBytes(bookDir.data() + 24, 8);
+            statement.bindBytes(*cursor);
```
cjcobb23, line 610: We can avoid having to fetch a ledger object to get the book directory by changing the cursor to an object, and including the book directory as a field of the cursor. Something like:
```
"cursor":
  {
    "book":[blob]
    "offer":[blob]
  }
```
or something. You can pick better field names than the ones here if you want.

reporting/CassandraBackend.cpp
```diff
@@ -573,49 +573,87 @@ CassandraBackend::fetchBookOffers(
...
+    uint32_t upper = sequence;
+    auto lastPage = rng->maxSequence - (rng->maxSequence % 256);
+
+    if (sequence != rng->minSequence)
     {
-        keys.push_back(result.getUInt256());
-    } while (result.nextRow());
+        upper = (sequence >> 8) << 8;
+        if (upper != sequence)
+            upper += (1 << 8);
```
cjcobb23, line 590: Use `BackendIndexer::getShift()` instead of hardcoding 256 and 8 here. Also, I'm adding a helper function that I'm going to push soon so you don't have to write this code here. It will be a member of `BackendInterface`. You can copy and paste it if you want:
```
    std::optional<uint32_t>
    getIndexOfSeq(uint32_t seq) const
    {
        if (!fetchLedgerRange())
            return {};
        if (fetchLedgerRange()->minSequence == seq)
            return seq;
        uint32_t shift = indexer_.getShift();
        uint32_t incr = (1 << shift);
        if ((seq % incr) == 0)
            return seq;
        return (seq >> shift << shift) + incr;
    }

```
Or just wait until I push this.

reporting/CassandraBackend.cpp
```diff
@@ -573,49 +573,87 @@ CassandraBackend::fetchBookOffers(
...
+            statement.bindBytes(zero);
+        }
+
+        // statement.bindUInt(limit);
+        CassandraResult result = executeSyncRead(statement);
+
+        BOOST_LOG_TRIVIAL(debug) << __func__ << " - got keys";
+        if (!result)
+        {
+            std::cout << "could not sync read" << std::endl;
```
cjcobb23, line 625: Let's change these `std::cout` print statements to log messages using `BOOST_LOG_TRIVIAL`. You can add them at whatever log level you think is appropriate.

reporting/CassandraBackend.h
```diff
@@ -1248,20 +1276,21 @@ class CassandraBackend : public BackendInterface
...
+    //     CassandraStatement statement{
+    //         (data.isCreated ? insertBook_ : deleteBook_)};
+    //     statement.bindBytes(*data.book);
+    //     statement.bindBytes(data.key);
+    //     statement.bindInt(data.sequence);
+    //     if (data.isCreated)
+    //         statement.bindInt(INT64_MAX);
+    //     executeAsyncWrite(statement, flatMapWriteBookCallback, data, isRetry);
+    // }
+
```
cjcobb23, line 1293: You can just delete this whole function.

reporting/CassandraBackend.cpp
```diff
@@ -1556,80 +1575,73 @@ CassandraBackend::open(bool readOnly)
...
-                 " ORDER BY key ASC LIMIT ? ALLOW FILTERING";
-        if (!getBook_.prepareStatement(query, session_.get()))
-            continue;
-        query = {};
-        query << "SELECT key FROM " << tablePrefix << "books2 "
-              << " WHERE book = ? AND sequence = ? AND "
-                 " key > ? "
-                 " ORDER BY key ASC LIMIT ?";
+        query.str("");
+        query << "SELECT quality_key FROM " << tablePrefix << "books2 "
```
cjcobb23, line 1586: Let's change everything that is called some form of `books2` to just `books`, and remove the old `books` stuff.

cjcobb23 (changes requested):
Looks mostly good. I left some comments.

The data model looks good to me, so let's implement this in Postgres too. Need to update the table schema, the `writeBooks` function, and the `fetchBookOffers` handler.

natenichols:
Impliment `taker` request field.

reporting/CassandraBackend.cpp
natenichols: For the time being, I will modify this such that it does not use `cursor` to mirror rippled API.

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
+            if(!object)
+                return {{}, {}};
+
+            ripple::SerialIter it{object->data(), object->size()};
+            ripple::SLE offer{it, *cursor};
+            ripple::uint256 bookDir = offer.getFieldH256(ripple::sfBookDirectory);
+
+            statement.bindBytes(bookDir.data() + 24, 8);
+            statement.bindBytes(*cursor);
+        }
```
cjcobb23, line 540: Let's remove this, since we are punting on cursors for book_offers for now. You can keep the query the same, and just bind 0, like the else clause does.

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
-        std::vector<Blob> objs = fetchLedgerObjects(keys, sequence);
-        for (size_t i = 0; i < objs.size(); ++i)
+        if (objs[i].size() != 0)
         {
-            if (objs[i].size() != 0)
-                page.offers.push_back({keys[i], objs[i]});
+            if (results.size() == limit)
+                return {results, keys[i], warning}; 
+
+            results.push_back({keys[i], objs[i]});
```
cjcobb23, line 621: return the warning

reporting/PostgresBackend.cpp
```diff
@@ -378,70 +378,155 @@ PostgresBackend::fetchBookOffers(
...
+    pairs.reserve(upperResults.size() + lowerResults.size());
+    std::merge(upperResults.begin(), upperResults.end(),
+               lowerResults.begin(), lowerResults.end(),
+               std::back_inserter(pairs), 
+               [](bookKeyPair pair1, bookKeyPair pair2) -> bool
+               {
+                   return pair1.first < pair2.first;
+               });
+
+    return fetchObjects(pairs, ledgerSequence, limit);
```
cjcobb23: let's return the warning here

reporting/PostgresBackend.cpp
```diff
@@ -378,70 +378,155 @@ PostgresBackend::fetchBookOffers(
...
+
+        bool complete;
+        PgQuery pgQuery(this->pgPool_);
         auto res = pgQuery(sql.str().data());
-        sql << " ORDER BY offer_key ASC"
-            << " LIMIT " << std::to_string(limit);
         if (size_t numRows = checkResult(res, 1))
+            complete = res.asInt(0, 0) != 0;
+        else 
+            return {true, {}};
```
cjcobb23, line 412: I think we want to return false here, since we don't know if the flag ledger is complete.

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
-                            << " book = " << ripple::strHex(book);
-    BookOffersPage page;
-    ripple::uint256 zero = {};
+    auto rng = fetchLedgerRange();
+    
+    if(!rng)
+        return {{},{}};
+
+    uint32_t upper = sequence;
+    auto lastPage = rng->maxSequence - (rng->maxSequence % 256);
```
cjcobb23, line 510: this variable is not used as far as I can tell

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
+
+    std::vector<std::pair<std::uint64_t, ripple::uint256>> quality_keys;
+    if (sequence != rng->minSequence)
     {
-        statement.bindBytes(zero);
+        upper = (sequence >> 8) << 8;
+        if (upper != sequence)
+            upper += (1 << 8);
+
+        quality_keys = readBooks(book, upper, cursor);
```
cjcobb23, line 577: I think this line is supposed to be outside of the `if` body. Otherwise we don't fetch anything if `sequence == rng->minSequence`

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
+        return keys;
+    };
+
+    std::vector<std::pair<std::uint64_t, ripple::uint256>> quality_keys;
+    if (sequence != rng->minSequence)
     {
-        statement.bindBytes(zero);
+        upper = (sequence >> 8) << 8;
+        if (upper != sequence)
+            upper += (1 << 8);
```
cjcobb23, line 575: You can just use `getBookIndexOfSeq()` for this logic.

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
-    if (keys.size() && keys.size() == limit)
+    BOOST_LOG_TRIVIAL(debug)
+        << __func__ << " - populated keys. num keys = " << quality_keys.size();
+
+    std::cout << "KEYS SIZE: " << quality_keys.size() << std::endl;
+    if (!quality_keys.size())
+        return {{}, {}};
+
+    std::optional<std::string> warning = {};
+    if (quality_keys[0].second.isZero())
```
cjcobb23, line 588: I don't think this is the correct way to check for completeness. We need to query the zero book to test for completeness, whereas this is looking at the offer keys that correspond to the book passed in as an argument to this function.

reporting/CassandraBackend.cpp
```diff
@@ -501,68 +501,128 @@ CassandraBackend::fetchBookOffers(
...
     {
-        page.cursor = keys.back();
-        keys.pop_back();
+        warning = "Data may be incomplete";
+
+        std::uint32_t lower = (sequence >> 8) << 8;
```
cjcobb23, line 592: use `getBookShift()` instead of hardcoding 8. Also, this will return sequence if (sequence % 2^8) == 0.

cjcobb23 (changes requested):
Left some comments, mostly about cursors and warnings.

Let's make sure we always return the warning if the flag ledger is not complete, regardless of the book or if there are offers on that book. Since we check lower everytime though, let's remove the warning from the actual JSON response. We will just use the warning internally to check if the flag ledger is complete, to know if we need to finish writing that flag ledger or not. 

Also, let's remove cursors from the interface for book_offers. Let's remove it from the JSON response, remove it from `BookPage` and remove it as an argument to `fetchBookOffers()`.

cjcobb23 (approved):
LGTM. Great work!

PR #5, opened 2021-05-06
Proxies transactions proposed and RPCs

### LINKED ISSUE #6: Proxy RPCs to a rippled node
RPCs like `fee` and `submit` need to be proxied to a rippled node

natenichols:
Proxies `accounts_proposed`, `transactions_proposed`, and RPCs that are handled by the P2P network (currently just `submit` and `fee`).

Still need to implement proxy-ing for `"ledger_index": "closed"` and `"ledger_index": "current"`.

fixes #6 

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/xrpl-reporting/5)


cjcobb23 (changes requested):
The code looks good, and you mostly followed the pattern that exists in rippled. I think though we can refactor this code and make the design cleaner. This is what I am thinking:

`ReportingETL` is sort of a master class in the code, and a lot of components are instantiated and managed by `ReportingETL`. I think we should elevate `BackendInterface`, `ETLLoadBalancer` and `SubscriptionManager` to a higher level and let them manage themselves. We can construct each of these in `main()` as a `shared_ptr`, and have any component that needs them contain them as a `shared_ptr` member. We would also remove any `start()` and `stop()` methods and just have each component start and stop itself in the constructor and destructor.

I think this design would be cleaner, as the components are less tightly coupled. For instance, `BackendInterface` doesn't need `ReportingETL`, and is used outside of ETL. Therefore, it should exist independently of the ETL process, so that way it can be used whether ETL is running or not. The same goes for `ETLLoadBalancer`. It's used by ETL, but also for forwarding, and therefore can exist independent of the ETL process. Also, we can rely on RAII and shared_ptr semantics for stopping things in the right order.

Let me know what you think of this.

reporting/server/session.h
```diff
@@ -93,21 +104,54 @@ class session : public std::enable_shared_from_this<session>
     boost::beast::websocket::stream<boost::beast::tcp_stream> ws_;
     boost::beast::flat_buffer buffer_;
     std::string response_;
-    BackendInterface const& backend_;
-    SubscriptionManager& subscriptions_;
+
+    std::shared_ptr<BackendInterface> backend_;
+    std::weak_ptr<SubscriptionManager> subscriptions_;
```
natenichols, line 109: I _think_ that `session::subscribers_` needs to be a `std::weak_ptr` to prevent a cyclic dependency when between SubscriptionManager and session.

We pass `subscribers_.lock()` into `buildResponse()`, so at any point during the handler the SubscriptionManager should exist.

reporting/ETLSource.h
```diff
@@ -262,32 +305,54 @@ class ETLSource
...
+    
 };
 /// This class is used to manage connections to transaction processing processes
 /// This class spawns a listener for each etl source, which listens to messages
 /// on the ledgers stream (to keep track of which ledgers have been validated by
 /// the network, and the range of ledgers each etl source has). This class also
 /// allows requests for ledger data to be load balanced across all possible etl
 /// sources.
-class ETLLoadBalancer
+class ETLLoadBalancer : std::enable_shared_from_this<ETLLoadBalancer>
```
cjcobb23, line 323: I don't think we need `enable_shared_from_this` here.

reporting/server/session.h
cjcobb23: You are correct that this needs to be a weak ptr to prevent a cyclic dependency

reporting/server/session.h
```diff
@@ -251,10 +311,14 @@ class session : public std::enable_shared_from_this<session>
...
             {
+                if (subscriptions_.expired())
+                    return;
+
                 response = buildResponse(
                     request, 
                     backend_,
-                    subscriptions_,
+                    subscriptions_.lock(),
+                    balancer_,
```
cjcobb23, line 321: The proper way to do this is to call `subscriptions_.lock()`, check the result to make sure it's not `nullptr`, and then pass that in to `buildResponse`. The way it is now, `expired()` could return false, then all of the shared_ptr instances get deleted, then we call `lock`, which is going to return `nullptr`.

cjcobb23 (commented):
LGTM. I merged this before I approved

PR #23, opened 2021-05-24
Tx subs tmp

natenichols:
I messed up the rebase for the different function handlers. This commit should fix it.


cjcobb23 (approved):
LGTM

PR #28, opened 2021-06-14
Rpc refactor

natenichols:
Here is my RPC refactor for that I've been working on to get Manoj's automation lib passing. 

So far, only the `ledger` RPC passes. I'm going to finish the refactor and then tweak the rest of the tests to get them passing. 

Let me know what you think, and note that this is still a WIP/draft.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/28)


natenichols:
Also, we should squash the commit messages. I did not do a good job of changing commit messages while rebasing.

src/backend/README.md
```diff
@@ -1,26 +1,108 @@
-The backend is clio's view into the database. The database could be either PostgreSQL or Cassandra.
```
cjcobb23, line 1: Unless I'm reading this wrong, I think this file is being reverted back to a previous state. Let's fix that.

src/etl/ReportingETL.cpp
```diff
@@ -138,57 +139,12 @@ ReportingETL::loadInitialLedger(uint32_t startingSequence)
...
-            transactions = backend_->fetchAllTransactionsInLedger(lgrInfo.seq);
-            break;
-        }
-        catch (Backend::DatabaseTimeout const&)
-        {
-            BOOST_LOG_TRIVIAL(warning) << "Read timeout fetching transactions";
-        }
-    }
+    auto fees = backend_->fetchFees(lgrInfo.seq);
+    auto transactions = backend_->fetchAllTransactionsInLedger(lgrInfo.seq);
```
cjcobb23, line 147: This stuff has to be inside a try catch, since this is on the write path. We need to catch the database exception, else the program will crash

src/handlers/Handlers.h
```diff
@@ -0,0 +1,44 @@
...
+constexpr unsigned int ApiMaximumSupportedVersion = 1;
+constexpr unsigned int APINumberVersionSupported =
+    ApiMaximumSupportedVersion - ApiMinimumSupportedVersion + 1;
+
+static_assert(ApiMinimumSupportedVersion >= APIVersionIfUnspecified);
+static_assert(ApiMaximumSupportedVersion >= ApiMinimumSupportedVersion);
+
+// Indicates the level of administrative permission to grant.
+// NOTE, this does not currently affect RPCs in clio, and will be implimented later.
+enum class Role { GUEST, USER, IDENTIFIED, ADMIN, PROXY, FORBID };
```
cjcobb23, line 27: Since the roles are not being used, let's not add them now

src/handlers/RPCHelpers.cpp
```diff
@@ -1,5 +1,6 @@
-#include <backend/BackendInterface.h>
 #include <handlers/RPCHelpers.h>
+#include <handlers/Status.h>
```
cjcobb23, line 2: I think there are two files called RPCHelpers.cpp. Is that intentional?

src/handlers/Handlers.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+        innerTable[HandlerImpl::name()] = h;
+    }
+
+    inline unsigned
+    versionToIndex(unsigned version) const
+    {
+        return version - RPC::ApiMinimumSupportedVersion;
+    }
+};
```
cjcobb23, line 98: I know this code comes from rippled. I don't really like this code. I think a lot of it is overkill. I like what we had before, where we just have a map and a switch statement. Is there a reason we can't do that?

src/handlers/Status.h
```diff
@@ -0,0 +1,213 @@
...
+    // unused                  24,
+    // unused                  25,
+    // unused                  26,
+    // unused                  27,
+    // unused                  28,
+    rpcTXN_NOT_FOUND = 29,
+    // unused                  30,
+
+    // Malformed command
+    rpcINVALID_PARAMS = 31,
```
cjcobb23, line 84: This is a lot of copy paste. Luckily, the error codes are part of `xrpl_core`, so we can just include `ripple/protocol/ErrorCodes.h` and use them. Otherwise, every time rippled changes something, we have to change it here manually as well.

src/handlers/methods/Account.h
```diff
@@ -0,0 +1,207 @@
...
+
+#include <handlers/Context.h>
+#include <handlers/Status.h>
+#include <handlers/Handlers.h>
+#include <boost/json.hpp>
+
+namespace RPC
+{
+    
+class AccountInfo
```
cjcobb23, line 31: I would like to stick to the old pattern, where handlers are just functions. We can do whatever checking necessary inside the handler. I think all of these classes makes the code harder to read. I know rippled does it like this, but I think it's hard to read, and we can do it a lot simpler.

src/webserver/Listener.h
```diff
@@ -0,0 +1,352 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef LISTENER_H
```
cjcobb23, line 18: Why does git think this is a new file, as opposed to just a renaming? It would be ideal if we could fix that, since then the diff will show just the changed lines.

cjcobb23 (changes requested):
First pass of comments

src/handlers/Handlers.h
cjcobb23: I think there are only two roles, and the `DOSGuard` handles them. There are clients who are whitelisted, and clients who aren't.

src/handlers/Status.h
natenichols: Ah cool, I didn't think any of the RPC stuff was included. I can change these to use rippled rpc codes

src/handlers/Handlers.cpp
natenichols: No, we can do that, I just liked how rippled did it better. 

I was also considering adding RPC versioning in this PR. This also will be much more extendable to versioning, versus adding a `switch` for every new API version we want to add.

src/handlers/RPCHelpers.cpp
natenichols: Um, i dont see two files, What are the relative paths to both of them?

src/webserver/Listener.h
natenichols: I dont know, let me look

src/backend/README.md
natenichols: So, the state of my branch is correct, and looks like its the same as `README.md` at `master`.  However, I have no idea why its showing these as changes...

cjcobb23: This is master in your fork: https://github.com/natenichols/xrpl-reporting/blob/master/src/backend/README.md
This is the rpc-refactor branch: https://github.com/natenichols/xrpl-reporting/blob/rpc-refactor/src/backend/README.md

src/handlers/methods/Account.h
```diff
@@ -0,0 +1,51 @@
...
+doAccountCurrencies(Context const& context);
+
+Result
+doAccountLines(Context const& context);
+
+Result  
+doAccountObjects(Context const& context);
+
+Result
+doAccountOffers(Context const& context);
```
cjcobb23, line 47: I think we should just declare all of the handlers in one file, and then define each in a separate file

src/backend/README.md
```diff
@@ -1,26 +1,108 @@
...
-Multiple clio servers can share access to the same database.
```
cjcobb23, line 2: Let's reset this file. If your local master branch is up to date with your remote master branch, you can run:
```
git checkout master -- src/backend/README.md
```

cjcobb23 (changes requested):
A few small comments. Otherwise this looks good

PR #52, opened 2021-11-23
Refactors subscription manager

natenichols:
Refactors the Subscription forwarding to use `boost::asio::io_context` instead of one very large mutex.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/52)

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -75,7 +75,6 @@ ETLSourceImpl<Derived>::reconnect(boost::beast::error_code ec)
     if (ec.category() == boost::asio::error::get_ssl_category()) {
         err = std::string(" (")
                 +boost::lexical_cast<std::string>(ERR_GET_LIB(ec.value()))+","
-                +boost::lexical_cast<std::string>(ERR_GET_FUNC(ec.value()))+","
                 +boost::lexical_cast<std::string>(ERR_GET_REASON(ec.value()))+") "
         ;
         //ERR_PACK /* crypto/err/err.h */
@@ -335,11 +334,7 @@ ETLSourceImpl<Derived>::onHandshake(boost::beast::error_code ec)
     {
         boost::json::object jv{
             {"command", "subscribe"},
-            {"streams", {
-                "ledger", 
-                "transactions_proposed",
-                "manifests",
-                "validations"}
+            {"streams", {"ledger", "manifests", "validations", "transactions_proposed"}
             }};
         std::string s = boost::json::serialize(jv);
         BOOST_LOG_TRIVIAL(trace) << "Sending subscribe stream message";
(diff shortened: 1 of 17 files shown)


src/subscriptions/SubscriptionManager.h
```diff
@@ -0,0 +1,290 @@
...
+    void
+    unsubscribe(
+        std::shared_ptr<WsBase> const& session,
+        ripple::Book const& book);
+
+    void
+    publish(
+        std::string const& message,
+        ripple::Book const& book);
+};
```
cjcobb23, line 129: Let's make these a template class. `BookSubscription` and `AccountSubscription` are easily templatized, where the map key is the template parameter. They can call template helper functions, which `Subscription` would also call. Consider something like this:
```
template <class T>
void
sendAll(std::string const& pubMsg, T& subscribers)
{
    for (auto it = subscribers.begin(); it != subscribers.end();)
    {
        auto& session = *it;
        if (session->dead())
        {
            it = subscribers.erase(it);
        }
        else
        {
            session->send(pubMsg);
            ++it;
        }
    }
}

template <class T>
void
subscribe(std::shared_ptr<WsBase> const& session, T& subscribers, boost::asio::io_context::strand& strand)
{
    boost::asio::post(strand, [subscribers, session](){
        subscribers.emplace(session);
    });
}

template <class T>
void
unsubscribe(std::shared_ptr<WsBase> const& session, T& subscribers, boost::asio::io_context::strand& strand)
{
    boost::asio::post(strand, [subscribers, session](){
        subscribers.erase(session);
    });
}

```

Does this make sense?

cjcobb23 (changes requested):

cjcobb23:
This is a lot nicer than the previous code btw.

src/etl/ETLSource.cpp
```diff
@@ -75,7 +75,6 @@ ETLSourceImpl<Derived>::reconnect(boost::beast::error_code ec)
     if (ec.category() == boost::asio::error::get_ssl_category()) {
         err = std::string(" (")
                 +boost::lexical_cast<std::string>(ERR_GET_LIB(ec.value()))+","
-                +boost::lexical_cast<std::string>(ERR_GET_FUNC(ec.value()))+","
```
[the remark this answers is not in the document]
natenichols: I was getting a compiler error here on MacOS.

src/subscriptions/SubscriptionManager.h
```diff
@@ -0,0 +1,290 @@
...
+
+class SubscriptionManager
+{
+    using subscription_ptr = std::unique_ptr<Subscription>;
```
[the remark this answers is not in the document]
natenichols: I don't think so, we can just instantiate these as members.

natenichols:
> This is a lot nicer than the previous code btw.

To be fair, the old code was pretty bad, so that's not saying much :)

src/subscriptions/SubscriptionManager.h
natenichols: Yup, implemented something similar and testing it out.

src/rpc/handlers/LedgerData.cpp
```diff
@@ -20,6 +20,8 @@
 
 namespace RPC {
 
+using boost::json::value_to;
+
```
cjcobb23, line 24: We should change remove this, but it's fine for this PR.

cjcobb23 (approved):
Looks good.

PR #54, opened 2021-11-23
parallelizes download

natenichols:
Parallelizes the initial database download in the same manner as reporting mode.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/54)

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -484,7 +484,6 @@ class AsyncCallData
     grpc::Status status_;
 
     unsigned char nextPrefix_;
-
 public:
     AsyncCallData(
         uint32_t seq,
@@ -522,7 +521,7 @@ public:
     process(
         std::unique_ptr<org::xrpl::rpc::v1::XRPLedgerAPIService::Stub>& stub,
         grpc::CompletionQueue& cq,
-        ThreadSafeQueue<std::shared_ptr<ripple::SLE>>& queue,
+        BackendInterface const& backend,
         bool abort = false)
     {
         BOOST_LOG_TRIVIAL(debug) << "Processing calldata";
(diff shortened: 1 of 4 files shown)


cjcobb23 (commented):
I don't think we even need the queue, because `writeLedgerObject` should be threadsafe. In reporting mode, we need the queue because we are inserting into a SHAMap, which is not thread safe.

natenichols:
It is not because of the BackendIndexer.

I'm happy to put this off until we have that chunk rewritten w/ the in memory validated ledger, but we'd have to ensure that that is threadsafe too. Here the loads into scyllaDB/Cassandra are done in a single threaded manner. 

Also, I'm pretty sure the initial download is single threaded, its just able to make more than one request at a time.

cjcobb23:
Oh true, I forgot about BackendIndexer. But you are right, the writes are single threaded still. It's just the requests to rippled that are happening in parallel. If that's the case, then we don't need the queue, right?

natenichols:
I dont think the requests even happen in parallel. 

They just happen asynchronously.

cjcobb23:
True, the parallelism occurs on the rippled side, and over the network, but clio processes everything sequentially.

src/etl/ETLSource.h
```diff
@@ -466,9 +466,11 @@ class ETLLoadBalancer
 private:
     std::vector<std::unique_ptr<ETLSource>> sources_;
 
+    std::uint32_t downloadRanges_ = 16;
```
cjcobb23, line 469: Can we just keep this function local and pass it around?

src/etl/ETLSource.cpp
```diff
@@ -669,14 +701,22 @@ ETLSourceImpl<Derived>::fetchLedger(uint32_t ledgerSequence, bool getObjects)
...
-    boost::json::array const& config,
+    boost::json::object const& config,
     boost::asio::io_context& ioContext,
     std::optional<std::reference_wrapper<boost::asio::ssl::context>> sslCtx,
     std::shared_ptr<BackendInterface> backend,
     std::shared_ptr<SubscriptionManager> subscriptions,
     std::shared_ptr<NetworkValidatedLedgers> nwvl)
 {
-    for (auto& entry : config)
+    if (config.contains("download_ranges") &&
```
cjcobb23, line 711: Can we call this `download_parallelism` or `num_ranges` or `num_markers`?

cjcobb23 (changes requested):
Two small things

src/etl/ETLSource.h
natenichols: Not super easily, since we only access the configuration in the constructor. 

The other option would be to save the configuration as a member

cjcobb23 (approved):
LGTM

PR #72, opened 2022-01-04
Report RPC counts in server_info

natenichols:
This will report RPC counts in `server_info`. The fields are as follows.

`started`. Total number of RPCs started. This counts both failed and successful RPC calls.

`ended`. Total number of successful RPCs.

`errored`. Total number of RPC calls that errored out. 

`forwarded`. Total number of RPCs that are forwarded to p2p servers.

`duration_us`. Total duration in microseconds of successful RPCs. The average duration, then is `duration_us`/`ended`

Example response
```json
{
  "result": {
    "info": {
      "complete_ledgers": "68793658-68794322",
      "counters": {
        "rpc": {
          "server_info": {
            "started": "17",
            "finished": "17",
            "errored": "0",
            "forwarded": "0",
            "duration_us": "3438564"
          }
        }
      }
    },
```

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/72)

### DIFF
--- a/src/rpc/Counters.h
+++ b/src/rpc/Counters.h
@@ -1,0 +1,53 @@
+#ifndef RPC_COUNTERS_H
+#define RPC_COUNTERS_H
+
+#include <chrono>
+#include <cstdint>
+#include <string>
+#include <shared_mutex>
+#include <boost/json.hpp>
+
+namespace RPC
+{
+
+class Counters
+{
+private:
+    struct MethodInfo
+    {
+        MethodInfo() = default;
+
+        std::atomic_uint started{0};
+        std::atomic_uint finished{0};
+        std::atomic_uint errored{0};
+        std::atomic_uint forwarded{0};
+        std::atomic_uint duration{0};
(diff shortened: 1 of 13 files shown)


src/rpc/Counters.h
```diff
@@ -0,0 +1,50 @@
...
+        MethodInfo() = default;
+
+        std::uint32_t started{0};
+        std::uint32_t finished{0};
+        std::uint32_t errored{0};
+        std::uint32_t forwarded{0};
+        std::chrono::microseconds duration{0};
+    };
+
+    std::mutex mutex_;
```
cjcobb23, line 27: Can we do this with atomics and remove the mutex? Might have to prepopulate `methodInfo_`

cjcobb23: If we can't remove the mutex, we could probably at least change the counts to atomic variables, change the mutex to a shared_mutex, and thus reduce the time spent under a unique lock.

PR #76, opened 2022-01-05
Make Clio use async database reads

### LINKED ISSUE #44: Make database fetches truly async
Currently, we use cassandra's async API to fetch records, but we really just submit a batch of async requests and then wait on a condition variable until they all complete. It would be better to actually free up the thread that makes the async requests, so that way the thread can handle additional RPC calls while waiting for the async requests to complete. This will give us better throughput overall for reads.

We need to figure out how to make async postgres queries.

It would be best to change all database reads to be asynchronous, even the ones that currently use the synchronous cassandra API.

natenichols:
Uses `Boost.Asio` to make database reads (and writes in postgres) async. Forwarding also uses async reads/writes

fixes #44 

Lets start w/ some high level stuff.

Whenever a websocket or http server is handling a request, instead of handling that request synchronously, Clio will now spawn a co-routine with `boost::asio::spawn` to handle the request. This provides us with the yield context we will use to yield the coroutine when awaiting database responses.

**Cassandra:**
In cassandra, I'm only changing the database reads, the writes are already async, so we'll just leave those as is. This is pretty simple, all `CassandraBackend` methods use three methods to use do database reads `executeSyncRead`, `fetchTransactions`, and `fetchLedgerObjects`. These are the three we'll have to make async.

For `executeSyncRead`, its pretty straightforward, we rewrite `cass_future_error_code` using an async_result, which waits asynchronously. 

For `fetchTransactions` and `fetchLedgerObjects`, we just replace the `condition_variable` with an `async_result::completion_handler_type` to be called when `outstanding == 0`.

**Postgres:**
Postgres is a lot harder, since we use `PgPool` to manage connections. This leads to a lot of open connections in postgres (which is fine, postgres can handle it). I rewrote the `Pg` class to be completely async, which means the number of `max_connections` in the PgPool must be set to be high. 

We don't have to use `async_result` here, since libPQ can expose the socket underlying each postgres connection, so we can async_wait on the socket for either `read-ready` or `write-ready` depending on the status of the connection.

Because the PgClass is now async, this means both reads and writes are now async, so I had to spawn coroutines to handle the ETL processes.

Note: I also added a `PgQuerySync` class, that does sync queries, for setting up the databases when the `io_context` is not running. This should not be used outside of setting up the database.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/76)


src/backend/Pg.cpp
```diff
@@ -61,90 +61,291 @@ PgResult::msg() const
...
-    pg_result_type ret{nullptr, [](PGresult* result) { PQclear(result); }};
+    pg_result_type ret {nullptr, [](PGresult* result){ PQclear(result); }};
     // Connect then submit query.
-    while (true)
+    try
     {
+        connect(yield);
+
+        int sent;
+        if (nParams)
```
mtrippled, line 262: I'm pretty sure PQsendQueryParams() with 0 nParams and null values is the same as PQsendQuery(). It's handled the same as PQexec & PQexecParams (the synchronous version): https://www.postgresql.org/docs/12/libpq-exec.html

So there should be no reason for calling the separate functions, nor for keeping the sent variable since the status can be checked for a single function call.

src/backend/Pg.cpp
```diff
@@ -61,90 +61,291 @@ PgResult::msg() const
...
+            if (PQisBusy(conn_.get()))
+                continue;
+            
+            pg_result_type res{PQgetResult(conn_.get()),
+                               [](PGresult *result)
+                               { PQclear(result); }};
+
+            if (!res)
+                break;
+
```
mtrippled, line 314: What do you want to do about multiple results? One possibility is to check if there's already a result, and then throw:

            if (ret)
            {
                Throw<std::runtime_error>("multiple results returned");
            }

The way it stands now is that if there are multiple commands submitted in the command*, such as "SELECT * FROM foo; SELECT * FROM bar", then this will return the results from the 2nd query.

Not a big deal either way. One way forces a single SQL statement per command submissions, the other allows multiple but silently loses the response to all but the last.

src/backend/Pg.cpp
```diff
@@ -311,7 +570,9 @@ Pg::clear()
 
 //-----------------------------------------------------------------------------
 
-PgPool::PgPool(boost::json::object const& config)
+PgPool::PgPool(
+    boost::asio::io_context& ioc,
+    boost::json::object const& config) : ioc_(ioc)
```
mtrippled, line 575: I'm actually looking at the code already implemented, and it looks like it forces the user to connect with username, password, and database name. In reporting mode, if the user does not specify these things, the connection defaults to using postgres on the local system as the same username/dbname as the rippled process.

Just a suggestion to make things easier for users running this on a single server.

Upon further thought, I'm not sure if asio would work with postgres on a local UNIX domain socket. So this requires the DB to be accessed over the network (even localhost), which requires specific configuration.

src/backend/Pg.h
```diff
@@ -406,13 +436,19 @@ class PgPool
...
         onStop();
     }
 
+    PgConfig&
+    config()
+    {
+        return config_;
+    }
+
     /** Initiate idle connection timer.
```
mtrippled, line 452: The idleSweeper() is not called by anything.

unittests/main.cpp
```diff
@@ -2077,5 +2123,11 @@ TEST(Backend, CacheIntegration)
...
     }
-}
 
+    done = true;
+    work.reset();
+
+    });
+
+    ioc.run();
+}
```
mtrippled, line 2133: Not in this PR, but in the code anyway, when doing checkin(), in addition to clearing the results, the asio socket should be cancelled. This aborts any pending asio async operations.

    /** Cancel asynchronous operations on underlying postgres socket. */
    void
    cancel()
    {
        if (socket_)
            try { socket_->cancel(); } catch (...) {}
    }

mtrippled (changes requested):
This is a first pass at the PR. I looked only at Pg.h and Pg.cpp, and also included some comments about the pre-existing code there.

src/backend/Pg.cpp
natenichols: So I think we want to allow multiple queries for setting up the database and creating tables, so we do want to allow this behavior.

natenichols: In that case, if one errors out, we should break and return the error. Otherwise a final success message should imply that all the queries were successful.

mtrippled: That's fine.

src/backend/Pg.h
```diff
@@ -255,12 +258,25 @@ class Pg
     friend class PgQuery;
 
     PgConfig const& config_;
+    boost::asio::io_context::strand strand_;
```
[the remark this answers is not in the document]
cjcobb23: Is there a problem if a different thread completes the async activity? i.e. Thread A starts, Thread B finishes

src/backend/Pg.cpp
```diff
@@ -238,38 +448,87 @@ Pg::bulkInsert(char const* table, std::string const& records)
...
 
-    if (PQputCopyData(conn_.get(), records.c_str(), records.size()) == -1)
+    auto& socket = getSocket();
+
+    std::int32_t putCopy = 0;
+    while (putCopy == 0)
+    {
+        putCopy = PQputCopyData(conn_.get(), records.c_str(), records.size());
+        
+        if (putCopy == -1)
```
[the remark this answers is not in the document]
cjcobb23: There are times when we don't want copy to be synchronous. There are multiple tables that all need to be copied to, but if copy is synchronous, we need to wait for one to finish before submitting the next. Also, during the initial ledger load, we flush the buffer via copy every 100,000 or million records or so.

src/etl/ETLSource.h
```diff
@@ -53,10 +53,14 @@ class ETLSource
     loadInitialLedger(
         uint32_t sequence,
         std::uint32_t numMarkers,
-        bool cacheOnly = false) = 0;
+        bool cacheOnly,
+        boost::asio::yield_context& yield) = 0;
```
cjcobb23, line 57: Can we make the yield_context a member variable?

src/backend/Pg.h
```diff
@@ -255,12 +258,25 @@ class Pg
...
     bool& stop_;
     std::mutex& mutex_;
 
+    int socketNum_ = -1;
```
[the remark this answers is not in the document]
natenichols: So I actually need to refactor this. According to the docs 
```
Obtains the file descriptor number of the connection socket to the server. A valid descriptor will be greater than or equal to 0; a result of -1 indicates that no server connection is currently open. (This will not change during normal operation, but could change during connection setup or reset.)
```
So it looks like we only need to acquire the socket during connection setup or reset.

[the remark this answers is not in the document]
mtrippled: Yes, same with connect()

src/backend/Pg.cpp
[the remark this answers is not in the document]
mtrippled: Are they all in the same transaction, such as in reporting mode?

Anyway, see the comment above this for Pg.cpp:454. Namely, this function looks broken to me, or at least not working as postgres documents it. Fixing it by making it synchronous would be easier.

src/backend/Pg.cpp
```diff
@@ -238,38 +448,87 @@ Pg::bulkInsert(char const* table, std::string const& records)
...
+            ss << "bulkInsert to " << table
+            << ". PQputCopyData error: " << PQerrorMessage(conn_.get());
+            disconnect();
+            BOOST_LOG_TRIVIAL(error) << __func__ << " " << records;
+            throw std::runtime_error(ss.str());
+        }
+
+        // If the value is zero, wait for write-ready and try again.
+        if(putCopy == 0)
+            socket.async_wait(boost::asio::ip::tcp::socket::wait_write, yield);
```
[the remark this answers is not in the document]
natenichols: Cool, I'll wrap all async_waits w/ try/catch blocks.

src/backend/Pg.cpp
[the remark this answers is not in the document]
cjcobb23: yeah they are all in the same transaction

src/backend/Pg.h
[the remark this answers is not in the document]
natenichols: So I have made the change to 
1) only acquire socket during connection, since it will not change during normal operation.
According to the docs:
"If PQconnectStart succeeds, the next stage is to poll libpq so that it can proceed with the connection sequence. Use PQsocket(conn) to obtain the descriptor of the socket underlying the database connection"
So as long as PQConnectStart succeeds we should be good to obtain the underlying socket.

2) Use RAII to manage the socket, calling `socket_->cancel()` upon destruction.

src/backend/Pg.cpp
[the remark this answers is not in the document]
natenichols: My understanding is that synchronous `COPY` should work similarly to `query`, but calling `PQputCopyData` or `PQputCopyEnd` instead of `PQsendQueryParams`.

This does _work_, I'm gonna look through the documentation to try to find how to correctly implement it.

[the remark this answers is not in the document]
natenichols: So after looking at some docs, this looks correct 
" To prevent this possibility and achieve completely nonblocking database operation, the following additional functions can be used.", and then lists `PQFlush`.

We flush after we send each command. 
"After sending any command or data on a nonblocking connection, call PQflush. If it returns 1, wait for the socket to become read- or write-ready. If it becomes write-ready, call PQflush again. If it becomes read-ready, call PQconsumeInput , then call PQflush again. Repeat until PQflush returns 0. (It is necessary to check for read-ready and drain the input with PQconsumeInput , because the server can block trying to send us data, e.g., NOTICE messages, and won't read our data until we read its.) Once PQflush returns 0, wait for the socket to be read-ready and then read the response as described above." 

Using PQflush here should eliminate the issue described above, if I'm RTFM-ing correctly.

src/backend/Pg.cpp
```diff
@@ -238,38 +448,87 @@ Pg::bulkInsert(char const* table, std::string const& records)
         throw std::runtime_error(ss.str());
     }
 
-    if (PQputCopyData(conn_.get(), records.c_str(), records.size()) == -1)
+    auto& socket = getSocket();
+
+    std::int32_t putCopy = 0;
+    while (putCopy == 0)
```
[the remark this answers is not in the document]
mtrippled: Duh, I'm almost as bad at reading code as I am at writing it.

src/backend/BackendInterface.h
```diff
@@ -21,9 +21,16 @@ class BackendInterface
 protected:
     std::optional<LedgerRange> range;
     SimpleCache cache_;
+    boost::asio::io_context& ioContext_;
```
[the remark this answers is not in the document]
mtrippled: The problem then is manually synchronizing. strands fix that, but by necessity they require the same io_context. Just have 1 big threadpool and tune it. It's not like if cassandra or postgres timeout all over the place everything else works. If they break, it's all broken.

src/backend/Pg.h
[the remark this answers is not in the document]
natenichols: No, there shouldn't be a problem if Thread A suspends a coro, and Thread B resumes it, as long as Thread A does not hold  any locks.

src/backend/CassandraBackend.h
```diff
@@ -639,10 +677,11 @@ class CassandraBackend : public BackendInterface
         ripple::AccountID const& account,
         std::uint32_t limit,
         bool forward,
-        std::optional<AccountTransactionsCursor> const& cursor) const override;
+        std::optional<AccountTransactionsCursor> const& cursor,
+        boost::asio::yield_context& yield) const override;
 
     bool
-    doFinishWrites() override
+    doFinishWrites(boost::asio::yield_context& yield) const override
```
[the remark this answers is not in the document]
natenichols: Yeah, the postgres impl needs it

src/backend/CassandraBackend.h
```diff
@@ -777,44 +835,59 @@ class CassandraBackend : public BackendInterface
...
         uint32_t seq,
         uint32_t date,
         std::string&& transaction,
-        std::string&& metadata) override;
+        std::string&& metadata,
+        boost::asio::yield_context& yield) override;
 
     void
-    startWrites() override
+    startWrites(boost::asio::yield_context& yield) const override
```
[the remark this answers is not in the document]
natenichols: Same

src/backend/PostgresBackend.cpp
```diff
@@ -393,153 +450,99 @@ PostgresBackend::doFetchSuccessorKey(
...
 std::vector<TransactionAndMetadata>
 PostgresBackend::fetchTransactions(
-    std::vector<ripple::uint256> const& hashes) const
+    std::vector<ripple::uint256> const& hashes,
+    boost::asio::yield_context& yield) const
 {
+    if (!hashes.size())
+        return {};
+
     std::vector<TransactionAndMetadata> results;
```
undertome, line 459: If we're anticipating that `results` will contain roughly `hashes.size()` entries, consider increasing the vector's capacity with `reserve`.

src/backend/PostgresBackend.cpp
```diff
@@ -393,153 +450,99 @@ PostgresBackend::doFetchSuccessorKey(
...
 PostgresBackend::doFetchLedgerObjects(
     std::vector<ripple::uint256> const& keys,
-    uint32_t sequence) const
+    uint32_t sequence,
+    boost::asio::yield_context& yield) const
 {
+    if (!keys.size())
+        return {};
+
+    std::vector<Blob> results = {};
```
undertome, line 502: If we're anticipating that `results` will contain roughly `keys.size()` entries, consider increasing the vector's capacity with `reserve`.

src/backend/PostgresBackend.cpp
```diff
@@ -393,153 +450,99 @@ PostgresBackend::doFetchSuccessorKey(
...
+        auto const& hash = hashes[i];
+        sql << "SELECT transaction,metadata,ledger_seq,date FROM "
+               "transactions "
+               "WHERE HASH = \'\\x"
+            << ripple::strHex(hash) << "\'";
+        if (i + 1 < hashes.size())
+            sql << " UNION ALL ";
     }
-    else
+    auto start = std::chrono::system_clock::now();
```
undertome, line 473: This may exceed the scope of this pull request, and I'm fine with that. However, there are several places in the code where we time the execution of various functions. Consider using [`boost::cpu_timer`](https://www.boost.org/doc/libs/1_75_0/libs/timer/doc/cpu_timers.html) to avoid manually calculating the duration.

src/backend/PostgresBackend.cpp
```diff
@@ -393,153 +450,99 @@ PostgresBackend::doFetchSuccessorKey(
...
+            results.push_back(res.asUnHexedBlob(i, 0));
+    }
+
     return results;
 }
+
 std::vector<LedgerObject>
-PostgresBackend::fetchLedgerDiff(uint32_t ledgerSequence) const
+PostgresBackend::fetchLedgerDiff(
+    uint32_t ledgerSequence,
```
undertome, line 536:
```suggestion
    std::uint32_t const ledgerSequence,
```

src/backend/PostgresBackend.cpp
```diff
@@ -706,7 +709,9 @@ PostgresBackend::doFinishWrites()
 }
 
 bool
-PostgresBackend::doOnlineDelete(uint32_t numLedgersToKeep) const
+PostgresBackend::doOnlineDelete(
+    uint32_t numLedgersToKeep,
```
undertome, line 713:
```suggestion
    std::uint32_t const numLedgersToKeep,
```

src/backend/PostgresBackend.h
```diff
@@ -16,87 +16,118 @@ class PostgresBackend : public BackendInterface
...
+    fetchLedgerByHash(
+        ripple::uint256 const& hash,
+        boost::asio::yield_context& yield) const override;
 
     std::optional<Blob>
-    doFetchLedgerObject(ripple::uint256 const& key, uint32_t sequence)
-        const override;
+    doFetchLedgerObject(
+        ripple::uint256 const& key,
+        uint32_t sequence,
```
undertome, line 43: Consider changing to `std::uint32_t const` wherever possible. I won't highlight every instance.

src/etl/ETLHelpers.h
```diff
@@ -179,4 +179,22 @@ getMarkers(size_t numMarkers)
...
+    boost::asio::spawn(ctx, [&f, &work](boost::asio::yield_context yield) {
+        f(yield);
+
+        work.reset();
+    });
+
+    ctx.run();
+}
+
+#endif  // RIPPLE_APP_REPORTING_ETLHELPERS_H_INCLUDED
```
undertome, line 200: I'm in favor of replacing include guards with `#pragma once`. This is not necessary for this pull request, but we could note it as technical debt if everyone is on board with switching over. However, this does conflict with the convention in rippled. @cjcobb23 - thoughts?

src/rpc/RPCHelpers.cpp
```diff
@@ -754,16 +767,17 @@ accountHolds(
     ripple::AccountID const& account,
     ripple::Currency const& currency,
     ripple::AccountID const& issuer,
-    bool zeroIfFrozen)
+    bool zeroIfFrozen,
```
undertome, line 770:
```suggestion
    bool const zeroIfFrozen,
```

src/rpc/RPCHelpers.cpp
```diff
@@ -819,17 +835,18 @@ postProcessOrderBook(
     ripple::Book const& book,
     ripple::AccountID const& takerID,
     Backend::BackendInterface const& backend,
-    uint32_t ledgerSequence)
+    uint32_t ledgerSequence,
```
undertome, line 838:
```suggestion
    std::uint32_t const ledgerSequence,
```

undertome (changes requested):
Great job on this - I've submitted my initial review.

undertome:
FWIW, try to avoid force-pushing while the pull request is being reviewed. This deletes history and can mangle pending reviews. The rippled approach is to include "fold" commits for changes that occur during the review: "[FOLD] parallelize db fetches". Once the pull request is approved you can squash the commits and force-push.

src/etl/ETLHelpers.h
cjcobb23: Are there any downsides to `#pragma once`?

src/backend/BackendInterface.h
```diff
@@ -173,39 +217,49 @@ class BackendInterface
...
     // Tell the database we have finished writing all data for a particular
     // ledger
     bool
-    finishWrites(uint32_t ledgerSequence);
+    finishWrites(uint32_t ledgerSequence, boost::asio::yield_context& yield);
 
     virtual bool
-    doOnlineDelete(uint32_t numLedgersToKeep) const = 0;
+    doOnlineDelete(uint32_t numLedgersToKeep, boost::asio::yield_context& yield)
+        const = 0;
```
cjcobb23, line 262: Let's get rid of the yield context for writes. Only Postgres needs it, and thus Postgres can do that internally.

cjcobb23 (changes requested):

src/backend/Pg.cpp
```diff
@@ -61,90 +61,296 @@ PgResult::msg() const
...
+        disconnect();
+        std::stringstream ss;
+        ss << "polling connection error: " << e.what();
+        throw std::runtime_error(ss.str());
+    }
+
+    /* Enable asynchronous writes. */
+    if (PQsetnonblocking(conn_.get(), 1) == -1)
+    {
+        std::stringstream ss;
```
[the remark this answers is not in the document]
mtrippled: Every new query does a connect() that ensures that there is a healthy connection (or throws). In this case, it's throwing so if the caller happens to retry with the same Pg object, then the connection status will again be checked and a reconnection attempt will be made.

src/backend/Pg.cpp
```diff
@@ -61,90 +61,296 @@ PgResult::msg() const
...
+        char* err = PQerrorMessage(conn_.get());
+        if (err)
+            ss << "Error setting connection to non-blocking: " << err;  
+        else
+            ss << "Unknown error setting connection to non-blocking";
+        throw std::runtime_error(ss.str());
+    }
+
+    if (PQstatus(conn_.get()) != CONNECTION_OK)
```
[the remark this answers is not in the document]
mtrippled: see above

mtrippled (approved):
LGTM :+1:

src/backend/BackendInterface.h
```diff
@@ -173,39 +217,49 @@ class BackendInterface
...
 
     virtual void
     writeTransaction(
         std::string&& hash,
         uint32_t seq,
         uint32_t date,
         std::string&& transaction,
-        std::string&& metadata) = 0;
+        std::string&& metadata,
+        boost::asio::yield_context& yield) = 0;
```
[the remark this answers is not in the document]
natenichols: I'm gonna remove `yield_context` from all writes. They will be synchronous.

src/backend/PostgresBackend.cpp
```diff
@@ -244,7 +258,42 @@ PostgresBackend::fetchLedgerByHash(ripple::uint256 const& hash) const
 std::optional<LedgerRange>
 PostgresBackend::hardFetchLedgerRange() const
 {
-    auto range = PgQuery(pgPool_)("SELECT complete_ledgers()");
+    auto range = PgQuerySync(pgPool_->config())("SELECT complete_ledgers()");
```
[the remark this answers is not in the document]
natenichols: I refactored this w/ the `synchronous`, and put it in `BackendInterface`
```cpp
    std::optional<LedgerRange>
    hardFetchLedgerRange() const
    {
        std::optional<LedgerRange> range = {};
        synchronous([&](boost::asio::yield_context yield)
        {
            range = hardFetchLedgerRange(yield);
        });

        return range;
    }
```

Which should call the underlying implementation for `hardFetchLedgerRange(yield_context)`

src/backend/CassandraBackend.h
```diff
@@ -1034,6 +1108,55 @@ class CassandraBackend : public BackendInterface
         cass_future_free(fut);
         return {res};
     }
+
+    CassandraResult
```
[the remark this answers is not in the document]
natenichols: I actually just got rid of `syncRead`, it was only used twice.

src/backend/PostgresBackend.cpp
```diff
@@ -332,14 +305,15 @@ PostgresBackend::doFetchLedgerObject(
     boost::asio::yield_context& yield) const
```
undertome, line 305: Unresolving this request since it wasn't fixed.

src/backend/BackendInterface.h
```diff
@@ -217,45 +246,40 @@ class BackendInterface
...
     // Tell the database we are about to begin writing data for a particular
     // ledger.
     virtual void
-    startWrites(boost::asio::yield_context& yield) const = 0;
+    startWrites() const = 0;
 
     // Tell the database we have finished writing all data for a particular
     // ledger
     bool
-    finishWrites(uint32_t ledgerSequence, boost::asio::yield_context& yield);
```
undertome: Please change to std::uint32_t const for this instance and elsewhere in this file/PR.

src/backend/BackendInterface.h
```diff
@@ -217,45 +246,40 @@ class BackendInterface
...
+    finishWrites(uint32_t const ledgerSequence);
 
     virtual bool
     doOnlineDelete(uint32_t numLedgersToKeep, boost::asio::yield_context& yield)
```
undertome, line 285: Please change to `std::uint32_t const` for this instance and elsewhere in this file/PR.

undertome (changes requested):
Thanks for addressing my changes. Just left a few comments. Almost there

src/etl/ETLHelpers.h
undertome: I mean there are some issues, we can forget this for now. Resolving.

cjcobb23 (approved):
LGTM; great work!

PR #77, opened 2022-01-05
Include client IP in header when forwarding to rippled

cjcobb23:
This allows rippled to rate limit based on the client IP passed in the header

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/77)

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -9,6 +9,7 @@
 #include <backend/DBHelpers.h>
 #include <etl/ETLSource.h>
 #include <etl/ReportingETL.h>
+#include <thread>
 
 // Create ETL source without grpc endpoint
 // Fetch ledger and load initial ledger will fail for this source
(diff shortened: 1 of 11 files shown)


src/etl/ETLSource.cpp
```diff
@@ -1026,17 +983,17 @@ ETLSourceImpl<Derived>::forwardToRippled(
...
         // https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg
         ws->set_option(websocket::stream_base::decorator(
-            [&request](websocket::request_type& req) {
+            [&request,&clientIp] (websocket::request_type& req) {
                 req.set(
                     http::field::user_agent,
                     std::string(BOOST_BEAST_VERSION_STRING) +
                         " websocket-client-coro");
                 req.set(
                     http::field::forwarded,
```
natenichols, line 992: I'm assuming this is using the `Forwarded` request header w/ `"for=1.2.3.4".

Does rippled support this, or does it only support `X-Forwarded-For`?

cjcobb23: This is how we do it in reporting mode: https://github.com/ripple/rippled/blob/89766c5f218e49f038e30aee05c1be9216f78bf5/src/ripple/app/reporting/ETLSource.cpp#L858

PR #86, opened 2022-01-13
clang-format

legleux:
Ran clang-format-10


cjcobb23:
This looks great, but let's add two things:
1. Let's add a file to remove this commit from git blame. rippled does that like so: https://github.com/ripple/rippled/blob/develop/.git-blame-ignore-revs
2. Can we add this as a github action? rippled has this as well: https://github.com/ripple/rippled/blob/develop/.github/workflows/clang-format.yml

cjcobb23:
I think you need to reformat again @legleux . You can force push.

cjcobb23 (approved):

PR #89, opened 2022-01-31
Build fix: specify template parameters

undertome:
Clio doesn't build on Ubuntu/g++ because of failed template deduction. This PR contains the fix.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/89)


cjcobb23:
Can you run clang-format on this? We're using version 10

PR #93, opened 2022-02-15
Link postgres statically

legleux:
Whoops, rippled has the same issue.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/93)


cjcobb23:
@legleux  get rid of this merge commit and then we're all good

cjcobb23 (approved):

legleux:
Ok, let's just build it all the time. Some distros have the necessary libs but most don't and the ones that do will be EOL soon anyway.

PR #97, opened 2022-02-16
Add a flag to defer cache reads while the cache is being updated

cjcobb23:
The cache is updated atomically every ledger. The size of this update depends on the size of the ledger. I think reads should just hit the database while the cache is updating.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/97)

### DIFF
--- a/src/backend/SimpleCache.cpp
+++ b/src/backend/SimpleCache.cpp
@@ -7,6 +7,7 @@ SimpleCache::update(
     uint32_t seq,
     bool isBackground)
 {
+    deferReads_ = true;
     std::unique_lock lck{mtx_};
     if (seq > latestSeq_)
     {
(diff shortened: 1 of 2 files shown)


natenichols (approved):

natenichols:
When deferring reads, this will appear behave just like a cache miss, right?

cjcobb23:
> When deferring reads, this will appear behave just like a cache miss, right?

Yes, it'll behave just like a miss. So everything will just miss cache while the cache is being updated.

undertome (approved):
My only concern is that we'll lacking any insight into how often cache attempts end up getting deferred or any other such stats. Otherwise LGTM.

cjcobb23:
> My only concern is that we'll lacking any insight into how often cache attempts end up getting deferred or any other such stats. Otherwise LGTM.

yeah we should probably keep track of such stats

PR #100, opened 2022-02-18
add max connection to test

natenichols:
Adds `max_connection: 100` to postgres configurations in unittests.

We were exceeding the default number of connections in postgres and that was causing the tests to halt.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/100)

### DIFF
--- a/unittests/main.cpp
+++ b/unittests/main.cpp
@@ -45,6 +45,7 @@ TEST(BackendTest, Basic)
                     {"database", keyspace.c_str()},
                     {"password", "postgres"},
                     {"indexer_key_shift", 2},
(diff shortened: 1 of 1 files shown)


cjcobb23:
How were we exceeding the max?

natenichols:
Now that we use coroutines, the max defaults to `numeric_limit<std::uint32_t>::max()`, because we want to be able to open as many connections as possible. 

Postgres container defaults to 100, so we were exceeding the max. Obviously when we run this in prod, we'll want to set the max_connections in postgres way higher.

PR #103, opened 2022-02-23
Make DOSGuard config loading more resilient

undertome:
Don't fail when the config is malformed

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/103)

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -29,8 +29,14 @@ class DOSGuard
         std::string const& key,
         std::uint32_t const fallback) const
     {
-        if (auto const c = getConfig(config); c)
-            return c->at(key).as_int64();
+        try
+        {
+            if (auto const c = getConfig(config))
+                return c->at(key).as_int64();
+        }
+        catch (std::exception const& e)
+        {
+        }
 
         return fallback;
     }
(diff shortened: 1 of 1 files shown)


src/webserver/DOSGuard.h
```diff
@@ -40,22 +46,26 @@ class DOSGuard
...
+            auto const transform = [](auto const& elem) {
+                return std::string(elem.as_string().c_str());
+            };
 
-        return T(
-            boost::transform_iterator(w.begin(), transform),
-            boost::transform_iterator(w.end(), transform));
+            return T(
+                boost::transform_iterator(w.begin(), transform),
+                boost::transform_iterator(w.end(), transform));
```
cjcobb23, line 63: I know this is not actually part of this PR, but can we do this with the standard library instead of boost?

cjcobb23 (approved):
LGTM. Left a non blocking comment

src/webserver/DOSGuard.h
undertome: It depends on what you mean by that. There's no `transform_iterator` in the standard library.

undertome: I don't think there's a concise way to do this without using boost.

PR #111, opened 2022-03-04
Add NFT RPC infrastructure

undertome:
Add infrastructure to handle three new RPCs relating to NFTs:

- `account_nfts`
- `nft_buy_offers`
- `nft_sell_offers`

This request also features CMake changes that modify the repository (and branch) used for building rippled.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/111)


src/rpc/RPCHelpers.cpp
```diff
@@ -476,6 +476,107 @@ traverseOwnedNodes(
...
 
+bool
+forEachItemAfter(
+    ripple::LedgerInfo const& lgrInfo,
+    ripple::Keylet const& root,
+    ripple::uint256 const& after,
+    std::uint64_t const hint,
+    std::uint64_t limit,
+    Context const& context,
+    std::function<bool(std::shared_ptr<ripple::SLE const> const&)> const f)
```
cjcobb23, line 487: Can we use `traverseOwnedNodes` for this?

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -23,7 +24,141 @@ std::unordered_map<std::string, ripple::LedgerEntryType> types{
...
+
+    ripple::uint256 cursor;
+    if (request.contains("marker"))
+    {
+        if (!request.at("marker").is_string())
+            return Status{Error::rpcINVALID_PARAMS, "markerNotString"};
+
+        if (!cursor.parseHex(request.at("marker").as_string().c_str()))
+            return Status{Error::rpcINVALID_PARAMS, "malformedCursor"};
+    }
```
cjcobb23, line 83: Out of scope for this PR, but we should find some way to simplify this code with helper functions. All of the handlers have this boilerplate code.

src/rpc/RPCHelpers.cpp
undertome: Just ran some tests - seems like it will work with a few snips.

src/rpc/handlers/AccountObjects.cpp
undertome: I agree - since this has to wait anyway maybe I can slip it into this PR.

src/rpc/RPCHelpers.cpp
undertome: Done

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -23,7 +24,141 @@ std::unordered_map<std::string, ripple::LedgerEntryType> types{
...
+
+    auto const first = ripple::keylet::nftpage(
+        ripple::keylet::nftpage_min(*accountID), cursor);
+    auto const last = ripple::keylet::nftpage_max(*accountID);
+
+    auto const key =
+        context.backend
+            ->fetchSuccessorKey(first.key, lgrInfo.seq, context.yield)
+            .value_or(last.key);
```
cjcobb23, line 70: Do we know that `first.key` actually exists? `fetchSuccessorKey` will return nothing if `first.key` is not actually in the ledger, and the cache is not full.

CMake/deps/rippled.cmake
```diff
@@ -1,5 +1,5 @@
-set(RIPPLED_REPO "https://github.com/cjcobb23/rippled.git")
-set(RIPPLED_BRANCH "clio")
+set(RIPPLED_REPO "https://github.com/undertome/rippled.git")
```
natenichols: Reminder to change this to rippled develop before merging.

src/rpc/RPCHelpers.cpp
```diff
@@ -97,6 +101,112 @@ getString(
         return dfault;
 }
 
+Status
+getMarker(boost::json::object const& request, ripple::uint256& marker)
+{
+    if (request.contains(JS(marker)))
+    {
+        if (!request.at(JS(marker)).is_string())
```
natenichols, line 184: I think markers for Account RPCs and account_tx are slightly different. 

For `account_tx`, I believe they're an object, and in the Account RPCs they contain a directory hint. Are we using this method in those handler?

src/rpc/RPCHelpers.cpp
```diff
@@ -97,6 +101,112 @@ getString(
...
+            return Status{Error::rpcINVALID_PARAMS, "markerNotString"};
+
+        if (!marker.parseHex(request.at(JS(marker)).as_string().c_str()))
+            return Status{Error::rpcINVALID_PARAMS, "malformedMarker"};
+    }
+
+    return {};
+}
+
+Status
```
natenichols, line 194: This is much better than what I did in #135, I'm gonna modify the `getLimit` helper in that PR to more closely resemble this one.

src/rpc/RPCHelpers.cpp
```diff
@@ -468,12 +597,29 @@ traverseOwnedNodes(
...
+read(
+    ripple::Keylet const& keylet,
+    ripple::LedgerInfo const& lgrInfo,
+    Context const& context)
+{
+    if (auto const blob = context.backend->fetchLedgerObject(
+            keylet.key, lgrInfo.seq, context.yield);
+        blob)
+    {
+        return std::make_shared<ripple::SLE const>(
```
natenichols, line 825: Is there an advantage to making this a `shared_ptr` over just returning a `std::optional<ripple::SLE>`?

src/rpc/RPCHelpers.h
```diff
@@ -14,6 +14,13 @@
 #include <backend/BackendInterface.h>
 #include <rpc/RPC.h>
 
+// Useful macro for borrowing from ripple::jss
+// static strings. (J)son (S)trings
+#define JS(x) ripple::jss::x.c_str()
```
natenichols, line 19: :heart:

src/rpc/RPCHelpers.cpp
natenichols: If we're not using it in those RPCs, can we call this `getHexMarker` or something like that.

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -37,55 +39,141 @@ doAccountObjects(Context const& context)
...
+            auto const nextKey = ripple::Keylet(ripple::ltNFTOKEN_PAGE, *npm);
+            auto const nextBlob = context.backend->fetchLedgerObject(
+                nextKey.key, lgrInfo.seq, context.yield);
+
+            cp.emplace(ripple::SLE{
+                ripple::SerialIter{nextBlob->data(), nextBlob->size()},
+                nextKey.key});
+        }
+        else
+            cp = std::nullopt;
```
natenichols: Can we just `break;` here?

src/rpc/RPCHelpers.cpp
```diff
@@ -411,18 +521,37 @@ traverseOwnedNodes(
...
+    return traverseOwnedNodes(
+        backend,
+        ripple::keylet::ownerDir(accountID),
+        sequence,
+        marker,
+        yield,
+        atOwnedNode);
+}
+
+std::optional<ripple::uint256>
```
natenichols: I originally wrote this function in a very naive way. In this PR, it pulls every object each time you query. So if you're paging through `n` times, you actually read the first owned node `n` times. This is _really_ stupid, and totally on me. :facepalm: 

I modified `traverseOwnedNodes` to take a hint in the `cursor` such that we didn't have to traverse from the first directory every time.

Can we rebase on develop?

natenichols (changes requested):
This looks great. My only real change is that we should rebase this PR onto develop, as I have changed `traverseOwnedNodes` since this was introduced. I'd like to look at it again once we rebase.

Overall, this looks great. The added RPCHelplers are a **huge** improvement over what we currently have.

src/rpc/RPCHelpers.cpp
undertome: Renamed. Done

src/rpc/RPCHelpers.cpp
undertome: Sure. Will rebase before merging.

src/rpc/RPCHelpers.cpp
undertome: There was an advantage while implementing this feature since the rippled source this is mimicking is also using `shared_ptr`

src/rpc/handlers/AccountObjects.cpp
undertome: Done

src/rpc/RPCHelpers.cpp
undertome: @natenichols please double check to make sure I'm using the modified function correctly.

natenichols (approved):

CMake/deps/rippled.cmake
```diff
@@ -1,5 +1,5 @@
-set(RIPPLED_REPO "https://github.com/ripple/rippled.git")
-set(RIPPLED_BRANCH "develop")
```
cjcobb23: Change this back now

cjcobb23 (changes requested):
change rippled.cmake back, now that the NFT PR has been merged into rippled

CMake/deps/rippled.cmake
cjcobb23: actually forget this, we have to change this in develop to point to a tag anyways

cjcobb23 (approved):

cjcobb23:
Merged into `develop`

PR #115, opened 2022-03-10
notifies rippled when it can_delete or when it needs a ledger

natenichols:
Adds two pieces of functionality.
- Clio will notify rippled that it `can_delete` once a ledger is extracted with `buildNextLedger`.
- Clio will attempt to have rippled `request_ledger` when an ETLSource has gaps.


natenichols:
I need to revisit this. 

We need a way to detect if we have `admin` role at our ETLSources to optionally enable this functionality. No point in running these jobs if if we don't have the correct permissions.

src/etl/ETLSource.h
```diff
@@ -232,6 +234,12 @@ class ETLSourceImpl : public ETLSource
         std::lock_guard lck(mtx_);
         validatedLedgers_ = std::move(pairs);
         validatedLedgersRaw_ = range;
+
+        if (backfilling_ || validatedLedgers_.size() == 1)
+            return;
+
```
natenichols, line 240: I should move 
```cpp
backfilling_ = true;
```
to here.

natenichols:
Closing this for now. I want to revisit this, but we should do it in a different way

PR #116, opened 2022-03-15
populate cache using coroutines. optionally disable cache

cjcobb23:
Use coroutines instead of threads so we can yield on the database reads.

---
This change is [Reviewable](https://reviewable.io/reviews/cjcobb23/clio/116)


src/etl/ReportingETL.cpp
```diff
@@ -933,38 +937,55 @@ ReportingETL::loadCacheAsync(uint32_t seq)
     {
         std::optional<ripple::uint256> start = cursors[i];
         std::optional<ripple::uint256> end = cursors[i + 1];
```
natenichols, line 939: is `cursors.back()` always going to be `{}`

src/etl/ReportingETL.cpp
```diff
@@ -933,38 +937,55 @@ ReportingETL::loadCacheAsync(uint32_t seq)
...
+                else
+                {
+                    BOOST_LOG_TRIVIAL(info)
+                        << "Finished a cursor. num remaining = "
+                        << *numRemaining;
+                }
+            });
     }
+    // If loading synchronously, poll cache until full
+    while (cacheLoadStyle_ == CacheLoadStyle::SYNC &&
```
natenichols, line 979: would it be better to wait on a `condition_variable` checking if cache is full?

natenichols (changes requested):

src/etl/ReportingETL.cpp
cjcobb23: I thought about that, and I think this loop is probably much simpler

src/etl/ReportingETL.cpp
cjcobb23: yes, we always push an empty optional to the front and back

src/etl/ReportingETL.cpp
```diff
@@ -933,38 +937,55 @@ ReportingETL::loadCacheAsync(uint32_t seq)
...
+            cacheDownloadStrands_.back(),
+            [this, seq, start, end, numRemaining](
+                boost::asio::yield_context yield) {
+                std::optional<ripple::uint256> cursor = start;
+                while (true)
+                {
+                    auto res =
+                        Backend::retryOnTimeout([this, seq, &cursor, &yield]() {
+                            return backend_->fetchLedgerPage(
+                                cursor, seq, 256, yield);
```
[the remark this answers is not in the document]
cjcobb23: It's sort of arbitrary. When it was 4096, it could take a long time to fetch a single page over a WAN, so I changed it to 256 to see progress more often. I feel that 256 is better though, since 4096 could lead to `4096 * num_cursors` objects being fetched at once. There are usually over 100 cursors.

cjcobb23:
I added some functionality to `ledger_data`. If you pass `"out_of_order":true` in the request, the ledger can be paged through much faster, but some objects will be out of order and you could receive duplicates. Also, the marker can be a string or an integer when `out_of_order` is `true`.

The way it does this is by looping through the cache, using the cache to get successors, and reading the object itself out of cache or the db, depending on how new the object in cache is. The cache is constantly being updated though, so some objects from the target ledger will be missed when looping through the cache. To remedy this, the code then walks backwards down the ledger diffs, and returns any deleted objects.

This can only be done when the cache is full. When it's not full, `out_of_order` will hit the database for successor info, and not be any faster. However, it will seamlessly transition to the new algorithm when the cache becomes full.

I've verified this algorithm is correct with the verifier tool. On s2, this brings the time needed to iterate a full ledger down to about 10 mins.

natenichols (approved):

PR #133, opened 2022-03-23
Update Build Instructions

jscottbranson:
I previously added an incorrect sentence to README.md indicating protobuf v2 is required. This is not the case - building works with protobuf v2 and v3. This PR also includes minimum version info for GCC & cmake.

I've also updated the build instructions to include `cd ~`, so that users don't accidentally build clio inside the boost directory.


README.md
```diff
@@ -42,11 +42,13 @@ Use these instructions to build a clio executable from source. These instruction
...
   source ~/.profile
-5. git clone https://github.com/XRPLF/clio.git
-6. mkdir build
-7. cd build
-8. cmake ..
-9. cmake --build . -- -j$((`nproc`+1))
+5. cd ~
+6. git clone https://github.com/XRPLF/clio.git
+7. cd clio
+8. mkdir build
```
legleux: Modern CMake allows a build/binary directory to be created when specified if it doesn't exist.
Steps 8,9,10,11 can just be
`cmake -B build && cmake --build build --parallel $(nproc)`

legleux (commented):
It's been decided. All "clio"s --> "Clio"
Also, the step numbers irk me. Why?

Thanks for getting rid of the protobuf stuff btw.

README.md
```diff
@@ -24,15 +24,15 @@ from which data can be extracted. The rippled node does not need to be running o
 2. Access to one or more rippled nodes. Can be local or remote.
 
 ## Building
-clio is built with cmake. clio requires c++20, and boost 1.75.0 or later. protobuf v2 is required and must be manually built on systems that ship with v3, including Debian 11 and Ubuntu 21.10.
+clio is built with cmake. clio requires c++20, and boost 1.75.0 or later. protobuf v2 or v3 can both be used, however, users with older systems will need to update to GCC 8 or later and cmake 3.16.3 or later.
 Use these instructions to build a clio executable from source. These instructions were tested on Ubuntu 20.04 LTS.
 
 ```
 1. sudo apt-get update
 2. sudo apt-get -y upgrade
```
legleux: Why upgrade?

README.md
```diff
@@ -42,11 +42,13 @@ Use these instructions to build a clio executable from source. These instruction
   # -------------------------------
   export BOOST_ROOT=/home/my_user/boost_1_75_0
   source ~/.profile
```
legleux: What is this trying to accomplish?
It's unnecessary.

jscottbranson: I think it's necessary to set an env variable defining where boost is, before calling cmake.

For consistency with the rest of the steps, we could change it to: `echo "export BOOST_ROOT=/home/$USER/boost_1_75_0" >> ~/.profile && source ~/.profile`

README.md
jscottbranson: That makes sense to me to simplify those steps. I also think eliminating step numbers (per your below comment) would make it easier to copy and run as a script.

README.md
jscottbranson: Are you talking about step 2? We could combine steps 1 & 2 into a single line: `sudo apt update && sudo apt -y upgrade`. There probably isn't a reason to upgrade before installing the dependencies, though, other than "it's just good practice"?

crypticrabbit:
I've updated the instructions again per the above comments. Thanks for the input, Michael!

crypticrabbit:
I've updated the instructions again per the above comments. Thanks for the input, Michael!

README.md
legleux: There's no reason these instructions _require_ an upgrade so let the user upgrade their system. If it ain't broke, don't fix it?

If any version of protobuf can be used and if you don't have it installed it gets built locally, why even mention it?
I know there was a few weird issues identifying protobuf deps but that's fixed.

I'm for succinctness here obviously ;)

README.md
legleux: These comments are referring to the line immediately above them (the sourcing of `.profile` that didn't change.)
You're right setting `BOOST_ROOT` is most definitely necessary.
Why `/home/$USER` then `~`? Maybe just cruft?
`$HOME` is nice.

jscottbranson: I don't see a reason to leave the line with `apt update/upgrade` in - I agree users can/should be able to figure that out on their own.

What about just using `$HOME` consistently? I suppose `~` would be more concise, though?

```                                                                                                                                                                                                                
# Install dependencies                                                                                                                                                                           
  sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget build-essential bison flex autoconf cmake                                                                               
                                                                                                                                                                                                                   
# Compile Boost                                                                                                                                                                                                    
  wget -O $HOME/boost_1_75_0.tar.gz https://boostorg.jfrog.io/artifactory/main/release/1.75.0/source/boost_1_75_0.tar.gz                                                                                           
  tar xvzf $HOME/boost_1_75_0.tar.gz                                                                                                                                                                               
  cd $HOME/boost_1_75_0                                                                                                                                                                                            
  ./bootstrap.sh                                                                                                                                                                                                   
  ./b2 -j$((`nproc`+1))                                                                                                                                                                                            
  echo "export BOOST_ROOT=$HOME/boost_1_75_0" >> $HOME/.profile && source $HOME/.profile                                                                                                                           
                                                                                                                                                                                                                   
# Clone the Clio Git repository & build Clio                                                                                                                                                                       
  cd $HOME                                                                                                                                                                                                         
  git clone https://github.com/XRPLF/clio.git                                                                                                                                                                      
  cd $HOME/clio                                                                                                                                                                                                    
  cmake -B build && cmake --build build --parallel $(nproc)                                                                                                                                                        
```

README.md
jscottbranson: I personally find it a tiny bit easier to have some version info in the README, even though it isn't strictly necessary. It makes it easier when I'm trying to build on different systems. I agree that protobuf doesn't need to be mentioned explicitly, aside from it being helpful for the few people who are used to the inaccurate info that was in the rippled README.

README.md
legleux: Beautiful.
``./b2 -j$((`nproc`+1))`` is still there.

jscottbranson: Remind me again what to replace that call to b2 with? I can't seem to find your original comment about it.

legleux: Oh I just meant the mixed backticks and $() in one place and just $() in another.

jscottbranson: I suppose it's probably safer to build boost w/ `nproc`, not `nproc + 1`? So change it to `./b2 -j$(nproc)`?

README.md
jscottbranson: I've removed references to protobuf.

README.md
jscottbranson: Ok, I think I've pushed all those changes now, if you want to review. Thanks again!

legleux: I love it.

jscottbranson:
I think we should be good to merge, then. I just don't want to merge my own PR :-)

PR #135, opened 2022-03-28
specify limit hi, low, and defaults in HandlerTable

natenichols:
Rewrites `handlerTable` to specify high, low, and default values for each `limit`.

### DIFF
--- a/src/rpc/RPC.cpp
+++ b/src/rpc/RPC.cpp
@@ -1,6 +1,7 @@
 #include <boost/asio/spawn.hpp>
 #include <etl/ETLSource.h>
 #include <rpc/Handlers.h>
+#include <rpc/RPCHelpers.h>
 #include <unordered_map>
 
 namespace RPC {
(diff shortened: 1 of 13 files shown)


cjcobb23:
Not sure we need default. What do you think about just defaulting to low?

src/rpc/RPC.cpp
```diff
@@ -118,31 +118,80 @@ make_error(Status const& status)
...
+
+static HandlerTable handlerTable
+{
+    {"account_channels",   &doAccountChannels,   LimitRange{10, 50, 100}},
+    {"account_currencies", &doAccountCurrencies, {}},
+    {"account_info",       &doAccountInfo,       {}},
+    {"account_lines",      &doAccountLines,      LimitRange{10, 50, 100}},
+    {"account_objects",    &doAccountObjects,    LimitRange{10, 50, 100}},
+    {"account_offers",     &doAccountOffers,     LimitRange{10, 50, 100}},
+    {"account_tx",         &doAccountTx,         LimitRange{10, 20, 40}},
```
cjcobb23: I think we can bump the hi here. What's rippleds max?

src/rpc/RPC.cpp
```diff
@@ -118,31 +118,80 @@ make_error(Status const& status)
...
+    {"gateway_balances",   &doGatewayBalances,   {}},
+    {"noripple_check",     &doNoRippleCheck,     {}},
+    {"book_offers",        &doBookOffers,        LimitRange{10, 20, 40}},
+    {"channel_authorize",  &doChannelAuthorize,  {}},
+    {"channel_verify",     &doChannelVerify,     {}},
+    {"ledger",             &doLedger,            {}},
+    {"ledger_data",        &doLedgerData,        LimitRange{10, 100, 256}},
```
cjcobb23: For ledger_data, we usually allow a max of 2048 when binary is set to true. Not sure what's best here.

src/rpc/RPC.cpp
```diff
@@ -118,31 +118,80 @@ make_error(Status const& status)
...
+    {"account_channels",   &doAccountChannels,   LimitRange{10, 50, 100}},
+    {"account_currencies", &doAccountCurrencies, {}},
+    {"account_info",       &doAccountInfo,       {}},
+    {"account_lines",      &doAccountLines,      LimitRange{10, 50, 100}},
+    {"account_objects",    &doAccountObjects,    LimitRange{10, 50, 100}},
+    {"account_offers",     &doAccountOffers,     LimitRange{10, 50, 100}},
+    {"account_tx",         &doAccountTx,         LimitRange{10, 20, 40}},
+    {"gateway_balances",   &doGatewayBalances,   {}},
+    {"noripple_check",     &doNoRippleCheck,     {}},
+    {"book_offers",        &doBookOffers,        LimitRange{10, 20, 40}},
```
cjcobb23: Let's bump the hi here to 100

cjcobb23:
I'm not even sure we need a low for each handler. We can probably just make sure the limit is positive, and just enforce a hi.

src/rpc/RPC.cpp
natenichols: I'm unsure, but I don't think its super relevant here, since rippled uses a tx queue and we use coroutines.

We should tweak these based on latencies that we gather from graphana.

natenichols: The big gain here is that limits are now in a central location in the codebase, and are easily changed.

src/rpc/RPC.cpp
natenichols: I bumped all of the existing handlers for now.

src/rpc/RPC.cpp
natenichols: I just added an extra clamp in `LedgerData.cpp` if `!binary`...

natenichols:
>I'm not even sure we need a low for each handler. We can probably just make sure the limit is positive, and just enforce a hi.

I like the `low` parameter. For things like the account RPCs where we traverse based on objects, it doesn't make a lot of sense to traverse one object at a time.

natenichols:
Lets merge #111 Before we merge this, @undertome's `getLimit` helper is much better than mine. I'll want to modify this PR based on those changes.

cjcobb23 (approved):
LGTM

PR #136, opened 2022-03-30
use 64 bit counters

natenichols:
the 32 bit counters would wrap back around to zero after hitting a couple hundred seconds. 

64 bit counters give us 580,000 years worth of RPC time, which should be sufficient.


cjcobb23 (approved):
LGTM

PR #140, opened 2022-04-13
assorted server improvements

natenichols:
save error on failure

execute sends on websocket executor

### DIFF
--- a/src/webserver/Listener.h
+++ b/src/webserver/Listener.h
@@ -61,8 +61,8 @@ public:
         if (ec == net::ssl::error::stream_truncated)
             return;
 
-        std::cerr << "Detector failed: " << message << ec.message()
-                  << std::endl;
+        BOOST_LOG_TRIVIAL(info)
+            << "Detector failed: " << message << ec.message() << std::endl;
     }
 
     // Launch the detector
--- a/src/webserver/Listener.h
+++ b/src/webserver/Listener.h
@@ -55,6 +55,15 @@ public:
     {
     }
 
+    inline void
+    fail(boost::system::error_code ec, char const* message)
+    {
+        if (ec == net::ssl::error::stream_truncated)
+            return;
+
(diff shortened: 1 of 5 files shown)


src/webserver/Listener.h
```diff
@@ -55,6 +55,16 @@ class Detector
     {
     }
 
+    inline void
+    fail(boost::system::error_code ec, char const* message)
+    {
+        if (ec == net::ssl::error::stream_truncated)
+            return;
+
+        std::cerr << "Detector failed: " << message << ec.message()
```
cjcobb23: Change this to log, not cerr

src/webserver/SslWsSession.h
```diff
@@ -146,6 +146,8 @@ class SslWsUpgrader : public std::enable_shared_from_this<SslWsUpgrader>
         boost::beast::get_lowest_layer(https_).expires_after(
             std::chrono::seconds(30));
 
+        auto self = shared_from_this();
```
cjcobb23: Is this used?

src/webserver/WsBase.h
```diff
@@ -91,8 +85,21 @@ class WsSession : public WsBase,
...
+    bool sending_ = false;
     std::queue<std::string> messages_;
 
+    void
+    wsFail(boost::beast::error_code ec, char const* what)
+    {
+        if (!ec_ && ec != boost::asio::error::operation_aborted)
+        {
+            ec_ = ec;
+            std::cerr << "wsFail: " << what << ": " << ec.message();
```
cjcobb23: Change to log

cjcobb23 (changes requested):
So we no longer need locking because everything is executing on the same strand now?

natenichols:
> So we no longer need locking because everything is executing on the same strand now?

That is correct.

cjcobb23 (approved):

PR #142, opened 2022-04-22
return null when successor not in object table

natenichols:
This is a valid case if a book no longer has any offers. 

We check that this object is > `bookEnd`, but we never actually reach that branch because we de-reference a `std::nullopt` optional. (which segfaults).

This fixes that behavior.


cjcobb23 (approved):
We need to add the assert back in somewhere to catch this case when we are iterating a full ledger (successor should always exist). But this is fine for now.

PR #144, opened 2022-04-22
fix validated_ledger location in server_info

natenichols:
* puts validated_ledger in "info"
* returns internal error if db reads fail
* reorganizes doServerInfo()

### DIFF
--- a/src/rpc/handlers/ServerInfo.cpp
+++ b/src/rpc/handlers/ServerInfo.cpp
@@ -47,7 +47,7 @@ doServerInfo(Context const& context)
 
     auto serverInfoRippled = context.balancer->forwardToRippled(
         {{"command", "server_info"}}, context.clientIp, context.yield);
-    
+
     info["load_factor"] = 1;
     if (serverInfoRippled && !serverInfoRippled->contains("error"))
     {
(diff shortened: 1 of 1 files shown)


src/rpc/handlers/ServerInfo.cpp
```diff
@@ -42,30 +69,7 @@ doServerInfo(Context const& context)
...
-
-    validatedLgr["age"] = age;
-    validatedLgr["hash"] = ripple::strHex(lgrInfo->hash);
-    validatedLgr["seq"] = lgrInfo->seq;
-    auto fees = context.backend->fetchFees(lgrInfo->seq, context.yield);
-    assert(fees.has_value());
-    validatedLgr["base_fee_xrp"] = fees->base.decimalXRP();
-    validatedLgr["reserve_base_xrp"] = fees->reserve.decimalXRP();
-    validatedLgr["reserve_inc_xrp"] = fees->increment.decimalXRP();
+        info["load_factor"] = 1;
```
cjcobb23: Isn't this supposed to use the value returned by rippled?

cjcobb23 (approved):

PR #146, opened 2022-04-28
cleanup subscriptions when session errors or is closed

natenichols:
Subscription manager keeps track of how to clean up sessions. It does this by capturing `unsub` function for Books and Accounts in lambdas, and calling those functions when a session goes out of scope.

This is by default threadsafe, because we're just calling Subscription::unsubBook, which posts work to an underlying strand.

This also adds counters for the subscriptions manager, and unsubscribe w/ `book`

### DIFF
--- a/src/backend/SimpleCache.h
+++ b/src/backend/SimpleCache.h
@@ -17,6 +17,7 @@ class SimpleCache
         uint32_t seq = 0;
         Blob blob;
     };
+
     std::map<ripple::uint256, CacheEntry> map_;
     mutable std::shared_mutex mtx_;
     uint32_t latestSeq_ = 0;
--- a/src/subscriptions/SubscriptionManager.cpp
+++ b/src/subscriptions/SubscriptionManager.cpp
@@ -5,25 +5,24 @@
 template <class T>
 inline void
 sendToSubscribers(
-    std::shared_ptr<Message>& message,
+    std::shared_ptr<Message> const& message,
     T& subscribers,
-    boost::asio::io_context::strand& strand)
+    std::atomic_uint64_t& counter)
 {
-    boost::asio::post(strand, [&subscribers, message]() {
-        for (auto it = subscribers.begin(); it != subscribers.end();)
+    for (auto it = subscribers.begin(); it != subscribers.end();)
+    {
+        auto& session = *it;
+        if (session->dead())
         {
-            auto& session = *it;
-            if (session->dead())
-            {
-                it = subscribers.erase(it);
-            }
-            else
-            {
-                session->send(message);
-                ++it;
-            }
+            it = subscribers.erase(it);
+            --counter;
         }
-    });
+        else
+        {
+            session->send(message);
+            ++it;
+        }
+    }
 }
 
 template <class T>
@@ -31,11 +30,13 @@ inline void
 addSession(
     std::shared_ptr<WsBase> session,
     T& subscribers,
-    boost::asio::io_context::strand& strand)
+    std::atomic_uint64_t& counter)
 {
-    boost::asio::post(strand, [&subscribers, s = std::move(session)]() {
-        subscribers.emplace(s);
-    });
+    if (!subscribers.contains(session))
+    {
+        subscribers.insert(session);
+        ++counter;
+    }
 }
 
 template <class T>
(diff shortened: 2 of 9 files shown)


src/subscriptions/SubscriptionManager.cpp
```diff
@@ -7,65 +7,77 @@ inline void
...
 addSession(
     std::shared_ptr<WsBase> session,
     T& subscribers,
-    boost::asio::io_context::strand& strand)
+    std::atomic_uint64_t& counter)
 {
-    boost::asio::post(strand, [&subscribers, s = std::move(session)]() {
-        subscribers.emplace(s);
-    });
+    std::cout << "ADDING SESSION" << std::endl;
```
cjcobb23: remove cout

src/backend/SimpleCache.h
```diff
@@ -16,7 +16,10 @@ class SimpleCache
     {
         uint32_t seq = 0;
         Blob blob;
+
+        ~CacheEntry() = default;
```
cjcobb23: why do we need this?

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -7,65 +7,77 @@ inline void
...
+        subscribers.erase(session);
+        --counter;
+    }
 }
 
 void
 Subscription::subscribe(std::shared_ptr<WsBase> const& session)
 {
-    addSession(session, subscribers_, strand_);
+    std::cout << "SUBBING" << std::endl;
```
cjcobb23: remove

src/webserver/HttpBase.h
```diff
@@ -349,7 +351,15 @@ handle_request(
...
-            yc, request, backend, nullptr, balancer, etl, *range, counters, ip);
+            yc,
+            request,
+            backend,
+            subscriptions,
+            balancer,
+            etl,
+            *range,
+            counters,
+            ip);
```
cjcobb23, line 362: what is this change?

cjcobb23, line 362: what is this change?

src/subscriptions/SubscriptionManager.h
```diff
@@ -195,26 +206,43 @@ class SubscriptionManager
...
+    }
 
 private:
     void
-    sendAll(
-        std::string const& pubMsg,
-        std::unordered_set<std::shared_ptr<WsBase>>& subs);
+    sendAll(std::string const& pubMsg, std::unordered_set<session_ptr>& subs);
+
+    using Cleanup = std::function<void(session_ptr)>;
```
cjcobb23: I'd add a comment explaining what this is for

cjcobb23: I'd add a comment explaining what this is for

src/subscriptions/SubscriptionManager.h
```diff
@@ -195,26 +206,43 @@ class SubscriptionManager
...
+    std::unordered_map<session_ptr, std::vector<Cleanup>> cleanup_ = {};
```
cjcobb23: nit, I'd call this `cleaners_` or `cleanupFuncs_` . a noun, instead of a verb. or even `cleanups_`

cjcobb23: nit, I'd call this `cleaners_` or `cleanupFuncs_` . a noun, instead of a verb. or even `cleanups_`

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -332,51 +368,65 @@ SubscriptionManager::forwardValidation(boost::json::object const& response)
...
 
 void
-SubscriptionManager::unsubProposedTransactions(std::shared_ptr<WsBase>& session)
+SubscriptionManager::unsubProposedTransactions(std::shared_ptr<WsBase> session)
 {
     txProposedSubscribers_.unsubscribe(session);
 }
+
+void
+SubscriptionManager::cleanup(std::shared_ptr<WsBase> session)
```
cjcobb23, line 428: why doesn't this function need a lock? I'd add a comment explaining why

cjcobb23, line 428: why doesn't this function need a lock? I'd add a comment explaining why

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -180,15 +214,17 @@ SubscriptionManager::unsubAccount(
 void
 SubscriptionManager::subBook(
     ripple::Book const& book,
-    std::shared_ptr<WsBase>& session)
+    std::shared_ptr<WsBase> session)
 {
     bookSubscribers_.subscribe(session, book);
+    cleanup_[session].emplace_back(
```
cjcobb23: Is this threadsafe? Isn't this being called outside of the strand?

cjcobb23: Is this threadsafe? Isn't this being called outside of the strand?

cjcobb23 (commented):
Added some comments about thread safety. It's a bit hard to reason about if things are safe, because the `post` is usually called deep within some call stack. I think we should add some comments about thread safety here, so readers can understand and don't accidentally change things

src/backend/SimpleCache.h
natenichols: We don't, will remove

src/subscriptions/SubscriptionManager.cpp
natenichols: It does need a lock, I'm just stupid :)

src/subscriptions/SubscriptionManager.cpp
natenichols: Nope, will use a mutex. Nice catch

src/webserver/HttpBase.h
natenichols: Passing `subscriptions` instead of `nullptr`. This is so `server_info` can report the number of subs.

cjcobb23 (approved):

PR #147, opened 2022-04-28
send messages to subscribers w/ shared_ptr

natenichols:
Defines a message type, which we create shared ptrs to. When publishing, WsSessions just hold a shared_ptr to the message, which will go out of scope once the last subscription sends the message

### DIFF
--- a/src/subscriptions/Message.h
+++ b/src/subscriptions/Message.h
@@ -1,0 +1,40 @@
+#ifndef CLIO_SUBSCRIPTION_MESSAGE_H
+#define CLIO_SUBSCRIPTION_MESSAGE_H
+
+#include <string>
+
+// This class should only be constructed once, then it can
+// be read from in parallel by many websocket senders
+class Message
+{
+    std::string message_;
+
+public:
+    Message() = delete;
+    Message(std::string&& message) : message_(std::move(message))
+    {
+    }
+
+    Message(Message const&) = delete;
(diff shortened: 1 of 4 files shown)


src/subscriptions/Message.h
```diff
@@ -0,0 +1,38 @@
...
+
+// This class should only be constructed once, then it can
+// be read from in parallel by many websocket senders
+class Message
+{
+    std::string message_;
+
+public:
+    Message() = delete;
+    Message(std::string&& message) : message_(message)
```
cjcobb23: Shouldn't it be `message_(std::move(message))`? to call the move constructor.

src/webserver/WsBase.h
```diff
@@ -168,18 +169,25 @@ class WsSession : public WsBase,
...
+    send(std::shared_ptr<Message> msg) override
     {
         net::dispatch(
             derived().ws().get_executor(),
             [this,
              self = derived().shared_from_this(),
-             msg = std::string(msg)]() {
-                messages_.push(std::move(msg));
+             msg = std::move(msg)]() {
+                messages_.push(msg);
```
cjcobb23: We can still move here, I think

src/subscriptions/Message.h
natenichols: duh, yes it should

cjcobb23 (approved):
LGTM

PR #149, opened 2022-05-03
return entryNotFound instead of ledgerNotFound

natenichols:
Return the correct error code when ledger entry is not found in `doLedgerEntry()`


cjcobb23 (approved):

ckniffen (commented):
Would `lgrNotFound` be ever thrown now?

natenichols:
> Would lgrNotFound be ever thrown now?

It will still be thrown, when the server does not have info about a given ledger. This is the correct error code for this case.

This is the only incorrect usage of `lgrNotFound` I found in the code base.

The request in question (negative UNL request in the explorer) should never return `lgrNotFound` when it is not called w/ a `ledger_index` parameter.

PR #150, opened 2022-05-04
Log cleanup

cjcobb23:
* log duration of rpc requests and request itself
* clean up book_offers logging

---
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/150)


natenichols (approved):

natenichols:
We should consider doing something similar in `getJson`, where we log the duration of JSON de-serialization if its > 1s or something, and what we were de-serializing.

PR #151, opened 2022-05-06
refactor main.cpp into an application class

natenichols:
This PR does two things. 

First, it adds `config.cpp` and `config.h`, which parse the config on startup, and throw an error if it finds a mismatched type. Currently I'm probably missing some config options, but since it compiles, and I cant find any other usages of `boost::json::object const& config`, I think I have everything. This also grants you compile time checks when working w/ a configuration. 

Second, anywhere where we pass around more than one `std::shared_ptr<Backend>, std:shared_ptr<SubscriptionManager>`, etc, we pass an `Application const& app` instead. This allows people extending the codebase to add new components to be able to access them on `Application`, instead of having to add new parameters in 80 different places. 

This is a big PR, and I'm sure I've done some things sub-optimally, so I'd love to have some folks who are better at C++ take a look (specifically w/ how I'm parsing config options).

### DIFF
--- a/src/backend/SimpleCache.cpp
+++ b/src/backend/SimpleCache.cpp
@@ -53,11 +53,13 @@ SimpleCache::getSuccessor(ripple::uint256 const& key, uint32_t seq) const
     if (!full_)
         return {};
     std::shared_lock{mtx_};
+    successorReqCounter_++;
     if (seq != latestSeq_)
         return {};
     auto e = map_.upper_bound(key);
     if (e == map_.end())
         return {};
+    successorHitCounter_++;
     return {{e->first, e->second.blob}};
 }
 
@@ -81,11 +83,13 @@ SimpleCache::get(ripple::uint256 const& key, uint32_t seq) const
     if (seq > latestSeq_)
         return {};
     std::shared_lock lck{mtx_};
+    objectReqCounter_++;
     auto e = map_.find(key);
     if (e == map_.end())
         return {};
     if (seq < e->second.seq)
         return {};
+    objectHitCounter_++;
     return {e->second.blob};
 }
 
@@ -116,5 +120,19 @@ SimpleCache::size() const
 {
     std::shared_lock lck{mtx_};
     return map_.size();
+}
+float
+SimpleCache::getObjectHitRate() const
+{
+    if (!objectReqCounter_)
+        return 1;
+    return ((float)objectHitCounter_) / objectReqCounter_;
+}
+float
+SimpleCache::getSuccessorHitRate() const
+{
+    if (!successorReqCounter_)
+        return 1;
+    return ((float)successorHitCounter_) / successorReqCounter_;
 }
 }  // namespace Backend
--- a/src/backend/SimpleCache.h
+++ b/src/backend/SimpleCache.h
@@ -18,6 +18,13 @@ class SimpleCache
         Blob blob;
     };
 
+    // counters for fetchLedgerObject(s) hit rate
+    mutable std::atomic_uint32_t objectReqCounter_;
+    mutable std::atomic_uint32_t objectHitCounter_;
+    // counters for fetchSuccessorKey hit rate
+    mutable std::atomic_uint32_t successorReqCounter_;
+    mutable std::atomic_uint32_t successorHitCounter_;
+
     std::map<ripple::uint256, CacheEntry> map_;
     mutable std::shared_mutex mtx_;
     uint32_t latestSeq_ = 0;
@@ -62,6 +69,12 @@ public:
 
     size_t
     size() const;
+
+    float
+    getObjectHitRate() const;
+
+    float
+    getSuccessorHitRate() const;
 };
 
 }  // namespace Backend
--- a/src/rpc/handlers/ServerInfo.cpp
+++ b/src/rpc/handlers/ServerInfo.cpp
@@ -90,6 +90,9 @@ doServerInfo(Context const& context)
     cache["is_full"] = context.backend->cache().isFull();
     cache["latest_ledger_seq"] =
         context.backend->cache().latestLedgerSequence();
+    cache["object_hit_rate"] = context.backend->cache().getObjectHitRate();
+    cache["successor_hit_rate"] =
+        context.backend->cache().getSuccessorHitRate();
 
     if (admin)
     {


natenichols:
Also, I still need to uncomment and fix the unittests.

src/main/config.h
```diff
@@ -0,0 +1,152 @@
+#ifndef CLIO_CONFIG_H
```
natenichols: Note, change:
`config.h/cpp` -> `Config.h/cpp`
`application.h/cpp` -> `Application.h/cpp`

CMake/ClioVersion.cmake
```diff
@@ -2,7 +2,7 @@
    read version from source
 #]===================================================================]
 
-file (STRINGS src/main/impl/Build.cpp BUILD_INFO)
+file (STRINGS src/clio/main/impl/Build.cpp BUILD_INFO)
```
cjcobb23: Why do we have another layer of nesting now? We already have `src` and `unittests`

CMake/settings.cmake
```diff
@@ -1 +1 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-narrowing -Wall -Werror -Wno-dangling-else")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-narrowing -Wall -Wno-error=deprecated-declarations -Wno-error=unused-function -Wno-error=deprecated-declarations -Wno-dangling-else")
```
cjcobb23: what are these changes for?

src/clio/backend/BackendInterface.cpp
```diff
@@ -56,7 +64,7 @@ BackendInterface::fetchLedgerObject(
     boost::asio::yield_context& yield) const
 {
     auto obj = cache_.get(key, sequence);
-    if (obj)
+    if (false)
```
cjcobb23: seems like a bug?

src/clio/backend/LayeredCache.cpp
```diff
@@ -0,0 +1,110 @@
+#include <clio/backend/LayeredCache.h>
```
cjcobb23: We deleted this file some time ago, remove

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+    reporting() const = 0;
+
+    virtual RPC::WorkQueue&
+    workers() const = 0;
+
+    virtual void
+    start() = 0;
+};
+
+class ApplicationImp : public Application
```
cjcobb23: What do we need an impl class for? Is there more than one subclass of `Application`?

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+    virtual ~Application(){};
+
+    virtual Config const&
+    config() const = 0;
+
+    virtual boost::asio::io_context&
+    rpc() const = 0;
+
+    virtual boost::asio::io_context&
+    etl() const = 0;
```
cjcobb23: These methods should probably be named something else, like `rpcIOCtx()` or something

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+    balancer() const = 0;
+
+    virtual SubscriptionManager&
+    subscriptions() const = 0;
+
+    virtual Backend::BackendInterface&
+    backend() const = 0;
+
+    virtual NetworkValidatedLedgers&
+    ledgers() const = 0;
```
cjcobb23: this needs a better name too, maybe just `netValidatedLedgers` or `p2pValidatedLedgers` or something

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+
+    virtual DOSGuard&
+    dosGuard() const = 0;
+
+    virtual ReportingETL&
+    reporting() const = 0;
```
cjcobb23: Needs a better name, we might as well just change the name of `ReportingETL` entirely actually.

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+
+    virtual RPC::WorkQueue&
+    workers() const = 0;
```
cjcobb23: I'd call this `workQueue`

src/clio/main/Application.h
```diff
@@ -0,0 +1,229 @@
...
+        initLogging(*config_);
+
+        std::vector<std::thread> rpcWorkers = {};
+        for (auto i = 0; i < config_->rpcWorkers; ++i)
+            rpcWorkers.emplace_back([this]() { rpcContext_.run(); });
+
+        std::vector<std::thread> etlWorkers = {};
+        for (auto i = 0; i < config_->etlWorkers; ++i)
+            rpcWorkers.emplace_back([this]() { etlContext_.run(); });
+
```
cjcobb23: Where do we start the http server now?

src/clio/subscriptions/SubscriptionManager.h
```diff
@@ -85,6 +87,8 @@ class SubscriptionManager
     boost::asio::io_context ioc_;
```
cjcobb23: Not in this PR, but does subscription manager need it's own io_context? Feel like we should just have two io_contexts. One for handling RPCs, and one for everything else.

src/test/cache/Cache.basicCache.cpp
```diff
@@ -0,0 +1,253 @@
+#include <gtest/gtest.h>
```
cjcobb23: These filenames are a bit weird, I'd use underscores instead of dots.

src/clio/backend/PostgresBackend.h
```diff
@@ -1,7 +1,7 @@
 #ifndef RIPPLE_APP_REPORTING_POSTGRESBACKEND_H_INCLUDED
 #define RIPPLE_APP_REPORTING_POSTGRESBACKEND_H_INCLUDED
 #include <boost/json.hpp>
-#include <backend/BackendInterface.h>
+#include <clio/backend/BackendInterface.h>
 
 namespace Backend {
 class PostgresBackend : public BackendInterface
```
cjcobb23: We should consider just removing all of this postgres code. If we want to add it back in later, we can.

src/clio/main/Config.h
```diff
@@ -0,0 +1,159 @@
...
+boost::json::object
+parseConfig(std::string const& filename);
+
+class Config
+{
+    boost::json::object json_;
+    std::shared_ptr<DatabaseConfig> dbConfigStore;
+
+public:
+    DatabaseConfig& database;
```
cjcobb23: Why is this a reference?

cjcobb23 (commented):
Some comments

src/clio/main/Application.h
natenichols: This allows us to Mock application while testing.

src/test/cache/Cache.basicCache.cpp
natenichols: this is out of scope for this PR. I'll separate tests out into a testing PR.

src/clio/main/Config.h
natenichols: This has to be polymorphic to account for different types of database configs.

src/clio/backend/PostgresBackend.h
natenichols: I like the postgres code, I think we should still support it

Especially once online_delete becomes and we can support partial history.

src/clio/subscriptions/SubscriptionManager.h
natenichols: No, really it should probably use the `WorkQueue`. Or at least it should use the `workers` io_context.

It'd be nice to use the WorkQueue, because then publishing contributes to the count of the work the server is doing.

PR #152, opened 2022-05-10
disable cache when `CacheLoadStyle::NONE`

natenichols:
Adds `atomic_bool disabled_` which lets the user disable the cache.

This is only called when `CacheLoadStyle::NONE`

### DIFF
--- a/src/backend/SimpleCache.h
+++ b/src/backend/SimpleCache.h
@@ -21,6 +21,7 @@ class SimpleCache
     mutable std::shared_mutex mtx_;
     uint32_t latestSeq_ = 0;
     std::atomic_bool full_ = false;
+    std::atomic_bool disabled_ = false;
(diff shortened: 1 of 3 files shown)


src/backend/SimpleCache.h
```diff
@@ -45,6 +46,9 @@ class SimpleCache
     std::optional<LedgerObject>
     getPredecessor(ripple::uint256 const& key, uint32_t seq) const;
 
+    void
+    setDisabled();
```
cjcobb23, line 50: are we actually calling this anywhere?

natenichols: We are _now_

natenichols: Sorry, accidentally `git stash` the function call.

cjcobb23 (approved):

PR #153, opened 2022-05-11
cache commands that dont take parameters

natenichols:


CMake/settings.cmake
```diff
@@ -1 +1 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-narrowing -Wall -Werror -Wno-dangling-else")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-narrowing -Wall -Werror -Wno-error=deprecated-declarations -Wno-dangling-else")
```
cjcobb23: remove

src/etl/ETLSource.h
```diff
@@ -24,6 +25,76 @@ class SubscriptionManager;
...
+
+    boost::asio::io_context::strand strand_;
+    boost::asio::steady_timer timer_;
+    ETLSource const& source_;
+    std::uint32_t duration_ = 10;
+
+    void
+    scheduleFreshen()
+    {
+        timer_.expires_after(boost::asio::chrono::seconds(duration_));
```
cjcobb23: Change this to milliseconds

src/etl/ETLSource.h
```diff
@@ -24,6 +25,76 @@ class SubscriptionManager;
...
+    using response_type = std::optional<boost::json::object>;
+
+    mutable std::atomic_bool stopping_ = false;
+    mutable std::shared_mutex mtx_;
+    std::unordered_map<std::string, response_type> latestForwarded_;
+
+    boost::asio::io_context::strand strand_;
+    boost::asio::steady_timer timer_;
+    ETLSource const& source_;
+    std::uint32_t duration_ = 10;
```
cjcobb23, line 39: I think we should default to every 100ms

src/etl/ETLSource.h
```diff
@@ -24,6 +25,76 @@ class SubscriptionManager;
...
+
+    void
+    scheduleFreshen()
+    {
+        timer_.expires_after(boost::asio::chrono::seconds(duration_));
+        timer_.async_wait([this](auto const& ec) { freshen(ec); });
```
cjcobb23: I'm wondering if instead of doing this on a timer, we do this every time we receive a proposed transaction

cjcobb23 (changes requested):

src/etl/ETLSource.h
natenichols: We can freshen after we hear `ledgerClosed` or something like that, but I think that this will run too often if run on every `transaction_proposed`...

natenichols: Or we can also do something like every `n` messages...

src/etl/ETLSource.h
natenichols: That seems a little too often

cjcobb23 (approved):
LGTM

Issue #145, opened 2022-04-24
cache JSON


cjcobb23:
When responding to a request, the bulk of our time is spent serializing to JSON. This is exacerbated even further for the cases where we serialize first to rippled JSON, then convert to boost JSON.

We should cache the JSON representation of commonly requested objects and transactions. For transactions, this is easy, but for objects, its a bit more difficult, because we don't have an identifier. The object ID is not sufficient, because an object changes over time but it's ID stays the same. 

There are a number of ways to go about this. We could just hash the whole blob. We could use uint256 and just invalidate modified entries on ledger update. We could store the ledger sequence with the uin256 and make sure we are getting the right object. etc.

natenichols:
We tried this in #155, but I went ahead and closed it due to increases in latency observed.. We should revisit this branch https://github.com/cjcobb23/clio/tree/json_cache_pr when the time comes to redesign it.

godexsoft:
Cache in general is part of our plan moving forward. There is no one true design at the moment. We will spend time investigating possibilities and trade offs. Will take the branch above into account. Closing this one for now.

PR #155, opened 2022-05-13
Add cache for JSON deserialization

cjcobb23:
* SimpleCache entries now have an optional JSON representation of the
object. By default, this object is an empty optional, and is written to
when that object is deserialized to JSON by a handler. The JSON objects
are only removed when the cache entry is updated as part of handling
newly validated ledgers. The cache in the worst case can grow to the
size of the full ledger, which is about 30GB at the time of writing.

---
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/155)

### DIFF
--- a/src/backend/SimpleCache.h
+++ b/src/backend/SimpleCache.h
@@ -3,6 +3,7 @@
 
 #include <ripple/basics/base_uint.h>
 #include <ripple/basics/hardened_hash.h>
+#include <boost/json.hpp>
 #include <backend/Types.h>
 #include <map>
 #include <mutex>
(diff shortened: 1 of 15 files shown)


src/backend/SimpleCache.h
```diff
@@ -16,11 +17,26 @@ class SimpleCache
...
+        std::atomic_bool hasJson = false;
+        std::atomic_bool updatingJson = false;
+        std::optional<boost::json::object> json;
+
+        CacheEntry&
+        operator=(CacheEntry&& other)
+        {
+            seq = other.seq;
+            blob = std::move(other.blob);
+            updatingJson = other.updatingJson.load();
```
cjcobb23, line 29: Need to also load `hasJson` here

cjcobb23:
Should not be merged

natenichols:
This requires more thought. We've seen a significant increase in latency under load.

We'll track this in #145 , but will close this for now. This likely requires significant redesign and won't go in version 1.0

PR #156, opened 2022-05-13
add clio version to repository

natenichols:
This mimics how rippled does versioning. Pretty much just copy and paste.

For release 0.2.0, we couldn't use string versions because of limitations to cmake. Now, when you want to use the version of clio in cmake, use the `clio_version` variable that is `SET` in ClioVersion.cmake.


legleux (commented):
I've got some changes to tack on here before this gets merged.

legleux:
LGTM!

PR #160, opened 2022-05-26
Added log rotation feature and console/file logging config options

Shoukozumi:
- Added log rotation feature, currently set to rotate for every 12h or if the log file size exceeds 2 Gb. If the log directory exceeds 50 Gb, old log files will be deleted.
- Added config options for toggling console and file logging.
- Changed config options for log file storage, now writing log files to a directory instead of a single file.

### DIFF
--- a/CMake/deps/cassandra.cmake
+++ b/CMake/deps/cassandra.cmake
@@ -10,7 +10,7 @@ if(NOT cassandra)
         ExternalProject_Add(zlib_src
             PREFIX ${nih_cache_path}
             GIT_REPOSITORY https://github.com/madler/zlib.git
-            GIT_TAG master
+            GIT_TAG v1.2.12
             INSTALL_COMMAND ""
             BUILD_BYPRODUCTS <BINARY_DIR>/${CMAKE_STATIC_LIBRARY_PREFIX}z.a
             )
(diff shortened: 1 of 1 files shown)


example-config.json
```diff
@@ -30,7 +30,9 @@
         "port":51233
     },
     "log_level":"debug",
-    "log_file":"./clio.log",
+    "log_to_console": true,
+    "log_to_file": true,
+    "log_directory":"./clio_log/",
```
cjcobb23: Let's make the rotation configurable as well, with the current parameters just used as defaults. So operators can specify how big the files can get before rotating, or how often to rotate, and the max directory size.

src/main.cpp
```diff
@@ -100,48 +100,71 @@ parse_certs(boost::json::object const& config)
     return ctx;
 }
 
+namespace src = boost::log::sources;
+namespace keywords = boost::log::keywords;
+namespace sinks = boost::log::sinks;
+namespace trivial = boost::log::trivial;
```
cjcobb23: Can we move these inside the `initLogging` function?

example-config.json
```diff
@@ -30,7 +30,12 @@
         "port":51233
     },
     "log_level":"debug",
-    "log_file":"./clio.log",
+    "log_to_console": true,
+    "log_to_file": true,
+    "log_rotation_size": 2147483648,
+    "log_directory_max_size": 53687091200,
+    "log_rotation_hour_interval": 12,
+    "log_directory":"./clio_log/",
```
legleux: This will produce this path for the log file by default:
`/var/log/clio/clio.log/clio.log`

PR #161, opened 2022-05-26
Added log rotation feature and console/file logging config options

Shoukozumi:
- Added log rotation feature, currently set to rotate for every 12h or if log file size exceeds 2 Gb. If the log directory exceeds 50 Gb, old log files will be deleted.
- Added config options for toggling console and file logging.
- Changed config options for log file storage, now writing log files to a directory instead of a single file.
- Updated config to allow log rotation size, log rotation interval, and log directory max size specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 report any issues they discover. Version 1.0 coming soon.
 
 # Clio
-Clio is an XRP Ledger API server. Clio is optimized for RPC calls, over websocket or JSON-RPC. Validated
-historical ledger and transaction data is stored in a more space efficient format,
+Clio is an XRP Ledger API server. Clio is optimized for RPC calls, over WebSocket or JSON-RPC. Validated
+historical ledger and transaction data are stored in a more space-efficient format,
 using up to 4 times less space than rippled. Clio can be configured to store data in Apache Cassandra or ScyllaDB,
 allowing for scalable read throughput. Multiple Clio nodes can share
 access to the same dataset, allowing for a highly available cluster of Clio nodes,
@@ -15,9 +15,9 @@ without the need for redundant data storage or computation.
 Clio offers the full rippled API, with the caveat that Clio by default only returns validated data.
 This means that `ledger_index` defaults to `validated` instead of `current` for all requests.
 Other non-validated data is also not returned, such as information about queued transactions.
-For requests that require access to the p2p network, such as `fee` or `submit`, Clio automatically forwards the request to a rippled node, and propagates the response back to the client. To access non-validated data for *any* request, simply add `ledger_index: "current"` to the request, and Clio will forward the request to rippled.
+For requests that require access to the p2p network, such as `fee` or `submit`, Clio automatically forwards the request to a rippled node and propagates the response back to the client. To access non-validated data for *any* request, simply add `ledger_index: "current"` to the request, and Clio will forward the request to rippled.
 
-Clio does not connect to the peer to peer network. Instead, Clio extracts data from a specified rippled node. Running Clio requires access to a rippled node
+Clio does not connect to the peer-to-peer network. Instead, Clio extracts data from a group of specified rippled nodes. Running Clio requires access to at least one rippled node
 from which data can be extracted. The rippled node does not need to be running on the same machine as Clio.
 
 
@@ -28,11 +28,11 @@ from which data can be extracted. The rippled node does not need to be running o
 
 ## Building
 
-Clio is built with cmake. Clio requires c++20, and boost 1.75.0 or later.
+Clio is built with CMake. Clio requires c++20, and boost 1.75.0 or later.
 
-Use these instructions to build a Clio executable from source. These instructions were tested on Ubuntu 20.04 LTS.
+Use these instructions to build a Clio executable from the source. These instructions were tested on Ubuntu 20.04 LTS.
 
-```
+```sh
 # Install dependencies
   sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget build-essential bison flex autoconf cmake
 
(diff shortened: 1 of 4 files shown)


legleux:
Don't merge this just yet.

example-config.json
```diff
@@ -30,7 +30,12 @@
         "port":51233
     },
     "log_level":"debug",
-    "log_file":"./clio.log",
+    "log_to_console": true,
+    "log_to_file": true,
+    "log_rotation_size": 2147483648,
```
natenichols: How hard would it be to be able to specify that in GB

Like
```json
"log_rotation_size": "2GB" // or "2MB"
```

If it's hard, don't worry about it, but it would be way more readable.

src/main.cpp
```diff
@@ -103,45 +103,75 @@ parse_certs(boost::json::object const& config)
...
+    namespace keywords = boost::log::keywords;
+    namespace sinks = boost::log::sinks;
+    namespace trivial = boost::log::trivial;
     boost::log::add_common_attributes();
     std::string format = "[%TimeStamp%] [%ThreadID%] [%Severity%] %Message%";
-    boost::log::add_console_log(
-        std::cout, boost::log::keywords::format = format);
-    if (config.contains("log_file"))
+    if (config.contains("log_to_console") &&
+        config.at("log_to_console") == true)
```
natenichols: nit: just say 

```cpp
&& config.at("log_to_console").as_bool()
```

src/main.cpp
```diff
@@ -103,45 +103,75 @@ parse_certs(boost::json::object const& config)
...
     {
-        boost::log::add_file_log(
-            config.at("log_file").as_string().c_str(),
-            boost::log::keywords::format = format,
-            boost::log::keywords::open_mode = std::ios_base::app);
+        boost::log::add_console_log(std::cout, keywords::format = format);
+    }
+    if (config.contains("log_to_file") && config.at("log_to_file") == true &&
```
natenichols: ditto
```cpp
config.at("log_to_file").as_bool()
```

src/main.cpp
```diff
@@ -103,45 +103,75 @@ parse_certs(boost::json::object const& config)
...
+        config.contains("log_directory"))
+    {
+        boost::filesystem::path dirPath{
```
natenichols: I would add a check here that is 

```cpp
if (!config.at("log_directory").is_string())
    throw std::runtime_error("log directory must be a string");
```

or something similar. that will be more helpful than the error boost.json throws, which is 
`bad_cast` or something.

natenichols: Our config parsing happens all over the place, which is bad. We should be parsing the config on startup and throwing errors immediately if its wrong. 

#151 fixes that, tho we have to get reviews on it before we can merge...

src/main.cpp
```diff
@@ -103,45 +103,75 @@ parse_certs(boost::json::object const& config)
...
+            keywords::format = format,
+            keywords::open_mode = std::ios_base::app,
+            keywords::rotation_size = rotationSize,
+            keywords::time_based_rotation =
+                sinks::file::rotation_at_time_interval(
+                    boost::posix_time::hours(rotationPeriod)));
+        fileSink->locked_backend()->set_file_collector(
+            sinks::file::make_collector(
+                keywords::target = dirPath, keywords::max_size = dirSize));
+        fileSink->locked_backend()->scan_for_files();
```
natenichols: So, does rotation happen when the `file > dirSize` or every `rotationPeriod` seconds, or both?

I.E. if we hit our limit, does that cancel out the time based rotation, or will we see something like this:
```
debug001.log -> 2GB
debug002.log -> 2GB
debug003.log -> 1GB // here we hit the time based rotation when the file is only 1GB
debug004.log -> 2GB
```

natenichols (changes requested):
Awesome work on this! 

This change make log rotation highly configurable, which will be super useful.

I have a couple of questions about how functionality behaves, and I can't actually test it out right now since I'm not on my workstation. I will test it out locally next tuesday.

src/main.cpp
Shoukozumi: The rotation happens whenever either of the two conditions is met (when the main log file exceeds the limit or enough time passes since the last rotation). So yes, when the log is rotated because of the file size limit, the time will reset.

src/main.cpp
Shoukozumi: Fixed

src/main.cpp
Shoukozumi: Fixed

README.md
```diff
@@ -17,7 +17,7 @@ This means that `ledger_index` defaults to `validated` instead of `current` for
 Other non-validated data is also not returned, such as information about queued transactions.
 For requests that require access to the p2p network, such as `fee` or `submit`, Clio automatically forwards the request to a rippled node, and propagates the response back to the client. To access non-validated data for *any* request, simply add `ledger_index: "current"` to the request, and Clio will forward the request to rippled.
 
-Clio does not connect to the peer to peer network. Instead, Clio extracts data from a specified rippled node. Running Clio requires access to a rippled node
+Clio does not connect to the peer to peer network. Instead, Clio extracts data from a group of specified rippled node. Running Clio requires access to at least one rippled node
```
natenichols: `rippled nodes`

README.md
```diff
@@ -143,3 +145,26 @@ are doing this, be aware that database traffic will be flowing across regions,
 which can cause high latencies. A possible alternative to this is to just deploy
 a database in each region, and the Clio nodes in each region use their region's database.
 This is effectively two systems.
+
```
natenichols, line 148: The config documentation section looks fantastic. Nice work

natenichols (approved):
I'll run your changes locally tomorrow, so don't merge until then (make sure @legleux is good to merge this too) but this looks good to me.

README.md
Shoukozumi: Fixed

src/main.cpp
```diff
@@ -103,45 +103,77 @@ parse_certs(boost::json::object const& config)
...
+    namespace keywords = boost::log::keywords;
+    namespace sinks = boost::log::sinks;
+    namespace trivial = boost::log::trivial;
     boost::log::add_common_attributes();
     std::string format = "[%TimeStamp%] [%ThreadID%] [%Severity%] %Message%";
-    boost::log::add_console_log(
-        std::cout, boost::log::keywords::format = format);
-    if (config.contains("log_file"))
+    if (config.contains("log_to_console") &&
+        config.at("log_to_console").as_bool())
```
cjcobb23: I think we should default to logging to console

Shoukozumi: Fixed

natenichols:
Tested this locally, seems to work as expected.

src/backend/README.md
```diff
@@ -1,174 +1,135 @@
-The data model used by clio is different than that used by rippled.
```
natenichols: Why did the `backend` README.md get added to this PR?

legleux:
LGTM! :shipit:

PR #163, opened 2022-05-27
dont report error.what() when returning rpcINTERNAL

natenichols:


cjcobb23:
why do we need this change?

natenichols:
`rpcINTERNAL` should never happen. If we want a nice error with an error message, it should be returned as a `Status`.

Its a security risk to tell users exactly what went wrong in this scenario. We log an error, but we shouldn't be telling users "hey, you got an out of bounds error reading some JSON".

cjcobb23 (approved):

PR #167, opened 2022-06-01
error when marker does not exist in ledger_data

natenichols:


cjcobb23:
Not sure about this one. There's a special flag called `out_of_order` that iterates through the cache, and might return markers that don't exist, but that's ok. Need to think about this.

It might be better if we somehow have a special return value to `fetchSuccessor` that represents the end of the ledger. We already have a marker for the beginning. But that would require a data model change.

What exactly is causing this anyways? The marker is supposed to be a value returned previously by this server. If the user is passing markers from a different server, that's user error. We might be able to get away with just some sort of warning. We could detect if the marker is small, but no objects are returned, and issue a warning, like "Markers are not interchangeable across servers".

src/rpc/handlers/LedgerData.cpp
```diff
@@ -73,6 +73,10 @@ doLedgerData(Context const& context)
         return *status;
 
     auto lgrInfo = std::get<ripple::LedgerInfo>(v);
+
+    if (!context.backend.fetchLedgerObject(cursor, lgrInfo.seq))
```
cjcobb23: I think `cursor` should be marker. Also, let's only do this check if `outOfOrder` is false. And lastly, we should add some additional message to the response warning about reusing markers from different servers

cjcobb23 (changes requested):

src/rpc/handlers/LedgerData.cpp
natenichols: Cool, I moved this around.

cjcobb23 (approved):
LGTM

PR #168, opened 2022-06-01
Use separate IO context for socket IO

cjcobb23:
* Keep track of number of requests currently being processed
* Reject new requests when number of in flight requests exceeds a
  configurable limit
* Track time spent between request arrival and start of request
  processing

---
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/168)

### DIFF
--- a/src/rpc/WorkQueue.cpp
+++ b/src/rpc/WorkQueue.cpp
@@ -1,0 +1,11 @@
+#include <rpc/WorkQueue.h>
+
+WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
+{
+    if (maxSize != 0)
+        maxSize_ = maxSize;
+    while (--numWorkers)
+    {
+        threads_.emplace_back([this] { ioc_.run(); });
+    }
+}
--- a/src/rpc/WorkQueue.h
+++ b/src/rpc/WorkQueue.h
@@ -1,0 +1,81 @@
+#ifndef CLIO_WORK_QUEUE_H
+#define CLIO_WORK_QUEUE_H
+
+#include <boost/asio.hpp>
+#include <boost/asio/spawn.hpp>
+#include <boost/json.hpp>
+#include <boost/log/trivial.hpp>
+
+#include <memory>
+#include <queue>
+#include <shared_mutex>
+#include <thread>
+
+class WorkQueue
+{
+    // these are cumulative for the lifetime of the process
+    std::atomic_uint64_t queued_ = 0;
+    std::atomic_uint64_t durationUs_ = 0;
+
+    std::atomic_uint64_t curSize_ = 0;
+    uint32_t maxSize_ = std::numeric_limits<uint32_t>::max();
+
+public:
+    WorkQueue(std::uint32_t numWorkers, uint32_t maxSize = 0);
+
+    template <typename F>
+    bool
+    postCoro(F&& f, bool isWhiteListed)
+    {
+        if (curSize_ >= maxSize_ && !isWhiteListed)
+        {
+            BOOST_LOG_TRIVIAL(warning)
+                << __func__
+                << " queue is full. rejecting job. current size = " << curSize_
+                << " max size = " << maxSize_;
+            return false;
+        }
+        ++curSize_;
+        auto start = std::chrono::system_clock::now();
+        // Each time we enqueue a job, we want to post a symmetrical job that
+        // will dequeue and run the job at the front of the job queue.
+        boost::asio::spawn(
+            ioc_,
+            [this, f = std::move(f), start](boost::asio::yield_context yield) {
+                auto run = std::chrono::system_clock::now();
+                auto wait =
+                    std::chrono::duration_cast<std::chrono::microseconds>(
+                        run - start)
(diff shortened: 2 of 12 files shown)


src/rpc/WorkQueue.cpp
```diff
@@ -0,0 +1,13 @@
+#include <ranges>
+#include <rpc/WorkQueue.h>
+
+WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
```
natenichols, line 3: Why not make `maxSize` optional?
```cpp
WorkQueue::WorkQueue(std::uint32_t numWorkers, std::optional<std::uint32_t> maxSize = {})
```

src/rpc/WorkQueue.h
```diff
@@ -0,0 +1,81 @@
...
+                auto run = std::chrono::system_clock::now();
+                auto wait =
+                    std::chrono::duration_cast<std::chrono::microseconds>(
+                        run - start)
+                        .count();
+                // increment queued_ here, in the same place we implement
+                // durationUs_
+                ++queued_;
+                durationUs_ += wait;
+                BOOST_LOG_TRIVIAL(debug) << "WorkQueue wait time = " << wait
```
natenichols, line 54: IMO this should be trace and will pollute our logs a lot (if it runs every job)

src/rpc/WorkQueue.h
```diff
@@ -0,0 +1,81 @@
...
+                                         << " queue size = " << curSize_;
+                f(yield);
+                --curSize_;
+            });
+        return true;
+    }
+
+    // TODO: this is not actually being called. Wait for application refactor
+    boost::json::object
```
natenichols, line 63: Can we just put this in `counters` for server_info? or are we waiting so we don't have to add a whole bunch of parameters?

src/main.cpp
```diff
@@ -184,16 +184,16 @@ main(int argc, char* argv[])
         ? std::optional<std::reference_wrapper<ssl::context>>{ctx.value()}
         : std::nullopt;
 
-    auto const threads = config->contains("workers")
-        ? config->at("workers").as_int64()
-        : std::thread::hardware_concurrency();
+    auto const threads = config->contains("io_threads")
+        ? config->at("io_threads").as_int64()
+        : 2;
```
natenichols: so now `workers` is the number of threads actually processing requests?

natenichols (approved):

src/rpc/WorkQueue.h
```diff
@@ -0,0 +1,81 @@
...
+            ioc_,
+            [this, f = std::move(f), start](boost::asio::yield_context yield) {
+                auto run = std::chrono::system_clock::now();
+                auto wait =
+                    std::chrono::duration_cast<std::chrono::microseconds>(
+                        run - start)
+                        .count();
+                // increment queued_ here, in the same place we implement
+                // durationUs_
+                ++queued_;
```
natenichols, line 52: We should increment queued right after we `spawn`, not in the function we pass to `spawn`. If we do it this way, I think we won't count a job as `queued` until it gets CPU time.

If you're observing something different lemme know, I've been known to be wrong.

cjcobb23: I originally had it like that, but the issue then is `queued_` gets incremented but there is a time gap before `durationUs_` is incremented, and so when the average time spent queued is calculated, the calculation could be off (less than the actual value). Right? Dummy check me here.

src/main.cpp
cjcobb23: yes, though I'm not tied to this change. I think these are the field names rippled uses, but we can use something else. What do you think?

src/rpc/WorkQueue.h
cjcobb23: I want to put it in that response, we just would need to add the `WorkQueue` to the `Context` right?

src/rpc/WorkQueue.h
cjcobb23: Agree

src/rpc/WorkQueue.cpp
cjcobb23: good idea

cjcobb23: Also, do you think maxSize should default to some low value? like `x*numWorkers` where `x` could be like 100 or something.

natenichols: Nah, just leave it as unlimited. If someone is running in prod, they can specify a reasonable value.

src/rpc/WorkQueue.h
natenichols: Yea, I think you'd only have to add it there. 

It'll be much easier when #151 is merged.

src/main.cpp
natenichols: Nope, I think `workers` is a good name.

src/rpc/WorkQueue.h
natenichols: `durationUs_` looks fine here. Its the time between the `start` and `run`.

It looks like you're not even using `queued_`, you're using `curSize_`.

PR #169, opened 2022-06-03
remove unused file in `src/clio`

natenichols:
`/src/clio` is used in my Application refector to allow for easier building of unittests.

It accidentally got included in #162


natenichols:
I'm probably not gonna wait for approval, because then people will notice how dumb I am.

Well, that and the fact that this is unused means it doesn't change anything :)

PR #172, opened 2022-06-07
fix ledger_index_min/max in account_tx response

### LINKED ISSUE #171: account_tx ledger_index_min returns 0
Request:
`AccountTransactionsRequestParams{account=r9m6MwViR4GnUNqoGXGa8eroBrZ9FAPHFS, ledgerSpecifier=Optional[LedgerSpecifier{ledgerIndex=72183480}], binary=false, forward=false}`
Response:
`{"result":{"account":"r9m6MwViR4GnUNqoGXGa8eroBrZ9FAPHFS","marker":{"ledger":71770866,"seq":20},"ledger_index_max":72183480,"ledger_index_min":0,"transactions":[],"validated":true,"status":"success"}}`

natenichols:
There are two issues with this
1) We had the branches for `forward` swapped. If !forward, you search from highest to lowest, so min should be the cursor, and vice versa.
2) we should never use `minReturnedIndex`. We verify that `minIndex` exists, so if we're searching forward, min index searched should be `minIndex`. Same with `maxIndex`

Fixes #171

### DIFF
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -65,7 +65,7 @@ doAccountTx(Context const& context)
             if (context.range.maxSequence < min.as_int64() ||
                 context.range.minSequence > min.as_int64())
                 return Status{
-                    Error::rpcINVALID_PARAMS, "ledgerSeqMaxOutOfRange"};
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):
LGTM

Issue #46, opened 2021-10-31
Return error when data is stale


cjcobb23:
If the database is not caught up with the network, we should return an error or at least a warning. Reasons for not being caught up are historical data is being loaded, there are no writer nodes deployed, or there are no in sync rippled nodes to talk to.

We can detect this by the recorded close time of the most recent ledger in the db.

cjcobb23:
Let's return a warning

cjcobb23:
Actually, if the user requests data from the most recently validated ledger, we should return an error, because we actually don't have it. If they are requesting data from a specific ledger, let's return that, but still append a warning saying clio is behind.

cjcobb23:
If database is over 1 minute behind, append a warning to any rpc response

PR #175, opened 2022-06-09
append warning for stale data (#46)

ethanlabelle:
This is a small PR to add a warning to all RPC responses when Clio has not published a ledger for 60 seconds. 
```
Manual Testing Strategy (HTTP requests):
- Start rippled, wait for its ledger to sync
- Start clio
- Send clio http requests: 
    - i.e.  curl localhost:51233 -d {"method": "server_info"}' | jq 
- Assert no warnings
- Stop rippled process
- Wait one minute
- Send clio http requests
- Assert warnings
- Turn on rippled, wait for sync
- Send requests to clio
- Verify warning disappears when clio starts publishing ledgers
```

The testing procedure for websocket connections is nearly identical. Instead of using curl to send http requests, create a ws connection using a tool such as wscat.

### DIFF
--- a/src/etl/ReportingETL.h
+++ b/src/etl/ReportingETL.h
@@ -130,19 +130,12 @@ private:
     /// server_info
     std::chrono::time_point<std::chrono::system_clock> lastPublish_;
 
-    mutable std::mutex publishTimeMtx_;
-
-    std::chrono::time_point<std::chrono::system_clock>
-    getLastPublish() const
-    {
-        std::unique_lock<std::mutex> lck(publishTimeMtx_);
-        return lastPublish_;
-    }
+    mutable std::shared_mutex publishTimeMtx_;
 
     void
     setLastPublish()
     {
-        std::unique_lock<std::mutex> lck(publishTimeMtx_);
+        std::unique_lock lck(publishTimeMtx_);
         lastPublish_ = std::chrono::system_clock::now();
     }
 
@@ -322,12 +315,24 @@ public:
         result["read_only"] = readOnly_;
         auto last = getLastPublish();
         if (last.time_since_epoch().count() != 0)
-            result["last_publish_age_seconds"] = std::to_string(
-                std::chrono::duration_cast<std::chrono::seconds>(
-                    std::chrono::system_clock::now() - getLastPublish())
-                    .count());
-
+            result["last_publish_age_seconds"] =
+                std::to_string(lastPublishAgeSeconds());
         return result;
+    }
+
+    std::chrono::time_point<std::chrono::system_clock>
+    getLastPublish() const
+    {
+        std::shared_lock lck(publishTimeMtx_);
+        return lastPublish_;
+    }
+
+    std::uint32_t
+    lastPublishAgeSeconds() const
+    {
+        return std::chrono::duration_cast<std::chrono::seconds>(
+                   std::chrono::system_clock::now() - getLastPublish())
+            .count();
     }
 };
 
--- a/src/webserver/HttpBase.h
+++ b/src/webserver/HttpBase.h
@@ -405,9 +405,25 @@ handle_request(
             responseStr = boost::json::serialize(response);
         }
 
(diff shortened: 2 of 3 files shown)


src/etl/ReportingETL.h
```diff
@@ -130,19 +130,12 @@ class ReportingETL
...
-        std::unique_lock<std::mutex> lck(publishTimeMtx_);
-        return lastPublish_;
-    }
+    mutable std::shared_mutex publishTimeMtx_;
 
     void
     setLastPublish()
     {
-        std::unique_lock<std::mutex> lck(publishTimeMtx_);
+        std::unique_lock<std::shared_mutex> lck(publishTimeMtx_);
```
natenichols: nit: lets remove this template. Let the compiler infer the type of the mutex.

```cpp
std::unique_lock lck(publishTimeMtx_);
```

src/etl/ReportingETL.h
```diff
@@ -329,6 +322,13 @@ class ReportingETL
 
         return result;
     }
+
+    std::chrono::time_point<std::chrono::system_clock>
+    getLastPublish() const
+    {
+        std::shared_lock<std::shared_mutex> lck(publishTimeMtx_);
```
natenichols: Same here.

src/webserver/HttpBase.h
```diff
@@ -405,8 +405,27 @@ handle_request(
             responseStr = boost::json::serialize(response);
         }
 
+        auto warningFlag = false;
+        result["warning"] = boost::json::array{};
```
natenichols: You could try something like:
```cpp
boost::json::array& warning = result["warning"].as_array();
....
warning.emplace_back("Too many requests");
```

These are pretty much functionally equivalent, so feel free to leave it as is, too.

src/webserver/HttpBase.h
```diff
@@ -405,8 +405,27 @@ handle_request(
...
+                context->etl->getLastPublish())
+                .count();
+        if (lastPublishAgeMinutes >= 1)
+        {
+            result["warning"].as_array().emplace_back(
+                "This server may be out of date");
+            warningFlag = true;
+        }
+        // reserialize only if a warning was appended.
+        if (warningFlag)
```
natenichols, line 422: Good catch on this bug.

src/webserver/WsBase.h
```diff
@@ -320,9 +321,28 @@ class WsSession : public WsBase,
...
+        result["warning"] = boost::json::array{};
         std::string responseStr = boost::json::serialize(response);
-        dosGuard_.add(*ip, responseStr.size());
+        auto warningFlag = false;
+        if (!dosGuard_.add(*ip, responseStr.size()))
+        {
+            result["warning"].as_array().emplace_back("Too many requests");
+            warningFlag = true;
+        }
+        auto lastPublishAgeMinutes =
```
natenichols: We may want to consider putting this as a member in `ReportingETL` so we don't have to do this `duration_cast` everywhere we want this info from the etl. 

```cpp
std::uint32_t 
ReportingETL::lastPublishAgeSeconds()
{
    return std::chrono::duration_cast<std::chrono::minutes>(
                 std::chrono::system_clock::now() - getLastPublish())
                 .count();
}
```

If you chose to go this route, also call this in `getInfo()`

natenichols (approved):
Nice work! None of my comments should hold up approval. 

I'm pretty indifferent towards my comments. If you like the suggestions, feel free to implement them. Else, leave a comment why you disagree.

src/webserver/HttpBase.h
ethanlabelle: I chose to take a slightly different approach that I believe is very clear.
```
boost::json::array warnings; // creates empty json array
...
warnings.emplace_back("Too many requests");
...
result["warning"] = warnings;
```
This way, the warnings array is not connected to the response json object until after all warnings, if any, have been added. Just need to make sure that result is a reference to response["result"] 😉

PR #177, opened 2022-06-14
Iterate account nft pages without using successor

cjcobb23:
NFTs are iterated in reverse order, starting from the max page, working towards the min page.

fetchSuccessorKey will not work if the input key does not exist, and clio's cache is not full. The alternative approach to this is to insert dummy entries into the successor table, but this is quite tedious. Instead, since the max page is guaranteed to exist if an account has any NFTs, we start from the max page and iterate backwards.

---
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/177)

### DIFF
--- a/src/rpc/handlers/AccountObjects.cpp
+++ b/src/rpc/handlers/AccountObjects.cpp
@@ -52,7 +52,9 @@ doAccountNFTs(Context const& context)
             ripple::keylet::account(accountID).key, lgrInfo.seq, context.yield))
         return Status{Error::rpcACT_NOT_FOUND, "accountNotFound"};
 
-    std::uint32_t limit = 200;
+    // limit controls the number of pages being returned. each page could have
+    // 32 nfts
+    std::uint32_t limit = 10;
     if (auto const status = getLimit(request, limit); status)
         return status;
 
@@ -60,46 +62,33 @@ doAccountNFTs(Context const& context)
     if (auto const status = getHexMarker(request, marker); status)
         return status;
 
-    auto const first =
-        ripple::keylet::nftpage(ripple::keylet::nftpage_min(accountID), marker);
-    auto const last = ripple::keylet::nftpage_max(accountID);
+    response[JS(account)] = ripple::toBase58(accountID);
+    response[JS(validated)] = true;
 
-    auto const key =
-        context.backend
-            ->fetchSuccessorKey(first.key, lgrInfo.seq, context.yield)
-            .value_or(last.key);
-    auto const blob = context.backend->fetchLedgerObject(
-        ripple::Keylet(ripple::ltNFTOKEN_PAGE, key).key,
-        lgrInfo.seq,
-        context.yield);
-
-    std::optional<ripple::SLE const> cp{
-        ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, key}};
-
-    std::uint32_t cnt = 0;
+    std::uint32_t numPages = 0;
     response[JS(account_nfts)] = boost::json::value(boost::json::array_kind);
     auto& nfts = response.at(JS(account_nfts)).as_array();
 
-    bool pastMarker = marker.isZero();
-    ripple::uint256 const maskedMarker = marker & ripple::nft::pageMask;
+    // if a marker was passed, start at the page specified in marker. Else,
+    // start at the max page
+    auto const pageKey =
+        marker.isZero() ? ripple::keylet::nftpage_max(accountID).key : marker;
 
-    // Continue iteration from the current page:
-    while (true)
+    auto const blob =
+        context.backend->fetchLedgerObject(pageKey, lgrInfo.seq, context.yield);
+    if (!blob)
+        return response;
+    std::optional<ripple::SLE const> page{
+        ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, pageKey}};
+
+    // Continue iteration from the current page
+    while (page)
     {
-        auto arr = cp->getFieldArray(ripple::sfNFTokens);
+        auto arr = page->getFieldArray(ripple::sfNFTokens);
 
         for (auto const& o : arr)
         {
             ripple::uint256 const nftokenID = o[ripple::sfNFTokenID];
-            ripple::uint256 const maskedNftokenID =
-                nftokenID & ripple::nft::pageMask;
-
-            if (!pastMarker && maskedNftokenID < maskedMarker)
-                continue;
-
-            if (!pastMarker && maskedNftokenID == maskedMarker &&
-                nftokenID <= marker)
-                continue;
 
             {
                 nfts.push_back(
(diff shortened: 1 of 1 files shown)


src/rpc/handlers/AccountObjects.cpp
```diff
@@ -84,21 +78,23 @@ doAccountNFTs(Context const& context)
...
-    while (true)
+    while (cp)
     {
         auto arr = cp->getFieldArray(ripple::sfNFTokens);
 
-        for (auto const& o : arr)
+        for (size_t i = 0; i < arr.size(); ++i)
         {
+            size_t idx = arr.size() - i - 1;
+            auto const& o = arr[idx];
```
undertome:
```suggestion
        for (auto itr = std::make_reverse_iterator(arr.end()); itr != std::make_reverse_iterator(arr.begin()); ++itr))
        {
            auto const& o = *itr;
```

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -60,21 +60,15 @@ doAccountNFTs(Context const& context)
...
     auto const blob = context.backend->fetchLedgerObject(
-        ripple::Keylet(ripple::ltNFTOKEN_PAGE, key).key,
-        lgrInfo.seq,
-        context.yield);
+        last.key, lgrInfo.seq, context.yield);
 
-    std::optional<ripple::SLE const> cp{
-        ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, key}};
+    std::optional<ripple::SLE const> cp;
+    if (blob)
```
undertome: Instead of this conditional, how about an early return the blob isn't good?

cjcobb23:
I changed the marker to be a pair of values, so that way we can jump directly to the nft page we need to continue iterating from

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -56,49 +56,50 @@ doAccountNFTs(Context const& context)
...
+    auto const last = ripple::keylet::nftpage_max(accountID);
+    bool hasMarker = !pageMarker.isZero();
+    auto const key = hasMarker ? pageMarker : last.key;
+
+    auto const blob =
+        context.backend->fetchLedgerObject(key, lgrInfo.seq, context.yield);
+    if (!blob)
+        return response;
+    std::optional<ripple::SLE const> cp{
+        ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, key}};
```
ledhed2222: why does this have to be optional now?

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -56,49 +56,50 @@ doAccountNFTs(Context const& context)
...
     response[JS(account_nfts)] = boost::json::value(boost::json::array_kind);
     auto& nfts = response.at(JS(account_nfts)).as_array();
 
-    bool pastMarker = marker.isZero();
-    ripple::uint256 const maskedMarker = marker & ripple::nft::pageMask;
+    // start iteration at the last page, which is guaranteed to exist if the
+    // account owns any NFTs
+    auto const last = ripple::keylet::nftpage_max(accountID);
+    bool hasMarker = !pageMarker.isZero();
+    auto const key = hasMarker ? pageMarker : last.key;
```
ledhed2222: nit:

```C++
auto const key = hasMarker ? pageMarker : ripple::keylet::nftpage_max(accountID);
```

just to get rid of one line and not take up stack memory space for `last`

src/rpc/handlers/AccountObjects.cpp
natenichols: Why not just use `arr.rbegin()` and `arr.rend()`?

src/rpc/RPCHelpers.cpp
```diff
@@ -175,6 +175,35 @@ getString(
...
+
+        boost::json::object const& marker = request.at(JS(marker)).as_object();
+
+        if (!marker.at("page_marker").is_string())
+            return Status{Error::rpcINVALID_PARAMS, "pageMarkerNotString"};
+
+        if (!pageMarker.parseHex(marker.at("page_marker").as_string().c_str()))
+            return Status{Error::rpcINVALID_PARAMS, "malformedPageMarker"};
+
+        if (!marker.at("token_marker").is_string())
```
natenichols: Can we add a check for `marker.contains("token_marker")`?

src/rpc/RPCHelpers.cpp
```diff
@@ -175,6 +175,35 @@ getString(
...
+    ripple::uint256& tokenMarker)
+{
+    if (request.contains(JS(marker)))
+    {
+        if (!request.at(JS(marker)).is_object())
+            return Status{Error::rpcINVALID_PARAMS, "markerNotObject"};
+
+        boost::json::object const& marker = request.at(JS(marker)).as_object();
+
+        if (!marker.at("page_marker").is_string())
```
natenichols: can we add a check for `marker.contains("page_marker")`?

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -139,10 +143,9 @@ doAccountNFTs(Context const& context)
                 nextKey.key});
         }
         else
-            break;
+            cp.reset();
```
natenichols: Why change this to a `reset()` over a `break`?

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -56,49 +56,50 @@ doAccountNFTs(Context const& context)
     if (auto const status = getLimit(request, limit); status)
         return status;
 
-    ripple::uint256 marker;
-    if (auto const status = getHexMarker(request, marker); status)
+    ripple::uint256 pageMarker;
+    ripple::uint256 tokenMarker;
```
scottschurr: I'm not excited about the default values of these being used as flags that they are not present.  Consider using `std::optional<ripple::uint256>`.

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -56,49 +56,50 @@ doAccountNFTs(Context const& context)
...
                 nftokenID & ripple::nft::pageMask;
 
-            if (!pastMarker && maskedNftokenID < maskedMarker)
+            if (hasMarker && maskedNftokenID < maskedMarker)
                 continue;
 
-            if (!pastMarker && maskedNftokenID == maskedMarker &&
-                nftokenID <= marker)
+            if (hasMarker && maskedNftokenID == maskedMarker &&
+                nftokenID <= tokenMarker)
```
scottschurr: Yup, these look right.

I'll mention that the reason I used `bool pastMarker` in my own implementation was to short circuit these `uint256` comparisons.  We know that the entries in the page are sorted (albeit oddly), so once we're past the marker we no longer need to do the comparisons.  In the code as currently written the `maskedNftokenID` comparisons will happen every time, even once you have passed the marker.  So the code as it stands is correct, but it's a little less efficient than it might be.

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -56,49 +56,50 @@ doAccountNFTs(Context const& context)
...
     std::uint32_t cnt = 0;
     response[JS(account_nfts)] = boost::json::value(boost::json::array_kind);
     auto& nfts = response.at(JS(account_nfts)).as_array();
 
-    bool pastMarker = marker.isZero();
-    ripple::uint256 const maskedMarker = marker & ripple::nft::pageMask;
+    // start iteration at the last page, which is guaranteed to exist if the
+    // account owns any NFTs
+    auto const last = ripple::keylet::nftpage_max(accountID);
+    bool hasMarker = !pageMarker.isZero();
```
scottschurr: Could be `bool const hasMarker`?

scottschurr (approved):
As far as I can see, by code inspection, this iteration code is correct.  The order in which NFT IDs are returned is a bit odd, but the iteration should return all NFTs.

The order is odd because we're running the pages backwards but running the page contents forward.  So there's kind of a saw-tooth jump at every page boundary.

You _could_ iterate forward, but there might be a high compute cost associated with returning the first value.  You could start out by walking the chain of  `Previous` links to the first page.  Once you've paid that price the passed in marker will keep you from needing to do that walk again.  However I'm wincing as I make the suggestion because the iteration to get to the first page could get long.  So I'm not saying that's the right choice.  But it is an option.

ledhed2222:
@scottschurr I mean, doing what you suggest still makes it a linear algorithm so it's not terrible. Originally, CJ had this backwards by page and backwards by token, but I asked him to change the latter to forwards for code readability. Do you think a consistently backwards approach is better than this one?

src/rpc/handlers/AccountObjects.cpp
undertome: Those methods aren't defined for `STArray`

src/rpc/handlers/AccountObjects.cpp
cjcobb23: done

src/rpc/handlers/AccountObjects.cpp
cjcobb23: this was optional before

src/rpc/handlers/AccountObjects.cpp
cjcobb23: Because I changed `while(true)` to `while(cp)`

natenichols (approved):

PR #182, opened 2022-06-17
append warning to the base level of the response

natenichols:


cjcobb23 (approved):
LGTM

cjcobb23 (changes requested):
We need to make this change for http responses as well. It's not a crasher, but it will be inconsistent. Warnings are not supposed to be nested inside `result`

cjcobb23 (approved):
LGTM

PR #184, opened 2022-06-17
send id in sendError

natenichols:


src/webserver/WsBase.h
```diff
@@ -315,15 +324,15 @@ class WsSession : public WsBase,
             catch (Backend::DatabaseTimeout const& t)
             {
                 BOOST_LOG_TRIVIAL(error) << __func__ << " Database timeout";
-                return sendError(RPC::Error::rpcNOT_READY);
+                return sendError(RPC::Error::rpcNOT_READY, nullptr);
```
cjcobb23: why are not sending the id here?

src/webserver/WsBase.h
```diff
@@ -315,15 +324,15 @@ class WsSession : public WsBase,
...
             }
         }
         catch (std::exception const& e)
         {
             BOOST_LOG_TRIVIAL(error)
                 << __func__ << " caught exception : " << e.what();
 
-            return sendError(RPC::Error::rpcINTERNAL);
+            return sendError(RPC::Error::rpcINTERNAL, nullptr);
```
cjcobb23: why are we not sending the id here?

natenichols: There may or may not an `id` in the request here. If someone sends garbage, and it causes an internal error, we may or may not have an `id`, so we just return a null id.

cjcobb23 (approved):

PR #186, opened 2022-06-17
always append clio warning

cjcobb23:
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/186)

### DIFF
--- a/src/webserver/HttpBase.h
+++ b/src/webserver/HttpBase.h
@@ -402,7 +402,6 @@ handle_request(
 
             result = error;
 
-            responseStr = boost::json::serialize(response);
             BOOST_LOG_TRIVIAL(debug)
                 << __func__ << " Encountered error: " << responseStr;
         }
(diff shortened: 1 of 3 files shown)


src/webserver/WsBase.h
```diff
@@ -322,26 +322,24 @@ class WsSession : public WsBase,
...
+
+        auto& result = response["result"].as_object();
+        result["warnings"] = warnings;
+        std::string responseStr = boost::json::serialize(response);
+        if (!dosGuard_.add(*ip, responseStr.size()))
         {
-            auto& result = response["result"].as_object();
-            result["warning"] = warnings;
+            warnings.emplace_back("Too many requests");
+            result["warnings"] = warnings;
```
natenichols: This will only actually add warnings to result if `!dosGuard.add()`

I would restructure to something like 

```cpp
result["warnings"] = boost::json::array{};
boost::json::array& warnings = result["warnings"].as_array();
```

src/webserver/HttpBase.h
```diff
@@ -401,27 +400,23 @@ handle_request(
...
-        }
-        // reserialize only if a warning was appended.
-        if (warningFlag)
+        result["warnings"] = warnings;
+        responseStr = boost::json::serialize(response);
+        if (!dosGuard.add(ip, responseStr.size()))
         {
-            result["warning"] = warnings;
+            warnings.emplace_back("Too many requests");
+            result["warnings"] = warnings;
```
natenichols: Same here, will warnings only be added when there are "Too many requests"?

natenichols (approved):

natenichols (changes requested):
My comment was meant to be "Request Changes" lol

src/webserver/WsBase.h
cjcobb23: Aren't they added on line 336?

src/webserver/HttpBase.h
cjcobb23: Aren't they added on line 413?

src/webserver/WsBase.h
ethanlabelle: does ```result["warnings"] = warnings;``` cause a copy of the warnings array? I am now starting to think that it may be better to modify a reference to the response json directly.

ethanlabelle (commented):
If possible, I think it would be best to directly modify the response json. If the warnings array gets copied during every response, this adds a small constant overhead.

src/webserver/WsBase.h
natenichols: I'm concerned about the same thing @ethanlabelle 

@cjcobb23 have you tested this?

natenichols: So I'm actually pretty sure this won't get copied, since `boost::json::object::operator[]` returns a `value&`. 

https://www.boost.org/doc/libs/1_75_0/libs/json/doc/html/json/ref/boost__json__object/operator_lb__rb_.html

natenichols: But it also depends on how the internals of `operator[]` handle this operation.

natenichols: I tested this locally, and it works as expected.

cjcobb23: I tested it, everything seems to check out

natenichols (approved):
As usual, I was wrong and this works totally fine :)

PR #187, opened 2022-06-17
concurrent request limit for http and WebSocket requests

ethanlabelle:
Testing strategy:

DOSGuard config:
```
"dos_guard":
    {
        "whitelist":[],
        "max_fetches" : -1,
        "max_concurrent_requests" : 4
    }
```
Http requests via bash script:
```
#!/bin/bash

for ((i=0; i < $1;i++))
do
    (curl localhost:51233 -d '{"method": "ledger_data"}' | jq )&
done

wait
```
WebSocket requests using javascript and Node.js:

```
const {WebSocket} = require('ws')
const ws = new WebSocket("ws://localhost:51233")
const requests = 1000
let counter = 0

ws.on('open', function open() {
    for (let i = 0; i < requests; i = i + 1)
        ws.send("{\"method\": \"ledger_data\"}")
})

ws.on('message', function message(data) {
    console.log("%s", data)
    counter = counter +  1
    if (counter == requests)
        ws.close()
})


ws.on('close', function close() {
    process.exit(0)
})
```
When I am on the whitelist, all of these requests are served with ledger data. When I am not on the whitelist, clio responds to almost all of the requests with rpc slow down.

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -218,8 +218,8 @@ public:
     {
         if (whitelist_.contains(ip))
             return true;
-        std::shared_lock lk(ftMtx_);
-        std::shared_lock lk(reqMtx_);
+        std::shared_lock flk(ftMtx_);
+        std::shared_lock rlk(reqMtx_);
         bool fetchesOk = maxFetches_ == -1 || getFetches(ip) < maxFetches_;
         bool requestsOk = maxConcurrentRequests_ == -1 ||
             getRequests(ip) < maxConcurrentRequests_;
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -2,6 +2,7 @@
 #define RIPPLE_REPORTING_DOS_GUARD_H
 
 #include <boost/asio.hpp>
+#include <atomic>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
@@ -9,11 +10,37 @@
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
-    std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::shared_mutex ftMtx_;   // protects ipFetchCount_
+    std::shared_mutex reqMtx_;  // protects ipRequestCount_
+    std::unordered_map<std::string, std::atomic_uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::atomic_uint32_t> ipRequestCount_;
     std::unordered_set<std::string> const whitelist_;
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConcurrentRequests_;
+
+    struct RPCScope
+    {
+        std::uint32_t count_;
+        std::string ip_;
+        DOSGuard& parent_;
+
+        RPCScope(std::string ip, DOSGuard& parent) : ip_(ip), parent_(parent)
+        {
+            count_ = parent_.incrementRequestCount(ip_);
+        }
+
+        ~RPCScope()
+        {
+            parent_.decrementRequestCount(ip_);
+        }
+
+        bool
+        isValid()
+        {
+            return parent_.isOk(ip_, count_);
+        }
+    };
 
     // Load config setting for DOSGuard
     std::optional<boost::json::object>
@@ -69,12 +96,99 @@ class DOSGuard
         }
     }
 
+    std::uint32_t
+    getFetches(std::string const& ip)
+    {
+        std::shared_lock lk(ftMtx_);
+        auto it = ipFetchCount_.find(ip);
+        if (it == ipFetchCount_.end())
+            return 0;
+        else
+            return it->second;
+    }
+
+    std::uint32_t
+    getRequests(std::string const& ip)
+    {
+        std::shared_lock lk(reqMtx_);
+        auto it = ipRequestCount_.find(ip);
+        if (it == ipRequestCount_.end())
+            return 0;
+        else
+            return it->second;
+    }
+
+    // increment atomic request counter for ip in a threadsafe manner.
+    // returns result (not a reference)
+    std::uint32_t
+    incrementRequestCount(std::string const& ip)
+    {
+        // reading ipReqiestCount_[ip] can potentially mutate the internal
+        // ipRequestCount_ data structure. One of two things will happen:
+        // 1) ipRequestCount_[ip] contains a reference to an atomic
+        // 2) ipRequestCount_[ip] creates a new reference
+        //      -this mutates the data structure
+
+        {
+            std::shared_lock read_lock(reqMtx_);
+            auto it = ipRequestCount_.find(ip);
+            if (it != ipRequestCount_.end())
+                return ipRequestCount_[ip]++;
+        }
+        {
+            // at this point the ip does not exist in the map. Must grab
+            // writer lock
+            std::unique_lock write_lock(reqMtx_);
+            // this will create and increment the atomic OR
+            // increment the atomic if another thread succeeded for this ip.
+            return ipRequestCount_[ip]++;
+        }
+    }
+
+    void
+    decrementRequestCount(std::string const& ip)
+    {
+        // assuming that ip is in the map
+        {
+            std::shared_lock read_lock(reqMtx_);
+            auto it = ipRequestCount_.find(ip);
+            if (it == ipRequestCount_.end())
+                assert(false);
+            if (--ipRequestCount_[ip])
+                return;
+        }
+        // erase if count is still zero
+        {
+            std::unique_lock write_lock(reqMtx_);
+            auto it = ipRequestCount_.find(ip);
+            if (it == ipRequestCount_.end())
+                return;  // another thread erased the reference
+
+            if (!it->second)  // if still zero
+                ipRequestCount_.erase(ip);
+        }
+        return;
+    }
+
+    bool
+    isOk(std::string const& ip, std::uint32_t requestCount)
+    {
+        if (whitelist_.contains(ip))
+            return true;
+        std::shared_lock lk(ftMtx_);
+        bool fetchesOk = maxFetches_ == -1 || getFetches(ip) < maxFetches_;
+        bool requestsOk = maxConcurrentRequests_ == -1 ||
+            requestCount < maxConcurrentRequests_;
+        return fetchesOk && requestsOk;
+    }
+
 public:
     DOSGuard(boost::json::object const& config, boost::asio::io_context& ctx)
         : ctx_(ctx)
         , whitelist_(getWhitelist(config))
         , maxFetches_(get(config, "max_fetches", 100))
         , sweepInterval_(get(config, "sweep_interval", 1))
+        , maxConcurrentRequests_(get(config, "max_concurrent_requests", 4))
     {
         createTimer();
     }
@@ -104,37 +218,41 @@ public:
     {
         if (whitelist_.contains(ip))
             return true;
-
-        std::unique_lock lck(mtx_);
-        auto it = ipFetchCount_.find(ip);
-        if (it == ipFetchCount_.end())
-            return true;
-
-        return it->second < maxFetches_;
+        std::shared_lock lk(ftMtx_);
+        std::shared_lock lk(reqMtx_);
+        bool fetchesOk = maxFetches_ == -1 || getFetches(ip) < maxFetches_;
+        bool requestsOk = maxConcurrentRequests_ == -1 ||
+            getRequests(ip) < maxConcurrentRequests_;
+        return fetchesOk && requestsOk;
     }
 
+    RPCScope
+    checkout(std::string const& ip)
+    {
+        return RPCScope(ip, *this);
+    }
+
+    // add numObjects bytes to the fetch count for ip
     bool
     add(std::string const& ip, uint32_t numObjects)
     {
-        if (whitelist_.contains(ip))
+        if (whitelist_.contains(ip) || maxFetches_ == -1)
             return true;
-
         {
-            std::unique_lock lck(mtx_);
+            std::unique_lock lck(ftMtx_);
             auto it = ipFetchCount_.find(ip);
             if (it == ipFetchCount_.end())
                 ipFetchCount_[ip] = numObjects;
             else
                 it->second += numObjects;
+            return ipFetchCount_[ip] < maxFetches_;
         }
-
-        return isOk(ip);
     }
 
     void
     clear()
     {
-        std::unique_lock lck(mtx_);
+        std::unique_lock lck(ftMtx_);
         ipFetchCount_.clear();
     }
 };
--- a/src/webserver/HttpBase.h
+++ b/src/webserver/HttpBase.h
@@ -318,11 +318,18 @@ handle_request(
         return res;
     };
 
+    // The DOS guard allocates tickets to concurrent requests.
+    // The resource is automatically checked in when ticket goes out of scope.
+    auto ticket = dosGuard.checkout(ip);
+    if (!ticket.isValid())
+        return send(httpResponse(
+            http::status::ok,
+            "application/json",
+            boost::json::serialize(RPC::make_error(RPC::Error::rpcSLOW_DOWN))));
+
     if (req.method() == http::verb::get && req.body() == "")
-    {
-        send(httpResponse(http::status::ok, "text/html", defaultResponse));
-        return;
-    }
+        return send(
+            httpResponse(http::status::ok, "text/html", defaultResponse));
 
     if (req.method() != http::verb::post)
         return send(httpResponse(
--- a/src/webserver/WsBase.h
+++ b/src/webserver/WsBase.h
@@ -263,6 +263,13 @@ public:
             send(boost::json::serialize(e));
         };
 
+        // The DOS guard allocates tickets to concurrent requests.
+        // The resource is automatically checked in when ticket goes out of
+        // scope.
+        auto ticket = dosGuard_.checkout(*ip);
+        if (!ticket.isValid())
+            return sendError(RPC::Error::rpcSLOW_DOWN, id);
+            
         try
         {
             BOOST_LOG_TRIVIAL(debug) << " received request : " << request;
--- a/src/webserver/WsBase.h
+++ b/src/webserver/WsBase.h
@@ -268,8 +268,8 @@ public:
         // scope.
         auto ticket = dosGuard_.checkout(*ip);
         if (!ticket.isValid())
-            return sendError(RPC::Error::rpcSLOW_DOWN, id);
-            
+            return sendError(RPC::Error::rpcSLOW_DOWN);
+
         try
         {
             BOOST_LOG_TRIVIAL(debug) << " received request : " << request;


src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,47 @@
...
+            }
+        }
+
+        bool
+        isValid()
+        {
+            return isValid_;
+        }
+
+        ticket(ticket const& other) = delete;
```
natenichols: Lets also delete copy assignment 
```cpp
ticket&
operator=(ticket const& ticket) = delete;
```

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,47 @@
...
-    std::mutex mtx_;  // protects ipFetchCount_
+    std::recursive_mutex mtx_;  // protects ipFetchCount_, ipRequestCount_
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::uint32_t> ipRequestCount_;
     std::unordered_set<std::string> const whitelist_;
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConcurrentRequests_;
+
+    struct ticket
```
natenichols: Can we name this something like `RPCScope` or something a little more descriptive.

If you like ticket, lets name this `Ticket`, class names are generally capitalized.

src/webserver/WsBase.h
```diff
@@ -255,6 +255,11 @@ class WsSession : public WsBase,
         auto sendError = [this](auto error) {
             send(boost::json::serialize(RPC::make_error(error)));
         };
+
+        auto ticket = dosGuard_.checkout(*ip);
```
natenichols, line 269: This is much cleaner than trying to close a session every time you exit scope. Nice work.

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,47 @@
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
+    std::recursive_mutex mtx_;  // protects ipFetchCount_, ipRequestCount_
```
natenichols: Would you make a note about why you chose `recursive_mutex`? I know we discussed that it allows people to directly call `getRequest` or `getFetches` outside of `checkout`, but just make a comment.

src/webserver/HttpBase.h
```diff
@@ -312,22 +312,21 @@ handle_request(
         return res;
     };
 
+    auto ticket = dosGuard.checkout(ip);
```
natenichols, line 323: Would you add a comment above `.checkout` that states that the resource is automatically checked in when `ticket` goes out of scope

natenichols (commented):
Did a quick pass before I hop off for the weekend. 

Overall this looks great. I think the RAII approach is a really solid way to do this.

cjcobb23:
I think we should get rid of the existing rate limiting mechanism and only use concurrent requests. It seems simpler and fairer.

ethanlabelle:
> I think we should get rid of the existing rate limiting mechanism and only use concurrent requests. It seems simpler and fairer.

I think both DOS prevention mechanisms are needed. The concurrent request limiting does not necessarily provide any rate limiting, as the total response bandwidth is dependent on the throughput of the server and the types of messages.

The concurrent request limit was primarily designed to prevent users from exceeding the existing rate limit. Users could send many requests at the same time, and they would all get served as the DOS Guard does book keeping after it has built the body of rpc responses.

cjcobb23:
> > I think we should get rid of the existing rate limiting mechanism and only use concurrent requests. It seems simpler and fairer.
> 
> I think both DOS prevention mechanisms are needed. The concurrent request limiting does not necessarily provide any rate limiting, as the total response bandwidth is dependent on the throughput of the server and the types of messages.
> 
> The concurrent request limit was primarily designed to prevent users from exceeding the existing rate limit. Users could send many requests at the same time, and they would all get served as the DOS Guard does book keeping after it has built the body of rpc responses.

Let's talk about this in the huddle. I would generally like to minimize rate limiting logic in clio and push that to an outer layer, and really like the simplicity of concurrent request limiting.

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,48 @@
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
+    std::recursive_mutex
```
natenichols: Not in the scope for this PR, but we should consider how to do more granular locking for the DOSGuard.

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,48 @@
...
+            if (isValid_)
+            {
+                std::unique_lock lk(mtx_);
+                count_++;
+            }
+        }
+
+        ~RPCScope()
+        {
+            if (isValid_)
```
natenichols: Should we count an RPC towards the `count_` even if it is invalid? 

Up to you, functionally I don't think it changes a lot.

natenichols (approved):
Lets discuss our overarching rate limiting goals in the Clio huddle before we merge this. I agree w/ Ethan that we likely should rate limit on both concurrent limits and bytes / second.

src/webserver/DOSGuard.h
```diff
@@ -103,15 +163,26 @@ class DOSGuard
     {
         if (whitelist_.contains(ip))
             return true;
-
         std::unique_lock lck(mtx_);
-        auto it = ipFetchCount_.find(ip);
-        if (it == ipFetchCount_.end())
-            return true;
+        bool fetchesOk =
+            (maxFetches_ == -1) ? true : getFetches(ip) < maxFetches_;
```
cjcobb23: I think `maxFetches_ == 1 || getFetches(ip) < maxFetches_` is cleaner, but up to you

src/webserver/DOSGuard.h
```diff
@@ -103,15 +163,26 @@ class DOSGuard
...
+        bool requestsOk = (maxConcurrentRequests_ == -1)
+            ? true
+            : getRequests(ip) < maxConcurrentRequests_;
```
cjcobb23: `maxConcurrentRequests_ == -1 || getRequests(ip) < maxConcurrentRequests_`

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,48 @@
...
+    std::unordered_map<std::string, std::uint32_t> ipRequestCount_;
     std::unordered_set<std::string> const whitelist_;
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConcurrentRequests_;
+
+    struct RPCScope
+    {
+        std::uint32_t& count_;
+        std::recursive_mutex& mtx_;
```
cjcobb23: Why does this need to be recursive?

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,48 @@
...
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::uint32_t> ipRequestCount_;
     std::unordered_set<std::string> const whitelist_;
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConcurrentRequests_;
+
+    struct RPCScope
+    {
+        std::uint32_t& count_;
```
cjcobb23: we can probably make this atomic and remove the need for the mutex

src/webserver/DOSGuard.h
cjcobb23: Actually, let's try to do that here. I think we can switch the values of the map to atomic, and then switch this to a shared_mutex. Then, we only need a unique lock when we are adding to one of the maps. @ethanlabelle

src/webserver/DOSGuard.h
```diff
@@ -9,11 +9,48 @@
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
+    std::recursive_mutex
+        mtx_;  // protects ipFetchCount_, ipRequestCount_
+               // recursive mutex allows reentrant mutex acquisitions
+               // this makes isOk(ip) and checkout(ip) atomic
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::uint32_t> ipRequestCount_;
```
cjcobb23: This map needs to be cleared at some point. Might be a bit tricky, since `RPCScope` contains a reference a value in the map, and thus we need to be careful not to invalidate the reference. Maybe use `shared_ptrs`?

cjcobb23 (changes requested):
Let's make the locking in this class more granular while we are here. Also, we need to clear `ipRequestCount_` at some point.

src/webserver/DOSGuard.h
ethanlabelle: An RPCScope is constructed with a reference to the DOSGuard mutex. The choice of recursive mutex allows for reentrant mutex acquisitions, which allow sequences of function calls to be atomic. For example, isOk calls both getFetches and getRequests. These methods both acquire the mutex such to avoid data races. The problem is that if getFetches is called and then getRequests after, a different thread could perform some work in between the calls. The recursive mutex allows us to use a single lock instead of an ordered locking scheme.

cjcobb23: I'd rather avoid `recursive_mutex` as I feel it makes the concurrency pattern harder to reason about. Are `getFetches()` and `getRequests()` called anywhere besides `isOk()`?

ethanlabelle: I agree with you here. I do not think `getFetches()` and `getRequests()` are called outside of `isOk()`. Currently, I would like to use a `shared_mutex` with `shared_lock` acquisitions in all of these methods. This would make `isOk()` atomic.

src/webserver/DOSGuard.h
```diff
@@ -2,18 +2,45 @@
...
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConcurrentRequests_;
+
+    friend struct RPCScope;
+    struct RPCScope
+    {
+        std::uint32_t count_;
+        std::string ip_;
+        DOSGuard* parent_;
```
natenichols: Can this just be a `DOSGuard& parent_`?

src/webserver/DOSGuard.h
```diff
@@ -68,12 +95,101 @@ class DOSGuard
...
+    {
+        std::shared_lock lk(mtx_);
+        auto it = ipRequestCount_.find(ip);
+        if (it == ipRequestCount_.end())
+            return 0;
+        else
+            return it->second;
+    }
+
+    // increment atomic request counter for ip in a hopefully threadsafe manner.
```
natenichols: Probably remove `hopefully` from this comment :)

src/webserver/DOSGuard.h
```diff
@@ -68,12 +95,101 @@ class DOSGuard
...
+    decrementRequestCount(std::string const& ip)
+    {
+        // assuming here that ip is in the map
+        std::shared_lock read_lock(mtx_);
+        if (!--ipRequestCount_[ip])  // decrement and compare to zero
+        {
+            read_lock.unlock();
+            std::unique_lock write_lock(mtx_);
+            auto it = ipRequestCount_.find(ip);
+            if (it == ipRequestCount_.end())
```
natenichols, line 155: This should be an error, and indicative that something is wrong. If you go to decrement an atomic that's not actually there, this seems likely to be a race condition.

Can we add a 
```cpp
assert(false);
```

to this branch?

src/webserver/DOSGuard.h
```diff
@@ -68,12 +95,101 @@ class DOSGuard
...
+    {
+        // ipReqiestCount_[ip] can potentially mutate the internal
+        // ipRequestCount_ data structure one of two things will happen:
+        // 1) ipRequestCount_[ip] contains a reference to an atomic
+        // 2) ipRequestCount_[ip] creates a new reference
+        //      -this mutates the data structure
+
+        while (1)
+        {
+            std::shared_lock read_lock(mtx_);
```
natenichols: I would prefer this lock acquisition to be RAII
```cpp
    {
            std::shared_lock read_lock(mtx_);
            auto it = ipRequestCount_.find(ip);
            if (it != ipRequestCount_.end())
                return ipRequestCount_[ip]++;
    }
     
    // at this point the ip does not exist in the map. Must grab
    // writer lock, as this thread _may_ have to create the entry
    {
            std::unique_lock write_lock(mtx_);
            // before creating a new reference, check if another thread
            // succeeded first
            it = ipRequestCount_.find(ip);
            if (it == ipRequestCount_.end())
                // ip is still not in map. It is now safe to create atomic
                // ref
                return ipRequestCount_[ip]++;
     }
```

src/webserver/DOSGuard.h
```diff
@@ -68,12 +95,101 @@ class DOSGuard
...
+    std::uint32_t
+    incrementRequestCount(std::string const& ip)
+    {
+        // ipReqiestCount_[ip] can potentially mutate the internal
+        // ipRequestCount_ data structure one of two things will happen:
+        // 1) ipRequestCount_[ip] contains a reference to an atomic
+        // 2) ipRequestCount_[ip] creates a new reference
+        //      -this mutates the data structure
+
+        while (1)
```
natenichols: What purpose does this `while (1)` serve?

src/webserver/DOSGuard.h
```diff
@@ -68,12 +95,101 @@ class DOSGuard
...
+    }
+
+    void
+    decrementRequestCount(std::string const& ip)
+    {
+        // assuming here that ip is in the map
+        std::shared_lock read_lock(mtx_);
+        if (!--ipRequestCount_[ip])  // decrement and compare to zero
+        {
+            read_lock.unlock();
```
natenichols: same here

natenichols (changes requested):

src/webserver/DOSGuard.h
ethanlabelle: The idea is that if another thread created a reference counter between releasing the shared lock and acquiring the unique lock, then this attempt to increment the counter failed and so we will try again. 

I think though that once we have the unique lock it is safe to just increment whatever is there.

src/webserver/DOSGuard.h
ethanlabelle: I do not think this is indicative of an error. Another thread could remove the reference before this thread is able to.  For example, two requests A, B


```
t ->>>>
|-------A---------|
      |----B----|
```

A could decrement the request counter before B starts. B could start and finish before A removes the reference from the map. At this point B will see that the counter is 0 and will erase ip. Now, ip is no longer in the map when A goes to erase it.

natenichols (dismissed):
LGTM. 

@cjcobb23, I'm not super concerned about cleaning up the map every time we hit zero. If we're only doing 1000 RPC per second, I don't think we'll see a huge performance hit from taking a couple nanoseconds to clean up a small map. I think the simplicity of this approach is better than requiring users reason about shared_ptrs being used in asynchronous handlers. 

If you'd like, I can perf test this to double check.

src/webserver/DOSGuard.h
```diff
@@ -2,18 +2,44 @@
...
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
 
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
-    std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::shared_mutex mtx_;  // protects ipFetchCount_, ipRequestCount_
```
cjcobb23: Would it be better if we had two mutexes, one for each map?

src/webserver/DOSGuard.h
```diff
@@ -68,12 +94,99 @@ class DOSGuard
...
+        {
+            std::unique_lock write_lock(mtx_);
+            auto it = ipRequestCount_.find(ip);
+            if (it == ipRequestCount_.end())
+                return;  // another thread erased the reference
+
+            if (!it->second)  // if still zero
+                ipRequestCount_.erase(ip);
+        }
+        return;
```
cjcobb23: I think this would be cleaner if we just used a unique_lock. I think most clients actually send one request at a time, and thus we will need to grab the unique lock anyways. Then, this function would be much cleaner and easier to reason about, imo. Same goes for `incrementRequestCount()`

cjcobb23 (changes requested):
I was thinking about this, and I am wondering if it would be more efficient to not actually delete entries from the map when the count goes to zero, and instead periodically clear the entire map. `decrementRequestCount()` can just do nothing if the ip doesn't exist.

src/webserver/DOSGuard.h
ethanlabelle: This is more efficient though for the case where a malicious client is sending many many requests. Using a unique lock also seems to defeat the purpose of making the counters atomic.

ethanlabelle:
Is it okay to just clear the map at set intervals, similar to the byte based rate limiting? It seems that in this case, actors that have outstanding requests will be able to send more traffic when they should still be rate listed. An alternative is to have a thread that periodically clears ip's from the map that have 0 concurrent requests, but removing a (key, value) from the map requires a unique lock.

An idea could be to hold a shared lock and construct a list of ip's with zero outstanding requests. Then, we can acquire a unique lock and remove all of the ip's that are still zero.

I think the main pitfall of the existing approach is that counters are potentially created and destroyed during every request if they are executed serially.

cjcobb23:
> Is it okay to just clear the map at set intervals, similar to the byte based rate limiting? It seems that in this case, actors that have outstanding requests will be able to send more traffic when they should still be rate listed. An alternative is to have a thread that periodically clears ip's from the map that have 0 concurrent requests, but removing a (key, value) from the map requires a unique lock.

I think its ok to clear the map at set intervals. If the interval is wide enough, few actors will be able to send more than one extra outstanding request. For example, if we clear every 10 seconds, its extremely unlikely any request is going to run for more than 10 seconds, so any client will only be able to exceed the limit by 1.

ethanlabelle:
> > Is it okay to just clear the map at set intervals, similar to the byte based rate limiting? It seems that in this case, actors that have outstanding requests will be able to send more traffic when they should still be rate listed. An alternative is to have a thread that periodically clears ip's from the map that have 0 concurrent requests, but removing a (key, value) from the map requires a unique lock.
> 
> I think its ok to clear the map at set intervals. If the interval is wide enough, few actors will be able to send more than one extra outstanding request. For example, if we clear every 10 seconds, its extremely unlikely any request is going to run for more than 10 seconds, so any client will only be able to exceed the limit by 1.

I can implement this, but it does allow for clients to temporarily double the configured max number of concurrent requests. After the map is cleared, any actor that has outstanding requests will be allowed to submit more 'valid' requests. This only happens immediately after the map is cleared, so in aggregate clients will not be able to send *much* more than the configured limits- dependent on clear/limit parameters.

cjcobb23:
> I can implement this, but it does allow for clients to temporarily double the configured max number of concurrent requests. After the map is cleared, any actor that has outstanding requests will be allowed to submit more 'valid' requests. This only happens immediately after the map is cleared, so in aggregate clients will not be able to send _much_ more than the configured limits- dependent on clear/limit parameters.

I think that's ok. We don't need to be really strict about the rate limiting. Just enough to not let a single client knock over the whole server. Allowing a client to temporarily have double the max number of concurrent requests is totally fine, in my opinion.

PR #193, opened 2022-06-22
allow user to specify no peer in doAccountLines

natenichols:
the account_lines handler was misbehaving when no peer was specified, due to reuse of the `getAccount` helper. 

When `peer` was empty, `traverseOwnedNodes` would attempt to look for lines w/ only the default constructed account ID, filtering out every peer account.

I added a `getOptionalAccount` helper, which assigns `account` to a `std::nullopt` if `!object.contains(peer)`, and otherwise errors if `peer` is not string.

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -222,6 +222,32 @@ getAccount(
 }
 
 Status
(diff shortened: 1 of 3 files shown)


cjcobb23 (approved):
LGTM

PR #194, opened 2022-06-23
Use ledger close times for stale data warning

ethanlabelle:
Small change to correct the stale data warning feature. Now, request handlers look at close time of last published ledger instead of time the last ledger was published.

I used the same testing strategy as (#175)


cjcobb23 (approved):
LGTM

natenichols (approved):

PR #201, opened 2022-06-29
Add headers for building with gcc-12

legleux:
I needed to add this to build with gcc-12

### DIFF
--- a/src/main/impl/Build.cpp
+++ b/src/main/impl/Build.cpp
@@ -2,6 +2,8 @@
 #include <boost/preprocessor/stringize.hpp>
 #include <algorithm>
 #include <main/Build.h>
+#include <optional>
+#include <stdexcept>
 
 namespace Build {
 


natenichols (commented):
According to slack, you said we also had to add <optional> to WorkQueue? Do you still have to do that?

```
diff --git a/src/main/impl/Build.cpp b/src/main/impl/Build.cpp
index ebb9842..3c13a3f 100644
--- a/src/main/impl/Build.cpp
+++ b/src/main/impl/Build.cpp
@@ -2,6 +2,8 @@
 #include <boost/preprocessor/stringize.hpp>
 #include <algorithm>
 #include <main/Build.h>
+#include <optional>
+#include <stdexcept>
 
 namespace Build {
 
diff --git a/src/rpc/WorkQueue.h b/src/rpc/WorkQueue.h
index 236dd9a..4a29208 100644
--- a/src/rpc/WorkQueue.h
+++ b/src/rpc/WorkQueue.h
@@ -10,6 +10,7 @@
 #include <queue>
 #include <shared_mutex>
 #include <thread>
+#include <optional>
 
 class WorkQueue
 {
 ```

legleux:
> According to slack, you said we also had to add to WorkQueue? Do you still have to do that?
> 

yes, that's needed also. I screwed up after clang-format ing.
Hold please...

natenichols:
I lied, its already in WorkQueue.h https://github.com/XRPLF/clio/blob/develop/src/rpc/WorkQueue.h/#L10

natenichols (approved):
Can you sign your commit so we can merge this?

PR #208, opened 2022-07-11
moved warnings array to outermost level of response json

ethanlabelle:


src/webserver/HttpBase.h
```diff
@@ -430,7 +430,7 @@ handle_request(
         if (!dosGuard.add(ip, responseStr.size()))
         {
             warnings.emplace_back("Too many requests");
-            response["warnings"] = warnings;
+            result["warnings"] = warnings;
```
cjcobb23: Are warnings supposed to be inside the `result` object? I thought they were at the top level: https://xrpl.org/response-formatting.html

cjcobb23 (approved):
LGTM

natenichols (approved):

cjcobb23:
Do we need to make this change for websocket as well?

ethanlabelle:
The warnings array is in the correct spot in WsBase.h.

PR #214, opened 2022-07-11
Fixed 503 response code

Shoukozumi:
JSON RPC now responds back with the right status code and error message when user is getting rate limited

### DIFF
--- a/src/webserver/HttpBase.h
+++ b/src/webserver/HttpBase.h
@@ -242,12 +242,16 @@ public:
                 },
                 dosGuard_.isWhiteListed(*ip)))
         {
+            // Non-whitelist connection rejected due to full connection queue
             http::response<http::string_body> res{
                 http::status::ok, req_.version()};
-            res.set(http::field::server, "clio-server-v0.0.0");
+            res.set(
+                http::field::server,
+                "clio-server-" + Build::getClioVersionString());
             res.set(http::field::content_type, "application/json");
             res.keep_alive(req_.keep_alive());
-            res.body() = "Server overloaded";
+            res.body() = boost::json::serialize(
+                RPC::make_error(RPC::Error::rpcTOO_BUSY));
             res.prepare_payload();
             lambda_(std::move(res));
         }
@@ -304,7 +308,9 @@ handle_request(
                                   std::string content_type,
                                   std::string message) {
         http::response<http::string_body> res{status, req.version()};
-        res.set(http::field::server, "xrpl-reporting-server-v0.0.0");
+        res.set(
+            http::field::server,
+            "clio-server-" + Build::getClioVersionString());
         res.set(http::field::content_type, content_type);
         res.keep_alive(req.keep_alive());
         res.body() = std::string(message);
@@ -324,9 +330,9 @@ handle_request(
 
     if (!dosGuard.isOk(ip))
         return send(httpResponse(
-            http::status::ok,
-            "application/json",
-            boost::json::serialize(RPC::make_error(RPC::Error::rpcSLOW_DOWN))));
+            http::status::service_unavailable,
+            "text/plain",
+            "Server is overloaded"));
 
     try
     {
@@ -349,13 +355,6 @@ handle_request(
                 boost::json::serialize(
                     RPC::make_error(RPC::Error::rpcBAD_SYNTAX))));
         }
-
-        if (!dosGuard.isOk(ip))
-            return send(httpResponse(
-                http::status::ok,
-                "application/json",
-                boost::json::serialize(
-                    RPC::make_error(RPC::Error::rpcSLOW_DOWN))));
 
         auto range = backend->fetchLedgerRange();
         if (!range)
@@ -429,8 +428,7 @@ handle_request(
         responseStr = boost::json::serialize(response);
         if (!dosGuard.add(ip, responseStr.size()))
         {
-            warnings.emplace_back("Too many requests");
-            response["warnings"] = warnings;
+            response["warning"] = "load";
             // reserialize when we need to include this warning
             responseStr = boost::json::serialize(response);
         }
--- a/src/webserver/WsBase.h
+++ b/src/webserver/WsBase.h
@@ -245,39 +245,23 @@ public:
     }
 
     void
-    handle_request(std::string const&& msg, boost::asio::yield_context& yc)
+    handle_request(
+        boost::json::object const&& request,
+        boost::json::value const& id,
+        boost::asio::yield_context& yield)
     {
         auto ip = derived().ip();
         if (!ip)
             return;
 
         boost::json::object response = {};
-        auto sendError = [this](auto error, boost::json::value id) {
+        auto sendError = [this, &request, id](auto error) {
             auto e = RPC::make_error(error);
-
             if (!id.is_null())
                 e["id"] = id;
-
+            e["request"] = request;
             send(boost::json::serialize(e));
         };
-
-        boost::json::value raw = [](std::string const&& msg) {
-            try
-            {
-                return boost::json::parse(msg);
-            }
-            catch (std::exception&)
-            {
-                return boost::json::value{nullptr};
-            }
-        }(std::move(msg));
-
-        if (!raw.is_object())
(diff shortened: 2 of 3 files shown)


src/webserver/HttpBase.h
```diff
@@ -242,12 +242,16 @@ class HttpBase
                 },
                 dosGuard_.isWhiteListed(*ip)))
         {
+            // Connection getting rate limited due to full connection queue
             http::response<http::string_body> res{
-                http::status::ok, req_.version()};
-            res.set(http::field::server, "clio-server-v0.0.0");
+                http::status::service_unavailable, req_.version()};
```
cjcobb23: I don't think this one is supposed to be `service_unavailable`. This is not the same as rate limiting, this is just a normal error.

src/webserver/WsBase.h
```diff
@@ -358,7 +358,9 @@ class WsSession : public WsBase,
         std::string responseStr = boost::json::serialize(response);
         if (!dosGuard_.add(*ip, responseStr.size()))
         {
-            warnings.emplace_back("Too many requests");
+            ripple::RPC::ErrorInfo const& info(
+                ripple::RPC::get_error_info(RPC::Error::rpcSLOW_DOWN));
+            warnings.emplace_back(info.message);
```
cjcobb23: The warning here is just supposed to be the string `"load"`

cjcobb23 (changes requested):
How did you test this?

cjcobb23 (commented):
I was looking at the documentation again, and apparently `"load"` is supposed to be inside a field called `"warning"`, and the rest of the warnings go inside an array called `"warnings"`. Very silly, I know, but we don't have the freedom to change this right now. https://xrpl.org/response-formatting.html#response-formatting

src/webserver/WsBase.h
```diff
@@ -358,8 +341,7 @@ class WsSession : public WsBase,
         std::string responseStr = boost::json::serialize(response);
         if (!dosGuard_.add(*ip, responseStr.size()))
         {
-            warnings.emplace_back("Too many requests");
-            response["warnings"] = warnings;
+            response["warning"] = "load";
```
natenichols, line 345: What is this doing?

src/webserver/WsBase.h
```diff
@@ -245,48 +245,31 @@ class WsSession : public WsBase,
...
 
         boost::json::object response = {};
-        auto sendError = [this](auto error, boost::json::value id) {
+        auto sendError = [this, &request, id](auto error) {
             auto e = RPC::make_error(error);
-
-            if (!id.is_null())
-                e["id"] = id;
-
+            e["id"] = id;
```
natenichols: What happens when `id` is null?

natenichols: if `id` is left off the request, it should also be left off the response

src/webserver/WsBase.h
Shoukozumi: This is the API convention for rate-limiting response stated on the doc here: https://xrpl.org/response-formatting.html#response-formatting

src/webserver/WsBase.h
Shoukozumi: Fixed.

src/webserver/WsBase.h
```diff
@@ -245,48 +245,32 @@ class WsSession : public WsBase,
     }
 
     void
-    handle_request(std::string const&& msg, boost::asio::yield_context& yc)
+    handle_request(
+        boost::json::object&& request,
+        const boost::json::value& id,
+        boost::asio::yield_context& yc)
```
cjcobb23: Why did we need to make this change?

Shoukozumi: Since we have to parse the `msg` into request objects in `on_read` in order to extract the id that will be included in the error response, we can then just pass the already-parsed request object directly to `handle_request` so we don't need to reparse the `msg` there.

src/webserver/WsBase.h
```diff
@@ -383,29 +366,57 @@ class WsSession : public WsBase,
...
         if (!dosGuard_.isOk(*ip))
         {
-            sendError("Too many requests. Slow down");
+            sendError(RPC::Error::rpcSLOW_DOWN, id, request);
         }
         else
         {
             if (!queue_.postCoro(
-                    [m = std::move(msg), shared_this = shared_from_this()](
+                    [shared_this = shared_from_this(), &request, id](
```
cjcobb23: I think this is actually unsafe. We capture `request` by reference, but then post to the job queue, after which this function returns and `request` goes out of scope. What do we need to parse the request for here? To get the id? Which we only need if we fail the DosGuard check? I think we can move the DosGuard check inside `handle_request` and send the error from there, and not have to parse the request here.

cjcobb23 (changes requested):

src/webserver/WsBase.h
Shoukozumi: Sounds good, I'll fix it

natenichols: This should fix the unsafe behavior
```cpp
[shared_this = shared_from_this(), request = std::move(request), id];
```

natenichols: The unsafe behavior is we capture `request` as a reference (just a memory address of `request`), which then goes out of scope (postCoro may be congested and take a while to handle the request). 

During the time that the request waits in the queue, `request` goes out of scope, and that reference points to memory that is now no longer allocated.

src/webserver/WsBase.h
```diff
@@ -246,7 +246,7 @@ class WsSession : public WsBase,
 
     void
     handle_request(
-        boost::json::object&& request,
+        boost::json::object const&& request,
```
natenichols, line 249: I don't think we need `const` here.

cjcobb23 (approved):
LGTM

src/webserver/WsBase.h
```diff
@@ -383,29 +366,57 @@ class WsSession : public WsBase,
...
+            }
+            catch (std::exception&)
+            {
+                return boost::json::value{nullptr};
+            }
+        }(std::move(msg));
+
+        if (!raw.is_object())
+            return sendError(
+                RPC::Error::rpcINVALID_PARAMS, nullptr, raw.as_object());
```
natenichols: This will throw, as raw is not an object. Just return a message that says something like "body is not json object"

src/webserver/WsBase.h
```diff
@@ -245,48 +245,32 @@ class WsSession : public WsBase,
     }
 
     void
-    handle_request(std::string const&& msg, boost::asio::yield_context& yc)
+    handle_request(
+        boost::json::object const&& request,
+        const boost::json::value& id,
```
natenichols: our code style prefers 
```cpp
boost::json::value const& id
```

src/webserver/WsBase.h
```diff
@@ -245,48 +245,32 @@ class WsSession : public WsBase,
...
+        boost::asio::yield_context& yc)
```
natenichols: can we rename this to `yield` to be consistant

src/webserver/WsBase.h
```diff
@@ -383,29 +366,57 @@ class WsSession : public WsBase,
...
             << __func__ << " received request from ip = " << *ip;
-        auto sendError = [&](auto&& msg) {
-            boost::json::object response;
-            response["error"] = std::move(msg);
-            std::string responseStr = boost::json::serialize(response);
 
-            BOOST_LOG_TRIVIAL(trace) << __func__ << " : " << responseStr;
+        auto sendError = [this, ip](
+                             auto error,
+                             const boost::json::value& id,
```
natenichols: prefer `const&` to be consistant

src/webserver/WsBase.h
```diff
@@ -383,29 +366,57 @@ class WsSession : public WsBase,
...
+                             boost::json::object& request) {
```
natenichols: I prefer `const& request` here, since we're not modifying it

natenichols (changes requested):
The only thing that _really_ needs to be changed is `raw.as_object()` being called in `!raw.is_object()`.

Other stuff is optional, but not necessary.

natenichols (approved):

PR #220, opened 2022-07-20
added cache hit rate to server info

ethanlabelle:


src/backend/SimpleCache.h
```diff
@@ -62,6 +62,12 @@ class SimpleCache
 
     size_t
     size() const;
+    
+    mutable std::atomic_uint32_t requestCounter_;
+    mutable std::atomic_uint32_t hitCounter_;
```
cjcobb23: Feel like maybe we should separate successor hits from regular object hits

src/backend/BackendInterface.cpp
```diff
@@ -89,8 +91,12 @@ BackendInterface::fetchLedgerObjects(
     for (size_t i = 0; i < keys.size(); ++i)
     {
         auto obj = cache_.get(keys[i], sequence);
+        cache_.objectReqCounter_++;
```
natenichols: Can we move this increment into `cache_.get()`?

src/backend/BackendInterface.cpp
```diff
@@ -121,9 +127,13 @@ BackendInterface::fetchSuccessorKey(
     boost::asio::yield_context& yield) const
 {
     auto succ = cache_.getSuccessor(key, ledgerSequence);
+    cache_.successorReqCounter_++;
```
natenichols: Same here.

natenichols: Can we also move this increment into `cache_.getSuccessor()`. 

Otherwise we should wrap this in something like
```cpp
cache_.hitObject();
```

natenichols (approved):

PR #226, opened 2022-07-26
ensure lgrInfo is in context.range

natenichols:
This fixes a data race where we write the ledger `n`, but the `range.maxSequence` is still `n-1`. In this case we should treat this the same as not finding the ledger to prevent a data race.

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -569,7 +569,7 @@ ledgerInfoFromRequest(Context const& ctx)
 
         auto lgrInfo = ctx.backend->fetchLedgerByHash(ledgerHash, ctx.yield);
 
-        if (!lgrInfo)
+        if (!lgrInfo || lgrInfo->seq > ctx.range.maxSequence)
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -569,7 +569,7 @@ ledgerInfoFromRequest(Context const& ctx)
 
         auto lgrInfo = ctx.backend->fetchLedgerByHash(ledgerHash, ctx.yield);
 
-        if (!lgrInfo)
+        if (!lgrInfo || lgrInfo->seq > ctx.range.maxSequence)
```
cjcobb23, line 572: Do we not perform this check earlier? If we don't, I think we need to. What if users request data from a specific ledger, but we don't actually fetch the ledger header?

natenichols: I think we always fetch the ledger header to get the ledger sequence

natenichols: `ledgerInfoFromRequest` is called in every RPC. This seems like the correct place to make this check

cjcobb23 (approved):

PR #229, opened 2022-07-28
Fixed warning message to be XRPL standard compliant

Shoukozumi:
All warnings now contain `warning objects` with the right format, which includes `id`, `message`, and `details`(optional) as fields

### DIFF
--- a/src/rpc/RPC.cpp
+++ b/src/rpc/RPC.cpp
@@ -91,6 +91,38 @@ make_HttpContext(
         clientIp};
 }
 
+constexpr static WarningInfo warningInfos[]{
+    {warnUNKNOWN, "Unknown warning"},
+    {warnRPC_CLIO,
+     "This is a clio server. clio only serves validated data. If you "
+     "want to talk to rippled, include 'ledger_index':'current' in your "
+     "request"},
+    {warnRPC_OUTDATED, "This server may be out of date"},
+    {warnRPC_RATE_LIMIT, "You are about to be rate limited"}};
+
+WarningInfo const&
+get_warning_info(warning_code code)
+{
+    for (WarningInfo const& info : warningInfos)
+    {
+        if (info.code == code)
+        {
+            return info;
+        }
+    }
+    throw(std::out_of_range("Invalid warning_code"));
+}
+
+boost::json::object
+make_warning(warning_code code)
+{
+    boost::json::object json;
+    WarningInfo const& info(get_warning_info(code));
+    json["id"] = code;
+    json["message"] = static_cast<std::string>(info.message);
+    return json;
+}
+
 boost::json::object
 make_error(Error err)
 {
--- a/src/rpc/RPC.h
+++ b/src/rpc/RPC.h
@@ -162,6 +162,33 @@ public:
     }
 };
 
+enum warning_code {
+    warnUNKNOWN = -1,
+    warnRPC_CLIO = 2001,
+    warnRPC_OUTDATED = 2002,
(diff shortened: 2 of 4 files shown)


src/rpc/RPC.h
```diff
@@ -162,6 +162,32 @@ class AccountNotFoundError : public std::exception
...
+    constexpr WarningInfo() : code(warnUNKNOWN), message("unknown warning")
+    {
+    }
+
+    constexpr WarningInfo(warning_code code_, char const* message_)
+        : code(code_), message(message_)
+    {
+    }
+    warning_code code;
+    Json::StaticString message;
```
natenichols: This can just be a string I think. We don't want to introduce new instances of ripple::Json

src/rpc/RPC.cpp
```diff
@@ -91,6 +91,37 @@ make_HttpContext(
         clientIp};
 }
 
+constexpr static WarningInfo warningInfos[]{
```
ethanlabelle, line 94: Does this need to be constexpr?

Shoukozumi: It doesn't for how it is used now, but the warning info is predetermined and therefore can be constexpr

src/rpc/RPC.cpp
```diff
@@ -91,6 +91,37 @@ make_HttpContext(
...
+    {warnUNKNOWN, "Unknown warning"},
+    {warnRPC_CLIO,
+     "This is a clio server. clio only serves validated data. If you "
+     "want to talk to rippled, include 'ledger_index':'current' in your "
+     "request"},
+    {warnRPC_OUTDATED, "This server may be out of date"}};
```
cjcobb23: While we are here, can we add a proper warning for when clients are about to be rate limited? I know we have the string `load` in a separate field, and we can leave that, but I think it would be easier for clients to be able to parse all warnings in one place.

Shoukozumi: Sounds good, I'll add that

src/rpc/RPC.cpp
```diff
@@ -97,7 +97,8 @@ constexpr static WarningInfo warningInfos[]{
      "This is a clio server. clio only serves validated data. If you "
      "want to talk to rippled, include 'ledger_index':'current' in your "
      "request"},
-    {warnRPC_OUTDATED, "This server may be out of date"}};
+    {warnRPC_OUTDATED, "This server may be out of date"},
+    {warnRPC_RATE_LIMIT, "Your connection is approaching rate limit!"}};
```
cjcobb23: I vote for this message being "You are placing too much load on the server". @natenichols what do you think?

natenichols: I like something like
```
"You are being rate limited"
```

natenichols: But I dont care about what we end up picking

Shoukozumi: Keep in mind that this is a warning sent before the client is actually being rate limited

cjcobb23: How about then `"You are about to be rate limited"`

Shoukozumi: Sounds good, I like it

natenichols (approved):

src/rpc/RPC.h
Shoukozumi: Fixed

cjcobb23 (approved):
LGTM

ethanlabelle (approved):

PR #232, opened 2022-08-04
Added a cache for TransactionAndMetadata objects

ethanlabelle:
This PR adds a cache for TransactionAndMetadata objects. The backend of clio maintains a cache of the `NUM_CACHED_LEDGERS` most recently published ledgers' transactions. I am posting this now mainly for visibility.

### DIFF
--- a/src/backend/BackendInterface.cpp
+++ b/src/backend/BackendInterface.cpp
@@ -327,4 +327,90 @@ BackendInterface::fetchFees(
     return fees;
 }
 
+std::optional<TransactionAndMetadata>
+BackendInterface::fetchTransaction(
+    ripple::uint256 const& hash,
+    boost::asio::yield_context& yield) const
+{
+    auto obj = txCache_.get(hash);
+    if (obj)
+    {
+        BOOST_LOG_TRIVIAL(trace)
+            << __func__ << " - tx_cache hit - " << ripple::strHex(hash);
+        return *obj;
+    }
+    else
+    {
+        BOOST_LOG_TRIVIAL(trace)
+            << __func__ << " - tx_cache miss - " << ripple::strHex(hash);
+        auto dbObj = doFetchTransaction(hash, yield);
+        if (!dbObj)
+            BOOST_LOG_TRIVIAL(trace)
+                << __func__ << " - missed tx_cache and missed in db";
+        else
+            BOOST_LOG_TRIVIAL(trace)
+                << __func__ << " - missed tx_cache but found in db";
+        return dbObj;
+    }
+}
+
+std::vector<TransactionAndMetadata>
+BackendInterface::fetchTransactions(
+    std::vector<ripple::uint256> const& hashes,
+    boost::asio::yield_context& yield) const
+{
+    std::vector<TransactionAndMetadata> results;
+    results.resize(hashes.size());
+    std::vector<ripple::uint256> misses;
+    for (size_t i = 0; i < hashes.size(); ++i)
+    {
+        auto obj = txCache_.get(hashes[i]);
+        if (obj)
+            results[i] = *obj;
+        else
+            misses.push_back(hashes[i]);
+    }
+    BOOST_LOG_TRIVIAL(trace)
+        << __func__ << " - tx_cache hits = " << hashes.size() - misses.size()
+        << " - tx_cache misses = " << misses.size();
+
+    if (misses.size())
+    {
+        auto objs = doFetchTransactions(misses, yield);
+        for (size_t i = 0, j = 0; i < results.size(); ++i)
+        {
+            if (results[i].transaction.size() == 0)
+            {
+                results[i] = objs[j];
+                ++j;
+            }
+        }
+    }
+
+    return results;
+}
+
+std::vector<TransactionAndMetadata>
+BackendInterface::fetchAllTransactionsInLedger(
+    std::uint32_t const ledgerSequence,
+    boost::asio::yield_context& yield) const
+{
+    auto obj = txCache_.getLedgerTransactions(ledgerSequence);
+    if (obj)
+        return *obj;
+    else
+        return doFetchAllTransactionsInLedger(ledgerSequence, yield);
+}
+
+std::vector<ripple::uint256>
+BackendInterface::fetchAllTransactionHashesInLedger(
+    std::uint32_t const ledgerSequence,
+    boost::asio::yield_context& yield) const
+{
+    auto obj = txCache_.getLedgerTransactionHashes(ledgerSequence);
+    if (obj)
+        return *obj;
+    else
+        return doFetchAllTransactionHashesInLedger(ledgerSequence, yield);
+}
 }  // namespace Backend
--- a/src/backend/TxCache.cpp
+++ b/src/backend/TxCache.cpp
@@ -1,0 +1,105 @@
+#include <backend/TxCache.h>
+namespace Backend {
+
+uint32_t
+TxCache::latestLedgerSequence() const
+{
+    std::shared_lock lck{mtx_};
+    return latestSeq_;
+}
+
+// adds an entire ledger's worth of transactions to the cache. Evicts oldest
+// ledger.
+void
+TxCache::update(
+    std::vector<ripple::uint256> const& hashes,
+    std::vector<TransactionAndMetadata> const& transactions,
+    uint32_t seq)
+{
+    std::unique_lock lck{mtx_};
+    cache_[tail_].clear();
+
+    assert(seq == latestSeq_ + 1 || latestSeq_ == 0);
+    latestSeq_ = seq;
+    for (std::size_t i = 0; i < hashes.size(); i++)
+    {
+        cache_[tail_][hashes[i]] = transactions[i];
+    }
+    tail_ = (tail_ + 1) % NUM_LEDGERS_CACHED;
+}
+
+std::optional<TransactionAndMetadata>
+TxCache::get(ripple::uint256 const& key) const
+{
+    std::shared_lock lck{mtx_};
+    txReqCounter_++;
+
+    for (auto& map : cache_)
+    {
+        auto e = map.find(key);
+        if (e != map.end())
+        {
+            txHitCounter_++;
+            return {e->second};
+        }
+    }
+    return {};
+}
+
+float
+TxCache::getHitRate() const
+{
+    if (!txReqCounter_)
+        return 0;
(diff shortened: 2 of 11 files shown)


src/backend/TxCache.h
```diff
@@ -0,0 +1,56 @@
...
+#include <ripple/basics/base_uint.h>
+#include <ripple/basics/hardened_hash.h>
+#include <backend/Types.h>
+#include <map>
+#include <mutex>
+#include <shared_mutex>
+#include <utility>
+#include <vector>
+
+#define NUM_LEDGERS_CACHED 10
```
natenichols, line 13: Can we change this to be a config option in `config.cache`?

src/etl/ReportingETL.cpp
```diff
@@ -208,6 +217,20 @@ ReportingETL::publishLedger(ripple::LedgerInfo const& lgrInfo)
...
+    {
+        BOOST_LOG_TRIVIAL(debug) << __func__ << " - Updating cache";
+
+        std::vector<Backend::LedgerObject> diff =
+            Backend::synchronousAndRetryOnTimeout([&](auto yield) {
+                return backend_->fetchLedgerDiff(lgrInfo.seq, yield);
+            });
+
+        backend_->cache().update(diff, lgrInfo.seq);
+        backend_->txCache().update(hashes, transactions, lgrInfo.seq);
```
natenichols, line 230: If we're updating the `txCache_` here, we don't need to also update it in `insertTransactions`. 

`publishLedger` is called in both readonly and etl modes. whereas insertTransactions is only called in etl mode.

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::unique_lock lck{mtx_};
+    cache_[tail_].clear();
+
+    assert(seq == latestSeq_ + 1 || latestSeq_ == 0);
+    latestSeq_ = seq;
+    for (std::size_t i = 0; i < hashes.size(); i++)
+    {
+        cache_[tail_][hashes[i]] = transactions[i];
+    }
+    tail_ = (tail_ + 1) % NUM_LEDGERS_CACHED;
```
natenichols, line 28: > So I actually think this stores duplicates when not in readonly mode. 

> We call txCache().update() twice, first in `updateTransactions`, then again in `publishLedger()`, meaning we increment tail_ twice for each ledger, storing a duplicate. I don't think this affects correctness, It just means it only stores `NUM_LEDGERS_CACHED / 2` ledgers worth of transactions in the cache.

> We can fix this by only updating the cache when publishing a ledger, see comments below. 
>
>I haven't run the code yet, so try to observe this behavior by adding some `trace` log statements before going to fix it.

This is an incorrect comment disregard

natenichols (changes requested):
First pass, and I have some questions about when we update `txCache_`.

src/etl/ReportingETL.cpp
ethanlabelle: The condition to enter this block is `if (!writing_)`

The intention is that for writing nodes the cache is updated in insertTransactions, while for reading nodes it is updated in publish ledger.

natenichols: Will `backend_->cache()` still get updated in etl mode with this change?

natenichols: @cjcobb23 I think that would slow down the extraction of successors a lot, right?

cjcobb23: `backend_->cache()` was already being updated under this conditional, if you look a little further up in the function. for writing nodes, its updated as part of building the ledger. Just like the tx cache is being updated as part of building the ledger (in `insertTransactions`).

src/etl/ReportingETL.cpp
```diff
@@ -60,13 +61,29 @@ ReportingETL::insertTransactions(
...
+        std::copy(raw->begin(), raw->end(), std::back_inserter(txBlob));
+        Backend::Blob mdBlob;
+        auto mdStr = txn.mutable_metadata_blob();
+        std::copy(mdStr->begin(), mdStr->end(), std::back_inserter(mdBlob));
+        transactions.push_back(
+            {txBlob,
+             mdBlob,
+             ledger.seq,
+             ledger.closeTime.time_since_epoch().count()});
     }
```
cjcobb23, line 85: This is a use after move bug, since we move from `raw` and `txn.mutable_metadata_blob()` a few lines above this, when we call `writeTransaction`. We need to make this copy before we call `writeTransaction`.

Also, I think this can be done in one line, like so (didn't test this):
```
transactions.push_back({
    {raw->data(),raw->size()},
    {txn.mutable_metadata_blob()->data(),txn.mutable_metadata_blob()->size()},
    ledger.seq,
    ledger.closeTime.time_since_epoch().count()}});
```

src/etl/ReportingETL.cpp
```diff
@@ -60,13 +61,29 @@ ReportingETL::insertTransactions(
         result.accountTxData.emplace_back(
             txMeta, sttx.getTransactionID(), journal);
         std::string keyStr{(const char*)sttx.getTransactionID().data(), 32};
+
+        ripple::uint256 hash{keyStr};
+        hashes.push_back(hash);
```
cjcobb23, line 66: I think you can do this in one line. Either:
```
hashes.emplace_back(keyStr);
```
or
```
hashes.push_back({keyStr});
```

src/etl/ReportingETL.cpp
```diff
@@ -208,6 +217,20 @@ ReportingETL::publishLedger(ripple::LedgerInfo const& lgrInfo)
...
+
+        std::vector<Backend::LedgerObject> diff =
+            Backend::synchronousAndRetryOnTimeout([&](auto yield) {
+                return backend_->fetchLedgerDiff(lgrInfo.seq, yield);
+            });
+
+        backend_->cache().update(diff, lgrInfo.seq);
+        backend_->txCache().update(hashes, transactions, lgrInfo.seq);
+        backend_->updateRange(lgrInfo.seq);
+    }
```
cjcobb23, line 232: Let's do all this before we publish the new ledger and transactions, which is when we call `subscriptions_->pubLedger()`

src/backend/TxCache.h
```diff
@@ -0,0 +1,56 @@
...
+    // basically a ring buffer
+    // array of maps of transaction hashes -> TransactionAndMetadata objects
+    // replace the oldest ledger in cache when inserting a new ledger
+    std::array<
+        std::map<ripple::uint256, TransactionAndMetadata>,
+        NUM_LEDGERS_CACHED>
+        cache_;
+    mutable std::shared_mutex mtx_;
+    uint32_t latestSeq_ = 0;
+    std::atomic_int tail_;
```
cjcobb23, line 27: I think this needs to be initialized

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+    std::shared_lock lck{mtx_};
+    return latestSeq_;
+}
+
+// adds an entire ledger's worth of transactions to the cache. Evicts oldest
+// ledger.
+void
+TxCache::update(
+    std::vector<ripple::uint256> const& hashes,
+    std::vector<TransactionAndMetadata> const& transactions,
```
cjcobb23, line 16: We probably want to `assert` that `transactions` and `hashes` are the same length

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+    for (auto& map : cache_)
+    {
+        auto e = map.find(key);
+        if (e != map.end())
+        {
+            txHitCounter_++;
+            return {e->second};
+        }
+    }
+    return {};
```
cjcobb23, line 46: so if we had a second map, that mapped a transaction hash to a ledger sequence, we wouldn't need to do this loop. The loop is probably not so bad for a small number of ledgers, but for a larger number of ledgers, maybe a second map would make sense. What do you think?

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+TxCache::getLedgerTransactions(std::uint32_t const ledgerSequence) const
+{
+    std::shared_lock lck{mtx_};
+    txReqCounter_++;
+    auto diff = latestSeq_ - ledgerSequence;
+    if (diff < NUM_LEDGERS_CACHED)
+    {
+        auto head = (tail_ - 1) % NUM_LEDGERS_CACHED;
+        auto ledgerCache = cache_[(head - diff) % NUM_LEDGERS_CACHED];
+        std::vector<TransactionAndMetadata> result;
```
cjcobb23, line 67: Probably should `reserve` here, based on the size of `ledgerCache`

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+        for (auto const& tx : ledgerCache)
+        {
+            result.push_back(tx.second);
+        }
+        if (result.size() > 0)
+        {
+            txHitCounter_++;
```
cjcobb23, line 74: I feel like this is a bit misleading, because we actually probably got a bunch of hits

src/backend/TxCache.cpp
```diff
@@ -0,0 +1,105 @@
...
+        {
+            result.push_back(tx.first);
+        }
+        if (result.size() > 0)
+        {
+            txHitCounter_++;
+            return {result};
+        }
+    }
+    return {};
```
cjcobb23, line 102: This is almost the exact same logic as `getLedgerTransactions`, we should make a helper function to encapsulate the shared logic

src/backend/BackendInterface.cpp
```diff
@@ -327,4 +327,90 @@ BackendInterface::fetchFees(
...
+            if (results[i].transaction.size() == 0)
+            {
+                results[i] = objs[j];
+                ++j;
+            }
+        }
+    }
+
+    return results;
+}
```
cjcobb23, line 391: Both of these functions have almost the exact same logic as `fetchLedgerObject` and `fetchLedgerObjects`. Can we create a helper function to encapsulate the shared logic? You can use templates, and pass in lambdas

cjcobb23 (changes requested):

src/backend/TxCache.h
ethanlabelle: Can you clarify what needs to be initialized? `tail_`?

cjcobb23: sorry, yes `tail_`

godexsoft:
I think it is safe to assume we are not going to be merging this so i'm closing it.

PR #242, opened 2022-08-08
Database read throttle

cjcobb23:
Add a way for the server to throttle the number of outstanding read requests. Needs a bit more testing, but wanted to get some eyes on this now.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/242)

### DIFF
--- a/example-config.json
+++ b/example-config.json
@@ -10,7 +10,7 @@
             "replication_factor":1,
             "table_prefix":"",
             "max_write_requests_outstanding":25000,
-            "max_read_requests_outstanding":1000,
+            "max_read_requests_outstanding":30000,
             "threads":8
         }
     },
--- a/example-config.json
+++ b/example-config.json
@@ -9,7 +9,8 @@
             "keyspace":"clio",
             "replication_factor":1,
             "table_prefix":"",
-            "max_requests_outstanding":25000,
+            "max_write_requests_outstanding":25000,
+            "max_read_requests_outstanding":1000,
             "threads":8
         }
     },
--- a/src/backend/BackendInterface.h
+++ b/src/backend/BackendInterface.h
@@ -13,10 +13,6 @@ namespace Backend {
 class DatabaseTimeout : public std::exception
 {
 public:
-    DatabaseTimeout()
-    {
-    }
-
     const char*
     what() const throw() override
     {
--- a/src/backend/BackendInterface.h
+++ b/src/backend/BackendInterface.h
@@ -57,13 +57,13 @@ retryOnTimeout(F func, size_t waitMs = 500)
         {
             BOOST_LOG_TRIVIAL(error)
                 << __func__
-                << " function timed out. Sleeping and retrying ... ";
+                << " Database request timed out. Sleeping and retrying ... ";
             std::this_thread::sleep_for(std::chrono::milliseconds(waitMs));
         }
         catch (DatabaseRequestThrottled& t)
         {
             BOOST_LOG_TRIVIAL(error)
-                << __func__ << " Request throttled. Retrying";
+                << __func__ << " Database request throttled. Retrying";
         }
     }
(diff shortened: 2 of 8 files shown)


example-config.json
```diff
@@ -9,7 +9,8 @@
             "keyspace":"clio",
             "replication_factor":1,
             "table_prefix":"",
-            "max_requests_outstanding":25000,
```
natenichols, line 12: I don't know how to avoid renaming this _particularly important_ config option. 

Could we leave `max_requests_outstanding` as an option, and add `max_write_requests_outstanding` as an alias for this config option?

src/backend/BackendInterface.h
```diff
@@ -7,10 +7,16 @@
 #include <backend/Types.h>
 #include <thread>
 #include <type_traits>
+
 namespace Backend {
 
 class DatabaseTimeout : public std::exception
 {
+public:
+    DatabaseTimeout()
```
natenichols: Why the addition of this constructor?

src/webserver/WsBase.h
```diff
@@ -266,59 +266,51 @@ class WsSession : public WsBase,
...
+            else
             {
-                BOOST_LOG_TRIVIAL(error) << __func__ << " Database timeout";
-                return sendError(RPC::Error::rpcNOT_READY);
+                counters_.rpcComplete(context->method, us);
+
+                response["result"] = std::get<boost::json::object>(v);
             }
         }
         catch (std::exception const& e)
```
natenichols, line 316: Is it ok to count `DatabaseTimeout`s as internal errors?

Database timeouts aren't exactly a _huge_ issue, do we retry reads in the `BackendInterface`, or anytime a read times out we'll respond with `rpcINTERNAL`.

src/rpc/RPC.cpp
```diff
@@ -324,6 +324,11 @@ buildResponse(Context const& ctx)
     if (ctx.method == "ping")
         return boost::json::object{};
 
+    if (ctx.backend->isTooBusy())
```
natenichols, line 345: Should/can this logic go in `WorkQueue`? It'd be nice to only handle `rpcTOO_BUSY` at one point in the code. 

IMO its better here, and then we only reject w/ `rpcTOO_BUSY` in WorkQueue when the queue is full. Putting this here still prevents us from sending any extra read traffic to the database.

example-config.json
cjcobb23: I feel like its ok to rename config options, because operators choose when they update, and the changes are in the release notes. I think two config options for the same thing is sloppy, and then its unclear which takes precedence. But I'm not totally sure.

src/backend/BackendInterface.h
cjcobb23: Not needed, will delete

src/webserver/WsBase.h
cjcobb23: we count them as `rpcTOO_BUSY`, look at the change in RPC.cpp, that's where we catch now

src/rpc/RPC.cpp
cjcobb23: We don't have access to the backend in `WorkQueue` as of now, so we would need to refactor and pass that in. I think its cleaner to just reject here.

I'm not sure if we can handle `rpcTOO_BUSY` at just one point, since there are multiple things that can cause too busy. Particularly the `DatabaseTimeout` can happen at any time during handler execution.

src/webserver/WsBase.h
natenichols: Noted, thanks

natenichols (dismissed):
LGTM :+1:

cjcobb23:
This is good to go

godexsoft (approved):
LGTM

PR #246, opened 2022-08-12
Allow server to download cache from another clio server

cjcobb23:
* Config takes an array of clio peers. If any of these peers have a
  full cache, clio picks a peer at random to download the cache from.
  Otherwise, fall back to downloading cache from the database

Needs a bit more testing and some cleanup, but ready for review

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/246)

### DIFF
--- a/example-config.json
+++ b/example-config.json
@@ -21,6 +21,7 @@
             "grpc_port":"50051"
         }
     ],
+    "clio_peers": [],
     "dos_guard":
     {
         "whitelist":["127.0.0.1"]
--- a/example-config.json
+++ b/example-config.json
@@ -21,7 +21,7 @@
             "grpc_port":"50051"
         }
     ],
-    "clio_peers": [],
+    "peers": [],
     "dos_guard":
     {
         "whitelist":["127.0.0.1"]
--- a/src/etl/ReportingETL.cpp
+++ b/src/etl/ReportingETL.cpp
@@ -1004,12 +1004,12 @@ ReportingETL::loadCacheFromClioPeer(
             BOOST_LOG_TRIVIAL(trace)
                 << __func__ << " Successfully parsed response " << parsed;
 
-            auto& response = parsed.as_object();
-            if (response.contains("error"))
+            if (auto const& response = parsed.as_object();
+                response.contains("error"))
             {
                 BOOST_LOG_TRIVIAL(error)
                     << __func__ << " Response contains error: " << response;
-                auto const& err = response["error"];
+                auto const& err = response.at("error");
                 if (err.is_string() && err.as_string() == "lgrNotFound")
                 {
                     BOOST_LOG_TRIVIAL(warning)
@@ -1020,7 +1020,8 @@ ReportingETL::loadCacheFromClioPeer(
                 }
                 return false;
             }
-            response = response["result"].as_object();
+            auto const& response = parsed.as_object()["result"].as_object();
+
             if (!response.contains("cache_full") ||
                 !response.at("cache_full").as_bool())
             {
@@ -1033,7 +1034,7 @@ ReportingETL::loadCacheFromClioPeer(
             else
                 marker = {};
 
-            auto const& state = response["state"].as_array();
+            auto const& state = response.at("state").as_array();
 
             std::vector<Backend::LedgerObject> objects;
             objects.reserve(state.size());
--- a/src/etl/ReportingETL.cpp
+++ b/src/etl/ReportingETL.cpp
@@ -912,6 +912,156 @@ ReportingETL::monitor()
         }
     }
 }
+bool
+ReportingETL::loadCacheFromClioPeer(
+    uint32_t ledgerIndex,
+    std::string const& ip,
+    std::string const& port,
+    boost::asio::yield_context& yield)
+{
+    namespace beast = boost::beast;          // from <boost/beast.hpp>
+    namespace http = beast::http;            // from <boost/beast/http.hpp>
+    namespace websocket = beast::websocket;  // from
+    namespace net = boost::asio;             // from
+    using tcp = boost::asio::ip::tcp;        // from
+    try
+    {
+        boost::beast::error_code ec;
+        // These objects perform our I/O
+        tcp::resolver resolver{ioContext_};
+
+        BOOST_LOG_TRIVIAL(trace) << "Creating websocket";
+        auto ws =
+            std::make_unique<websocket::stream<beast::tcp_stream>>(ioContext_);
+
+        // Look up the domain name
+        auto const results = resolver.async_resolve(ip, port, yield[ec]);
+        if (ec)
+            return {};
+
+        BOOST_LOG_TRIVIAL(trace) << "Connecting websocket";
+        // Make the connection on the IP address we get from a lookup
+        ws->next_layer().async_connect(results, yield[ec]);
+        if (ec)
+            return {};
+
+        BOOST_LOG_TRIVIAL(trace) << "Performing websocket handshake";
+        // Perform the websocket handshake
+        ws->async_handshake(ip, "/", yield[ec]);
+        if (ec)
+            return {};
+
+        std::optional<boost::json::value> marker;
+
+        BOOST_LOG_TRIVIAL(trace) << "Sending request";
+        auto getRequest = [&](auto marker) {
+            boost::json::object request = {
+                {"command", "ledger_data"},
+                {"ledger_index", ledgerIndex},
+                {"binary", true},
+                {"out_of_order", true},
+                {"limit", 2048}};
+
+            if (marker)
+                request["marker"] = *marker;
+            return request;
+        };
+
+        do
+        {
+            // Send the message
+            ws->async_write(
+                net::buffer(boost::json::serialize(getRequest(marker))),
+                yield[ec]);
+            if (ec)
+            {
+                BOOST_LOG_TRIVIAL(error) << "error writing = " << ec.message();
+                return {};
+            }
+
+            beast::flat_buffer buffer;
+            ws->async_read(buffer, yield[ec]);
+            if (ec)
+            {
+                BOOST_LOG_TRIVIAL(error) << "error reading = " << ec.message();
+                return {};
+            }
+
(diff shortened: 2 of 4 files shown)


src/etl/ReportingETL.cpp
```diff
@@ -1159,5 +1343,23 @@ ReportingETL::ReportingETL(
         if (cache.contains("page_fetch_size") &&
             cache.at("page_fetch_size").is_int64())
             cachePageFetchSize_ = cache.at("page_fetch_size").as_int64();
+        if (cache.contains("clio_peers") && cache.at("clio_peers").is_array())
```
cjcobb23: I kind of think this should just be called "peers"

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,156 @@ ReportingETL::monitor()
...
+        tcp::resolver resolver{ioContext_};
+
+        BOOST_LOG_TRIVIAL(trace) << "Creating websocket";
+        auto ws =
+            std::make_unique<websocket::stream<beast::tcp_stream>>(ioContext_);
+
+        // Look up the domain name
+        auto const results = resolver.async_resolve(ip, port, yield[ec]);
+        if (ec)
+            return {};
```
godexsoft, line 942: i think this is a copy-paste. the return type of the function is `bool`. while `{}` works it would probably be more readable to put `false` instead :)

godexsoft: there is more of the same below

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,156 @@ ReportingETL::monitor()
...
+            ws->async_read(buffer, yield[ec]);
+            if (ec)
+            {
+                BOOST_LOG_TRIVIAL(error) << "error reading = " << ec.message();
+                return {};
+            }
+
+            auto begin = static_cast<char const*>(buffer.data().data());
+            auto end = begin + buffer.data().size();
+            auto parsed = boost::json::parse(std::string(begin, end));
```
godexsoft: you might be able to use `beast::buffers_to_string(buffer)` here to get your string instead

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,156 @@ ReportingETL::monitor()
...
+
+            if (!parsed.is_object())
+            {
+                BOOST_LOG_TRIVIAL(error)
+                    << "Error parsing response: " << std::string{begin, end};
```
godexsoft: can reuse the string from above here i think

src/etl/ReportingETL.cpp
godexsoft: agree, simpler yet still makes total sense

godexsoft (dismissed):
LGTM. Leaving a few minor comments, feel free to ignore.

Other overall minor things:
- `clio` code in general can use `auto` more
- consider making things that don't change `auto const&`
- consider temporary variables vs long chain calls (readability, formatting) - compiler generates the same assembly anyways so there is no performance concern in doing that

cjcobb23:
@godexsoft can you give this one more review? This is ready to go now, did lots of testing

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,171 @@ ReportingETL::monitor()
...
+            if (!parsed.is_object())
+            {
+                BOOST_LOG_TRIVIAL(error)
+                    << __func__ << "Error parsing response: " << raw;
+                return false;
+            }
+            BOOST_LOG_TRIVIAL(trace)
+                << __func__ << "Successfully parsed response " << parsed;
+
+            auto response = parsed.as_object();
```
godexsoft: is copy intentional here? i'd probably use `auto const&` if not.

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,171 @@ ReportingETL::monitor()
...
+                {
+                    BOOST_LOG_TRIVIAL(warning)
+                        << __func__
+                        << " ledger not found. ledger = " << ledgerIndex
+                        << ". trying again";
+                    continue;
+                }
+                return false;
+            }
+            response = response["result"].as_object();
```
godexsoft: consider using `auto const& result` instead

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,171 @@ ReportingETL::monitor()
...
+                BOOST_LOG_TRIVIAL(error)
+                    << __func__ << " cache not full for clio node. ip = " << ip;
+                return false;
+            }
+            if (response.contains("marker"))
+                marker = response.at("marker");
+            else
+                marker = {};
+
+            boost::json::array& state = response["state"].as_array();
```
godexsoft: consider `auto const& state` instead

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,171 @@ ReportingETL::monitor()
...
+
+            std::vector<Backend::LedgerObject> objects;
+            objects.reserve(state.size());
+            for (auto const& ledgerObject : state)
+            {
+                boost::json::object const& obj = ledgerObject.as_object();
```
godexsoft: minor: can be `auto const&` as well

src/etl/ReportingETL.cpp
```diff
@@ -935,6 +1100,51 @@ ReportingETL::loadCache(uint32_t seq)
         assert(false);
         return;
     }
+
+    if (clioPeers.size() > 0)
+    {
+        boost::asio::spawn(
+            ioContext_, [this, seq](boost::asio::yield_context yield) {
+                for (auto& peer : clioPeers)
```
godexsoft: probably can be `auto const&`?

src/etl/ReportingETL.cpp
```diff
@@ -912,6 +912,171 @@ ReportingETL::monitor()
...
+        BOOST_LOG_TRIVIAL(trace) << __func__ << "Creating websocket";
+        auto ws =
+            std::make_unique<websocket::stream<beast::tcp_stream>>(ioContext_);
+
+        // Look up the domain name
+        auto const results = resolver.async_resolve(ip, port, yield[ec]);
+        if (ec)
+            return {};
+
+        BOOST_LOG_TRIVIAL(trace) << __func__ << "Connecting websocket";
```
godexsoft: this and many other lines seem to collapse `__func__` and the string after it (needs a space)

src/etl/ReportingETL.cpp
```diff
@@ -935,6 +1100,51 @@ ReportingETL::loadCache(uint32_t seq)
...
+        loadCacheFromDb(seq);
+    }
+    // If loading synchronously, poll cache until full
+    while (cacheLoadStyle_ == CacheLoadStyle::SYNC &&
+           !backend_->cache().isFull())
+    {
+        BOOST_LOG_TRIVIAL(debug)
+            << "Cache not full. Cache size = " << backend_->cache().size()
+            << ". Sleeping ...";
+        std::this_thread::sleep_for(std::chrono::seconds(10));
```
godexsoft, line 1132: seeing that we have an `ioContext_` here, maybe it would be better to use a deadline timer so that the thread does not have to freeze? i'm not sure if we are inside an `asio`-managed context here or our own thread tho so it's just an idea - it may be good as is.

src/etl/ReportingETL.cpp
```diff
@@ -1159,5 +1358,23 @@ ReportingETL::ReportingETL(
         if (cache.contains("page_fetch_size") &&
             cache.at("page_fetch_size").is_int64())
             cachePageFetchSize_ = cache.at("page_fetch_size").as_int64();
+        if (cache.contains("peers") && cache.at("peers").is_array())
+        {
+            boost::json::array peers = cache.at("peers").as_array();
```
godexsoft: probably should be `auto const&`

src/etl/ReportingETL.cpp
```diff
@@ -1159,5 +1358,23 @@ ReportingETL::ReportingETL(
...
+            for (auto& peer : peers)
```
godexsoft: consider `auto const&`

src/etl/ReportingETL.cpp
```diff
@@ -1159,5 +1358,23 @@ ReportingETL::ReportingETL(
...
+            {
+                auto& clio = peer.as_object();
```
godexsoft: consider `const` also

example-config.json
```diff
@@ -21,6 +21,7 @@
             "grpc_port":"50051"
         }
     ],
+    "clio_peers": [],
```
godexsoft: this should now be `peers` if i read the code right

godexsoft (changes requested):

src/etl/ReportingETL.cpp
cjcobb23: We are in our own thread here, not an `asio` managed context, so we actually have to block. The semantics of the function are to block until the cache is loaded in this case.

src/etl/ReportingETL.cpp
cjcobb23: I actually prefer this way, then no one accidentally uses `response` further down. There's no need for the original `response` object after this, and having `response` and `result` I find confusing

godexsoft: well, now that you changed response to `auto&` you are rewriting your original response inside of `parsed` so this line needs to change i think? i'd have either what you had before or split into two differently named `auto const&` instead.

src/etl/ReportingETL.cpp
godexsoft: sounds good then :+1:

godexsoft (changes requested):

src/etl/ReportingETL.cpp
cjcobb23: oh yeah good catch. After thinking about this, I ended up removing this reassignment entirely, since a copy here is going to be expensive, since these objects are very large.

godexsoft (approved):
Looks good!

PR #252, opened 2022-08-16
insert delivered_amount based on close time

cjcobb23:
Previously, we were only checking the ledger sequence to determine if we should insert delivered amount. This works on mainnet, but could fail on other networks that are newer

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/252)

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -296,7 +296,8 @@ std::optional<ripple::STAmount>
 getDeliveredAmount(
     std::shared_ptr<ripple::STTx const> const& txn,
     std::shared_ptr<ripple::TxMeta const> const& meta,
-    std::uint32_t const ledgerSequence)
+    std::uint32_t const ledgerSequence,
+    uint32_t date)
 {
     if (meta->hasDeliveredAmount())
         return meta->getDeliveredAmount();
@@ -312,7 +313,7 @@ getDeliveredAmount(
         // then its absence indicates that the amount delivered is listed in the
         // Amount field. DeliveredAmount went live January 24, 2014.
         // 446000000 is in Feb 2014, well after DeliveredAmount went live
-        if (ledgerSequence >= 4594095)
+        if (ledgerSequence >= 4594095 || date > 446000000)
         {
             return txn->getFieldAmount(ripple::sfAmount);
         }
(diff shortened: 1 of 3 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -458,19 +459,20 @@ toExpandedJson(Backend::TransactionAndMetadata const& blobs)
...
 
 bool
 insertDeliveredAmount(
     boost::json::object& metaJson,
     std::shared_ptr<ripple::STTx const> const& txn,
-    std::shared_ptr<ripple::TxMeta const> const& meta)
+    std::shared_ptr<ripple::TxMeta const> const& meta,
+    uint32_t date)
 {
     if (canHaveDeliveredAmount(txn, meta))
```
natenichols, line 473: Do we also need to change `canHaveDeliveredAmount`?

src/rpc/RPCHelpers.cpp
```diff
@@ -458,19 +459,20 @@ toExpandedJson(Backend::TransactionAndMetadata const& blobs)
...
     {
-        if (auto amt = getDeliveredAmount(txn, meta, meta->getLgrSeq()))
+        if (auto amt = getDeliveredAmount(txn, meta, meta->getLgrSeq(), date))
```
natenichols, line 475: Can we do a cpp17 style if statement here?

src/rpc/RPCHelpers.cpp
```diff
@@ -312,7 +313,7 @@ getDeliveredAmount(
         // then its absence indicates that the amount delivered is listed in the
         // Amount field. DeliveredAmount went live January 24, 2014.
         // 446000000 is in Feb 2014, well after DeliveredAmount went live
-        if (ledgerSequence >= 4594095)
+        if (ledgerSequence >= 4594095 || date > 446000000)
         {
             return txn->getFieldAmount(ripple::sfAmount);
```
natenichols, line 318: Is it _guaranteed_ that in this case `sfAmount` will be `delivered_amount`?

If a transaction is a partial payment, will it _always_ have `delivered_amount` in the metadata?

cjcobb23: According to the comment (which I copied from rippled), delivered amount will be in the metadata for partial payments after this ledger. The metadata is checked prior to this, at the beginning of the function.

src/rpc/RPCHelpers.cpp
cjcobb23: Don't think so, that function is just looking at the transaction to see if its the type of transaction that could have a delivered amount.

src/rpc/RPCHelpers.cpp
cjcobb23: Like : `if (auto amt = getDeliveredAmount(txn, meta, meta->getLgrSeq(), date); amt)`?

natenichols: Yes. I like that better, but if you hate it feel free to leave it

natenichols (approved):

PR #253, opened 2022-08-16
refactor main.cpp into an application class

natenichols:
Throwing out #151, and starting anew

By removing the testing framework, we cut the surface area of this PR in half

I still have that branch saved locally, and should be able to resurrect it when the time comes.


src/backend/BackendFactory.h
```diff
@@ -5,43 +5,28 @@
 #include <backend/BackendInterface.h>
 #include <backend/CassandraBackend.h>
 #include <backend/PostgresBackend.h>
+#include <main/Application.h>
+#include <main/Config.h>
 
 namespace Backend {
-std::shared_ptr<BackendInterface>
-make_Backend(boost::asio::io_context& ioc, boost::json::object const& config)
+static std::unique_ptr<BackendInterface>
```
cjcobb23, line 12: What do we need static here for?

src/backend/BackendInterface.cpp
```diff
@@ -108,6 +116,9 @@ BackendInterface::fetchLedgerObjects(
                 results[i] = objs[j];
                 ++j;
             }
+
+            if (j >= objs.size())
+                break;
```
cjcobb23, line 121: why do we need this?

src/backend/CassandraBackend.cpp
```diff
@@ -964,19 +965,8 @@ CassandraBackend::doOnlineDelete(
...
-            return {jsonStr.c_str(), jsonStr.size()};
-        }
-        return {""};
-    };
-    auto getInt = [this](std::string const& field) -> std::optional<int> {
-        if (config_.contains(field) && config_.at(field).is_int64())
-            return config_[field].as_int64();
-        return {};
-    };
+    auto const& dbConfig = app_.config().cassandra().cassandra;
```
cjcobb23, line 968: The `cassandra().cassandra` looks a little weird, do we need to access via a member variable?

src/etl/ETLSource.cpp
```diff
@@ -77,52 +79,32 @@ ForwardCache::get(boost::json::object const& request) const
...
+    ip_ = config.ip;
+    wsPort_ = config.wsPort;
+    grpcPort_ = config.grpcPort;
+
+    try
     {
-        auto portjs = config.at("ws_port").as_string();
-        wsPort_ = {portjs.c_str(), portjs.size()};
+        boost::asio::ip::tcp::endpoint endpoint{
+            boost::asio::ip::make_address(ip_), std::stoi(grpcPort_)};
```
cjcobb23: grpcPort_ is optional, are we just catching an exception if its not present?

src/main/Application.h
```diff
@@ -0,0 +1,246 @@
...
+    config() const = 0;
+
+    virtual boost::asio::io_context&
+    socketIoc() const = 0;
+
+    virtual boost::asio::io_context&
+    rpcIoc() const = 0;
+
+    virtual boost::asio::io_context&
+    etlIoc() const = 0;
```
cjcobb23, line 52: I think this should be called `backendIoc` because we use this for etl, for publishing ledgers, and for retrying db writes. I think backend is more general.

src/main/impl/Application.cpp
```diff
@@ -0,0 +1,189 @@
+#include <boost/asio/dispatch.hpp>
```
cjcobb23: What does having an `impl` folder do for us here?

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+#include <iostream>
+#include <optional>
+#include <string>
+#include <unordered_set>
+
+struct ETLSourceConfig
+{
+    std::string ip;
+    std::string wsPort;
+    std::string grpcPort;
```
cjcobb23: this is supposed to be optional. read only nodes don't need it

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    std::vector<std::string> cacheCommands;
+};
+
+enum class CacheLoadStyle { ASYNC, SYNC, NOT_AT_ALL };
+
+struct CacheConfig
+{
+    std::uint32_t numDiffs;
```
cjcobb23, line 23: We need numMarkers too

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    struct CassandraOptions
+    {
+        std::optional<std::string> secureConnectBundle;
+        std::optional<std::string> contactPoints;
+        std::optional<std::string> keyspace;
+        std::optional<std::string> username;
+        std::optional<std::string> password;
+        std::optional<std::string> certfile;
+        std::optional<std::uint32_t> port;
+        std::optional<std::uint32_t> ttl;
```
cjcobb23, line 60: I'd probably remove this for now, data deletion is not supported yet

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+        std::string tablePrefix;
```
cjcobb23, line 61: I think we should remove this parameter, this was a hack. People should use separate keyspaces.

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    PostgresConfig(boost::json::value& config);
+};
+
+boost::json::object
+parseConfig(std::string const& filename);
+
+class Config
+{
+    boost::json::object json_;
+    std::shared_ptr<DatabaseConfig> dbConfigStore;
```
cjcobb23: What is this member? There is another `DatabaseConfig` member on line 114

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    std::optional<CacheConfig> cache;
+    std::optional<ServerConfig> server;
+
+    bool readOnly;
+
+    std::optional<std::string> sslCertFile;
+    std::optional<std::string> sslKeyFile;
+
+    std::optional<std::string> logFile;
+    std::string logLevel;
```
cjcobb23, line 118: should this be an enum?

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    std::vector<ETLSourceConfig> etlSources;
+    std::optional<CacheConfig> cache;
+    std::optional<ServerConfig> server;
+
+    bool readOnly;
+
+    std::optional<std::string> sslCertFile;
+    std::optional<std::string> sslKeyFile;
+
+    std::optional<std::string> logFile;
```
cjcobb23: Do we even support this anymore? I thought we just use a log_directory now

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    std::uint32_t logRotationHourInterval;
+    std::uint32_t logDirectoryMaxSize;
+
+    std::uint32_t numMarkers;
+
+    std::uint32_t subscriptionWorkers;
+    std::uint32_t etlWorkers;
+    std::uint32_t rpcWorkers;
+    std::uint32_t socketWorkers;
+    std::uint32_t maxQueueSize;
```
cjcobb23, line 131: Why do we need to write `std::uint32_t` versus just `uint32_t`? What does the extra verbosity give us?

cjcobb23: Also I'd call these `numEtlWorkers` and the like

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+
+    std::optional<std::uint32_t> startSequence;
+    std::optional<std::uint32_t> finishSequence;
+    std::optional<std::uint32_t> onlineDelete;
```
cjcobb23, line 135: Let's remove this, its not supported yet

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+        std::optional<std::uint32_t> port;
+        std::optional<std::uint32_t> ttl;
+        std::string tablePrefix;
+        std::uint32_t replicationFactor;
+        std::uint32_t syncInterval;
+        std::uint32_t maxRequestsOutstanding;
+        std::uint32_t threads;
+    };
+
+    CassandraOptions cassandra;
```
cjcobb23, line 68: What is the purpose of the double nesting? All of the fields of this struct are inside another struct. Why two structs?

src/main/Config.h
```diff
@@ -0,0 +1,164 @@
...
+    std::uint32_t maxQueueSize;
+
+    std::optional<std::uint32_t> startSequence;
+    std::optional<std::uint32_t> finishSequence;
+    std::optional<std::uint32_t> onlineDelete;
+    std::uint32_t extractorThreads;
+    std::uint32_t txnThreshold;
+
+    // Be careful w/ these helper. They will throw if the underlying
+    // database configuration is of the wrong type.
```
cjcobb23, line 140: Why not just use a variant? Instead of inheritance

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+            contents << in.rdbuf();
+            in.close();
+            std::cout << contents.str() << std::endl;
+            boost::json::value value = boost::json::parse(contents.str());
+            return value.as_object();
+        }
+    }
+    catch (std::exception const& e)
+    {
+        std::cout << e.what() << std::endl;
```
cjcobb23: Should be logged, not cout

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    throw std::runtime_error(
+        "Could not parse JSON configuration. Verify it is valid JSON format");
```
cjcobb23: Maybe we should dump the json here?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+}
+
+struct cfgOPTIONAL;
+struct cfgREQUIRED;
```
cjcobb23: What's up with the funky styling here? Structs and classes are usually camel case, with the first letter capitalized.

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    {
+        throw std::runtime_error(
+            "Could not parse field " + boost::json::serialize(json) + "\n" +
+            e.what());
+    }
+}
+
+template <typename T, typename F, typename Def>
+void
+add(T& value, boost::json::value& json, F&& parser, Def def)
```
cjcobb23: I'd change `Def` to `V` and `def` to `default`.

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+{
+    std::optional<std::string> parsed = {};
+
+    if (value.is_string())
+        parsed = value.as_string().c_str();
+
+    return checkRequired<Required>(parsed);
+}
+
+template <typename Required>
```
cjcobb23: So `Required` is really just a bool? Why do this templating stuff, why not just pass a bool, or have two functions?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    std::unique_ptr<T>& store,
+    boost::json::value& json,
+    F&& parser)
+{
+    add(store, json, parser);
+    value = *store;
+}
+
+template <typename Required, typename T>
+auto
```
cjcobb23: Is the return value ever used?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    if (!value.is_object())
+        throw std::runtime_error("DOSGuard config must be a json object");
+
+    boost::json::object& config = value.as_object();
+
+    DOSGuardConfig dosGuard;
+    add(dosGuard.maxFetches,
+        config["max_fetches"],
+        &parseUInt32<cfgREQUIRED>,
+        100);
```
cjcobb23: This seems like a lot of indirection. Why not something like:
```
dosGuard.maxFetches = parseUInt32(config["max_fetches"],100);
```

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+    std::unordered_set<T> result = {};
+    std::transform(
+        collection.begin(),
+        collection.end(),
+        std::inserter(result, result.begin()),
+        [](boost::json::value& value) -> T { return value_to<T>(value); });
+
+    return result;
+}
```
cjcobb23: There's a lot of templating in these functions, and I don't see why its necessary. It's a bit hard to read. Can we do this in a more readable, simple way?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    throw std::runtime_error(
+        "Could not parse JSON configuration. Verify it is valid JSON format");
+}
+
+struct cfgOPTIONAL;
+struct cfgREQUIRED;
+
+template <typename T, typename F>
+void
+add(T& value, boost::json::value& json, F&& parser)
```
cjcobb23: Can we avoid using an out-param?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+    std::transform(
+        collection.begin(),
+        collection.end(),
+        std::inserter(result, result.begin()),
+        [](boost::json::value& value) -> T { return value_to<T>(value); });
+
+    return result;
+}
+
+static DOSGuardConfig
```
cjcobb23: Why do we need `static` here?

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+            config.cacheLoadStyle = CacheLoadStyle::ASYNC;
+        else if (entry == "none" || entry == "no")
+            config.cacheLoadStyle = CacheLoadStyle::NOT_AT_ALL;
+        else
+            throw std::runtime_error(
+                "Invalid cache load option: specify sync, async, or none");
+    }
+
+    add(config.numDiffs, cache["num_diffs"], &parseUInt32<cfgREQUIRED>, 1);
+
```
cjcobb23: missing `num_markers`

src/main/impl/Config.cpp
```diff
@@ -0,0 +1,422 @@
...
+        &parseUInt32<cfgREQUIRED>,
+        1000);
+}
+
+Config::Config(boost::json::object const& json)
+    : json_(json)
+    , dbConfigStore(parseDatabaseConfig(json_["database"]))
+    , database(*dbConfigStore)
+{
+    add(dosGuard, json_["dos_guard"], &parseDosGuardConfig);
```
cjcobb23: Why can't we do something like:
```
dosGuard_ = parseDosGuardConfig(json_["dos_guard"]);
```
I feel this would be much more readable and straightforward.

src/main/Application.h
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    ETLLoadBalancer&
+    balancer() const override
+    {
+        return *balancer_;
+    }
+
+    SubscriptionManager&
+    subscriptions() const override
```
cjcobb23, line 178: `subscriptionManager` would be a better name.

src/main/Application.h
```diff
@@ -0,0 +1,246 @@
...
+        return *ledgers_;
+    }
+
+    DOSGuard&
+    dosGuard() const override
+    {
+        return *dosGuard_;
+    }
+
+    ReportingETL&
```
cjcobb23, line 201: I feel like we should rename this class.

cjcobb23 (changes requested):
Did a first pass. I think we can simplify the config parsing implementation, its a bit hard to read as is

src/etl/ETLSource.cpp
natenichols: `grpcPort_` should be required. Will change.

src/backend/BackendFactory.h
natenichols: It is not needed. I was adding for consistency with other factory function.s

src/main/Config.h
natenichols: We don't need to. We use `std::uint32_t` everywhere else, so I figured I'd be consistent.

src/main/impl/Config.cpp
natenichols: Logging is not initialized at this point

natenichols: So we can't log yet. 

Logging is initialized based on the values in the configuration.

src/main/impl/Config.cpp
natenichols: `default` is a keyword in cpp

will use `dfault`.

src/main/impl/Config.cpp
natenichols: Because we want `parseUInt<cfgOptional>()` to return a `std::optional<std::uin32_t>` but we want `parseUInt<cfgRequired>()` to return a `std::uint32_t`

src/main/impl/Config.cpp
natenichols: The template-ing is because we want `parseUInt<cfgOptional>()` to return a `std::optional<std::uin32_t>` but we want `parseUInt<cfgRequired>()` to return a `std::uint32_t`

src/main/impl/Config.cpp
natenichols: Sure, lemme refactor this a bit.

src/main/impl/Config.cpp
natenichols: We don't, will remove

src/main/Application.h
natenichols: That's probably not in scope for this PR

We should rename to `ClioETL` or something.

src/main/Config.h
natenichols: It matches the config.json structure

PR #255, opened 2022-08-17
add nft_history

ledhed2222:
This PR does three things:

* Adds `nft_history` API which should work identically to `account_tx`. The one difference is that `nft_history` only responds with successful transactions at this point, since that is all we are storing in the index table.
* Adds `isClioOnly` field to RPC handlers and marks `nft_info` and `nft_history` as such. This will prevent these APIs from being forwarded to ripple. The benefit is not only to short circuit a forward that will not work, but also to improve error handling. In the event that someone specifies `ledger_index: "current"` on these APIs, and if they were to be forwarded, the resulting error just indicates that ripple doesn't know what the API command is. This is confusing. Instead, now that they are not forwarded, the error will be more sensible and indicate that the specified ledger was not found.
* Fixes an issue where the returned `ledger_index_min` and `ledger_index_max` values for `account_tx` were not correct

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 build/
 .vscode
 .python-version
+config.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,9 @@ target_sources(clio PRIVATE
   src/rpc/handlers/GatewayBalances.cpp
   src/rpc/handlers/NoRippleCheck.cpp
   # NFT
+  src/rpc/handlers/NFTHistory.cpp
   src/rpc/handlers/NFTInfo.cpp
+  src/rpc/handlers/NFTOffers.cpp
   # Ledger
   src/rpc/handlers/Ledger.cpp
   src/rpc/handlers/LedgerData.cpp
@@ -88,8 +90,6 @@ target_sources(clio PRIVATE
   # Dex
   src/rpc/handlers/BookChanges.cpp
   src/rpc/handlers/BookOffers.cpp
-  # NFT
-  src/rpc/handlers/NFTOffers.cpp
   # Payment Channel
   src/rpc/handlers/ChannelAuthorize.cpp
   src/rpc/handlers/ChannelVerify.cpp
--- a/src/backend/CassandraBackend.cpp
+++ b/src/backend/CassandraBackend.cpp
@@ -651,6 +651,9 @@ CassandraBackend::fetchNFTTransactions(
                 static_cast<std::uint32_t>(lgrSeq),
                 static_cast<std::uint32_t>(txnIdx)};
 
+            // Only modify if forward because forward query
+            // (selectNFTTxForward_) orders by ledger/tx sequence >= whereas
+            // reverse query (selectNFTTx_) orders by ledger/tx sequence <.
             if (forward)
                 ++cursor->transactionIndex;
         }
@@ -732,6 +735,9 @@ CassandraBackend::fetchAccountTransactions(
                 static_cast<std::uint32_t>(lgrSeq),
                 static_cast<std::uint32_t>(txnIdx)};
 
+            // Only modify if forward because forward query
+            // (selectAccountTxForward_) orders by ledger/tx sequence >= whereas
+            // reverse query (selectAccountTx_) orders by ledger/tx sequence <.
             if (forward)
                 ++cursor->transactionIndex;
         }
--- a/src/rpc/RPC.cpp
+++ b/src/rpc/RPC.cpp
@@ -168,6 +168,7 @@ struct Handler
     std::string method;
     std::function<Result(Context const&)> handler;
     std::optional<LimitRange> limit;
+    bool isClioOnly = false;
 };
 
 class HandlerTable
@@ -206,6 +207,12 @@ public:
 
         return handlerMap_[command].handler;
     }
+
+    bool
+    isClioOnly(std::string const& command)
+    {
+        return handlerMap_.contains(command) && handlerMap_[command].isClioOnly;
+    }
 };
 
 static HandlerTable handlerTable{
@@ -224,7 +231,8 @@ static HandlerTable handlerTable{
     {"ledger", &doLedger, {}},
     {"ledger_data", &doLedgerData, LimitRange{1, 100, 2048}},
     {"nft_buy_offers", &doNFTBuyOffers, LimitRange{1, 50, 100}},
-    {"nft_info", &doNFTInfo},
+    {"nft_history", &doNFTHistory, LimitRange{1, 50, 100}, true},
+    {"nft_info", &doNFTInfo, {}, true},
     {"nft_sell_offers", &doNFTSellOffers, LimitRange{1, 50, 100}},
     {"ledger_entry", &doLedgerEntry, {}},
     {"ledger_range", &doLedgerRange, {}},
@@ -250,6 +258,12 @@ bool
 validHandler(std::string const& method)
 {
     return handlerTable.contains(method) || forwardCommands.contains(method);
+}
+
+bool
+isClioOnly(std::string const& method)
+{
+    return handlerTable.isClioOnly(method);
 }
 
 Status
@@ -286,6 +300,9 @@ bool
 shouldForwardToRippled(Context const& ctx)
 {
     auto request = ctx.params;
+
+    if (isClioOnly(ctx.method))
+        return false;
 
     if (forwardCommands.find(ctx.method) != forwardCommands.end())
         return true;
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -268,7 +268,7 @@ getTaker(boost::json::object const& request, ripple::AccountID& takerID)
     if (request.contains(JS(taker)))
     {
         auto parsed = parseTaker(request.at(JS(taker)));
-        if (auto status = std::get_if<Status>(&parsed))
+        if (auto status = std::get_if<Status>(&parsed); status)
             return *status;
         else
             takerID = std::get<ripple::AccountID>(parsed);
@@ -808,15 +808,21 @@ traverseOwnedNodes(
     }
     auto end = std::chrono::system_clock::now();
 
-    BOOST_LOG_TRIVIAL(debug) << "Time loading owned directories: "
-                             << ((end - start).count() / 1000000000.0);
+    BOOST_LOG_TRIVIAL(debug)
+        << "Time loading owned directories: "
+        << std::chrono::duration_cast<std::chrono::milliseconds>(end - start)
+               .count()
+        << " milliseconds";
 
     start = std::chrono::system_clock::now();
     auto objects = backend.fetchLedgerObjects(keys, sequence, yield);
     end = std::chrono::system_clock::now();
 
-    BOOST_LOG_TRIVIAL(debug) << "Time loading owned entries: "
-                             << ((end - start).count() / 1000000000.0);
+    BOOST_LOG_TRIVIAL(debug)
+        << "Time loading owned entries: "
+        << std::chrono::duration_cast<std::chrono::milliseconds>(end - start)
+               .count()
+        << " milliseconds";
 
     for (auto i = 0; i < objects.size(); ++i)
     {
@@ -1478,6 +1484,213 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
         }
     }
     return false;
+}
+
+std::variant<ripple::uint256, Status>
+getNFTID(boost::json::object const& request)
+{
+    if (!request.contains(JS(nft_id)))
+        return Status{Error::rpcINVALID_PARAMS, "missingTokenID"};
+
+    if (!request.at(JS(nft_id)).is_string())
+        return Status{Error::rpcINVALID_PARAMS, "tokenIDNotString"};
+
+    ripple::uint256 tokenid;
+    if (!tokenid.parseHex(request.at(JS(nft_id)).as_string().c_str()))
+        return Status{Error::rpcINVALID_PARAMS, "malformedTokenID"};
+
+    return tokenid;
+}
+
+// TODO - this function is long and shouldn't be responsible for as much as it
+// is. Split it out into some helper functions.
+std::variant<Status, boost::json::object>
+traverseTransactions(
+    Context const& context,
+    std::function<Backend::TransactionsAndCursor(
+        std::shared_ptr<Backend::BackendInterface const> const& backend,
+        std::uint32_t const,
+        bool const,
+        std::optional<Backend::TransactionsCursor> const&,
+        boost::asio::yield_context& yield)> transactionFetcher)
+{
+    auto request = context.params;
+    boost::json::object response = {};
+
+    bool const binary = getBool(request, JS(binary), false);
+    bool const forward = getBool(request, JS(forward), false);
+
+    std::optional<Backend::TransactionsCursor> cursor;
+
+    if (request.contains(JS(marker)))
+    {
+        if (!request.at(JS(marker)).is_object())
+            return Status{Error::rpcINVALID_PARAMS, "invalidMarker"};
+        auto const& obj = request.at(JS(marker)).as_object();
+
+        std::optional<std::uint32_t> transactionIndex = {};
+        if (obj.contains(JS(seq)))
+        {
+            if (!obj.at(JS(seq)).is_int64())
+                return Status{
+                    Error::rpcINVALID_PARAMS, "transactionIndexNotInt"};
+
+            transactionIndex =
+                boost::json::value_to<std::uint32_t>(obj.at(JS(seq)));
+        }
+
+        std::optional<std::uint32_t> ledgerIndex = {};
+        if (obj.contains(JS(ledger)))
+        {
+            if (!obj.at(JS(ledger)).is_int64())
+                return Status{Error::rpcINVALID_PARAMS, "ledgerIndexNotInt"};
+
+            ledgerIndex =
+                boost::json::value_to<std::uint32_t>(obj.at(JS(ledger)));
+        }
+
+        if (!transactionIndex || !ledgerIndex)
+            return Status{Error::rpcINVALID_PARAMS, "missingLedgerOrSeq"};
+
+        cursor = {*ledgerIndex, *transactionIndex};
+    }
+
+    auto minIndex = context.range.minSequence;
+    if (request.contains(JS(ledger_index_min)))
+    {
+        auto& min = request.at(JS(ledger_index_min));
+
+        if (!min.is_int64())
+            return Status{Error::rpcINVALID_PARAMS, "ledgerSeqMinNotNumber"};
+
+        if (min.as_int64() != -1)
+        {
+            if (context.range.maxSequence < min.as_int64() ||
+                context.range.minSequence > min.as_int64())
+                return Status{
+                    Error::rpcINVALID_PARAMS, "ledgerSeqMinOutOfRange"};
+            else
+                minIndex = boost::json::value_to<std::uint32_t>(min);
+        }
+
+        if (forward && !cursor)
+            cursor = {minIndex, 0};
+    }
+
+    auto maxIndex = context.range.maxSequence;
+    if (request.contains(JS(ledger_index_max)))
+    {
+        auto& max = request.at(JS(ledger_index_max));
+
+        if (!max.is_int64())
+            return Status{Error::rpcINVALID_PARAMS, "ledgerSeqMaxNotNumber"};
+
+        if (max.as_int64() != -1)
+        {
+            if (context.range.maxSequence < max.as_int64() ||
+                context.range.minSequence > max.as_int64())
+                return Status{
+                    Error::rpcINVALID_PARAMS, "ledgerSeqMaxOutOfRange"};
+            else
+                maxIndex = boost::json::value_to<std::uint32_t>(max);
+        }
+
+        if (minIndex > maxIndex)
+            return Status{Error::rpcINVALID_PARAMS, "invalidIndex"};
+
+        if (!forward && !cursor)
+            cursor = {maxIndex, INT32_MAX};
+    }
+
+    if (request.contains(JS(ledger_index)) || request.contains(JS(ledger_hash)))
+    {
+        if (request.contains(JS(ledger_index_max)) ||
+            request.contains(JS(ledger_index_min)))
+            return Status{
+                Error::rpcINVALID_PARAMS, "containsLedgerSpecifierAndRange"};
+
+        auto v = ledgerInfoFromRequest(context);
+        if (auto status = std::get_if<Status>(&v); status)
+            return *status;
+
+        maxIndex = minIndex = std::get<ripple::LedgerInfo>(v).seq;
+    }
+
+    if (!cursor)
+    {
+        if (forward)
+            cursor = {minIndex, 0};
+        else
+            cursor = {maxIndex, INT32_MAX};
+    }
+
+    std::uint32_t limit;
+    if (auto const status = getLimit(context, limit); status)
+        return status;
+
+    if (request.contains(JS(limit)))
+        response[JS(limit)] = limit;
+
+    boost::json::array txns;
+    auto [blobs, retCursor] = transactionFetcher(
+        context.backend, limit, forward, cursor, context.yield);
+    auto serializationStart = std::chrono::system_clock::now();
+
+    if (retCursor)
+    {
+        boost::json::object cursorJson;
+        cursorJson[JS(ledger)] = retCursor->ledgerSequence;
+        cursorJson[JS(seq)] = retCursor->transactionIndex;
+        response[JS(marker)] = cursorJson;
+    }
+
+    for (auto const& txnPlusMeta : blobs)
+    {
+        if (txnPlusMeta.ledgerSequence < minIndex ||
+            txnPlusMeta.ledgerSequence > maxIndex)
+        {
+            BOOST_LOG_TRIVIAL(debug)
+                << __func__
+                << " skipping over transactions from incomplete ledger";
+            continue;
+        }
+
+        boost::json::object obj;
+
+        if (!binary)
+        {
+            auto [txn, meta] = toExpandedJson(txnPlusMeta);
+            obj[JS(meta)] = meta;
+            obj[JS(tx)] = txn;
+            obj[JS(tx)].as_object()[JS(ledger_index)] =
+                txnPlusMeta.ledgerSequence;
+            obj[JS(tx)].as_object()[JS(date)] = txnPlusMeta.date;
+        }
+        else
+        {
+            obj[JS(meta)] = ripple::strHex(txnPlusMeta.metadata);
+            obj[JS(tx_blob)] = ripple::strHex(txnPlusMeta.transaction);
+            obj[JS(ledger_index)] = txnPlusMeta.ledgerSequence;
+            obj[JS(date)] = txnPlusMeta.date;
+        }
+        obj[JS(validated)] = true;
+
+        txns.push_back(obj);
+    }
+
+    response[JS(ledger_index_min)] = minIndex;
+    response[JS(ledger_index_max)] = maxIndex;
+
+    response[JS(transactions)] = txns;
+
+    BOOST_LOG_TRIVIAL(info)
+        << __func__ << " serialization took "
+        << std::chrono::duration_cast<std::chrono::milliseconds>(
+               std::chrono::system_clock::now() - serializationStart)
+               .count()
+        << " milliseconds";
+
+    return response;
 }
 
 }  // namespace RPC
--- a/src/rpc/handlers/NFTInfo.cpp
+++ b/src/rpc/handlers/NFTInfo.cpp
@@ -89,24 +89,15 @@ doNFTInfo(Context const& context)
     auto request = context.params;
     boost::json::object response = {};
 
-    if (!request.contains("nft_id"))
-        return Status{Error::rpcINVALID_PARAMS, "Missing nft_id"};
-
-    auto const& jsonTokenID = request.at("nft_id");
-    if (!jsonTokenID.is_string())
-        return Status{Error::rpcINVALID_PARAMS, "nft_id is not a string"};
-
-    ripple::uint256 tokenID;
-    if (!tokenID.parseHex(jsonTokenID.as_string().c_str()))
-        return Status{Error::rpcINVALID_PARAMS, "Malformed nft_id"};
-
-    // We only need to fetch the ledger header because the ledger hash is
-    // supposed to be included in the response. The ledger sequence is specified
-    // in the request
-    auto v = ledgerInfoFromRequest(context);
-    if (auto status = std::get_if<Status>(&v))
+    auto const maybeTokenID = getNFTID(request);
+    if (auto const status = std::get_if<Status>(&maybeTokenID); status)
         return *status;
-    ripple::LedgerInfo lgrInfo = std::get<ripple::LedgerInfo>(v);
+    auto const tokenID = std::get<ripple::uint256>(maybeTokenID);
+
+    auto const maybeLedgerInfo = ledgerInfoFromRequest(context);
+    if (auto status = std::get_if<Status>(&maybeLedgerInfo); status)
+        return *status;
+    auto const lgrInfo = std::get<ripple::LedgerInfo>(maybeLedgerInfo);
 
     std::optional<Backend::NFT> dbResponse =
         context.backend->fetchNFT(tokenID, lgrInfo.seq, context.yield);
@@ -130,10 +121,10 @@ doNFTInfo(Context const& context)
     {
         auto const maybeURI = getURI(*dbResponse, context);
         // An error occurred
-        if (Status const* status = std::get_if<Status>(&maybeURI))
+        if (Status const* status = std::get_if<Status>(&maybeURI); status)
             return *status;
         // A URI was found
-        if (std::string const* uri = std::get_if<std::string>(&maybeURI))
+        if (std::string const* uri = std::get_if<std::string>(&maybeURI); uri)
             response["uri"] = *uri;
         // A URI was not found, explicitly set to null
         else
--- a/src/rpc/Handlers.h
+++ b/src/rpc/Handlers.h
@@ -61,6 +61,9 @@ doNFTSellOffers(Context const& context);
 Result
 doNFTInfo(Context const& context);
 
+Result
+doNFTHistory(Context const& context);
+
 // ledger methods
 Result
 doLedger(Context const& context);
--- a/src/rpc/RPC.h
+++ b/src/rpc/RPC.h
@@ -226,6 +226,9 @@ buildResponse(Context const& ctx);
 bool
 validHandler(std::string const& method);
 
+bool
+isClioOnly(std::string const& method);
+
 Status
 getLimit(RPC::Context const& context, std::uint32_t& limit);
 
--- a/src/rpc/RPCHelpers.h
+++ b/src/rpc/RPCHelpers.h
@@ -24,6 +24,7 @@
 namespace RPC {
 std::optional<ripple::AccountID>
 accountFromStringStrict(std::string const& account);
+
 std::optional<ripple::AccountID>
 accountFromSeed(std::string const& account);
 
@@ -253,6 +254,21 @@ getChannelId(boost::json::object const& request, ripple::uint256& channelId);
 
 bool
 specifiesCurrentOrClosedLedger(boost::json::object const& request);
+
+std::variant<ripple::uint256, Status>
+getNFTID(boost::json::object const& request);
+
+// This function is the driver for both `account_tx` and `nft_tx` and should
+// be used for any future transaction enumeration APIs.
+std::variant<Status, boost::json::object>
+traverseTransactions(
+    Context const& context,
+    std::function<Backend::TransactionsAndCursor(
+        std::shared_ptr<Backend::BackendInterface const> const& backend,
+        std::uint32_t const,
+        bool const,
+        std::optional<Backend::TransactionsCursor> const&,
+        boost::asio::yield_context& yield)> transactionFetcher);
 
 }  // namespace RPC
 #endif
--- a/src/rpc/handlers/NFTHistory.cpp
+++ b/src/rpc/handlers/NFTHistory.cpp
@@ -3,24 +3,26 @@
 namespace RPC {
 
 Result
-doAccountTx(Context const& context)
+doNFTHistory(Context const& context)
 {
-    ripple::AccountID accountID;
-    if (auto const status = getAccount(context.params, accountID); status)
-        return status;
+    auto const maybeTokenID = getNFTID(context.params);
+    if (auto const status = std::get_if<Status>(&maybeTokenID); status)
+        return *status;
+    auto const tokenID = std::get<ripple::uint256>(maybeTokenID);
 
     constexpr std::string_view outerFuncName = __func__;
     auto const maybeResponse = traverseTransactions(
         context,
-        [&accountID, &outerFuncName](
+        [&tokenID, &outerFuncName](
             std::shared_ptr<Backend::BackendInterface const> const& backend,
             std::uint32_t const limit,
             bool const forward,
             std::optional<Backend::TransactionsCursor> const& cursorIn,
-            boost::asio::yield_context& yield) {
+            boost::asio::yield_context& yield)
+            -> Backend::TransactionsAndCursor {
             auto const start = std::chrono::system_clock::now();
-            auto const txnsAndCursor = backend->fetchAccountTransactions(
-                accountID, limit, forward, cursorIn, yield);
+            auto const txnsAndCursor = backend->fetchNFTTransactions(
+                tokenID, limit, forward, cursorIn, yield);
             BOOST_LOG_TRIVIAL(info)
                 << outerFuncName << " db fetch took "
                 << std::chrono::duration_cast<std::chrono::milliseconds>(
@@ -34,7 +36,7 @@ doAccountTx(Context const& context)
         return *status;
     auto response = std::get<boost::json::object>(maybeResponse);
 
-    response[JS(account)] = ripple::to_string(accountID);
+    response[JS(nft_id)] = ripple::to_string(tokenID);
     return response;
 }
 
--- a/src/rpc/handlers/NFTOffers.cpp
+++ b/src/rpc/handlers/NFTOffers.cpp
@@ -129,26 +129,10 @@ enumerateNFTOffers(
     return response;
 }
 
-std::variant<ripple::uint256, Status>
-getTokenid(boost::json::object const& request)
-{
-    if (!request.contains(JS(nft_id)))
-        return Status{Error::rpcINVALID_PARAMS, "missingTokenid"};
-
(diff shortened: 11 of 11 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+    if (!tokenid.parseHex(request.at(JS(nft_id)).as_string().c_str()))
+        return Status{Error::rpcINVALID_PARAMS, "malformedCursor"};
+
+    return tokenid;
+}
+
+// TODO - this function is long and shouldn't be responsible for as much as it
+// is. Split it out into some helper functions.
+std::variant<Status, boost::json::object>
+traverseTransactions(
```
ledhed2222, line 1508: copied (nearly) the entire body of the format `account_tx` implementation, but now the actual DB hit is done by the function passed as argument. fixed an issue with returned `ledger_index_min` and `ledger_index_max` values w/in here as well.

.gitignore
```diff
@@ -1,3 +1,4 @@
 *clio*.log
 build/
 .python-version
+config.json
```
natenichols, line 5: :+1:

src/rpc/RPC.cpp
```diff
@@ -206,6 +207,12 @@ class HandlerTable
 
         return handlerMap_[command].handler;
     }
+
+    bool
+    getIsClioOnly(std::string const& command)
```
natenichols: nit: I prefer 
```cpp
isClioOnly(std::string const& command)
```

`getIsClioOnly` sounds weird IMO.

src/rpc/RPC.cpp
```diff
@@ -286,6 +300,9 @@ shouldForwardToRippled(Context const& ctx)
 {
     auto request = ctx.params;
 
+    if (isClioOnly(ctx.method))
+        return false;
```
natenichols, line 305: Nice, this is a good addition.

src/rpc/handlers/NFTTx.cpp
```diff
@@ -0,0 +1,51 @@
...
+Result
+doNFTTx(Context const& context)
+{
+    auto const maybeTokenID = getNFTID(context.params);
+    if (auto const status = std::get_if<Status>(&maybeTokenID))
+        return *status;
+    auto const tokenID = std::get<ripple::uint256>(maybeTokenID);
+
+    auto const maybeResponse = traverseTransactions(
+        context,
```
natenichols: I don't love that we both capture context and pass context as a parameter.

It seems like (in my head) you don't actually need access to anything other than the backend. IMO you should also pass the yield_context as a parameter, not a capture. IMO that it is more consistent with how database reads are structured. 

You can do one of two things instead:

1) Just capture a reference to the backend. The context will stay in scope until `traverseTransactions()` returns:

```cpp
[&tokenID, &backend = *context.backend](
    std::uint32_t const limit,
    bool const forward,
    std::optional<Backend::TransactionsCursor>  const & in,
    boost::asio::yield_context& yield)
{
    .....
    backend.fetchNFTTransactions(...., yield);
    .....
}
```

2) or capture it as a shared_ptr, which is trivial to copy.

```cpp
[&tokenID, backend = context.backend] (
    std::uint32_t const limit,
    bool const forward,
    std::optional<Backend::TransactionsCursor>  const& in,
    boost::asio::yield_context& yield
)
{
    .....
    backend->fetchNFTTransactions(...., yield);
    .....
}
```

Either way works for me, and functionally there won't be much of a difference. 

then you'd call `getter(...., context.yield);`

src/rpc/handlers/NFTTx.cpp
```diff
@@ -0,0 +1,51 @@
+#include <rpc/RPCHelpers.h>
+
+// {
+//   nft_id: <ident>
+//   ledger_hash: <ledger>
+//   ledger_index: <ledger_index>
+//   ledger_index_min: <OPTIONAL -1 or ledger_index>
+//   ledger_index_max: <OPTIONAL -1 or ledger_index>
```
natenichols: While these comments are nice, I don't think they're needed. I would much prefer a more human-readable version over this JSON version that looks like its for some docgen script.

Unless we plan to setup docgen, which would also be nice.

src/rpc/handlers/NFTInfo.cpp
```diff
@@ -89,24 +89,15 @@ doNFTInfo(Context const& context)
...
-    auto v = ledgerInfoFromRequest(context);
-    if (auto status = std::get_if<Status>(&v))
+    auto const maybeTokenID = getNFTID(request);
+    if (auto const status = std::get_if<Status>(&maybeTokenID))
         return *status;
-    ripple::LedgerInfo lgrInfo = std::get<ripple::LedgerInfo>(v);
+    auto const tokenID = std::get<ripple::uint256>(maybeTokenID);
+
+    auto const maybeLedgerInfo = ledgerInfoFromRequest(context);
+    if (auto status = std::get_if<Status>(&maybeLedgerInfo))
```
natenichols: I know I haven't been doing this elsewhere, and is prolly out of scope, but I would like to change all of these to `cpp-17` style if statements

```cpp
if (auto status = std::get_if<Status>(&maybeLedgerInfo); status)
    return *status
```

src/rpc/handlers/AccountTx.cpp
```diff
@@ -2,211 +2,53 @@
...
-    auto [blobs, retCursor] = context.backend->fetchAccountTransactions(
-        accountID, limit, forward, cursor, context.yield);
-
-    auto end = std::chrono::system_clock::now();
-    BOOST_LOG_TRIVIAL(info) << __func__ << " db fetch took "
-                            << ((end - start).count() / 1000000000.0)
-                            << " num blobs = " << blobs.size();
+    auto const maybeResponse = traverseTransactions(
+        context,
+        [&accountID, &context](
```
natenichols: Same here, just capture `backend`

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+        {
+            obj[JS(meta)] = ripple::strHex(txnPlusMeta.metadata);
+            obj[JS(tx_blob)] = ripple::strHex(txnPlusMeta.transaction);
+            obj[JS(ledger_index)] = txnPlusMeta.ledgerSequence;
+            obj[JS(date)] = txnPlusMeta.date;
+        }
+        obj[JS(validated)] = true;
+
+        txns.push_back(obj);
+        if (!minReturnedIndex || txnPlusMeta.ledgerSequence < *minReturnedIndex)
```
natenichols: Do we use `minReturnedIndex` or `maxReturnedIndex` at all any more?

If not can we remove?

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+{
+    auto request = context.params;
+    boost::json::object response = {};
+
+    bool const binary = getBool(request, JS(binary), false);
+    bool const forward = getBool(request, JS(forward), false);
+
+    std::optional<Backend::TransactionsCursor> cursor;
+
+    if (request.contains(JS(marker)))
```
natenichols, line 1525: I'm just noticing this now, but can we parse out the marker _after_ we parse out `ledger_index_min/max`, and verify that the `marker.ledgerIndex` is in fact between `ledger_index_min` and `ledger_index_max`?

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+            cursor = {minIndex, 0};
+        else
+            cursor = {maxIndex, INT32_MAX};
+    }
+
+    std::uint32_t limit;
+    if (auto const status = getLimit(context, limit); status)
+        return status;
+
+    if (request.contains(JS(limit)))
```
natenichols, line 1631: I think we add this _only_ if `txns.size() == limit`.

Otherwise, don't include `response["limit"]`.

@cjcobb23 double check me? We should only add `response.limit` if we also add `response.marker`.

src/rpc/handlers/NFTTx.cpp
natenichols: This of course, is totally up for debate, and I'm more than willing to have a conversation about it.

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+
+// TODO - this function is long and shouldn't be responsible for as much as it
+// is. Split it out into some helper functions.
+std::variant<Status, boost::json::object>
+traverseTransactions(
+    Context const& context,
+    std::function<Backend::TransactionsAndCursor(
+        std::uint32_t const,
+        bool const,
+        std::optional<Backend::TransactionsCursor> const&)> getter)
```
natenichols: nit: I don't love the name `getter`.

Maybe `fetchTransactions`?

src/rpc/RPC.cpp
```diff
@@ -223,8 +230,9 @@ static HandlerTable handlerTable{
     {"ledger", &doLedger, {}},
     {"ledger_data", &doLedgerData, LimitRange{1, 100, 2048}},
     {"nft_buy_offers", &doNFTBuyOffers, LimitRange{1, 50, 100}},
-    {"nft_info", &doNFTInfo},
+    {"nft_info", &doNFTInfo, {}, true},
     {"nft_sell_offers", &doNFTSellOffers, LimitRange{1, 50, 100}},
+    {"nft_tx", &doNFTTx, LimitRange{1, 50, 100}, true},
```
natenichols: can we call this `nft_txs`, or `account_nft_tx` or something that implies that this will return _multiple_ txs.

When I read `nft_tx`, I think `tx`, which only returns one transaction.

natenichols (changes requested):
Awesome work duplicating code between `account_tx` and `nft_tx`.

Made a first pass at this PR. Something about the way `traverseTransactions` is structured gives me a slightly uneasy feeling, but I cant exactly put my finger on what it is.

Give me a day or two for it to think about it, and I might have more suggestions in a second volley.

As always, nice work.

src/rpc/RPC.cpp
ledhed2222: well the idea is to mimic `account_tx`, which gets you transactions related to one account. this is the same but for one nft :)

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+{
+    if (!request.contains(JS(nft_id)))
+        return Status{Error::rpcINVALID_PARAMS, "missingTokenid"};
+
+    if (!request.at(JS(nft_id)).is_string())
+        return Status{Error::rpcINVALID_PARAMS, "tokenidNotString"};
+
+    ripple::uint256 tokenid;
+    if (!tokenid.parseHex(request.at(JS(nft_id)).as_string().c_str()))
+        return Status{Error::rpcINVALID_PARAMS, "malformedCursor"};
```
cjcobb23: `malformedCursor` should be `malformedTokenID` or some such

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+
+    return tokenid;
+}
+
+// TODO - this function is long and shouldn't be responsible for as much as it
+// is. Split it out into some helper functions.
```
cjcobb23, line 1506: You could split this into pre-processing and post-processing

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+
+    response[JS(ledger_index_min)] = minIndex;
+    response[JS(ledger_index_max)] = maxIndex;
+
+    response[JS(transactions)] = txns;
+
+    auto serializationEnd = std::chrono::system_clock::now();
+    BOOST_LOG_TRIVIAL(info)
+        << __func__ << " serialization took "
+        << ((serializationEnd - dbFetchEnd).count() / 1000000000.0);
```
cjcobb23: Let's remove the hard coded number here, and change all of these to:
`std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()`

I think this is the syntax, but double check me.

src/rpc/handlers/AccountTx.cpp
```diff
@@ -2,211 +2,53 @@
...
+            bool const forward,
+            std::optional<Backend::TransactionsCursor> const& cursorIn) {
+            auto const start = std::chrono::system_clock::now();
+            auto const txnsAndCursor =
+                context.backend->fetchAccountTransactions(
+                    accountID, limit, forward, cursorIn, context.yield);
+            BOOST_LOG_TRIVIAL(info)
+                << __func__ << " db fetch took "
+                << ((std::chrono::system_clock::now() - start).count() /
+                    1000000000.0)
```
cjcobb23: Remove the hardcoded number. See my other comment

src/rpc/handlers/NFTTx.cpp
```diff
@@ -0,0 +1,51 @@
...
+            std::optional<Backend::TransactionsCursor> const& cursorIn)
+            -> Backend::TransactionsAndCursor {
+            auto const start = std::chrono::system_clock::now();
+            auto const txnsAndCursor = context.backend->fetchNFTTransactions(
+                tokenID, limit, forward, cursorIn, context.yield);
+            BOOST_LOG_TRIVIAL(info)
+                << __func__ << " db fetch took "
+                << ((std::chrono::system_clock::now() - start).count() /
+                    1000000000.0)
+                << " num blobs = " << txnsAndCursor.txns.size();
```
cjcobb23: Remove the hardcoded number. See my other comment

src/rpc/RPC.cpp
cjcobb23: I concur

cjcobb23 (commented):
Did a first pass, will follow up with more comments later

cjcobb23:
Do we have any unit tests for this at the database level?

src/rpc/RPC.cpp
cjcobb23: This is a tough one. I agree `nft_txs` is a better name, but it's also good to mimic `account_tx`. We could also just call it `nft_history` or something similar

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+    std::uint32_t limit;
+    if (auto const status = getLimit(context, limit); status)
+        return status;
+
+    if (request.contains(JS(limit)))
+        response[JS(limit)] = limit;
+
+    boost::json::array txns;
+    auto [blobs, retCursor] = getter(limit, forward, cursor);
+    auto dbFetchEnd = std::chrono::system_clock::now();
```
cjcobb23: I'd call this `serializationStart` or some such

src/rpc/RPCHelpers.cpp
cjcobb23, line 1525: You could probably split this into its own helper function

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+                boost::json::value_to<std::uint32_t>(obj.at(JS(ledger)));
+        }
+
+        if (!transactionIndex || !ledgerIndex)
+            return Status{Error::rpcINVALID_PARAMS, "missingLedgerOrSeq"};
+
+        cursor = {*ledgerIndex, *transactionIndex};
+    }
+
+    auto minIndex = context.range.minSequence;
```
cjcobb23, line 1558: You could probably split this min/max parsing into its own function

src/rpc/RPCHelpers.cpp
cjcobb23: I'd call it `dbFunc` or some such

src/rpc/handlers/AccountTx.cpp
```diff
@@ -2,211 +2,53 @@
 #include <backend/Pg.h>
 #include <rpc/RPCHelpers.h>
 
-namespace RPC {
+// {
```
undertome: It would be nice to have something explaining what this comment is describing

src/rpc/handlers/AccountTx.cpp
```diff
@@ -2,211 +2,53 @@
...
+            std::uint32_t const limit,
+            bool const forward,
+            std::optional<Backend::TransactionsCursor> const& cursorIn) {
+            auto const start = std::chrono::system_clock::now();
+            auto const txnsAndCursor =
+                context.backend->fetchAccountTransactions(
+                    accountID, limit, forward, cursorIn, context.yield);
+            BOOST_LOG_TRIVIAL(info)
+                << __func__ << " db fetch took "
+                << ((std::chrono::system_clock::now() - start).count() /
```
undertome: This will say "db fetch took xxx" - let's add a unit of time for clarity.

src/rpc/handlers/NFTTx.cpp
```diff
@@ -0,0 +1,51 @@
...
+            std::uint32_t const limit,
+            bool const forward,
+            std::optional<Backend::TransactionsCursor> const& cursorIn)
+            -> Backend::TransactionsAndCursor {
+            auto const start = std::chrono::system_clock::now();
+            auto const txnsAndCursor = context.backend->fetchNFTTransactions(
+                tokenID, limit, forward, cursorIn, context.yield);
+            BOOST_LOG_TRIVIAL(info)
+                << __func__ << " db fetch took "
+                << ((std::chrono::system_clock::now() - start).count() /
```
undertome: This will say "db fetch took xxx" - let's add a unit of time for clarity.

CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ target_sources(clio PRIVATE
   src/rpc/handlers/NoRippleCheck.cpp
   # NFT
   src/rpc/handlers/NFTInfo.cpp
+  src/rpc/handlers/NFTTx.cpp
+  src/rpc/handlers/NFTOffers.cpp
```
undertome, line 80: I think the standard is to order these alphabetically (though clio doesn't always uphold that standard); you could at least rearrange the NFT section.

src/backend/CassandraBackend.cpp
```diff
@@ -651,6 +651,8 @@ CassandraBackend::fetchNFTTransactions(
                 static_cast<std::uint32_t>(lgrSeq),
                 static_cast<std::uint32_t>(txnIdx)};
 
+            // Only modify if forward because forward query is >= seq_idx,
```
undertome: This comment might be confusing or unclear to someone new to the repo. There isn't even a `seq_idx` parameter in scope. Please rewrite this to be more helpful.

src/rpc/RPCHelpers.cpp
```diff
@@ -1478,4 +1478,210 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+    boost::json::object response = {};
+
+    bool const binary = getBool(request, JS(binary), false);
+    bool const forward = getBool(request, JS(forward), false);
+
+    std::optional<Backend::TransactionsCursor> cursor;
+
+    if (request.contains(JS(marker)))
+    {
+        auto const& obj = request.at(JS(marker)).as_object();
```
undertome, line 1529: `as_object()` can throw - is the exception handled gracefully in each case?

src/rpc/RPCHelpers.cpp
ledhed2222: oh nice catch

src/rpc/RPCHelpers.cpp
ledhed2222: totally - there's a lot of room to improve this. this is mostly just the entirety of the old account_tx driver. what do you think about leaving such a refactor for later? I want to keep the footprint of this PR small :)

src/rpc/RPCHelpers.cpp
ledhed2222: i just tested this. the limit is returned on `account_tx` as the user input it, even if:

a) the limit wasn't respected (like with a limit of 0)
b) the limit was larger than the total number of transactions (so there is no marker)

In other words - this is correct :)

ledhed2222: FWIW - here's the examples:

this one is with a limit of 0 (IE when the limit is ignore). Scenario a) from above:

```json
{
  "id": 2,
  "result": {
    "account": "rV3WAvwwXgvPrYiUgSoytn9w3mejtPgLo",
    "ledger_index_max": 73786431,
    "ledger_index_min": 32570,
    "limit": 0,
    "transactions": [
      {
        "ledger_index": 71887087,
        "meta": "201C00000035F8E5110061250448E8EB553F81263D63398D1B7F6F0B28EE1C04EFD8FC9DF056B4A34926A13F31EA1295905643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E62403FC738962400000000EF5F4A8E1E722000000002403FC738A2D0000000041BEAE1D4E275FCB5135999224BBD4DCB062400000000EF5F49C7710677265677765697362726F642E636F6D81140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000322800000002403FC7389201B0448E91868400000000000000C7321033E6BA2570EE7649CAD27CC0E66FD67B0EDB0BF7100642F3EA34A116952DDDEFE7447304502210087D945474A2E28DA27372B3ED7351ACA313F2883D9CAA54A4814C667E353977A02200834243B3BD97520B1D38FE0A07D60823D508CF1D2554FDCA23B91BE297278107710677265677765697362726F642E636F6D81140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 71887083,
        "meta": "201C00000035F8E51100612504075E8B555B1266AFC2B283845272D90760E73580AB52BFE3F8B5D7CB687A6DC26D67F6745643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E62403FC738862400000000EF5F4B4E1E722000000002403FC73892D0000000041BEAE1D4E275FCB5135999224BBD4DCB062400000000EF5F4A881140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000322800000002403FC7388201B0448E91341BEAE1D4E275FCB5135999224BBD4DCB068400000000000000C7321033E6BA2570EE7649CAD27CC0E66FD67B0EDB0BF7100642F3EA34A116952DDDEFE74463044022065114D8C5D71BC9946086C89A5E8C7DCA676600639322FB10B79D5F59BC6C87C022038B4E5996CE59C6BB240609AEA855DFD8C8C8146CEDAB9B48BFCDBB5B32B4AB681140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 67591819,
        "meta": "201C00000006F8E51100612503FC7435559616B70A793DBD8CB0C0D9BDFEF44C3BCDD670378C7F590729158AC7162FD0445643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E6220001000062400000000EE6B274E1E722000000002403FC73882D0000000062400000000EF5F4B481140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1E5110061250404B88A5503798C9E683C6D2495A64CFD05A7514A1511CB119F79739FD16C6E704B4E0AB256812A475FC6783D6274DC513A8DB4F7CD4E1DB9E1211C28D04FA7BA50F0D9433FE62403D0252962400000012A05F1CDE1E722000000002403D0252A2D00000001624000000129F6AF8181146635422CE87C6A39B6E9059B49D7E427CE9745E9E1E1F1031000",
        "tx_blob": "12000022800000002403D025296140000000000F424068400000000000000C7321021398DFE916A7E389E9E1DA2BD1B9B4CEA61AC39291247DB0C0B365A97059A47574473045022100FEE7FB433F9082467EC5100DEA2435BABEDB7782509642EA88B2558C77D9D91D0220368230CADA239DC6C3B612A480D8CB10B26A34AB2AF884134E26A9B076F1C41381146635422CE87C6A39B6E9059B49D7E427CE9745E983140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 66876469,
        "meta": "201C00000017F8E51100612503FC73875517B0879CCB323DB10895A884117B27FE8DB17367ED6B69871E69D3340B866DF75643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E622000000002403FC738762400000000EE6B280E1E722000100002403FC73882D0000000062400000000EE6B27481140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000522800000002403FC7387201B03FC745E68400000000000000C732103AA5DBD640350A5E922406393CBDA79CFBFAD1A708B2FB80F4547A6EBBB13FDCD74473045022100EC66D5388C7ECB185B770CE89B6169DE9A0463A05FC447EC643D48918E72017802200CCC26C08CB85D9CCF21CC13A597458EE02911FD4535407D1E2BD1B0B6BFD89081140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01B",
        "validated": true
      },
      {
        "ledger_index": 66876295,
        "meta": "201C00000023F8E51100612503FC736055F9521DEBA37013ADDDC209B39B7A7796EBFAABD9E3D3CD420828A33E60B75FA9561B8A0A4C8456AE1C3B1384772E503F3EFC97E6E64D749AC2907963C4D3924FEBE624000000066240000007BEB99B36E1E7220000000024000000072D000000006240000007AFD2E8AC722102000000000000000000000000A5D8B15C88EF95C0016E0AA8FA0FE6DCA01A12EA8114BCF5F32E42CF502E79E75BAC7DE5E27D4EA48DC1E1E1E31100615643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E82403FC738762400000000EE6B28081140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AEE1E1F1031000",
        "tx_blob": "12000022800000002400000006201B03FC738F61400000000EE6B28068400000000000000A7321028FBB8E2C06E43484F8D393CFAAF8B8170C68F7AED21D70B093C1BDF34AE3364574463044022071F621F61B071CDB3F8C087BB490373CA7F94CE399E08306BEA8644AD614EE6C0220601BBF36222ED23F68526007C0690AD23146D874632795CAD3070457EF70602A8114BCF5F32E42CF502E79E75BAC7DE5E27D4EA48DC183140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      }
    ],
    "used_postgres": true,
    "validated": true,
    "warnings": [
      {
        "id": 1004,
        "message": "This is a reporting server.  The default behavior of a reporting server is to only return validated data. If you are looking for not yet validated data, include \"ledger_index : current\" in your request, which will cause this server to forward the request to a p2p node. If the forward is successful the response will include \"forwarded\" : \"true\""
      }
    ]
  },
  "status": "success",
  "type": "response"
}
```

ledhed2222: and here's scenario b) 

```json
{
  "id": 2,
  "result": {
    "account": "rV3WAvwwXgvPrYiUgSoytn9w3mejtPgLo",
    "ledger_index_max": 73786425,
    "ledger_index_min": 32570,
    "limit": 2000000000,
    "transactions": [
      {
        "ledger_index": 71887087,
        "meta": "201C00000035F8E5110061250448E8EB553F81263D63398D1B7F6F0B28EE1C04EFD8FC9DF056B4A34926A13F31EA1295905643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E62403FC738962400000000EF5F4A8E1E722000000002403FC738A2D0000000041BEAE1D4E275FCB5135999224BBD4DCB062400000000EF5F49C7710677265677765697362726F642E636F6D81140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000322800000002403FC7389201B0448E91868400000000000000C7321033E6BA2570EE7649CAD27CC0E66FD67B0EDB0BF7100642F3EA34A116952DDDEFE7447304502210087D945474A2E28DA27372B3ED7351ACA313F2883D9CAA54A4814C667E353977A02200834243B3BD97520B1D38FE0A07D60823D508CF1D2554FDCA23B91BE297278107710677265677765697362726F642E636F6D81140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 71887083,
        "meta": "201C00000035F8E51100612504075E8B555B1266AFC2B283845272D90760E73580AB52BFE3F8B5D7CB687A6DC26D67F6745643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E62403FC738862400000000EF5F4B4E1E722000000002403FC73892D0000000041BEAE1D4E275FCB5135999224BBD4DCB062400000000EF5F4A881140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000322800000002403FC7388201B0448E91341BEAE1D4E275FCB5135999224BBD4DCB068400000000000000C7321033E6BA2570EE7649CAD27CC0E66FD67B0EDB0BF7100642F3EA34A116952DDDEFE74463044022065114D8C5D71BC9946086C89A5E8C7DCA676600639322FB10B79D5F59BC6C87C022038B4E5996CE59C6BB240609AEA855DFD8C8C8146CEDAB9B48BFCDBB5B32B4AB681140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 67591819,
        "meta": "201C00000006F8E51100612503FC7435559616B70A793DBD8CB0C0D9BDFEF44C3BCDD670378C7F590729158AC7162FD0445643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E6220001000062400000000EE6B274E1E722000000002403FC73882D0000000062400000000EF5F4B481140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1E5110061250404B88A5503798C9E683C6D2495A64CFD05A7514A1511CB119F79739FD16C6E704B4E0AB256812A475FC6783D6274DC513A8DB4F7CD4E1DB9E1211C28D04FA7BA50F0D9433FE62403D0252962400000012A05F1CDE1E722000000002403D0252A2D00000001624000000129F6AF8181146635422CE87C6A39B6E9059B49D7E427CE9745E9E1E1F1031000",
        "tx_blob": "12000022800000002403D025296140000000000F424068400000000000000C7321021398DFE916A7E389E9E1DA2BD1B9B4CEA61AC39291247DB0C0B365A97059A47574473045022100FEE7FB433F9082467EC5100DEA2435BABEDB7782509642EA88B2558C77D9D91D0220368230CADA239DC6C3B612A480D8CB10B26A34AB2AF884134E26A9B076F1C41381146635422CE87C6A39B6E9059B49D7E427CE9745E983140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      },
      {
        "ledger_index": 66876469,
        "meta": "201C00000017F8E51100612503FC73875517B0879CCB323DB10895A884117B27FE8DB17367ED6B69871E69D3340B866DF75643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E622000000002403FC738762400000000EE6B280E1E722000100002403FC73882D0000000062400000000EE6B27481140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1F1031000",
        "tx_blob": "12000522800000002403FC7387201B03FC745E68400000000000000C732103AA5DBD640350A5E922406393CBDA79CFBFAD1A708B2FB80F4547A6EBBB13FDCD74473045022100EC66D5388C7ECB185B770CE89B6169DE9A0463A05FC447EC643D48918E72017802200CCC26C08CB85D9CCF21CC13A597458EE02911FD4535407D1E2BD1B0B6BFD89081140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01B",
        "validated": true
      },
      {
        "ledger_index": 66876295,
        "meta": "201C00000023F8E51100612503FC736055F9521DEBA37013ADDDC209B39B7A7796EBFAABD9E3D3CD420828A33E60B75FA9561B8A0A4C8456AE1C3B1384772E503F3EFC97E6E64D749AC2907963C4D3924FEBE624000000066240000007BEB99B36E1E7220000000024000000072D000000006240000007AFD2E8AC722102000000000000000000000000A5D8B15C88EF95C0016E0AA8FA0FE6DCA01A12EA8114BCF5F32E42CF502E79E75BAC7DE5E27D4EA48DC1E1E1E31100615643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E82403FC738762400000000EE6B28081140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AEE1E1F1031000",
        "tx_blob": "12000022800000002400000006201B03FC738F61400000000EE6B28068400000000000000A7321028FBB8E2C06E43484F8D393CFAAF8B8170C68F7AED21D70B093C1BDF34AE3364574463044022071F621F61B071CDB3F8C087BB490373CA7F94CE399E08306BEA8644AD614EE6C0220601BBF36222ED23F68526007C0690AD23146D874632795CAD3070457EF70602A8114BCF5F32E42CF502E79E75BAC7DE5E27D4EA48DC183140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      }
    ],
    "used_postgres": true,
    "validated": true,
    "warnings": [
      {
        "id": 1004,
        "message": "This is a reporting server.  The default behavior of a reporting server is to only return validated data. If you are looking for not yet validated data, include \"ledger_index : current\" in your request, which will cause this server to forward the request to a p2p node. If the forward is successful the response will include \"forwarded\" : \"true\""
      }
    ]
  },
  "status": "success",
  "type": "response"
}
```

ledhed2222: FWIW these both apparently came from a reporting server. Would that make a difference here? I presume not.

src/rpc/handlers/NFTInfo.cpp
ledhed2222: ah no - this is new code let's do it.

src/rpc/RPCHelpers.cpp
cjcobb23: docs say `limit` is always present, and is just the value from the request:
https://xrpl.org/account_tx.html#response-format

src/rpc/RPCHelpers.cpp
ledhed2222: super interestingly (and not correctly IMO) the current implementation that I'm seeing for `account_tx` will ignore `ledger_index_min` and `ledger_index_max` if there is a `marker`. Here's proof:

Request:

```json
{
  "id": 2,
  "command": "account_tx",
  "account": "rV3WAvwwXgvPrYiUgSoytn9w3mejtPgLo",
  "ledger_index_min": 67591817,
  "ledger_index_max": 67591818,
  "binary": true,
  "limit": 2,
  "forward": false,
  "marker": {
    "ledger": 67591819,
    "seq": 6
   }
}
```

And response:

```json
{
  "id": 2,
  "result": {
    "account": "rV3WAvwwXgvPrYiUgSoytn9w3mejtPgLo",
    "ledger_index_max": 67591818,
    "ledger_index_min": 67591817,
    "limit": 2,
    "transactions": [
      {
        "ledger_index": 67591819,
        "meta": "201C00000006F8E51100612503FC7435559616B70A793DBD8CB0C0D9BDFEF44C3BCDD670378C7F590729158AC7162FD0445643815338DA88D7A8B068D1BCE5AD605B49139FCA0987CE7E99BD365E79F9BA58E6220001000062400000000EE6B274E1E722000000002403FC73882D0000000062400000000EF5F4B481140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE8814B33434F7BA42002CEDC9B52C257C71D2AAC9D01BE1E1E5110061250404B88A5503798C9E683C6D2495A64CFD05A7514A1511CB119F79739FD16C6E704B4E0AB256812A475FC6783D6274DC513A8DB4F7CD4E1DB9E1211C28D04FA7BA50F0D9433FE62403D0252962400000012A05F1CDE1E722000000002403D0252A2D00000001624000000129F6AF8181146635422CE87C6A39B6E9059B49D7E427CE9745E9E1E1F1031000",
        "tx_blob": "12000022800000002403D025296140000000000F424068400000000000000C7321021398DFE916A7E389E9E1DA2BD1B9B4CEA61AC39291247DB0C0B365A97059A47574473045022100FEE7FB433F9082467EC5100DEA2435BABEDB7782509642EA88B2558C77D9D91D0220368230CADA239DC6C3B612A480D8CB10B26A34AB2AF884134E26A9B076F1C41381146635422CE87C6A39B6E9059B49D7E427CE9745E983140551EBD684BF2ADE0EF093A92B6E2C55D15BD9AE",
        "validated": true
      }
    ],
    "used_postgres": true,
    "validated": true,
    "warnings": [
      {
        "id": 1004,
        "message": "This is a reporting server.  The default behavior of a reporting server is to only return validated data. If you are looking for not yet validated data, include \"ledger_index : current\" in your request, which will cause this server to forward the request to a p2p node. If the forward is successful the response will include \"forwarded\" : \"true\""
      }
    ]
  },
  "status": "success",
  "type": "response"
}
```

again this is a reporting server, and i'm not sure if there is a difference there.

if this all looks right, we could not even parse the `ledger_index_min` and `ledger_index_max` if there is a `marker`. I think that's the way.

ledhed2222: FWIW - I agree that if the marker and the index ranges don't align it should error. But I think we should strive to be exactly equal to rippled here.

undertome (dismissed):

ledhed2222:
> Do we have any unit tests for this at the database level?

Yup - that was included in the prior work for `nft_info` 🤘

src/rpc/RPCHelpers.cpp
ledhed2222: nice call - it is not (this was existing code but going to fix it)

ledhed2222:
NOTE - commit history is going to get full of "addressed comments" commits. I'll squash it all down prior to merge.

src/rpc/RPCHelpers.cpp
ledhed2222: so, i've fixed this. but interestingly i just checked what rippled does in this case and it hard errors just like we did - IE it's just an "internalError". That's obviously not right but @cjcobb23 to what extent should we be _exactly_ like rippled here?

ledhed2222: my $0.02 is that this is an example of where clio should be different from rippled. the shape of the happy path response is the same, but we are improving errors. rippled should later have a project to incorporate these improvements there.

src/rpc/handlers/AccountTx.cpp
ledhed2222: removed the comment

cjcobb23:
@ledhed2222 can you share your testing plan for this? I assume its some form of running this on nft devnet, since we have no analogous method in ripple to compare the output too. Also, we do need some unit tests for this, testing the database functions themselves.

cjcobb23:
>     * Fixes an issue where the returned `ledger_index_min` and `ledger_index_max` values for `account_tx` were not correct
Can you explain more what bug was fixed?

src/rpc/RPCHelpers.cpp
cjcobb23: We should avoid `internalError` at all costs, and I don't think it's documented that a non-object marker results in `internalError`. We actually are striving to adhere to the documented API exactly, as opposed to rippled itself. So we have some freedom here, since I don't think documentation says anything about non-object markers. I think we should improve the errors and return something like `markerNotObject`

cjcobb23:
> > Do we have any unit tests for this at the database level?
> 
> Yup - that was included in the prior work for `nft_info` metal

ok great, disregard my comment about this elsewhere

src/rpc/RPCHelpers.cpp
cjcobb23: It's up to you whether you want to refactor now or later. This function is a bit hairy and the hairiness often leads to bugs, especially since we don't have unit tests for the handlers. If we refactor into helpers, we could unit test the helpers. But up to you, will let you decide.

ledhed2222:
NOTE - test existing behavior of account_tx WRT `ledger_index_max/min` on a standalone node to confirm it's the same as w reporting mode

src/rpc/RPCHelpers.cpp
ledhed2222: thinking more - shouldn't error actually. you want to be able to not change your input when iterating w/ a marker. it's just that the docs are confusing, and perhaps the returned keys. "parsed_ledger_index_min` could have been a superior name

src/rpc/RPC.cpp
natenichols: Up to y'all on this one.

src/rpc/handlers/AccountTx.cpp
```diff
@@ -1,212 +1,41 @@
...
-        accountID, limit, forward, cursor, context.yield);
-
-    auto end = std::chrono::system_clock::now();
-    BOOST_LOG_TRIVIAL(info) << __func__ << " db fetch took "
-                            << ((end - start).count() / 1000000000.0)
-                            << " num blobs = " << blobs.size();
+    constexpr std::string_view outerFuncName = __func__;
+    auto const maybeResponse = traverseTransactions(
+        context,
+        [&accountID, &outerFuncName](
```
natenichols, line 15: Nice, I like this a lot more than the previous iteration.

natenichols (dismissed):
Nice work @ledhed2222! All of my comments have been addressed and `traverseTransactions` is much cleaner.

up to y'all about the name of the RPC, I don't have strong feelings.

cjcobb23:
think im leaning towards `nft_txs` for the name. `nft_tx` is a bit too confusing, does seem like it should return a single transaction about an nft. or maybe `nft_history` or `nft_tx_history`. `account_tx` really should be `account_txs` or `account_tx_history`.

godexsoft (approved):

PR #256, opened 2022-08-19
Dockerfiles and push packages to GitHub's "Releases" page

legleux:
This adds a final step to push the packages to Github's Releases page when a release is published.
This requires a review from people (currently @cjcobb23, @legleux @natenichols) in the ["publishToRelease"](https://github.com/XRPLF/clio/settings/environments/596989701/edit) environment. If that's too much, it can just do it automatically. 

Added a Dockerfile that builds Clio, a Docker compose file that sets up a network with Clio, rippled and Cassandra running in concert.
Theses images are built and pushed to Docker Hub and ghcr.io on release also.


legleux:
Publishing the release itself could also be automated, with the release notes gleaned from a `RELEASENOTES.md` file or something committed with the version bump.
One pain point is if you make a release where the tag points to multiple branches, the other branches images need to manually be uploaded. I'll fix that

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,13 @@
+runs:
+  using: composite
+  steps:
+    # Github's ubuntu-20.04 image already has clang-format-11 installed
+    - run: |
+        find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format-11 -i
```
godexsoft: Maybe we should use `clang-format` alias instead, WDYT? I doubt anyone would have anything older than `v11` either way and most will have `clang-format-14` on a modern system.

docker/clio_docker/centos/install_cmake.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
godexsoft: You can use the `/usr/bin/env` trick here too or just use `/bin/bash` everywhere but I'd rather not mix the two

docker/ubuntu/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+FROM ubuntu:20.04 AS build
+ENV BOOST_ROOT=/boost
+COPY --from=boost /boost /boost
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install --no-install-recommends -y build-essential software-properties-common pkg-config libssl-dev wget curl gpg git zlib1g-dev bison flex autoconf lsb-release
+RUN apt install -y gpg-agent
+RUN wget https://apt.llvm.org/llvm.sh
+RUN chmod +x llvm.sh && ./llvm.sh 14 && ./llvm.sh 15
+# COPY . /clio
```
godexsoft: if not needed - let's remove this line?

godexsoft (dismissed):
Leaving a couple of suggestions. Overall it would not hurt to remove some more commented lines that are not needed anymore unless you want to keep them for some future effort you have in mind. 
Otherwise, LGTM!

.dockerignore
```diff
@@ -1 +1,5 @@
+<<<<<<< HEAD
```
godexsoft: This looks like part of a diff that accidentally got committed?

.github/workflows/build.yml
```diff
@@ -67,6 +68,64 @@ jobs:
...
+      id: boost
+      uses: actions/cache@v3
+      with:
+        path: boost
+        key: ${{ runner.os }}-boost
+
+    - name: Build boost
+      if: steps.boost.outputs.cache-hit != 'true'
+      run: |
+        curl -s -OJL "https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/boost_1_77_0.tar.gz"
```
godexsoft: Are we for sure compatible with 1.77? I know we are not with 1.80 but did not test 1.77 myself.

CMake/docs.cmake
```diff
@@ -0,0 +1,22 @@
+option(BUILD_DOC "Build documentation" ON)
+
+# check if Doxygen is installed
+find_package(Doxygen)
```
godexsoft, line 5: We have #421 for this, please coordinate there that you are implementing it here

docker/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# ENV DEBIAN_FRONTEND=noninteractive
+
+# RUN apt-get update && apt-get install -y build-essential software-properties-common pkg-config libssl-dev curl gpg git zlib1g-dev  bison flex autoconf
+RUN apt-get update && apt-get install -y build-essential
+# COPY install_compilers.sh .
+# # ARG GCC_VERSION=11
+# RUN ./install_compilers.sh gcc-11
+
+ARG BOOST_VERSION_=1_75_0
+ARG BOOST_VERSION=1.75.0
```
godexsoft: Here it's boost 1.75, why is it 1.77 in the other one?

PR #281, opened 2022-08-24
Fix bug causing some ledgers to not be published

cjcobb23:
close time in the ledger header is not accurate to the second. Sometimes, the close time is a second or two in the future, and so the age was being calculated as negative, which converts to a very large unsigned integer. This caused us to skip publishing ledgers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/281)


cjcobb23:
We don't publish ledgers that are over 10 minutes old

natenichols (approved):

Issue #195, opened 2022-06-27
account_tx returns `marker` after all transactions have been received


mukulljangid:
{account=r9m6MwViR4GnUNqoGXGa8eroBrZ9FAPHFS, ledgerIndexMinimum=61400000, ledgerIndexMaximum=61437000, binary=false, forward=false}
lengths:
50
50
50
50
50
34
0
Clio returned marker in the response with 34 transactions, hence the last one has 0.

sappenin:
Just for clarity here on this issue, Clio is returning a `Marker` even when there are no more results to be returned. Reporting mode does _not_ do this, and [per the docs](https://xrpl.org/markers-and-pagination.html#markers-and-pagination) the marker should be `null` when the # of transactions is 0.

PR #282, opened 2022-08-24
bugfix/account_tx-returns-marker

officialfrancismendoza:
**Issue** (#195): account_tx API method returns marker with "0" when paging has been exhausted
**Fix**: After AccountTx.cpp functionality was moved to RPCHelpers.cpp from develop, core logic on paging, skipping transactions, and ingress/egress paging direction based on window of transactions was modified to exit when remaining transactions < limit

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1646,8 +1646,13 @@ traverseTransactions(
 
     for (auto const& txnPlusMeta : blobs)
     {
-        if (txnPlusMeta.ledgerSequence < minIndex ||
-            txnPlusMeta.ledgerSequence > maxIndex)
+        if ((txnPlusMeta.ledgerSequence < minIndex && !forward) ||
+            (txnPlusMeta.ledgerSequence > maxIndex && forward))
+        {
+            response.erase(JS(marker));
+            break;
+        }
+        else if (txnPlusMeta.ledgerSequence > maxIndex && !forward)
         {
             BOOST_LOG_TRIVIAL(debug)
                 << __func__
@@ -1674,13 +1679,11 @@ traverseTransactions(
             obj[JS(date)] = txnPlusMeta.date;
         }
         obj[JS(validated)] = true;
-
         txns.push_back(obj);
     }
 
     response[JS(ledger_index_min)] = minIndex;
     response[JS(ledger_index_max)] = maxIndex;
-
     response[JS(transactions)] = txns;
 
     BOOST_LOG_TRIVIAL(info)


cjcobb23 (commented):
Does this actually fix the issue? Issue was a cursor was returned when there were actually no more transactions to page through. But it doesn't look like we've changed the logic of when a cursor is returned. We still return a cursor on line 148, and then proceed to possibly filter out some of the transactions. How does this fix the issue?

src/rpc/handlers/AccountTx.cpp
```diff
@@ -152,8 +152,15 @@ doAccountTx(Context const& context)
     std::optional<size_t> minReturnedIndex;
     for (auto const& txnPlusMeta : blobs)
     {
-        if (txnPlusMeta.ledgerSequence < minIndex ||
-            txnPlusMeta.ledgerSequence > maxIndex)
+        if (txnPlusMeta.ledgerSequence < minIndex && !forward)
```
natenichols: I think this is wrong. We want to continue if this is entirely out of range. 

I think the change we need to make to fix this is flagging if the returned transactions are out of the range of `minIndex/maxIndex`. 

I'm thinking this will break w/ `forward:true`.

natenichols (dismissed):

officialfrancismendoza:
Closed as there is a new PR for this issue

cjcobb23:
Please continue using this PR, so we don't lose track of the comment history. No need to make a new PR for the same issue.

src/rpc/handlers/AccountTx.cpp
```diff
@@ -180,12 +191,16 @@ doAccountTx(Context const& context)
             obj[JS(date)] = txnPlusMeta.date;
         }
         obj[JS(validated)] = true;
-
         txns.push_back(obj);
-        if (!minReturnedIndex || txnPlusMeta.ledgerSequence < *minReturnedIndex)
+
+        if (!minReturnedIndex)
```
cjcobb23: I don't think this change is correct. We need to make sure minReturnedIndex is the min we see during the entire iteration. We need to check it every time, to see if it should be updated.

cjcobb23 (changes requested):

src/rpc/handlers/AccountTx.cpp
officialfrancismendoza: @cjcobb23 will change it, but do we want similar logic for L202 for maxReturnedIndex (basically restoring txnPlusMeta.ledgerSequence > *maxReturnedIndex)

src/rpc/handlers/AccountTx.cpp
```diff
@@ -180,12 +191,16 @@ doAccountTx(Context const& context)
...
         txns.push_back(obj);
+
         if (!minReturnedIndex || txnPlusMeta.ledgerSequence < *minReturnedIndex)
+        {
             minReturnedIndex = txnPlusMeta.ledgerSequence;
+        }
         if (!maxReturnedIndex || txnPlusMeta.ledgerSequence > *maxReturnedIndex)
+        {
             maxReturnedIndex = txnPlusMeta.ledgerSequence;
+        }
```
cjcobb23: Can we remove the curly braces here? They are not needed, since the body is only 1 line

src/rpc/handlers/AccountTx.cpp
```diff
@@ -152,8 +152,19 @@ doAccountTx(Context const& context)
...
+        {
+            response.erase(JS(marker));
+            break;
+        }
+
+        if (txnPlusMeta.ledgerSequence > maxIndex && forward)
+        {
+            response.erase(JS(marker));
+            break;
+        }
```
cjcobb23: We could combine this logic into one if statement, since the body of each is the same

ledhed2222:
@officialfrancismendoza would you describe the original issue in the description please?

officialfrancismendoza:
@cjcobb23 build and test successful, Note that as per prior conversations with @natenichols, "undefined" instead of "0" is what we want. Modified RPCHelpers.cpp from develop to have equivalent functionality as initial modified file from first rebase failed to build

If it looks good to you, will squash first so you can merge to develop

image

src/rpc/RPCHelpers.cpp
```diff
@@ -1674,23 +1683,38 @@ traverseTransactions(
...
+        if (!forward)
+        {
+            response[JS(ledger_index_min)] = cursor->ledgerSequence;
+            response[JS(ledger_index_max)] = maxIndex;
+        }
+        else
+        {
+            response[JS(ledger_index_max)] = cursor->ledgerSequence;
+            response[JS(ledger_index_min)] = minIndex;
+        }
```
cjcobb23: Since all of this was removed from develop, I would just remove it here as well

src/rpc/RPCHelpers.cpp
```diff
@@ -1644,10 +1644,19 @@ traverseTransactions(
         response[JS(marker)] = cursorJson;
     }
 
+    std::optional<size_t> maxReturnedIndex;
+    std::optional<size_t> minReturnedIndex;
```
cjcobb23: I don't think these are being used. We can remove them.

src/rpc/RPCHelpers.cpp
```diff
@@ -1674,23 +1683,38 @@ traverseTransactions(
...
-               .count()
-        << " milliseconds";
+        response[JS(transactions)] = txns;
+
+        BOOST_LOG_TRIVIAL(info)
+            << __func__ << " serialization took "
+            << std::chrono::duration_cast<std::chrono::milliseconds>(
+                   std::chrono::system_clock::now() - serializationStart)
+                   .count()
+            << " milliseconds";
```
cjcobb23: This is supposed to happen outside of the loop

src/rpc/RPCHelpers.cpp
officialfrancismendoza: @cjcobb23 I should've elaborated further: those lines did not exist in RPCHelpers.cpp but were a part of the working AccountTx.cpp you approved in a prior PR before the merge conflict. (remember, as we moved any functionality from AccountTx.cpp into there). 

To replicate the same functionality based on the changes in develop, had to move these new lines into the loop.

officialfrancismendoza: @cjcobb23 Update: Tested with suggested changes and successfully replicated same result 
image

cjcobb23 (dismissed):
LGTM, nice job!

src/rpc/RPCHelpers.cpp
```diff
@@ -1674,13 +1680,9 @@ traverseTransactions(
             obj[JS(date)] = txnPlusMeta.date;
         }
         obj[JS(validated)] = true;
-
         txns.push_back(obj);
     }
 
-    response[JS(ledger_index_min)] = minIndex;
-    response[JS(ledger_index_max)] = maxIndex;
```
cjcobb23: Why did we remove this? Sorry I didn't flag this before.

cjcobb23 (changes requested):

src/rpc/RPCHelpers.cpp
officialfrancismendoza: @cjcobb23 mistakenly removed when accounting for changes regarding deletions for maxReturnedIndex and minReturnedIndex. Restored in latest push, successful result
image

cjcobb23 (dismissed):

src/rpc/RPCHelpers.cpp
```diff
@@ -1690,7 +1694,8 @@ traverseTransactions(
                .count()
         << " milliseconds";
 
+    /*! @brief Put return response out of for-loop */
```
cjcobb23: Remove this comment please

cjcobb23 (changes requested):
Please squash to one commit, no merge commits

cjcobb23 (dismissed):

src/rpc/RPCHelpers.cpp
```diff
@@ -1646,8 +1646,14 @@ traverseTransactions(
...
     {
-        if (txnPlusMeta.ledgerSequence < minIndex ||
-            txnPlusMeta.ledgerSequence > maxIndex)
+        if ((txnPlusMeta.ledgerSequence < minIndex && !forward) ||
+            (txnPlusMeta.ledgerSequence > maxIndex && forward))
+        {
+            response.erase(JS(marker));
+            break;
+        }
+
```
godexsoft: Please remove empty line here

src/rpc/RPCHelpers.cpp
```diff
@@ -1693,4 +1697,4 @@ traverseTransactions(
     return response;
 }
 
-}  // namespace RPC
+}  // namespace RPC
```
godexsoft, line 1699: You may want to mark the checkbox for `Render Final Newline` and `Insert Final Newline` and uncheck `Trim Final Newlines` in VSCode. This way you won't be getting these complaints from github for missing newline at end of file.

godexsoft (dismissed):

cjcobb23 (approved):
LGTM

Issue #251, opened 2022-08-16
Don't require ssl certs in config to connect to rippled via ssl


cjcobb23:
Currently, clio needs to be passed an SSL keyfile and certfile to connect to rippled via SSL. This is unnecessary and should be changed, because clio doesn't actually use those certs when talking to rippled as a client.

PR #292, opened 2022-08-31
Probing ETL Source

godexsoft:
Fixing #251

- Remove the requirement for an SSL context to be passed to create an `SslETLSource`
- Implement a probing source that automatically detects SSL/Plain Websocket
- Add recreation of `ws_` for `PlainETLsource` the same way we do for `SslETLSource` (probably was a bug)

### DIFF
--- a/src/etl/ETLSource.h
+++ b/src/etl/ETLSource.h
@@ -147,27 +147,8 @@ struct ETLSourceHooks
 {
     enum class Action { STOP, PROCEED };
 
-    std::reference_wrapper<std::mutex> mtx_;
-    std::function<Action(boost::beast::error_code)> onConnected_ = [](auto) {
-        return Action::PROCEED;  // nop
-    };
-    std::function<Action(boost::beast::error_code)> onDisconnected_ = [](auto) {
-        return Action::STOP;  // nop
-    };
-
-    Action
-    onConnected(boost::beast::error_code ec) const
-    {
-        std::lock_guard lck(mtx_.get());
-        return onConnected_(ec);
-    }
-
-    Action
-    onDisconnected(boost::beast::error_code ec) const
-    {
-        std::lock_guard lck(mtx_.get());
-        return onDisconnected_(ec);
-    }
+    std::function<Action(boost::beast::error_code)> onConnected;
+    std::function<Action(boost::beast::error_code)> onDisconnected;
 };
 
 template <class Derived>
--- a/src/etl/ETLSource.h
+++ b/src/etl/ETLSource.h
@@ -16,6 +16,7 @@
 
 class ETLLoadBalancer;
 class ETLSource;
+class ProbingETLSource;
 class SubscriptionManager;
 
 /// This class manages a connection to a single ETL source. This is almost
@@ -96,6 +97,12 @@ public:
     virtual void
     run() = 0;
 
+    virtual void
+    pause() = 0;
+
+    virtual void
+    resume() = 0;
+
     virtual std::string
     toString() const = 0;
 
@@ -126,12 +133,41 @@ public:
 
 private:
     friend ForwardCache;
+    friend ProbingETLSource;
 
     virtual std::optional<boost::json::object>
     requestFromRippled(
         boost::json::object const& request,
         std::string const& clientIp,
         boost::asio::yield_context& yield) const = 0;
+};
+
+/// Hooks are executed with respect to the mutex provided
+struct ETLSourceHooks
+{
+    enum class Action { stop, proceed };
+
+    std::reference_wrapper<std::mutex> mtx_;
+    std::function<Action(boost::beast::error_code)> onConnected_ = [](auto) {
+        return Action::proceed;  // nop
+    };
+    std::function<Action(boost::beast::error_code)> onDisconnected_ = [](auto) {
+        return Action::proceed;  // nop
+    };
+
+    Action
+    onConnected(boost::beast::error_code ec) const
+    {
+        std::lock_guard lck(mtx_.get());
+        return onConnected_(ec);
+    }
+
+    Action
+    onDisconnected(boost::beast::error_code ec) const
+    {
+        std::lock_guard lck(mtx_.get());
+        return onDisconnected_(ec);
+    }
 };
 
 template <class Derived>
@@ -199,6 +235,10 @@ protected:
 
     std::atomic_bool closing_{false};
 
+    std::atomic_bool paused_{false};
+
+    ETLSourceHooks hooks_;
+
     void
     run() override
     {
@@ -215,7 +255,7 @@ protected:
 public:
     ~ETLSourceImpl()
     {
-        close(false);
+        derived().close(false);
     }
 
     bool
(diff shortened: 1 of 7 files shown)


src/etl/ETLSource.h
```diff
@@ -134,6 +142,34 @@ class ETLSource
         boost::asio::yield_context& yield) const = 0;
 };
 
+/// Hooks are executed with respect to the mutex provided
+struct ETLSourceHooks
+{
+    enum class Action { stop, proceed };
+
+    std::reference_wrapper<std::mutex> mtx_;
```
natenichols: Why use a reference wrapper here and not a `std::mutex&`?

src/etl/ProbingETLSource.cpp
```diff
@@ -0,0 +1,187 @@
...
+          balancer,
+          make_SSLHooks(mtx_))}
+    , plainSrc_{make_shared<PlainETLSource>(
+          config,
+          ioc,
+          backend,
+          subscriptions,
+          nwvl,
+          balancer,
+          make_PlainHooks(mtx_))}
```
natenichols: Do we need to pass this `mtx_` as a parameter?

`make_SSLHooks` and `make_plainHooks` are private members of `ProbingETLSource`, so they should have access to `mtx_`.

Is there another place where we construct the hooks w/ a different mutex?

src/etl/ProbingETLSource.h
```diff
@@ -0,0 +1,91 @@
...
+
+public:
+    ProbingETLSource(
+        boost::json::object const& config,
+        boost::asio::io_context& ioc,
+        std::shared_ptr<BackendInterface> backend,
+        std::shared_ptr<SubscriptionManager> subscriptions,
+        std::shared_ptr<NetworkValidatedLedgers> nwvl,
+        ETLLoadBalancer& balancer,
+        boost::asio::ssl::context sslCtx = boost::asio::ssl::context{
```
natenichols, line 33: Is there ever a scenario where this is not `boost::asio::ssl::context::tlsv12`?

If thats the case can we just declare this as a member variable?

natenichols (approved):
Nice work @godexsoft! Left a couple comments about only about a few small things. 

Pulled your code down + ran it + tested it. It works on my machine.

src/etl/ProbingETLSource.h
godexsoft: This is just a simple dependency injection. Basically future-proofing a bit. We could totally have it as a member variable and/or even hardcoded but i think there is no harm done by allowing the clients of this class to specify a custom context if needed. wdyt?

src/etl/ProbingETLSource.cpp
godexsoft: You are totally right. I think originally those factory functions were `static` but since we needed to capture `this` i did not have much choice but convert them into members. They are also `private` so there is no risk that someone else would want to call them with a different mutex.

src/etl/ETLSource.h
godexsoft: Storing references as members is considered bad style and it can lead to side-effects when copying. `std::reference_wrapper` is essentially a `pointer` (internally) that gives us reference semantics. I don't like using raw pointers in general so `std::reference_wrapper` is probably the best option here imo.

src/etl/ProbingETLSource.h
natenichols: I'm okay with leaving it as is, then.

src/etl/ETLSource.h
```diff
@@ -96,6 +97,12 @@ class ETLSource
     virtual void
     run() = 0;
 
+    virtual void
+    pause() = 0;
+
+    virtual void
+    resume() = 0;
+
```
cjcobb23, line 105: Maybe add a short comment describing what these do?

src/etl/ETLSource.h
```diff
@@ -134,6 +142,34 @@ class ETLSource
...
+{
+    enum class Action { STOP, PROCEED };
+
+    std::reference_wrapper<std::mutex> mtx_;
+    std::function<Action(boost::beast::error_code)> onConnected_ = [](auto) {
+        return Action::PROCEED;  // nop
+    };
+    std::function<Action(boost::beast::error_code)> onDisconnected_ = [](auto) {
+        return Action::STOP;  // nop
+    };
```
cjcobb23: Do we actually need to initialize these with some value here? I doubt any class would actually use this default value, right?

src/etl/ProbingETLSource.cpp
```diff
@@ -0,0 +1,187 @@
...
+        ledgerSequence, getObjects, getObjectNeighbors);
+}
+
+std::optional<boost::json::object>
+ProbingETLSource::forwardToRippled(
+    boost::json::object const& request,
+    std::string const& clientIp,
+    boost::asio::yield_context& yield) const
+{
+    assert(currentSrc_);  // expected to be connected
```
cjcobb23: I think we should have the same if condition here as in the above 2 functions, right?

src/etl/ProbingETLSource.cpp
```diff
@@ -0,0 +1,187 @@
...
+    return currentSrc_->requestFromRippled(request, clientIp, yield);
+}
+
+ETLSourceHooks
+ProbingETLSource::make_SSLHooks() noexcept
+{
+    return {
+        std::ref(mtx_),
+        // onConnected
+        [this](auto ec) {
```
cjcobb23: We could just capture the mutex in the lambdas here. Not sure if thats cleaner. I don't like the reference wrapper to be honest. And then the locking can be unique to the lambda body. It could be more fine grained, or lock free entirely, now or in the future.

src/etl/ProbingETLSource.h
```diff
@@ -0,0 +1,91 @@
...
+        boost::json::object const& request,
+        std::string const& clientIp,
+        boost::asio::yield_context& yield) const override;
+
+private:
+    std::optional<boost::json::object>
+    requestFromRippled(
+        boost::json::object const& request,
+        std::string const& clientIp,
+        boost::asio::yield_context& yield) const override;
```
cjcobb23, line 82: What's the difference between these two functions again?

src/etl/ProbingETLSource.cpp
```diff
@@ -0,0 +1,187 @@
...
+        // onDisconnected
+        [this](auto ec) {
+            if (currentSrc_)
+            {
+                currentSrc_ = nullptr;
+                sslSrc_->resume();
+            }
+            return ETLSourceHooks::Action::STOP;
+        }};
+}
```
cjcobb23: These two functions are almost identical. You could probably refactor so we don't have this duplicated code.

src/etl/ETLSource.h
cjcobb23: We can just capture the `mutex` in the lambdas down below and not need a member variable at all.

src/etl/ProbingETLSource.cpp
cjcobb23: Something like this (didn't test):
```
auto makeOnConected = [this](auto& me, auto& other) {
        return [this,&me,&other](auto ec) {
           std::lock_guard lck(mtx_);
            if (currentSrc_)
                return ETLSourceHooks::Action::STOP;

            if (!ec)
            {
                other->pause();
                currentSrc_ = me;
            }
            return ETLSourceHooks::Action::PROCEED;
        };
};

auto onConnectedPlain = makeOnConnected(plainSrc_,sslSrc_);
auto onConnectedSsl = makeOnConnected(sslSrc_,plainSrc_);
```
This also captures the `mutex`. I removed the log statement, but we can still add that in somehow. Either pass a string to `makeOnConnected`, or include relevant info in `ETLSource::toString()`

src/etl/ETLSource.h
godexsoft: The comments were not present for other abstract methods so i added them only to the implementation.. do you think we should copy and paste to here as well? 

In general I of course agree that we should have doxygen comments everywhere - maybe a separate issue to document (in code) everything we can and have a target for doxygen perhaps?

src/etl/ETLSource.h
godexsoft: I agree. good call - less code and no way to misuse. Will do

src/etl/ProbingETLSource.cpp
godexsoft: That's a good question that i can't really answer. I had these asserts on each method initially to see if things crash or not. Then i started removing one by one as it crashed. With the current asserts in place i could not get it to crash but it's entirely possible it still could in some edge case.
Ideally we would not be calling methods that require a connection while we are not connected. This would require too much refactoring and verifying tho so maybe the `if` way is better. I'm not 100% sure on this one.

src/etl/ProbingETLSource.cpp
godexsoft: That's a good idea. I'll change

src/etl/ProbingETLSource.h
godexsoft: I believe `forward` goes thru the cache first and calls `request` if not found in forward cache.

src/etl/ProbingETLSource.cpp
godexsoft: I quite like your lambda-based implementation, might try to do this.

In general tho, you are guided by the `DRY principle` (i assume). It's good to remember that `DRY` is not about "never repeat code" but rather about "what will be easier when I need to change this in the future". 
So if the logic will change (and it likely will eventually) then your approach actually will require more changes to the code than two separate functions with _sortof_ copied code. It's always a trade off..

godexsoft: I tried to implement this in a few different ways but since hooks rely on both `ssl` and `plain` sources and are injected in the constructor in the initialization list of `probing source` there is currently not really a good non-hacky way to do this. 
I think it's not worth to do the refactor at this moment, especially because after v1 we wanted to refactor the `etl` subsystem anyways. I'd leave this as is for now except for the mutex which is now captured as you suggested.

cjcobb23 (approved):
LGTM

PR #295, opened 2022-09-06
Fmendoz7#195

officialfrancismendoza:
Changed AccountTx.cpp on filtering functionality, window progression, and empty cursor return + deletion, tested successfully on forward:false and forward:true cases. Will squash commits when looks good from reviewer's end.


cjcobb23:
Please use the old PR

PR #297, opened 2022-09-07
Reconnect after timeout in ETLSource websockets

### LINKED ISSUE #289: Reconnect to rippled after a timeout
There should not be a situation in which `rippled` does not send `clio` updates for longer than a few seconds. Having this in mind, there are reports that sometimes `clio` stops receiving updates for arbitrarily long time. When this happens it appears to never go back to normal again. Looks like the socket just stops working for some unknown reason (needs more research).
 
For the time being tho it should be easy enough to implement a fail-safe mechanism: detect no activity and recreate the `ETLSource` when the configurable timeout is exceeded.

godexsoft:
Fixes #289 (hopefully).

* Added a 20 second timeout for both `Plain` and `Ssl` `ETLSource`s
* Reset websocket on reconnect for `PlainETLSource` in case this might have caused the original problem

Please note that I could not reproduce the original bug using `netem` so this is more of an attempt to see if it helps or not.

From the research I did it seems like adding a big delay (i.e. 20 seconds) using `netem` did not trigger a reconnect (unsurprisingly - we had no timeout whatsoever).
However with those 20 seconds delay we would sometimes go into a loop of death (2 second delay -> error message) but `clio` would _eventually_ recover. With the timeout option added it seems to reconnect more eagerly but this may also have been a lucky coincidence. Time will show.

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -72,6 +72,29 @@ ForwardCache::get(boost::json::object const& request) const
     return {latestForwarded_.at(*command)};
 }
 
+static boost::beast::websocket::stream_base::timeout
+make_TimeoutOption()
+{
+    // See #289 for details.
+    // TODO: investigate the issue and find if there is a solution other than
+    // introducing artificial timeouts.
+    if (true)
+    {
+        // The only difference between this and the suggested client role is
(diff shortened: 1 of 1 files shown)


natenichols (approved):
LGTM :+1:

PR #300, opened 2022-09-09
Implement book_changes RPC

### LINKED ISSUE #284: implement book_changes subscription
https://github.com/XRPLF/rippled/pull/4212/files

godexsoft:
Fixes #284

- Ported `book_changes` rpc call from `rippled`.
- Implemented `book_changes` support in `test.py`

To try you can execute something like this: `./test.py --ip 127.0.0.1 --port 51233 --ledger 74256240 book_changes`
Example output:
```
{
    "result": {
        "changes": [
            {
                "close": "9240",
                "currency_a": "XRP_drops",
                "currency_b": "raq7pGaYrLZRa88v6Py9V5oWMYEqPYr8Tz/5377697373546563680000000000000000000000",
                "high": "9240",
                "low": "9199.998196515558",
                "open": "9200",
                "volume_a": "96285669",
                "volume_b": "10444.094543578"
            },
            {
                "close": "379325.64751315",
                "currency_a": "XRP_drops",
                "currency_b": "rsoLo2S1kiGeCcn6hCUXVrCpGMWLrRrLZz/534F4C4F00000000000000000000000000000000",
                "high": "379325.64751315",
                "low": "379325.64751315",
                "open": "379325.64751315",
                "volume_a": "113941838",
                "volume_b": "300.38"
            }
        ],
        "ledger_hash": "F4FC90CC918C4E4EC33A015CE39603D83CBB44A7E97EE06DAA5CFAC8B835CD1F",
        "ledger_index": 74256240,
        "ledger_time": 715968871,
        "type": "bookChanges",
        "validated": true
    },
    "status": "success",
    "type": "response",
    "warnings": [
        {
            "id": 2001,
        }
    ]
}
```

> Please note: despite the `warning` suggesting to set `ledger_index` to `current` - this will not work as this call is not forwarding to `rippled` but rather uses `clio`'s local database.

There is also currently no unit-tests for this. We should add it at some point when we have refactored for testability.

### DIFF
--- a/src/rpc/handlers/BookChanges.cpp
+++ b/src/rpc/handlers/BookChanges.cpp
@@ -52,7 +52,7 @@ public:
     {
         reset();
 
-        auto transactions =
+        auto const transactions =
             context_.get().backend->fetchAllTransactionsInLedger(
                 ledger.seq, context_.get().yield);
         for (auto const& tx : transactions)
@@ -79,7 +79,7 @@ private:
     handleAffectedNode(STObject const& node)
     {
         auto const& metaType = node.getFName();
-        auto nodeType = node.getFieldU16(sfLedgerEntryType);
+        auto const nodeType = node.getFieldU16(sfLedgerEntryType);
 
         // we only care about ltOFFER objects being modified or
         // deleted
@@ -115,13 +115,13 @@ private:
 
         // compute the difference in gets and pays actually
         // affected onto the offer
-        auto deltaGets = finalFields.getFieldAmount(sfTakerGets) -
+        auto const deltaGets = finalFields.getFieldAmount(sfTakerGets) -
             previousFields.getFieldAmount(sfTakerGets);
-        auto deltaPays = finalFields.getFieldAmount(sfTakerPays) -
+        auto const deltaPays = finalFields.getFieldAmount(sfTakerPays) -
             previousFields.getFieldAmount(sfTakerPays);
 
-        auto g = to_string(deltaGets.issue());
-        auto p = to_string(deltaPays.issue());
+        auto const g = to_string(deltaGets.issue());
+        auto const p = to_string(deltaPays.issue());
 
         auto const noswap =
             isXRP(deltaGets) ? true : (isXRP(deltaPays) ? false : (g < p));
@@ -133,7 +133,7 @@ private:
         if (second == beast::zero)
             return;
 
-        auto rate = divide(first, second, noIssue());
+        auto const rate = divide(first, second, noIssue());
 
         if (first < beast::zero)
             first = -first;
@@ -141,7 +141,7 @@ private:
         if (second < beast::zero)
             second = -second;
 
-        auto key = noswap ? (g + '|' + p) : (p + '|' + g);
+        auto const key = noswap ? (g + '|' + p) : (p + '|' + g);
         if (tally_.contains(key))
         {
             auto& entry = tally_.at(key);
@@ -229,13 +229,13 @@ tag_invoke(
 Result
 doBookChanges(Context const& context)
 {
-    auto request = context.params;
-    auto v = ledgerInfoFromRequest(context);
-    if (auto status = std::get_if<Status>(&v))
+    auto const request = context.params;
+    auto const v = ledgerInfoFromRequest(context);
+    if (auto const status = std::get_if<Status>(&v))
         return *status;
 
-    auto lgrInfo = std::get<ripple::LedgerInfo>(v);
-    auto changes = BookChangesHandler{context}.handle(lgrInfo);
+    auto const lgrInfo = std::get<ripple::LedgerInfo>(v);
+    auto const changes = BookChangesHandler{context}.handle(lgrInfo);
     auto response = json::object{};
 
     response[JS(type)] = "bookChanges";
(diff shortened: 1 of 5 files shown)


src/rpc/handlers/BookChanges.cpp
```diff
@@ -0,0 +1,250 @@
...
+    STAmount sideBVolume;
+    STAmount highRate;
+    STAmount lowRate;
+    STAmount openRate;
+    STAmount closeRate;
+};
+
+class BookChangesHandler
+{
+    std::reference_wrapper<Context const> context_;
```
natenichols, line 27: If you've deleted copy assignment, you can probably just use `Context const&`

src/rpc/handlers/BookChanges.cpp
```diff
@@ -0,0 +1,250 @@
...
+            std::back_inserter(changes),
+            [](auto obj) { return obj.second; });
+        return changes;
+    }
+
+private:
+    inline void
+    reset() noexcept
+    {
+        tally_.clear();
```
natenichols, line 78: Why do we need `reset()` .

If the class is instantiated each time we handle, couldn't we just put this functionality in the constructor.

src/rpc/handlers/BookChanges.cpp
```diff
@@ -0,0 +1,250 @@
...
+    STAmount highRate;
+    STAmount lowRate;
+    STAmount openRate;
+    STAmount closeRate;
+};
+
+class BookChangesHandler
+{
+    std::reference_wrapper<Context const> context_;
+    std::map<std::string, BookChange> tally_;
```
natenichols: These aren't initialized until we call reset. 

can we just initialize these here with `= {}`. I don't think it will change anything, I just prefer explicitly instantiating member variables.

src/rpc/handlers/BookChanges.cpp
```diff
@@ -0,0 +1,250 @@
...
+            return;
+
+        // if either FF or PF are missing we can't compute
+        // but generally these are cancelled rather than crossed
+        // so skipping them is consistent
+        if (!node.isFieldPresent(sfFinalFields) ||
+            !node.isFieldPresent(sfPreviousFields))
+            return;
+
+        auto& finalFields = (const_cast<STObject&>(node))
```
natenichols: Vetoing `const_cast`. 

Why do we need const cast here. There's got to be a way to do this without requiring this. Look in STObject.h in rippled to see if there's a way that doesn't require this.

natenichols (changes requested):
Only requesting changes because I'd like to look into a way to do this without using `const_cast`. 

If it turns out that we _have_ to do that, I won't die on this hill, but I'd prefer it if we found a different way.

src/rpc/handlers/BookChanges.cpp
godexsoft: Oh god i did not even notice it tbh. - this was copied as is from the original implementation in `rippled`. I'll definitely try to find a way around this - thanks

src/rpc/handlers/BookChanges.cpp
godexsoft: I'd rather not have const/ref members ever, no matter what

src/rpc/handlers/BookChanges.cpp
godexsoft: Just because in my code i don't reuse it does not mean one could not reuse it some day later. So this is for correctness sake and it also does nothing on the first call to it as far as i understand

src/rpc/handlers/BookChanges.cpp
godexsoft: Why would that be?
Just because i'm not explicitly setting a value in initialization list nor in the members section does not mean default constructor will not be called for a user defined type.
See https://godbolt.org/z/ovnW6erx3

godexsoft: Having that said we can absolutely do the `= {}` thingy for good measure :)

src/rpc/handlers/BookChanges.cpp
cjcobb23: Why do we want to avoid ref members?

godexsoft: @cjcobb23 for me personally it's enough that [it's a core guideline](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c12-dont-make-data-members-const-or-references). I understand it's ok if you know what you are doing but i'd try to avoid it most of the time. using `reference_wrapper` is totally fine - it's just a pointer inside and basically a `zero-cost abstraction`.

src/rpc/handlers/BookChanges.cpp
natenichols: Yea, these are exactly the same functionally. 

I prefer explicitly calling the default constructor w/ `= {}` or `Foo foo{}` because its more explicit, but that's just me.

natenichols: Up to you here.

src/rpc/handlers/BookChanges.cpp
natenichols: How did `const_cast` get into the rippled implementation. Has this been merged yet.

godexsoft: Yea it was. I'll submit a patch to `rippled` later - no worries.

natenichols (approved):

PR #301, opened 2022-09-09
Fixes issue with assigning values to offers objects

ledhed2222:
Fixes an issue with using `nft_buy_offers` or `nft_sell_offers` when there are actually offers to return. Previously these RPC calls would fail with internal error when building the offer object for the serialized response.


godexsoft:
LGTM

godexsoft (approved):

cjcobb23 (dismissed):
Remove the merge commit, then this is good to go

PR #303, opened 2022-09-12
Add tsan/asan and time-trace support

### LINKED ISSUE #302: Add asan/tsan/time-trace for clang
We need a way to easily run `clio` with `asan`/`tsan` enabled. Once our "unit-tests" are properly implemented and covering a decent part of `clio` we can enable  the sanitizers only on `unit-tests` but right now we need to run `clio_server` with `tsan` enabled.

Note: I have already ran it locally and it generates many race warnings. Will have to fix those but that's outside of the scope of this issue.
`time-trace` is just a bonus that is easy to add. It will allow us to get insights on `clio`'s build times.

godexsoft:
Fixes #302 

@legleux please feel free to change the method of integration if you think it can be done in a better way.


CMake/deps/asan.cmake
```diff
@@ -0,0 +1,9 @@
+if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
```
legleux, line 1: Aren't the sanitizer options available in GCC?
[rippled also uses these](https://github.com/XRPLF/rippled/blob/fe05b8c4feb815e1aa6afd7ff8c85dbc92bcd651/Builds/CMake/RippledSettings.cmake#L62) (not saying it's necessarily a better way to include it.)

CMake/deps/asan.cmake
```diff
@@ -0,0 +1,9 @@
...
+  message(STATUS "Enabling address sanitizer")
+  target_compile_options(clio_tests PRIVATE -g -fsanitize=address -fno-omit-frame-pointer) 
+  target_link_options (clio_tests PRIVATE -g -fsanitize=address -fno-omit-frame-pointer)
```
legleux, line 4: Are these required to be link options also?

CMake/deps/asan.cmake
godexsoft: Seems like you are right! i usually used `clang` and did not know `gcc` had them too. That's awesome.
Also, yes, the way `rippled` adds them is probably better than my quick and dirty way.

Would you take the time to add them properly instead of using what i did?

CMake/deps/asan.cmake
godexsoft: Chances are they are not. I have not tried without but it could work just fine too. In some old project of mine i had them in the link options so i thought maybe i have done that for a reason before - can't really remember.

godexsoft:
@legleux, let's revisit this some time in the future. Closing for now.

PR #306, opened 2022-09-12
Fix compiler warnings

godexsoft:
Clang produced a bunch of warnings which this pr fixes.
If some of the code removed was unfinished and should therefore remain in the codebase - let me know. We can use `[[maybe_unused]]` in the right places and add a few comments with refs to issues in which said functionality will be finally implemented.


cjcobb23 (approved):
LGTM

Issue #285, opened 2022-08-29
Clio + Rippled Architecture Diagram & API Comments


officialfrancismendoza:
Add comments for BackendInterface.h and append new clio architecture diagram

PR #307, opened 2022-09-14
Add doxygen documentation

officialfrancismendoza:
**Issue** (#285): Lack of doxygen documentation
**Fix**: Included simplified doxyfile and initial set of comments for BackendInterface.h

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,0 +1,3 @@
+PROJECT_NAME           = "Clio"
+INPUT                  = src
+RECURSIVE              = YES
--- a/src/backend/BackendInterface.h
+++ b/src/backend/BackendInterface.h
@@ -18,6 +18,12 @@
 
 namespace Backend {
 
+/**
+ * @brief Throws an error when database read time limit is exceeded.
+ *
+ * This class is throws an error when read time limit is exceeded but
+ * is also paired with a separate class to retry the connection.
+ */
 class DatabaseTimeout : public std::exception
 {
 public:
@@ -28,6 +34,14 @@ public:
     }
 };
 
+/**
+ * @brief Separate class that reattempts connection after time limit.
+ *
+ * @tparam F Represents a class of handlers for Cassandra database.
+ * @param func Instance of Cassandra database handler class.
+ * @param waitMs Is the arbitrary time limit of 500ms.
+ * @return auto
+ */
 template <class F>
 auto
 retryOnTimeout(F func, size_t waitMs = 500)
@@ -48,19 +62,48 @@ retryOnTimeout(F func, size_t waitMs = 500)
     }
 }
 
+/**
+ * @brief Passes in serialized handlers in an asynchronous fashion.
+ *
+ * Note that the synchronous auto passes handlers critical to supporting
+ * the Clio backend. The coroutine types are checked if same/different.
+ *
+ * @tparam F Represents a class of handlers for Cassandra database.
+ * @param f R-value instance of Cassandra handler class.
+ * @return auto
+ */
 template <class F>
 auto
 synchronous(F&& f)
 {
+    /** @brief Serialized handlers and their execution.
+     *
+     * The ctx class is converted into a serialized handler, also named
+     * ctx, and is used to pass a stream of data into the method.
+     */
     boost::asio::io_context ctx;
     boost::asio::io_context::strand strand(ctx);
     std::optional<boost::asio::io_context::work> work;
 
+    /*! @brief Place the ctx within the vector of serialized handlers. */
     work.emplace(ctx);
 
+    /**
+     * @brief If/else statements regarding coroutine type matching.
+     *
+     * R is the currently executing coroutine that is about to get passed.
+     * If corountine types do not match, the current one's type is stored.
+     */
     using R = typename std::result_of<F(boost::asio::yield_context&)>::type;
     if constexpr (!std::is_same<R, void>::value)
     {
+        /**
+         * @brief When the coroutine type is the same
+         *
+         * The spawn function enables programs to implement asynchronous logic
+         * in a synchronous manner. res stores the instance of the currently
+         * executing coroutine, yield. The different type is returned.
+         */
         R res;
         boost::asio::spawn(
             strand, [&f, &work, &res](boost::asio::yield_context yield) {
@@ -73,6 +116,7 @@ synchronous(F&& f)
     }
     else
     {
+        /*! @brief When the corutine type is different, run as normal. */
         boost::asio::spawn(
             strand, [&f, &work](boost::asio::yield_context yield) {
                 f(yield);
@@ -83,6 +127,13 @@ synchronous(F&& f)
     }
 }
 
+/**
+ * @brief Reestablishes synchronous connection on timeout.
+ *
+ * @tparam Represents a class of handlers for Cassandra database.
+ * @param f R-value instance of Cassandra database handler class.
+ * @return auto
+ */
 template <class F>
 auto
 synchronousAndRetryOnTimeout(F&& f)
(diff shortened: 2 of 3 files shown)


src/backend/BackendInterface.h
```diff
@@ -7,8 +7,13 @@
 #include <backend/Types.h>
 #include <thread>
 #include <type_traits>
+
 namespace Backend {
 
+/**
```
godexsoft, line 21: If we are using doxygen, may as well use the full capabilities of the format:
```
/**
 * @brief Responsible for throwing an error when the database read time limit is exceeded.
 * 
 * Longer description if needed here...
 */
```

some example for a template function:
```
/**
 * @brief Sums two values
 * 
 * This function sums two values of specified `Type`.
 * Note that it's a requirement for `Type` to support the `+` operation.
 * 
 * @tparam Type The type of data
 * @param a The first value
 * @param b The second value
 * @return Type The sum of `a` and `b` 
 */
template<typename Type>
static Type sum(Type a, Type b) { return a + b; } 
```

i'd also pick one style and use it everywhere. so either the above or 
```
/// @brief ...
///
/// Text
/// @return ...
```

what i personally like to do is for short one-line doc-comments i do this:
```
class A {
    int member_;    /*! @brief Describe this member */
};
```
Because we have 80 character limit for clang-format.. the above might not look good. In which case it's also ok to put it above the `member_`.

To appreciate how the documentation looks you really should setup doxygen and actually generate docs from your code. Just experiment with it and see how it looks and what is better or worse in your opinion.

godexsoft (changes requested):

src/backend/BackendInterface.h
cjcobb23: yes lets use the full capabilities of doxygen

godexsoft: This now looks better @officialfrancismendoza! I'd still like to see explanations for the return, tparam, etc. in some places.

Also, it seems like in the final output (html) this file was not included somehow. Please check doxygen configuration and make sure this file is part of the resulting html doc :+1:

src/backend/BackendInterface.h
```diff
@@ -96,8 +118,11 @@ synchronous(F&& f)
 }
 
 /**
- * This template is responsible for reestablishing the synchronous connection on
- * timeout.
+ * @brief Reestablishes synchronous connection on timeout.
+ *
+ * @tparam F
```
godexsoft: these 3 could use some explanation what they are and what they do

src/backend/BackendInterface.h
```diff
@@ -274,55 +342,74 @@ class BackendInterface
...
+    /*! @brief Virtual function version of fetchLedgerObject */
     virtual std::optional<Blob>
     doFetchLedgerObject(
         ripple::uint256 const& key,
         std::uint32_t const sequence,
         boost::asio::yield_context& yield) const = 0;
 
-    /// A virtual function version of fetchLedgerObjects, taking in the exact
-    /// same parameters.
+    /*! @brief Virtual function version of fetchLedgerObjects */
```
godexsoft, line 389: you don't need to specify "virtual" in the doc comment - the function signature already should contain this information (you need to confirm this is true after generating the doc tho).

src/backend/BackendInterface.h
```diff
@@ -411,52 +529,94 @@ class BackendInterface
...
     open(bool readOnly) = 0;
 
-    // Close the database, releasing any resources
+    /*! @brief Closes the database, releasing any resources. */
     virtual void
     close(){};
 
-    // *** private helper methods
+    /**
+     * @brief Private helper methods.
```
godexsoft: i'm not sure this will come out as you expect in the doc. you need to verify once you fix the doc output to actually contain this file's documentation.

godexsoft (changes requested):

officialfrancismendoza:
@godexsoft note that in a thread in #clio, Michael requested that we do not generate the documentation within the repo, but the PR above demonstrates it's possible. If you think we should not include it in the repo, lmk and I will roll back and we can just generate what we need manually + copy-paste to the doc CDN

godexsoft:
@officialfrancismendoza Please remove generated docs and their assets from the repo - @legleux is right, we should probably not store the generated doc in the repo but rather generate the doc for each release and publish it somewhere accessible to everyone. Otherwise it looks good to me! thanks for making all the adjustments :+1:

Doxyfile
```diff
@@ -829,7 +829,7 @@ WARN_LOGFILE           =
 # spaces. See also FILE_PATTERNS and EXTENSION_MAPPING
 # Note: If this tag is empty the current directory is searched.
 
-INPUT                  = src/backend/BackendInterface.h
+INPUT                  =
```
godexsoft: hmm.. i think you want to leave the paths here. just don't generate html docs before you commit/push. 
i think in fact we need to specify something like src/* in here but please go read the doxy docs on how to specify a path for ALL your headers and cpp files. i'd prefer not to have to add a path to a new header every time we add a new one to the codebase.

godexsoft (changes requested):

Doxyfile
legleux:
```
INPUT = src
RECURSIVE = YES
```

Doxyfile
```diff
@@ -0,0 +1,2579 @@
...
+                         *.vhdl \
+                         *.ucf \
+                         *.qsf \
+                         *.ice
+
+# The RECURSIVE tag can be used to specify whether or not subdirectories should
+# be searched for input files as well.
+# The default value is: NO.
+
+RECURSIVE              = NO
```
legleux: `RECURSIVE = YES` as [Alex mentioned above](https://github.com/XRPLF/clio/pull/307/files#r976860481)

Doxyfile
```diff
@@ -0,0 +1,2579 @@
...
+# This tag requires that the tag HAVE_DOT is set to YES.
+
+GENERATE_LEGEND        = YES
+
+# If the DOT_CLEANUP tag is set to YES, doxygen will remove the intermediate dot
+# files that are used to generate the various graphs.
+# The default value is: YES.
+# This tag requires that the tag HAVE_DOT is set to YES.
+
+DOT_CLEANUP            = YES
```
godexsoft: I think we can remove ~2576 lines from this file, wdyt? I doubt we need most of these default config options right?

officialfrancismendoza: @godexsoft I think we should keep them there just in case someone needs to change something more specific in the future. Otherwise they will get confused when they want to generate something very specific and the parameter doesn't exist + might affect document generation

godexsoft: Who is they? It's gonna be us most likely. Users of clio will not be generating "custom" docs - at best they would generate exactly the doc we want them to generate.. but realistically users will read the doc we generate and put up on a website for any release from now on. 
IMHO if you want to change something specific you go and look up the config parameter in doxy documentation or example configuration file. I would really prefer not to have 2500+ lines of default config values in our codebase. 
I'm pretty sure that most if not all of the options in this file are set to their defaults, meaning that if you don't specify them they will end up the same value anyway. 

Let's try to remove everything that's not customized along with all stock comments. Feel free to replace with your own comments if you find it helpful. You can verify it still generates the same doc by running it locally after modifying the configuration file. Thanks!

godexsoft:
I'm seeing a lot of generated files in `html` folder. Drop the whole folder please.
Second thing is, currently in your `html` folder you are missing the doc for BackendInterface etc. So when i open `html/index.html` i follow to `Classes` and try to open the doc for BackendInterface it goes 'file not found'. Now i don't know if that's because you deleted only this part of `html` folder by hand or it's doxy fault for generating it like that. But it needs fixing :) 
And after it's fixed and you verify that `html` folder actually works fine - please delete the whole thing so that in the PR there is no `html` folder at all.

godexsoft:
Ah, last thing @officialfrancismendoza , can you please also remove the PDF diagram? we don't need it in the codebase either - it may be however useful for the website/docs later. Otherwise everything looks alright to me :) Thanks

officialfrancismendoza:
@godexsoft can you review? Do note that there was critical change in BackendInterface.h, as I removed the mutex that was in my branch in an older version but not in the latest of develop (which caused a merge conflict that I had to resolve manually). If looks good, please approve and merge

godexsoft (dismissed):
LGTM

godexsoft (dismissed):

godexsoft:
> LGTM

This last commit is not signed again.. you used `-s` instead of `-S` so it's `signed-off` not `pgp signed`.

godexsoft (dismissed):

godexsoft (approved):

PR #309, opened 2022-09-15
Throw error if server bind or listen fails

cjcobb23:
Previously, if the server failed to bind to the specified endpoint, it would continue to run, but not handle any requests. Now, the server will fail to start if it can't bind.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/309)

### DIFF
--- a/src/webserver/Listener.h
+++ b/src/webserver/Listener.h
@@ -241,7 +241,8 @@ public:
         if (ec)
         {
             BOOST_LOG_TRIVIAL(error)
-                << "Failed to bind to endpoint: " << endpoint;
+                << "Failed to bind to endpoint: " << endpoint
+                << ". message: " << ec.message();
             throw std::runtime_error("Failed to bind to specified endpoint");
         }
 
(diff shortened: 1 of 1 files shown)


src/webserver/Listener.h
```diff
@@ -239,12 +239,20 @@ class Listener
         // Bind to the server address
         acceptor_.bind(endpoint, ec);
         if (ec)
-            return;
+        {
+            BOOST_LOG_TRIVIAL(error)
+                << "Failed to bind to endpoint: " << endpoint;
+            throw std::runtime_error("Failed to bind to specified endpoint");
```
godexsoft, line 246: Do you think it may be worth including `ec.message()` somewhere? 
I'm approving anyway and will let you decide if you want it or not.

godexsoft (dismissed):

godexsoft (changes requested):
Looks great 🤟

godexsoft (approved):

PR #311, opened 2022-09-16
Implement unique taging of incoming requests

### LINKED ISSUE #212: Unique logging identifier
We need a unique logging identifier to keep track of client's request from the moment the request is received at the socket to after the response is sent. This will enable effective debugging and performance metrics parsing.

godexsoft:
Fixes #212

Implemented in a way that we can easily switch different ways we generate the `tag` because locking a mutex can be taxing on performance. Left UUID implementation as the current default because it's not only more readable/uniform in the logs but also because it will produce unique tags on each separate `clio` node.  

Please let's review what we should log with this unique tag once again. Is it enough as it is or should we also log something inside any of the rpc handlers perhaps? 
As far as i can tell with the current setup we can already find out how long which rpc exactly takes but we don't have detail about what operations were invoked (backend etc.) for each session. So the real question is do we need it now or should we add it gradually as we find some rpc calls that we need to improve performance of.


src/etl/ReportingETL.cpp
```diff
@@ -1008,7 +1008,7 @@ ReportingETL::loadCache(uint32_t seq)
                         backend_->cache().update(res.objects, seq, true);
                         if (!res.cursor || (end && *(res.cursor) > *end))
                             break;
-                        BOOST_LOG_TRIVIAL(debug)
+                        BOOST_LOG_TRIVIAL(trace)
```
cjcobb23, line 1011: I think we still want some sort of progress marker around the cache download at the debug level. It doesn't have to be as often as this was, but the download can take 45 minutes, and its nice to have a way to see in the logs how far along it is.

cjcobb23: Thinking about it more, how does the output look if we change this to trace, but the statement outside the loop is left at debug?

src/util/Taggable.h
```diff
@@ -0,0 +1,101 @@
...
+     */
+    class Tag final
+    {
+        std::string tag_ = Strategy::next();
+
+    public:
+        /**
+         * @return std::string const& The string representation of the tag
+         */
+        std::string const&
```
cjcobb23: I am wondering whether we should use a templated type here. In the case where taggable is `atomic_uint64_t`, we are using extra memory by storing a string. 8 bytes for atomic versus 32 bytes for string. Not sure how many of these `Taggable`s will be alive in RAM at once.

src/util/Taggable.h
```diff
@@ -0,0 +1,101 @@
...
+        return tag_;
+    }
+
+private:
+    Tag tag_;
+};
+
+using UUIDTaggable = Taggable<detail::UUIDTagStrategy>;
+using UIntTaggable = Taggable<detail::UIntTagStrategy>;
+using DefaultTaggable = UUIDTaggable;
```
cjcobb23: Let's make this a configuration parameter, and default to UInt, for performance reasons.

src/webserver/WsBase.h
```diff
@@ -44,20 +45,28 @@ getDefaultWsResponse(boost::json::value const& id)
     return defaultResp;
 }
 
-class WsBase
+class WsBase : public util::DefaultTaggable
```
cjcobb23: We need to be able to tag each individual request coming over a websocket. Websockets can handle multiple requests, either in parallel or sequentially, and we need to be able to distinguish each of those requests from each other.

src/webserver/HttpBase.h
```diff
@@ -161,6 +162,11 @@ class HttpBase
         , lambda_(*this)
         , buffer_(std::move(buffer))
     {
+        BOOST_LOG_TRIVIAL(info) << "[" << tag() << "] http session created";
```
cjcobb23: Info is probably a bit high for this. In production, these messages will flood the logs, as we get thousands of requests per second.

src/webserver/HttpBase.h
```diff
@@ -161,6 +162,11 @@ class HttpBase
...
+    }
+    virtual ~HttpBase()
+    {
+        BOOST_LOG_TRIVIAL(info) << "[" << tag() << "] http session closed";
```
cjcobb23: info level is too high

src/webserver/HttpBase.h
```diff
@@ -221,6 +227,9 @@ class HttpBase
         if (!ip)
             return;
 
+        BOOST_LOG_TRIVIAL(info) << "[" << tag() << "] http::" << __func__
+                                << " received request from ip = " << *ip;
```
cjcobb23: I think info is too high for this. I would also add something like "Posting to WorkQueue".

cjcobb23 (changes requested):

src/util/Taggable.h
godexsoft: I defaulted it to `null` (nop) if nothing is specified in the config. i'll set the example config file to `uint`.

src/etl/ReportingETL.cpp
godexsoft: It looks like every 3-4 seconds or so we get `Finished a cursor` followed by `Starting a cursor` (both are long lines). It's quite a lot of logs. The `Finished a cursor` is on `info` level and `Starting a cursor` is on `debug` as you suggested. I'd leave  this on trace and probably push `Finished a cursor` down to `debug` too but it's up to you. If we leave just the `Finished` on `info` maybe it's a good progress marker for the download. wdyt?

example-config.json
```diff
@@ -36,6 +36,7 @@
     "log_rotation_size": 2048,
     "log_directory_max_size": 51200,
     "log_rotation_hour_interval": 12,
+    "log_tag_style": "uint",
```
cjcobb23, line 39: All of this log stuff should be nested inside one `logging` object in my opinion, but im not sure best practices around changing the config file format.

src/rpc/RPC.cpp
```diff
@@ -30,7 +66,7 @@ make_WsContext(
 
     std::string command = commandValue.as_string().c_str();
 
-    return Context{
+    return std::make_optional<Context>(
```
cjcobb23, line 69: why do we need the `make_optional` now, whereas before we did not?

example-config.json
godexsoft: That's a very good point. I have no problem to change it if you think we can change the format for next version. I'm a bit hesitant to break backward compatibility but if it happens between releases then maybe that's ok. Especially if clio is not very wide spread at this moment. wdyt?

src/rpc/RPC.cpp
godexsoft: It's because we were relying on implicit conversion which relied on copy-constructor of `Context`. Now that `Context` is inheriting from `Taggable` its copy constructor is implicitly deleted since we are using `unique_ptr` deep inside (move only type). `std::make_optional` does not rely on copying anything and just creates `Context` in-place. In any case we should have been using `make_optional` in the first place.

src/rpc/RPC.cpp
```diff
@@ -1,11 +1,46 @@
...
+    std::shared_ptr<BackendInterface const> const& backend_,
+    std::shared_ptr<SubscriptionManager> const& subscriptions_,
+    std::shared_ptr<ETLLoadBalancer> const& balancer_,
+    std::shared_ptr<ReportingETL const> const& etl_,
+    std::shared_ptr<WsBase> const& session_,
+    util::TagDecoratorFactory const& tagFactory_,
+    Backend::LedgerRange const& range_,
+    Counters& counters_,
+    std::string const& clientIp_)
+    : Taggable(tagFactory_.make(std::cref(session_->tag())))
```
cjcobb23: I think `session_` is null when the request comes over regular http and not websocket. Did you test this case?

godexsoft: I just checked and you are totally right - nicely spotted! I will have to think about how to workaround the fact we can have an empty session. If i just don't use the session if it's `null` then we will have a random `id` for the context without any coupling to the underlying session be it ws or http.

example-config.json
cjcobb23: I'm not really sure. I don't know how much operators read release notes. I really don't know best practices around changing a config file.

src/rpc/RPC.cpp
cjcobb23: well when session is null, the request is over http, and I think then we can probably just use the same tag as the http session?

cjcobb23 (approved):
LGTM

PR #313, opened 2022-09-17
Make NFT URI lookup O(1)

ledhed2222:
* In cassandra, adds `nf_token_uris` table that maps an NFT ID to its URI. This table is _NOT_ sparse. Because of the re-minting of NFTs edge case, we have to:
  * Include a sequence number in this table
  * Write an empty URI if a token is minted without one, because it could have previously been minted with a different URI :(
  * (But, there should only be one row written to the table for each NFT, except in the case of a re-mint in which case you'll get another, and so on. So number of rows is as small as we can make it)
* Refactors NFT lookup code such that we hit this new table. This means that looking up an NFT is now O(1), whereas before it was O(n) where is is the number of NFTokenPages owned by the account that currently owns the NFT in question.
* Replaces the `issuer_nf_tokens` table with the `issuer_nf_tokens_v2` table. The new one also stores the taxon and includes it as part of a composite key. This new table will be used for the upcoming `issuer_nfts` RPC call to filter on issuer/taxon combo. NOTE - the old `issuer_nf_tokens` table was being written to, but not yet serving any data, so it is totally safe to delete this table.
* Refactors NFT writing such that the `issuer_nf_tokens_v2` table is only ever written to when there is a net new token. Ditto with the new `nf_token_uris`. Previously, the `issuer_nf_tokens` table would be clobbered with the same data. This never would have resulted in bad data, but would result in unnecessary writes.
* Adds logic to load the various NFT states when reading from a ledger instead of transaction-by-transaction. This allows us to accomplish all of the above without serving bad data. That is, when loading the initial ledger we now pull in all the NFT data form that ledger.

_NOTE_ - when we deploy this to an environment with existing data, the following needs to happen:

* Stop allowing incoming users requests to the clio instance. You can use another running instance of clio on the old version to continue serving requests.
* Upgrade clio to this version and restart. This will create all the new tables.
* To migrate existing data, run the utility [here](https://github.com/ledhed2222/clio_migrator) to migrate existing data. This tool _can_ be run while simultaneously writing new ledgers using clio.
* Once this migration is complete, you can resume serving requests.

_NOTE_ - this pr _technically_ has breaking changes to the `nft_info` API call:

1) the returned `uri` field is no longer hex decoded
2) the returned `nft_sequence` field is renamed `nft_serial` to match rippled `account_nfts`
3) the returned `transfer_fee` field is renamed `transfer_rate`
4) the returned `uri` field is no longer null when empty, but an empty string
5) `nft_info` can raise a new error if clio doesn't know the URI for the requested NFT. This should not happen unless there is a really unexpected race condition, in which case just trying the request again will work.

_NOTES_ on performance improvements:

nft_info API on standalone with 1 account that has 100K NFTs (to simulate worst cases)

Old Code (63872 requests):
    Worst Case: 10664 ms
    Best Case: 2 ms
    Average: 2983 ms
    Median: 2376 ms
New Code (99892 requests):
   Worst Case: 25 ms
   Best Case: 1 ms
   Average: 3 ms
   Median: 2 ms
In other words - it’s about 3 orders of magnitude  faster on average

----

Anecdotally with specific production examples:

Previously 000A1AF4DBDE606E6D2819D1E99AB9B086ABF7B01ABFFCF51320BAA100001849 at index 75471206 took 194ms

Now the same request takes 2ms.

### DIFF
--- a/src/backend/CassandraBackend.cpp
+++ b/src/backend/CassandraBackend.cpp
@@ -17,11 +17,12 @@
 */
 //==============================================================================
 
-#include <ripple/app/tx/impl/details/NFTokenUtils.h>
 #include <backend/CassandraBackend.h>
 #include <backend/DBHelpers.h>
 #include <log/Logger.h>
 #include <util/Profiler.h>
+
+#include <ripple/app/tx/impl/details/NFTokenUtils.h>
 
 #include <functional>
 #include <unordered_map>
@@ -405,17 +406,41 @@ CassandraBackend::writeNFTs(std::vector<NFTsData>&& data)
             },
             "nf_tokens");
 
-        makeAndExecuteAsyncWrite(
-            this,
-            std::make_tuple(record.tokenID),
-            [this](auto const& params) {
-                CassandraStatement statement{insertIssuerNFT_};
-                auto const& [tokenID] = params.data;
-                statement.bindNextBytes(ripple::nft::getIssuer(tokenID));
-                statement.bindNextBytes(tokenID);
-                return statement;
-            },
-            "issuer_nf_tokens");
+        // If `uri` is set (and it can be set to an empty uri), we know this
+        // is a net-new NFT. That is, this NFT has not been seen before by us
+        // _OR_ it is in the extreme edge case of a re-minted NFT ID with the
+        // same NFT ID as an already-burned token. In this case, we need to
+        // record the URI and link to the issuer_nf_tokens table.
+        if (record.uri)
+        {
+            makeAndExecuteAsyncWrite(
+                this,
+                std::make_tuple(record.tokenID),
+                [this](auto const& params) {
+                    CassandraStatement statement{insertIssuerNFT_};
+                    auto const& [tokenID] = params.data;
+                    statement.bindNextBytes(ripple::nft::getIssuer(tokenID));
+                    statement.bindNextInt(
+                        ripple::nft::toUInt32(ripple::nft::getTaxon(tokenID)));
+                    statement.bindNextBytes(tokenID);
+                    return statement;
+                },
+                "issuer_nf_tokens");
+
+            makeAndExecuteAsyncWrite(
+                this,
+                std::make_tuple(
+                    record.tokenID, record.ledgerSequence, record.uri.value()),
+                [this](auto const& params) {
+                    CassandraStatement statement{insertNFTURI_};
+                    auto const& [tokenID, lgrSeq, uri] = params.data;
+                    statement.bindNextBytes(tokenID);
+                    statement.bindNextInt(lgrSeq);
+                    statement.bindNextBytes(uri);
+                    return statement;
+                },
+                "nf_token_uris");
+        }
     }
 }
 
@@ -598,18 +623,35 @@ CassandraBackend::fetchNFT(
     std::uint32_t const ledgerSequence,
     boost::asio::yield_context& yield) const
 {
-    CassandraStatement statement{selectNFT_};
-    statement.bindNextBytes(tokenID);
-    statement.bindNextInt(ledgerSequence);
-    CassandraResult response = executeAsyncRead(statement, yield);
-    if (!response)
+    CassandraStatement nftStatement{selectNFT_};
+    nftStatement.bindNextBytes(tokenID);
+    nftStatement.bindNextInt(ledgerSequence);
+    CassandraResult nftResponse = executeAsyncRead(nftStatement, yield);
+    if (!nftResponse)
         return {};
 
     NFT result;
     result.tokenID = tokenID;
-    result.ledgerSequence = response.getUInt32();
-    result.owner = response.getBytes();
-    result.isBurned = response.getBool();
+    result.ledgerSequence = nftResponse.getUInt32();
+    result.owner = nftResponse.getBytes();
+    result.isBurned = nftResponse.getBool();
+
+    // now fetch URI. Usually we will have the URI even for burned NFTs, but
+    // if the first ledger on this clio included NFTokenBurn transactions
+    // we will not have the URIs for any of those tokens. In any other case
+    // not having the URI indicates something went wrong with our data.
+    //
+    // TODO - in the future would be great for any handlers that use this
+    // could inject a warning in this case (the case of not having a URI
+    // because it was burned in the first ledger) to indicate that even though
+    // we are returning a blank URI, the NFT might have had one.
+    CassandraStatement uriStatement{selectNFTURI_};
+    uriStatement.bindNextBytes(tokenID);
+    uriStatement.bindNextInt(ledgerSequence);
+    CassandraResult uriResponse = executeAsyncRead(uriStatement, yield);
+    if (uriResponse.hasResult())
+        result.uri = uriResponse.getBytes();
+
     return result;
 }
 
@@ -1388,17 +1430,37 @@ CassandraBackend::open(bool readOnly)
 
         query.str("");
         query << "CREATE TABLE IF NOT EXISTS " << tablePrefix
-              << "issuer_nf_tokens"
+              << "issuer_nf_tokens_v2"
               << "  ("
               << "    issuer blob,"
+              << "    taxon bigint,"
               << "    token_id blob,"
-              << "    PRIMARY KEY (issuer, token_id)"
+              << "    PRIMARY KEY (issuer, taxon, token_id)"
               << "  )";
         if (!executeSimpleStatement(query.str()))
             continue;
 
         query.str("");
-        query << "SELECT * FROM " << tablePrefix << "issuer_nf_tokens"
+        query << "SELECT * FROM " << tablePrefix << "issuer_nf_tokens_v2"
+              << " LIMIT 1";
+        if (!executeSimpleStatement(query.str()))
+            continue;
+
+        query.str("");
+        query << "CREATE TABLE IF NOT EXISTS " << tablePrefix << "nf_token_uris"
+              << "  ("
+              << "    token_id blob,"
+              << "    sequence bigint,"
+              << "    uri blob,"
+              << "    PRIMARY KEY (token_id, sequence)"
+              << "  )"
+              << "  WITH CLUSTERING ORDER BY (sequence DESC)"
+              << "    AND default_time_to_live = " << ttl;
+        if (!executeSimpleStatement(query.str()))
+            continue;
+
+        query.str("");
+        query << "SELECT * FROM " << tablePrefix << "nf_token_uris"
               << " LIMIT 1";
         if (!executeSimpleStatement(query.str()))
             continue;
@@ -1558,10 +1620,26 @@ CassandraBackend::open(bool readOnly)
             continue;
 
         query.str("");
-        query << "INSERT INTO " << tablePrefix << "issuer_nf_tokens"
-              << " (issuer,token_id)"
-              << " VALUES (?,?)";
+        query << "INSERT INTO " << tablePrefix << "issuer_nf_tokens_v2"
+              << " (issuer,taxon,token_id)"
+              << " VALUES (?,?,?)";
         if (!insertIssuerNFT_.prepareStatement(query, session_.get()))
+            continue;
+
+        query.str("");
+        query << "INSERT INTO " << tablePrefix << "nf_token_uris"
+              << " (token_id,sequence,uri)"
+              << " VALUES (?,?,?)";
+        if (!insertNFTURI_.prepareStatement(query, session_.get()))
+            continue;
+
+        query.str("");
+        query << "SELECT uri FROM " << tablePrefix << "nf_token_uris"
+              << " WHERE token_id = ? AND"
+              << " sequence <= ?"
+              << " ORDER BY sequence DESC"
+              << " LIMIT 1";
+        if (!selectNFTURI_.prepareStatement(query, session_.get()))
             continue;
 
         query.str("");
--- a/src/backend/CassandraBackend.h
+++ b/src/backend/CassandraBackend.h
@@ -655,6 +655,8 @@ private:
     CassandraPreparedStatement insertNFT_;
     CassandraPreparedStatement selectNFT_;
     CassandraPreparedStatement insertIssuerNFT_;
+    CassandraPreparedStatement insertNFTURI_;
+    CassandraPreparedStatement selectNFTURI_;
     CassandraPreparedStatement insertNFTTx_;
     CassandraPreparedStatement selectNFTTx_;
     CassandraPreparedStatement selectNFTTxForward_;
--- a/src/backend/DBHelpers.h
+++ b/src/backend/DBHelpers.h
@@ -85,11 +85,41 @@ struct NFTsData
     // final state of an NFT per ledger. Since we pull this from transactions
     // we keep track of which tx index created this so we can de-duplicate, as
     // it is possible for one ledger to have multiple txs that change the
-    // state of the same NFT.
-    std::uint32_t transactionIndex;
+    // state of the same NFT. This field is not applicable when we are loading
+    // initial NFT state via ledger objects, since we do not have to tiebreak
+    // NFT state for a given ledger in that case.
+    std::optional<std::uint32_t> transactionIndex;
     ripple::AccountID owner;
-    bool isBurned;
+    // We only set the uri if this is a mint tx, or if we are
+    // loading initial state from NFTokenPage objects. In other words,
+    // uri should only be set if the etl process believes this NFT hasn't
+    // been seen before in our local database. We do this so that we don't
+    // write to the the nf_token_uris table every
+    // time the same NFT changes hands. We also can infer if there is a URI
+    // that we need to write to the issuer_nf_tokens table.
+    std::optional<ripple::Blob> uri;
+    bool isBurned = false;
 
+    // This constructor is used when parsing an NFTokenMint tx.
+    // Unfortunately because of the extreme edge case of being able to
+    // re-mint an NFT with the same ID, we must explicitly record a null
+    // URI. For this reason, we _always_ write this field as a result of
+    // this tx.
+    NFTsData(
+        ripple::uint256 const& tokenID,
+        ripple::AccountID const& owner,
+        ripple::Blob const& uri,
+        ripple::TxMeta const& meta)
+        : tokenID(tokenID)
+        , ledgerSequence(meta.getLgrSeq())
+        , transactionIndex(meta.getIndex())
+        , owner(owner)
+        , uri(uri)
+    {
+    }
+
+    // This constructor is used when parsing an NFTokenBurn or
+    // NFTokenAcceptOffer tx
     NFTsData(
         ripple::uint256 const& tokenID,
         ripple::AccountID const& owner,
@@ -100,6 +130,24 @@ struct NFTsData
         , transactionIndex(meta.getIndex())
         , owner(owner)
         , isBurned(isBurned)
+    {
+    }
+
+    // This constructor is used when parsing an NFTokenPage directly from
+    // ledger state.
+    // Unfortunately because of the extreme edge case of being able to
+    // re-mint an NFT with the same ID, we must explicitly record a null
+    // URI. For this reason, we _always_ write this field as a result of
+    // this tx.
+    NFTsData(
+        ripple::uint256 const& tokenID,
+        std::uint32_t const ledgerSequence,
+        ripple::AccountID const& owner,
+        ripple::Blob const& uri)
+        : tokenID(tokenID)
+        , ledgerSequence(ledgerSequence)
+        , owner(owner)
+        , uri(uri)
     {
     }
 };
--- a/src/backend/Types.h
+++ b/src/backend/Types.h
@@ -83,6 +83,7 @@ struct NFT
     ripple::uint256 tokenID;
     std::uint32_t ledgerSequence;
     ripple::AccountID owner;
+    Blob uri;
     bool isBurned;
 
     // clearly two tokens are the same if they have the same ID, but this
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -27,6 +27,7 @@
 
 #include <backend/DBHelpers.h>
 #include <etl/ETLSource.h>
+#include <etl/NFTHelpers.h>
 #include <etl/ProbingETLSource.h>
 #include <etl/ReportingETL.h>
 #include <log/Logger.h>
@@ -680,6 +681,8 @@ public:
                         request_.ledger().sequence(),
                         std::string{obj.key()});
                 lastKey_ = obj.key();
+                backend.writeNFTs(getNFTDataFromObj(
+                    request_.ledger().sequence(), obj.key(), obj.data()));
                 backend.writeLedgerObject(
                     std::move(*obj.mutable_key()),
                     request_.ledger().sequence(),
--- a/src/etl/NFTHelpers.cpp
+++ b/src/etl/NFTHelpers.cpp
@@ -17,7 +17,6 @@
 */
 //==============================================================================
 
-#include <ripple/app/tx/impl/details/NFTokenUtils.h>
 #include <ripple/protocol/STBase.h>
 #include <ripple/protocol/STTx.h>
 #include <ripple/protocol/TxMeta.h>
@@ -123,7 +122,11 @@ getNFTokenMintData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
         return {
             {NFTTransactionsData(
                 tokenIDResult.front(), txMeta, sttx.getTransactionID())},
-            NFTsData(tokenIDResult.front(), *owner, txMeta, false)};
+            NFTsData(
+                tokenIDResult.front(),
+                *owner,
+                sttx.getFieldVL(ripple::sfURI),
+                txMeta)};
 
     std::stringstream msg;
     msg << " - unexpected NFTokenMint data in tx " << sttx.getTransactionID();
@@ -150,11 +153,11 @@ getNFTokenBurnData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
         // NFT burn can result in an NFTokenPage being modified to no longer
         // include the target, or an NFTokenPage being deleted. If this is
         // modified, we want to look for the target in the fields prior to
-        // modification. If deleted, it's possible that the page was modified
-        // to remove the target NFT prior to the entire page being deleted. In
-        // this case, we need to look in the PreviousFields. Otherwise, the
-        // page was not modified prior to deleting and we need to look in the
-        // FinalFields.
+        // modification. If deleted, it's possible that the page was
+        // modified to remove the target NFT prior to the entire page being
+        // deleted. In this case, we need to look in the PreviousFields.
+        // Otherwise, the page was not modified prior to deleting and we
+        // need to look in the FinalFields.
         std::optional<ripple::STArray> prevNFTs;
 
         if (node.isFieldPresent(ripple::sfPreviousFields))
@@ -360,7 +363,7 @@ getNFTokenCreateOfferData(
 }
 
 std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
-getNFTData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
+getNFTDataFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
 {
     if (txMeta.getResultTER() != ripple::tesSUCCESS)
         return {{}, {}};
@@ -385,4 +388,29 @@ getNFTData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
         default:
             return {{}, {}};
     }
+}
+
+std::vector<NFTsData>
+getNFTDataFromObj(
+    std::uint32_t const seq,
+    std::string const& key,
+    std::string const& blob)
+{
+    std::vector<NFTsData> nfts;
+    ripple::STLedgerEntry const sle = ripple::STLedgerEntry(
+        ripple::SerialIter{blob.data(), blob.size()},
+        ripple::uint256::fromVoid(key.data()));
+
+    if (sle.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltNFTOKEN_PAGE)
+        return nfts;
+
+    auto const owner = ripple::AccountID::fromVoid(key.data());
+    for (ripple::STObject const& node : sle.getFieldArray(ripple::sfNFTokens))
+        nfts.emplace_back(
+            node.getFieldH256(ripple::sfNFTokenID),
+            seq,
+            owner,
+            node.getFieldVL(ripple::sfURI));
+
+    return nfts;
 }
--- a/src/etl/NFTHelpers.h
+++ b/src/etl/NFTHelpers.h
@@ -1,0 +1,36 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <backend/DBHelpers.h>
+
+#include <ripple/protocol/STTx.h>
+#include <ripple/protocol/TxMeta.h>
+
+// Pulling from tx via ReportingETL
+std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
+getNFTDataFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx);
+
+// Pulling from ledger object via loadInitialLedger
+std::vector<NFTsData>
+getNFTDataFromObj(
+    std::uint32_t const seq,
+    std::string const& key,
+    std::string const& blob);
--- a/src/etl/ReportingETL.cpp
+++ b/src/etl/ReportingETL.cpp
@@ -21,6 +21,8 @@
 #include <ripple/beast/core/CurrentThreadName.h>
 
 #include <backend/DBHelpers.h>
+
+#include <etl/NFTHelpers.h>
 #include <etl/ReportingETL.h>
 #include <log/Logger.h>
 #include <subscriptions/SubscriptionManager.h>
@@ -73,7 +75,7 @@ ReportingETL::insertTransactions(
         ripple::TxMeta txMeta{
             sttx.getTransactionID(), ledger.seq, txn.metadata_blob()};
 
-        auto const [nftTxs, maybeNFT] = getNFTData(txMeta, sttx);
+        auto const [nftTxs, maybeNFT] = getNFTDataFromTx(txMeta, sttx);
         result.nfTokenTxData.insert(
             result.nfTokenTxData.end(), nftTxs.begin(), nftTxs.end());
         if (maybeNFT)
--- a/src/rpc/handlers/NFTInfo.cpp
+++ b/src/rpc/handlers/NFTInfo.cpp
@@ -24,89 +24,12 @@
 #include <backend/BackendInterface.h>
 #include <rpc/RPCHelpers.h>
 
-// {
-//   nft_id: <ident>
-//   ledger_hash: <ledger>
-//   ledger_index: <ledger_index>
-// }
-
 namespace RPC {
-
-std::variant<std::monostate, std::string, Status>
-getURI(Backend::NFT const& dbResponse, Context const& context)
-{
-    // Fetch URI from ledger
-    // The correct page will be > bookmark and <= last. We need to calculate
-    // the first possible page however, since bookmark is not guaranteed to
-    // exist.
-    auto const bookmark = ripple::keylet::nftpage(
-        ripple::keylet::nftpage_min(dbResponse.owner), dbResponse.tokenID);
-    auto const last = ripple::keylet::nftpage_max(dbResponse.owner);
-
-    ripple::uint256 nextKey = last.key;
-    std::optional<ripple::STLedgerEntry> sle;
-
-    // when this loop terminates, `sle` will contain the correct page for
-    // this NFT.
-    //
-    // 1) We start at the last NFTokenPage, which is guaranteed to exist,
-    // grab the object from the DB and deserialize it.
-    //
-    // 2) If that NFTokenPage has a PreviousPageMin value and the
-    // PreviousPageMin value is > bookmark, restart loop. Otherwise
-    // terminate and use the `sle` from this iteration.
-    do
-    {
-        auto const blob = context.backend->fetchLedgerObject(
-            ripple::Keylet(ripple::ltNFTOKEN_PAGE, nextKey).key,
-            dbResponse.ledgerSequence,
-            context.yield);
-
-        if (!blob || blob->size() == 0)
-            return Status{
-                RippledError::rpcINTERNAL,
-                "Cannot find NFTokenPage for this NFT"};
-
-        sle = ripple::STLedgerEntry(
-            ripple::SerialIter{blob->data(), blob->size()}, nextKey);
-
-        if (sle->isFieldPresent(ripple::sfPreviousPageMin))
-            nextKey = sle->getFieldH256(ripple::sfPreviousPageMin);
-
-    } while (sle && sle->key() != nextKey && nextKey > bookmark.key);
-
-    if (!sle)
-        return Status{
-            RippledError::rpcINTERNAL, "Cannot find NFTokenPage for this NFT"};
-
-    auto const nfts = sle->getFieldArray(ripple::sfNFTokens);
-    auto const nft = std::find_if(
-        nfts.begin(),
-        nfts.end(),
-        [&dbResponse](ripple::STObject const& candidate) {
-            return candidate.getFieldH256(ripple::sfNFTokenID) ==
-                dbResponse.tokenID;
-        });
-
-    if (nft == nfts.end())
-        return Status{
-            RippledError::rpcINTERNAL, "Cannot find NFTokenPage for this NFT"};
-
-    ripple::Blob const uriField = nft->getFieldVL(ripple::sfURI);
-
-    // NOTE this cannot use a ternary or value_or because then the
-    // expression's type is unclear. We want to explicitly set the `uri`
-    // field to null when not present to avoid any confusion.
-    if (std::string const uri = std::string(uriField.begin(), uriField.end());
-        uri.size() > 0)
-        return uri;
-    return std::monostate{};
-}
 
 Result
 doNFTInfo(Context const& context)
 {
-    auto request = context.params;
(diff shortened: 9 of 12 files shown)


ledhed2222:
Todo should test minting and burning same token within same ledger on a standalone to see if possible and what happens

src/etl/ETLSource.cpp
```diff
@@ -673,6 +674,8 @@ class AsyncCallData
                         request_.ledger().sequence(),
                         std::string{obj.key()});
                 lastKey_ = obj.key();
+                backend.writeNFTs(std::move(getNFTDataFromObj(
+                    request_.ledger().sequence(), obj.key(), obj.data())));
```
ledhed2222: what about pg?

src/backend/CassandraBackend.cpp
```diff
@@ -575,18 +588,26 @@ CassandraBackend::fetchNFT(
...
-    CassandraStatement statement{selectNFT_};
-    statement.bindNextBytes(tokenID);
-    statement.bindNextInt(ledgerSequence);
-    CassandraResult response = executeAsyncRead(statement, yield);
-    if (!response)
+    CassandraStatement nftStatement{selectNFT_};
+    nftStatement.bindNextBytes(tokenID);
+    nftStatement.bindNextInt(ledgerSequence);
+    CassandraResult nftResponse = executeAsyncRead(nftStatement, yield);
+    if (!nftResponse)
```
ledhed2222, line 630: reviewers - i have questions about the syntax of this function now

src/etl/ETLSource.cpp
```diff
@@ -673,6 +674,8 @@ class AsyncCallData
                         request_.ledger().sequence(),
                         std::string{obj.key()});
                 lastKey_ = obj.key();
+                backend.writeNFTs(std::move(getNFTDataFromObj(
```
godexsoft: Here `std::move` prevents copy elision and best be removed

src/etl/NFTHelpers.h
```diff
@@ -0,0 +1,15 @@
+#include <ripple/protocol/STTx.h>
+#include <ripple/protocol/TxMeta.h>
+
+#include <backend/DBHelpers.h>
+
+// Pulling from tx via ReportingETL
+std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
```
godexsoft, line 28: Have you considered using a `struct` instead?

src/etl/NFTHelpers.cpp
```diff
@@ -342,7 +346,7 @@ getNFTokenCreateOfferData(
 }
 
 std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
```
godexsoft, line 365: Hm, I see now that it was there before. Maybe we can refactor it to use a `struct` or at least use a `typedef`?

godexsoft (changes requested):
Leaving a couple comments. Overall looks good! Feel free to use `auto` more :+1: 
If you have questions re `yield` and the function that uses it - let's chat on slack.

src/etl/ETLSource.cpp
godexsoft: Postgres (if that's what you mean) support is no longer a thing (see develop branch) so you don't need to think about it no more :)

src/backend/Types.h
```diff
@@ -64,6 +64,7 @@ struct NFT
     ripple::uint256 tokenID;
     std::uint32_t ledgerSequence;
     ripple::AccountID owner;
+    std::optional<Blob> uri;
```
cjcobb23: why optional?

src/backend/CassandraBackend.cpp
cjcobb23: what questions?

src/backend/CassandraBackend.cpp
```diff
@@ -575,18 +593,26 @@ CassandraBackend::fetchNFT(
...
-    result.owner = response.getBytes();
-    result.isBurned = response.getBool();
+    result.ledgerSequence = nftResponse.getUInt32();
+    result.owner = nftResponse.getBytes();
+    result.isBurned = nftResponse.getBool();
+
+    // now fetch URI
+    CassandraStatement uriStatement{selectNFTURI_};
+    uriStatement.bindNextBytes(tokenID);
+    CassandraResult uriResponse = executeAsyncRead(uriStatement, yield);
```
cjcobb23: To reduce latency, you could really execute both async reads in parallel. Not sure how much of a difference that would make. How about we throw some timers in here, and log the duration, to see how much time each of these takes?

src/backend/CassandraBackend.cpp
```diff
@@ -1391,8 +1417,9 @@ CassandraBackend::open(bool readOnly)
               << "issuer_nf_tokens"
               << "  ("
               << "    issuer blob,"
+              << "    token_taxon bigint,"
               << "    token_id blob,"
-              << "    PRIMARY KEY (issuer, token_id)"
+              << "    PRIMARY KEY (issuer, token_taxon, token_id)"
```
cjcobb23: So now that we are changing this, what happens if the table already exists? Does clio crash? We probably need some easy way for people to update their database. Maybe a simple script?

src/etl/NFTHelpers.cpp
```diff
@@ -368,3 +372,30 @@ getNFTData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
...
+std::vector<NFTsData>
+getNFTDataFromObj(
+    std::uint32_t const seq,
+    std::string const& key,
+    std::string const& blob)
+{
+    std::vector<NFTsData> nfts;
+    ripple::STLedgerEntry const sle = ripple::STLedgerEntry(
+        ripple::SerialIter{blob.data(), blob.size()},
+        ripple::uint256::fromVoid(key.data()));
```
cjcobb23, line 402: We can probably avoid deserializing the whole object before checking the type. The type is usually encoded somewhere in the first few bytes

src/etl/NFTHelpers.cpp
```diff
@@ -368,3 +372,30 @@ getNFTData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
...
+
+    if (sle.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltNFTOKEN_PAGE)
+        return nfts;
+
+    auto const owner = ripple::AccountID::fromVoid(key.data());
```
cjcobb23, line 407: how does this work? isn't this the key to the nft token page?

src/rpc/handlers/NFTInfo.cpp
```diff
@@ -99,37 +23,28 @@ doNFTInfo(Context const& context)
...
-        if (Status const* status = std::get_if<Status>(&maybeURI); status)
-            return *status;
-        // A URI was found
-        if (std::string const* uri = std::get_if<std::string>(&maybeURI); uri)
-            response["uri"] = *uri;
-        // A URI was not found, explicitly set to null
-        else
-            response["uri"] = nullptr;
-    }
+    response[JS(nft_serial)] = ripple::nft::getSerial(dbResponse->tokenID);
```
cjcobb23, line 63: is this a field name change from before?

src/etl/ReportingETL.cpp
```diff
@@ -76,7 +78,7 @@ ReportingETL::insertTransactions(
         result.nfTokensData.end(),
         [](NFTsData const& a, NFTsData const& b) {
             return a.tokenID > b.tokenID &&
-                a.transactionIndex > b.transactionIndex;
+                a.transactionIndex.value_or(0) > b.transactionIndex.value_or(0);
```
cjcobb23: why this change?

src/etl/ETLSource.cpp
cjcobb23: I think we should probably check whether this is an nft token dir before calling this function. Look at `isOfferDir` as an example

cjcobb23 (changes requested):
Mostly looks good. Left some small comments, but main thing we need is a migration script for people to run to drop the old tables, and back up etl to right before nfts went live

src/backend/CassandraBackend.cpp
godexsoft: We need to support migrations.. it's kinda a bit late tho. Any good DB wrapper/library supports migrations in one way or the other. Worth checking if datastax provides anything at all in their cpp library. If they don't then we need a new table that will hold the versions of each table and some support in code to apply changes to the table structure when migrating from version X to version Y. and for each update of the table we increment the version and add a migration.

cjcobb23: We do, though I'm not sure if a db wrapper will always be able to do what we want. There are things that would require preprocessing the ledger data in a way the dbwrapper cannot do itself. Do you think a db wrapper would be able to do what we have done in this PR? It seems to me that we just need to rerun ETL here.

godexsoft: I'm not sure i understand your concern. If you support migrations you can run arbitrary queries (and c++ code) against the DB when launching first time with DB of version XXX and `clio` that needs DB of version YYY. So you can literally do anything with the DB and/or run any C++ code you need to support the migration.

cjcobb23: oh i see, I misunderstood. yes if the dbwrapper can just run arbitrary c++ code, we can probably support arbitrarily complex migrations, at least anything that can be done solely with the data already in the database. I was thinking the migration would need to happen exclusively server side on the database itself.

ledhed2222: see script above

src/backend/Types.h
ledhed2222: NFTs don't have to have a URI at all, and `nf_token_uris` only stores a uri if there is one - IE the table is sparse

src/etl/ETLSource.cpp
ledhed2222: why @cjcobb23? that check happens within the `getNFTDataFromObj` function, which will return an empty vector if it's not an NFT page. then `writeNFTs` will not loop over anything

src/etl/NFTHelpers.cpp
ledhed2222: i think we should tackle this later

src/etl/NFTHelpers.cpp
ledhed2222: oh nice - i was bummed about having to deserialize just to do this as well

src/etl/NFTHelpers.cpp
ledhed2222: this is how object IDs work for NFTokenPages. The first 160 bits are the owner's account, to save space

src/etl/NFTHelpers.h
ledhed2222: let's do this refactor later

src/etl/ReportingETL.cpp
ledhed2222: because now `transactionIndex` is optional, since we cannot determine that from NFTokenPages directly. In this case however we are parsing from transactions, so the transactionIndex should always be there, otherwise something weird has happened. but i thought it would be safer to just default to 0 rather than crash

src/rpc/handlers/NFTInfo.cpp
ledhed2222: yes, to match what we have in `account_nfts`. see PR description

src/backend/CassandraBackend.cpp
cjcobb23: the script doesn't have a concept of versions, maybe we can add that in

src/etl/NFTHelpers.cpp
```diff
@@ -368,3 +372,30 @@ getNFTData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
...
+    for (ripple::STObject const& node : sle.getFieldArray(ripple::sfNFTokens))
+    {
+        std::optional<ripple::Blob> uri;
+        if (node.isFieldPresent(ripple::sfURI))
+            uri = node.getFieldVL(ripple::sfURI);
+        nfts.emplace_back(
+            node.getFieldH256(ripple::sfNFTokenID), seq, owner, uri);
```
cjcobb23: you can probably move `owner` and `uri` to be more efficient (@godexsoft double check me please). This function will be called a lot during the initial ledger download. A little speed up will go a long way here.

src/rpc/handlers/NFTInfo.cpp
cjcobb23: I think this is a no-go at this point, nft_info is already live and being used by numerous people, lets not break their script.

src/etl/Migrations.cpp
```diff
@@ -0,0 +1,131 @@
...
+        throw std::runtime_error(ss.str());
+        // data.isBurned = true;
+        // toWrite.push_back(data);
+    }
+
+    // write the new data
+    backend->writeNFTs(std::move(toWrite));
+
+    // write the migration receipt
+    backend->writeMigrationReceipt(__func__);
```
godexsoft: Please make sure that `__func__` returns the same string on all supported compilers (it should, but better check).

src/etl/Migrations.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+void
+execute(boost::asio::io_context& ioc, std::shared_ptr<BackendInterface> backend)
+{
+    boost::asio::spawn(ioc, [backend](boost::asio::yield_context yield) {
+        nftMigration20221123(backend, yield);
```
godexsoft: Is the plan to just keep adding migration functions here preserving the order?

src/backend/CassandraBackend.cpp
```diff
@@ -1386,6 +1461,28 @@ CassandraBackend::open(bool readOnly)
         if (!executeSimpleStatement(query.str()))
             continue;
 
+        query.str("");
+        query << "CREATE INDEX IF NOT EXISTS " << tablePrefix
+              << "index_issuer_nf_tokens_taxon ON " << tablePrefix
+              << "issuer_nf_tokens (issuer, token_taxon)";
+        if (!executeSimpleStatement(query.str()))
+            continue;
```
cjcobb23: We should not have an index. Indexes in Cassandra have serious performance implications, because querying the index requires contacting every node in the cluster.

src/backend/CassandraBackend.cpp
```diff
@@ -1540,13 +1652,39 @@ CassandraBackend::open(bool readOnly)
         if (!selectNFT_.prepareStatement(query, session_.get()))
             continue;
 
+        query.str("");
+        query << "SELECT DISTINCT token_id FROM " << tablePrefix << "nf_tokens";
+        if (!selectAllNFTIDs_.prepareStatement(query, session_.get()))
+            continue;
```
cjcobb23: I think this query needs to be paginated. This could end up being a lot of token ids, and might be too large of a result set.

cjcobb23 (changes requested):

ledhed2222:
Index is already removed and query is already paginated - just haven’t
pushed in a while 🤘

src/backend/CassandraBackend.cpp
```diff
@@ -598,18 +624,30 @@ CassandraBackend::fetchNFT(
...
+    result.isBurned = nftResponse.getBool();
+
+    // now fetch URI
+    CassandraStatement uriStatement{selectNFTURI_};
+    uriStatement.bindNextBytes(tokenID);
+    uriStatement.bindNextInt(ledgerSequence);
+    CassandraResult uriResponse = executeAsyncRead(uriStatement, yield);
+    if (!uriResponse)
+    {
+        // TODO what should we do here?
```
ledhed2222: to discuss

src/backend/CassandraBackend.cpp
```diff
@@ -994,6 +1050,36 @@ CassandraBackend::isTooBusy() const
     return numReadRequestsOutstanding_ >= maxReadRequestsOutstanding;
 }
 
+// TODO need help getting this implementation defined on the backend interface
+// itself, where it should belong.
```
ledhed2222: to discuss

src/rpc/handlers/NFTInfo.cpp
```diff
@@ -119,37 +42,31 @@ doNFTInfo(Context const& context)
...
+    response[JS(nft_id)] = ripple::strHex(dbResponse->tokenID);
+    response[JS(ledger_index)] = dbResponse->ledgerSequence;
+    response[JS(owner)] = ripple::toBase58(dbResponse->owner);
     response["is_burned"] = dbResponse->isBurned;
+    if (dbResponse->uri)
+        response[JS(uri)] = ripple::strHex(dbResponse->uri.value());
+    else
+    {
+        // TODO what should we do here?
+        response[JS(uri)] = nullptr;
```
ledhed2222: discuss

src/backend/CassandraBackend.cpp
ledhed2222: probably can just leave this null. want to check with an empty string input in an NFT

src/backend/CassandraBackend.cpp
ledhed2222: try wrapping the iocontext in a strand? spawn once and loop inside

src/backend/BackendFactory.h
```diff
@@ -57,6 +57,10 @@ make_Backend(boost::asio::io_context& ioc, clio::Config const& config)
 
     log.info() << "Constructed BackendInterface Successfully";
 
+    log.info() << __func__ << ": Running migrations";
+    backend->runMigrations();
```
godexsoft: Not too sure this belongs in the factory function. Migrations should be attempted regardless of how the backend is created, right?

src/backend/BackendInterface.h
```diff
@@ -595,6 +603,9 @@ class BackendInterface
     virtual void
     startWrites() const = 0;
 
+    virtual void
```
godexsoft: The name `sync` does not really explain what this does. Consider adding doxygen comments to all public member functions, especially in the common interface. 
It also may be possible to have a more meaningful name for this function.. `sync` has a very broad meaning.

src/backend/CassandraBackend.cpp
```diff
@@ -17,15 +17,17 @@
...
 #include <backend/CassandraBackend.h>
 #include <backend/DBHelpers.h>
+#include <backend/migrations/cassandra/Cassandra.h>
 #include <log/Logger.h>
 #include <util/Profiler.h>
 
 #include <functional>
 #include <unordered_map>
 
+#include <ripple/app/tx/impl/details/NFTokenUtils.h>
```
godexsoft: If you wanna regroup the includes (which is good), use the following convention please:
1st group: local files (in the project)
2nd group: 3rd party libs (ripple/*, boost/*, etc.)
3rd group: standard libraries

src/backend/CassandraBackend.cpp
```diff
@@ -405,17 +407,41 @@ CassandraBackend::writeNFTs(std::vector<NFTsData>&& data)
...
-            },
-            "issuer_nf_tokens");
+        // If `uri` is set (and it can be set to an empty uri), we know this
+        // is a net-new NFT. That is, this NFT has not been seen before by us
+        // _OR_ it is in the extreme edge case of a re-minted NFT ID with the
+        // same NFT ID as an already-burned token. In this case, we need to
+        // record the URI and link to the issuer_nf_tokens table.
+        if (record.uri)
+        {
+            makeAndExecuteAsyncWrite(
```
godexsoft, line 416: Is it important or can it matter which record is inserted first?

Keep in mind that you are spawning two async operations and they (likely) could execute in any order. We don't seem to provide any safety for ordering and rely on cassandra driver's `future` deep inside.

1) We could use `executeSyncWrite` directly as we are already in CassandraBackend.
2) There is something that we could use (other than synchronizing by hand somehow) if needed [described here](https://datastax.github.io/cpp-driver/api/struct.CassBatch) but that would mean we need to implement more stuff in CassandraBackend as we currently don't have support for proper batches.

src/backend/CassandraBackend.cpp
```diff
@@ -994,6 +1050,36 @@ CassandraBackend::isTooBusy() const
     return numReadRequestsOutstanding_ >= maxReadRequestsOutstanding;
 }
 
+// TODO need help getting this implementation defined on the backend interface
+// itself, where it should belong.
+void
+CassandraBackend::runMigrations()
+{
+    for (auto const& migration : Migrations::Cassandra::MIGRATIONS)
```
godexsoft: I'd not stress right now about how this is here instead of BackendInterface - we are going to fully rewrite the whole `backend` and start from Cassandra pretty soon.

godexsoft: As long as the DB model is good, the implementation details can be later changed in any direction we'd like.

src/backend/CassandraBackend.cpp
```diff
@@ -994,6 +1050,36 @@ CassandraBackend::isTooBusy() const
...
+    {
+        // Note, we need to pass migration by value otherwise it will have
+        // been claimed off the stack by the time this runs
+        boost::asio::spawn(
+            ioContext_, [this, migration](boost::asio::yield_context yield) {
```
godexsoft: You will not have this problem with capturing `migration` if you run the whole loop inside of the lambda instead.

src/backend/CassandraBackend.cpp
```diff
@@ -994,6 +1050,36 @@ CassandraBackend::isTooBusy() const
...
+                    log_.info() << "Beginning migration " << migration.name();
+                    migration(*this, yield);
+                    sync();
+                    writeMigrationReceipt(migration.name());
+                    sync();
+                    log_.info() << "Completed migration " << migration.name();
+                }
+            });
+
+        ioContext_.run_one();
```
godexsoft: This probably makes the whole application hang for the duration of `one` handler running (assuming this is called from the main thread). Ideally we would not "hang" but rather "block" in a more sophisticated way (e.g. with some sort of flag/condition) so that other parts are at least responsive while paused.

src/backend/CassandraBackend.cpp
```diff
@@ -1388,17 +1474,37 @@ CassandraBackend::open(bool readOnly)
 
         query.str("");
         query << "CREATE TABLE IF NOT EXISTS " << tablePrefix
-              << "issuer_nf_tokens"
+              << "issuer_nf_tokens_v2"
```
godexsoft, line 1433: I'm not too sure about this design of migrations in general.

I'd keep the name of the table as before. If you already have one with an older schema - you will run the migration for it and update the schema (and populate the new fields etc.). Or you will create a tmp table, move data, populate whatever needed and rename to the old name while dropping the old table.

Migrations therefore should run somewhere between attempting to create all the tables (with the latest schema) and preparing the statements for later as statements will rely on the schema and fail to 'prepare' if the query and schema mismatch.
When you create the tables you specify `CREATE TABLE IF NOT EXISTS` so if the name is already taken we will keep the old schema until the migrations run (next step).

You also need some sort of `constexpr static` version of the schema that comes with the current version of `clio` code. Then you run all migrations that are relevant from whatever version of schema you have in a system db table and up to the current version in that `constexpr static` constant. 

I see your design is very different. If you are sure it will be able to do everything then we can try to keep it as is but you will have to explain to me that all paths are covered :)

src/backend/CassandraBackend.cpp
```diff
@@ -1423,6 +1529,21 @@ CassandraBackend::open(bool readOnly)
         if (!executeSimpleStatement(query.str()))
             continue;
 
+        query.str("");
+        query << "CREATE TABLE IF NOT EXISTS " << tablePrefix
+              << "clio_migrations"
```
godexsoft: We are in the `clio` namespace, right? We probably can call the tables `migrations` instead. Or even something like `__migrations` if you want to show that these are important system tables.

src/backend/CassandraBackend.cpp
```diff
@@ -1558,12 +1679,28 @@ CassandraBackend::open(bool readOnly)
             continue;
 
         query.str("");
-        query << "INSERT INTO " << tablePrefix << "issuer_nf_tokens"
-              << " (issuer,token_id)"
-              << " VALUES (?,?)";
+        query << "INSERT INTO " << tablePrefix << "issuer_nf_tokens_v2"
```
godexsoft, line 1623: I'm not too sure how to approach this. The table name here should be the old name. But the query depends on the current version of the schema.. and we did not yet apply migrations at this point so we would have old schema on first run here. The prepared statement would then fail (i assume). 

Basically it looks like these statements need to be prepared **before** we start working with the DB from `clio` (rpc etc.) but **after** we ran the migrations and got to the latest DB schema.

src/backend/CassandraBackend.cpp
godexsoft: In the `version` based design we also only need to store one value at all times, not a list of migrations that were applied. Perhaps there is already some common way to store this sort of thing.

src/backend/CassandraBackend.cpp
godexsoft: Very important is to think how we can do zero downtime migration with either of the designs. It's still an open question.

src/backend/CassandraBackend.h
```diff
@@ -152,6 +153,25 @@ class CassandraStatement
         curBindingIndex_++;
     }
 
+    void
+    bindNextString(std::string const& data)
+    {
+        if (!statement_)
```
godexsoft: Not your fault but this is basically duplicated in each function here.. we should extract this into a helper

src/backend/CassandraBackend.h
```diff
@@ -152,6 +153,25 @@ class CassandraStatement
...
+            throw std::runtime_error(
+                "CassandraStatement::bindNextString - statement_ is null");
+        CassError rc = cass_statement_bind_string(
+            statement_, curBindingIndex_, data.c_str());
+        if (rc != CASS_OK)
```
godexsoft: And this whole block of course too :/

src/backend/migrations/Migrations.h
```diff
@@ -0,0 +1,37 @@
+#ifndef MIGRATION_TEMPLATE_H_INCLUDED
```
godexsoft: Your branch go so far behind `develop` 🤣 
You really need to rebase on latest `develop` asap.
We use `#pragma once` instead of this everywhere since a long time already.

godexsoft: Also you will see when you rebase but we added a header comment with copyright info, like in rippled.

src/backend/migrations/Migrations.h
```diff
@@ -0,0 +1,37 @@
...
+#define MIGRATION_TEMPLATE_H_INCLUDED
+
+namespace Backend {
+namespace Migrations {
+
+template <typename R>
+class Migration
+{
+private:
```
godexsoft: We usually omit this `private:` in classes since default visibility is already `private`.

src/backend/migrations/cassandra/20221123_Issuer_NFTokens.cpp
```diff
@@ -0,0 +1,154 @@
...
+    cass_statement_free(nftTxQuery);
+
+    /*
+     * Step 2 - Pull every object from our initial ledger and load all NFTs
+     * found in any NFTokenPage object. Prior to this migration, we were not
+     * pulling out NFTs from the initial ledger, so all these NFTs would be
+     * missed. This will also record the URI of any NFTs minted prior to the
+     * start sequence.
+     */
+    // TODO need to test on a fresh DB. does it short circuit?
```
godexsoft: Did you test yet and what does it mean exactly?

src/backend/migrations/cassandra/Cassandra.h
```diff
@@ -0,0 +1,23 @@
+#include <backend/CassandraBackend.h>
+#include <backend/migrations/Migrations.h>
+
+namespace Backend {
+namespace Migrations {
+namespace Cassandra {
+
+// IMPORTANT! The names of each migration function cannot change. If they do, it
```
godexsoft: That's also one reason to use version number instead and have `static constexpr uint32_t version = 2;` or something like this here and a corresponding value in the database (representing the actual current schema).

src/backend/migrations/cassandra/Cassandra.h
```diff
@@ -0,0 +1,23 @@
...
+// will cause the migration to be run again.
+
+Migration<CassandraBackend>
+issuerNFTokens20221123();
+
+std::vector<Migration<CassandraBackend>> const MIGRATIONS = {
```
godexsoft: Very scary indeed.

src/etl/ReportingETL.cpp
godexsoft: You should check for yourself, but i think if you compare two optionals (without .value()) it would not crash. There are some rules how they are evaluated but the result should be pretty much same (confirm for yourself on godbolt.org if that's exactly what you would expect or not - don't take my word for it).

godexsoft (changes requested):
First pass as promised. A lot of comments, we can discuss on Monday probably.

src/backend/CassandraBackend.cpp
cjcobb23: just lurking here: you don't want to execute writes synchronously as part of ETL. Synchronous writes are a round trip to the database, which can be on the order of a few milliseconds. If you have a bunch of sync writes, that adds up, and would dramatically slow down ETL. The pattern for ETL is to submit a bunch of asynchronous writes, and wait for them all to finish, then a single synchronous write/update to mark the entire ledger as ready to be read.

ledhed2222: no, it doesn't matter which is inserted first. they are not tightly coupled when read, just that we know that whenever `nf_token_uris` is written we must also write to `issuer_nf_tokens_v2`. so writing async here for reach should definitely still be ok

src/backend/BackendFactory.h
ledhed2222: that's exactly why i put it in the factory. otherwise were you thinking of putting it in the BackendInterface constructor? In that case, isn't it called before the child constructor? that wouldn't work.

src/backend/BackendInterface.h
ledhed2222: yeah, but i think this can be done in another PR/is out of scope for this. agree?

src/backend/CassandraBackend.cpp
ledhed2222: i mean it's actually the `nf_tokens` and `nf_token_uris` that _are_ tightly coupled when read, but I'm not really worried about any user encountering some weird case there. though perhaps that means the `fetchNFT` should return an error/raise if it doesn't find a URI

src/backend/CassandraBackend.cpp
```diff
@@ -695,6 +733,24 @@ CassandraBackend::fetchNFTTransactions(
...
+    query.bindNextString(migration);
+    return executeAsyncRead(query, yield).hasResult();
+}
+
+void
+CassandraBackend::writeMigrationReceipt(std::string const& migration) const
+{
+    CassandraStatement statement(insertClioMigration_);
+    statement.bindNextString(migration);
+    executeSyncWrite(statement);
```
ledhed2222: perhaps this should be async too?

src/backend/CassandraBackend.cpp
```diff
@@ -994,6 +1050,36 @@ CassandraBackend::isTooBusy() const
...
+                if (fetchIsMigrationExecuted(migration.name(), yield))
+                    log_.info() << "Skipping migration " << migration.name()
+                                << " as it has already run";
+                else
+                {
+                    log_.info() << "Beginning migration " << migration.name();
+                    migration(*this, yield);
+                    sync();
+                    writeMigrationReceipt(migration.name());
+                    sync();
```
ledhed2222: this can be removed if we're sticking with a sync write for the recipt

src/backend/CassandraBackend.cpp
ledhed2222: yeah let's leave this as an open question. i'm not really sure what the best practice is for zero downtime migrations in cassandra specifically, as i'm used to postgres. adding columns to tables in postgres is trivial, but cassandra is pickier as you know

src/backend/CassandraBackend.cpp
ledhed2222: i called this out in the README changes. it's prefixed with `clio_` to indicate that it's data internal to the application. leading `_` would work too, assuming that's a legal table name in cassandra.

src/backend/CassandraBackend.cpp
ledhed2222: correct. this is why, unless you can call the BackendInterface constructor at the end of the CassandraBackend constructor, I think the best place to put the call to `runMigrations` is in the factory. I don't like the former approach even if it is possible, because it creates a trap for future developers of new backends who have to _remember_ to call the BackendInterface constructor in their constructor.

src/backend/CassandraBackend.h
ledhed2222: agreed and i think well out of scope for this pr

src/backend/migrations/cassandra/20221123_Issuer_NFTokens.cpp
ledhed2222: have not tested that condition yet. there are a few conditions i need to test, but it takes a really fucking long time to seed the old data using `develop` before it's ready for a test run of the migration. i'm in the process of seeding all  the NFT data back to october to test this. i think we should continue with the code review and other parts while i'm making progress there

src/backend/BackendFactory.h
godexsoft: I think we should leave migrations entirely to each implementation instead of artificially trying to go thru BackendInterface. 
Each DB will have completely different migration system, schema and everything else specific to the DB. Furthermore we support Cassandra today with Schema `1`, tomorrow it will be `2` and next year we support another Database entirely which will start with schema `0` (which already maps to schema `2` of Cassandra at that point). 
In any case - factory is never the right place for any logic other than creation of objects.

src/backend/BackendInterface.h
godexsoft: Disagree. Each PR should not make the codebase worse/less readable than it already is.

src/backend/CassandraBackend.cpp
godexsoft: I'm not sure anything should be async when performing migrations..

src/backend/CassandraBackend.cpp
godexsoft: I actually don't know. You were investigating this as part of migrations i assume so i believe you when you say that. I'd imagine we should be still able to alter a table and add some column, is that not the case?

src/backend/CassandraBackend.cpp
godexsoft: I don't mind. We just need to settle on something and move on :)

src/backend/CassandraBackend.cpp
godexsoft: No no no :D just leave migrations to the actual backend, not the interface.

src/rpc/handlers/NFTInfo.cpp
godexsoft: You could add both for now and later we can have a process to formally remove the former over a period of releases with deprecation warnings in changelog etc.

src/backend/CassandraBackend.cpp
cjcobb23: The user should never be able to see one of these writes and not the other. The server should not let users read from ledgers that are not fully written. This is what the ledger_range table is for, to be able to mark a ledger as fully written and ready to be read from.

src/backend/CassandraBackend.cpp
cjcobb23: This makes sense to be synchronous. It's a single write, done at the end of the migration. I'd imagine you also want the semantics of knowing that the receipt is viewable by all readers.

src/backend/CassandraBackend.cpp
ledhed2222: make new set of backend errors, with error for this, throw it, catch it in RPC and return something to user

src/backend/Types.h
ledhed2222: make it not optional and throw an error in the fetcher

src/rpc/handlers/NFTInfo.cpp
ledhed2222: spoke about this w alex. because there's already a breaking change in the API wrt the `uri` field (it's no longer hex-decoded to resolve an unrelated issue) we think we should just make these other API changes while we're here anyway.

src/backend/BackendFactory.h
ledhed2222: move this into CassandraBackend somewhere

src/backend/CassandraBackend.cpp
```diff
@@ -598,18 +625,29 @@ CassandraBackend::fetchNFT(
...
+    result.owner = nftResponse.getBytes();
+    result.isBurned = nftResponse.getBool();
+
+    // now fetch URI
+    CassandraStatement uriStatement{selectNFTURI_};
+    uriStatement.bindNextBytes(tokenID);
+    uriStatement.bindNextInt(ledgerSequence);
+    CassandraResult uriResponse = executeAsyncRead(uriStatement, yield);
+    if (!uriResponse)
+        throw UnexpectedDataError(
```
ledhed2222: note to reviewers - this would legitimately mean bad data. hence the error condition.

src/rpc/handlers/NFTInfo.cpp
```diff
@@ -119,37 +43,36 @@ doNFTInfo(Context const& context)
...
+    {
+        dbResponse =
+            context.backend->fetchNFT(tokenID, lgrInfo.seq, context.yield);
+    }
+    catch (Backend::UnexpectedDataError const& e)
+    {
+        return Status{
+            RippledError::rpcDB_DESERIALIZATION,
+            "NFT has no URI recorded. Try again in a moment and this error "
+            "still persists open an issue as this is unexpected"};
```
ledhed2222: note reviewers - please look at this wording. i don't expect this error so i would legitimately want an issue opened. but not sure what you think in terms of verbiage.

src/backend/Errors.h
```diff
@@ -0,0 +1,38 @@
...
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+namespace Backend {
+
+class UnexpectedDataError : public std::exception
```
ledhed2222: IMO reviewers we should have a baseline `Clio::Exception` that all our errors inherit from. that feels like good practice to me.

src/backend/CassandraBackend.cpp
ledhed2222: i checked the possibility of this again, for a third time lol. no, it is not possible to change the PK of an existing table (which makes sense fo course). the best you can do is exactly what I'm doing: create a new table, copy the data from the old table into the new table, and drop the old table.

if we wanted to we could then rename the new table to remove the `v2`, but honestly i think that's going to be more trouble than it's worth.

reference: https://stackoverflow.com/a/22140030/568390

src/backend/CassandraBackend.h
```diff
@@ -702,6 +723,12 @@ class CassandraBackend : public BackendInterface
     mutable std::uint32_t ledgerSequence_ = 0;
 
 public:
+    std::string
+    tablePrefix() const
+    {
+        return config_.valueOr<std::string>("table_prefix", "");
+    }
```
ledhed2222: i made this a public method back when i needed this for the migration. but i'm leaving it here because it seems totally reasonable to me for this to be a public method here.

src/backend/CassandraBackend.cpp
```diff
@@ -598,18 +624,46 @@ CassandraBackend::fetchNFT(
...
+                " has no URI recorded and is not burned");
+        auto const ledgerRange = hardFetchLedgerRange(yield);
+        if (result.ledgerSequence != ledgerRange->minSequence)
+            throw UnexpectedDataError(
+                "NFT " + ripple::strHex(tokenID) +
+                " has no URI recorded and was not burned in the first ledger "
+                "recorded on this server");
+
+        // In the special allowed case, we leave the URI field as an empty
+        // blob
```
ledhed2222: add TODO comment about injecting warning for RPC

src/backend/CassandraBackend.cpp
```diff
@@ -598,18 +624,46 @@ CassandraBackend::fetchNFT(
...
+    CassandraResult uriResponse = executeAsyncRead(uriStatement, yield);
+    if (!uriResponse)
+    {
+        if (!result.isBurned)
+            throw UnexpectedDataError(
+                "NFT " + ripple::strHex(tokenID) +
+                " has no URI recorded and is not burned");
+        auto const ledgerRange = hardFetchLedgerRange(yield);
+        if (result.ledgerSequence != ledgerRange->minSequence)
+            throw UnexpectedDataError(
```
ledhed2222: get rid of errors

src/backend/CassandraBackend.cpp
```diff
@@ -1164,7 +1206,7 @@ CassandraBackend::open(bool readOnly)
     }
 
     auto rf = config_.valueOr<int>("replication_factor", 3);
-    auto tablePrefix = config_.valueOr<std::string>("table_prefix", "");
+    auto const tablePrefix = this->tablePrefix();
```
godexsoft: minor: I don't mind adding this member function but since we don't use it other than here - consider bringing back the old code - less code for me to port to the new backend later on.

src/backend/CassandraBackend.h
```diff
@@ -152,6 +152,25 @@ class CassandraStatement
         curBindingIndex_++;
     }
 
+    void
+    bindNextString(std::string const& data)
```
godexsoft: Looks like dead code leftover from migrations?

src/backend/DBHelpers.h
```diff
@@ -99,9 +130,30 @@ struct NFTsData
         , ledgerSequence(meta.getLgrSeq())
         , transactionIndex(meta.getIndex())
         , owner(owner)
+        , uri({})
```
godexsoft: This should not be needed, uri is empty optional by default.
If you really want to explicitly set it please prefer `std::nullopt`.

src/backend/DBHelpers.h
```diff
@@ -85,11 +85,42 @@ struct NFTsData
...
+        ripple::uint256 const& tokenID,
+        ripple::AccountID const& owner,
+        ripple::Blob const& uri,
+        ripple::TxMeta const& meta)
+        : tokenID(tokenID)
+        , ledgerSequence(meta.getLgrSeq())
+        , transactionIndex(meta.getIndex())
+        , owner(owner)
+        , uri(uri)
+        , isBurned(false)
```
godexsoft: minor: You could set this at line 101: `bool isBurned = false;` for it to always be `false` unless specified in the initialization list.

src/backend/DBHelpers.h
```diff
@@ -99,9 +130,30 @@ struct NFTsData
...
+        ripple::uint256 const& tokenID,
+        std::uint32_t const ledgerSequence,
+        ripple::AccountID const& owner,
+        ripple::Blob const& uri)
+        : tokenID(tokenID)
+        , ledgerSequence(ledgerSequence)
+        , transactionIndex({})
+        , owner(owner)
+        , uri(uri)
+        , isBurned(false)
```
godexsoft: Can be omitted by setting it to `false` above.

src/etl/NFTHelpers.h
```diff
@@ -0,0 +1,34 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
godexsoft, line 19: Looks like we are missing a `#pragma once` here

godexsoft (changes requested):
Left a few minor comments and suggestions. Overall looks good 👍

godexsoft (dismissed):
LGTM 👍

godexsoft:
Looks like you need to squash everything and sign the commit. After that i'll have to approve again and should be good to go.

godexsoft (approved):
LGTM 🚀

PR #315, opened 2022-09-19
Implement subscription for book_changes

### LINKED ISSUE #312: Implement book_changes subscription
In #300 `book_changes` was implemented without subscription support.
Gotta add this rpc through the subscription manager.

godexsoft:
Fixes #312 

Add subscription for book_changes.
On every ledger publication compute the book changes and send as json to subscribers.
Output looks like this (each object is individual `book_changes` response in the stream):
```
{
    "changes": [
        {
            "close": "2805072.690304475",
            "currency_a": "XRP_drops",
            "currency_b": "r9MDUnXQfdKkVrZgDxbPiXUSgvjTPh35dS/USD",
            "high": "2805072.690304475",
            "low": "2805072.690304475",
            "open": "2805072.690304475",
            "volume_a": "150000000",
            "volume_b": "53.4745500601335"
        }
    ],
    "ledger_hash": "56816200AC24CB2109473BF1583788A07202B0BFC19F97967D2D25DFC8417698",
    "ledger_index": 74502099,
    "ledger_time": 716920241,
    "type": "bookChanges"
}
{
    "changes": [
        {
            "close": "2805072.688117325",
            "currency_a": "XRP_drops",
            "currency_b": "r9MDUnXQfdKkVrZgDxbPiXUSgvjTPh35dS/USD",
            "high": "2805072.690304464",
            "low": "2805072.688117325",
            "open": "2805072.690304464",
            "volume_a": "500000000",
            "volume_b": "178.248500283835"
        }
    ],
    "ledger_hash": "FB81477276391290726BA876488BC02DBBA4237EC1CD92550B9714B86C0FCAB3",
    "ledger_index": 74502100,
    "ledger_time": 716920250,
    "type": "bookChanges"
}
{
    "changes": [],
    "ledger_hash": "F6B53E2FFE39688939780DCA06CB8ADC6EB3FBD83587096EA3983075C22613E5",
    "ledger_index": 74502101,
    "ledger_time": 716920251,
    "type": "bookChanges"
}
```

### DIFF
--- a/src/rpc/handlers/BookChanges.cpp
+++ b/src/rpc/handlers/BookChanges.cpp
@@ -12,6 +12,9 @@ using namespace ripple;
 
 namespace RPC {
 
+/**
+ * @brief Represents an entry in the book_changes' changes array.
+ */
 struct BookChange
 {
     STAmount sideAVolume;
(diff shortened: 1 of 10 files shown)


src/subscriptions/SubscriptionManager.cpp
```diff
@@ -246,6 +262,14 @@ SubscriptionManager::pubLedger(
         getLedgerPubMessage(lgrInfo, fees, ledgerRange, txnCount)));
 
     ledgerSubscribers_.publish(message);
+
+    if (not bookChangesSubscribers_.empty())
```
cjcobb23: I like using `not` and `and`, but we don't use this elsewhere in the codebase. Should we start using this now? Is this supported on most linux platforms?

src/rpc/handlers/BookChanges.cpp
```diff
@@ -24,14 +24,15 @@ struct BookChange
 
 class BookChangesHandler
```
cjcobb23: What is the reason we want this to be a class? As far as I can tell, this is only used in `computeBookChanges`, where it is constructed locally and then immediately destroyed. It doesn't actually store any internal state, and I can't see a reason we would ever want to extend this class. The only argument to its constructor is the backend, which is a singleton inside the application. So all possible instances of `BookChangesHandler` are functionally identical. I feel this class is just a wrapper around the member functions, but the member functions could just be free functions. I feel we are adding a class here just for the sake of adding a class, when free functions would probably be simpler, achieve the same result, and be more appropriate for the problem. But its up to you, that's just my two cents.

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -246,6 +262,14 @@ SubscriptionManager::pubLedger(
...
+    {
+        auto const json = RPC::computeBookChanges(lgrInfo, backend_);
+        auto const bookChangesMsg =
+            std::make_shared<Message>(boost::json::serialize(json));
+        bookChangesSubscribers_.publish(bookChangesMsg);
```
cjcobb23: I think we should add a new function to `SubscriptionManager` called `pubBookChanges`, where we pass the transactions and ledger in as arguments, and then publish. Reason being, we already fetch all of the transactions once, in order to call `pubTransaction`, and it would be more efficient to just use those fetched transactions, instead of fetching the same transactions again to compute the book changes. We fetch the transactions in `ReportingETL::publishLedger`. Fetching transactions is a network call, so it's one of the most expensive operations we perform in the whole application.

cjcobb23 (changes requested):

src/rpc/handlers/BookChanges.cpp
```diff
@@ -53,10 +54,12 @@ class BookChangesHandler
...
 
-        for (auto const transactions =
-                 context_.get().backend->fetchAllTransactionsInLedger(
-                     ledger.seq, context_.get().yield);
-             auto const& tx : transactions)
+        std::vector<Backend::TransactionAndMetadata> const transactions =
+            Backend::synchronousAndRetryOnTimeout(
+                [seq = ledger.seq, this](auto yield) {
+                    return backend_->fetchAllTransactionsInLedger(seq, yield);
+                });
```
cjcobb23: So in the publishing codepath, we want to retry on timeout, but in the regular RPC handling codepath, we don't. For a regular RPC, if the function times out, we want to just return the error. We don't have to worry about returning that error, because the timeout throws an exception, and the exception is caught and an error is returned to the client.

However, in the publishing codepath, we don't actually need to fetch the transactions at all, since we already fetched them in order to publish the transactions themselves.

src/subscriptions/SubscriptionManager.cpp
godexsoft: actually, i was not aware this was in C/C++ forever and thought this was a recent feature of the language.. but it looks like it was always there: https://en.cppreference.com/w/cpp/language/operator_alternative
i usually use `not` when it's more readable (and i remember to use it) - like in this case. I don't mind using it or not but i do think this is more readable this way.

src/rpc/handlers/BookChanges.cpp
godexsoft: yea, this sounds reasonable - i'll see how i can change it so i still reuse code where possible.
initially this retryOnTimeout creeped in because i needed a `yield` which i now did not have anymore. It seemed like this is the way we do it in the codebase from what i have seen.

src/subscriptions/SubscriptionManager.cpp
godexsoft: yea, agree. I'll change this part

src/rpc/handlers/BookChanges.cpp
godexsoft: Initially the design was more fitting than it became now but in general it's for encapsulation and readability. It does store internal state - `offerCancel_` and `tally_` some of which are accessed by different member functions (private ones so that they are defo not called by someone else, like a free function would be able to).
 
But you are not wrong thinking this "is just a wrapper around member functions" - it is. This is better than 5 free functions or one giant function in most cases imo.
I don't have a problem with free function where they make more sense than wrappers/classes/encapsulation but i don't think this is the right use case for them. 

C++ offers us a lot of flexibility and tools in the language - IMHO, we should use them :)

src/rpc/handlers/BookChanges.cpp
cjcobb23: in the publish codepath, when we call this from `ReportingETL`, we do it like this. But when making database calls in response to client requests, we have a yield context already, so we can just use that.

src/rpc/handlers/BookChanges.cpp
cjcobb23: That's a good point that the class encapsulates all of the functions, and makes the private functions not callable. What do you think about just making the member functions static, and passing around the member variables as function parameters?

src/subscriptions/SubscriptionManager.cpp
cjcobb23: I think it's more readable as well, I just want to make sure its supported on all of the linux platforms. I feel like there was some issue with this not being supported on some platforms, but maybe I'm thinking of windows, which I don't think we need to support.

src/rpc/handlers/BookChanges.cpp
godexsoft: I'm not a huge fan of passing "state" around as arguments so what i did instead is i kept the temporary instance approach but i hid it behind a static function - this way we can't misuse the interface (by calling handle on the instance twice for example) and also the correct usage pattern is now obvious as there is now only one way to access it. As a bonus i was able to get rid of the `reset` function because now the implementation is a guaranteed `one-shot`.

I think i quite like this approach, thanks for giving me ideas :+1:

src/subscriptions/SubscriptionManager.cpp
godexsoft: if i read the link i sent correctly then it's part of c/c++ standard since forever ago.. in that sense it's not only supported by both windows and linux but also by just about any compiler in existence. I could be misunderstanding the reference but i'm pretty sure that's the case. Having that said, i refactored the code a bit and the `not` went away naturally so no worries until next time we use it :D

cjcobb23 (approved):
LGTM

PR #317, opened 2022-09-19
Mark package release's version string

legleux:
When Clio is packaged, `-release` is appended to version string. This will just make it obvious if a server if from a package.
If the executable is from a release, you can already know which git commit it was built from.
Built in debug mode, the git commit is part of the version string.
Another issue is actually building `Release` mode but that's another issue altogether that also affects rippled.


godexsoft (approved):
Sorry, i only now see you requested a review for this one. LGTM

PR #322, opened 2022-09-23
add work queue output to server_info

cjcobb23:
Previously, the length of the work queue was not in server_info. This made it impossible to monitor. The length of the work queue is a crucial metric that signals when the server is under heavy load. This change puts the work queue info in server_info.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/322)

### DIFF
--- a/src/rpc/Counters.cpp
+++ b/src/rpc/Counters.cpp
@@ -1,5 +1,6 @@
 #include <rpc/Counters.h>
 #include <rpc/RPC.h>
+#include <rpc/RPCHelpers.h>
 
 namespace RPC {
 
@@ -66,6 +67,8 @@ Counters::report()
 {
     std::shared_lock lk(mutex_);
     boost::json::object obj = {};
+    obj[JS(rpc)] = boost::json::object{};
+    auto& rpc = obj[JS(rpc)].as_object();
 
     for (auto const& [method, info] : methodInfo_)
     {
@@ -76,8 +79,9 @@ Counters::report()
         counters["forwarded"] = std::to_string(info.forwarded);
         counters["duration_us"] = std::to_string(info.duration);
 
-        obj[method] = std::move(counters);
+        rpc[method] = std::move(counters);
     }
+    obj["work_queue"] = workQueue_.get().report();
 
     return obj;
 }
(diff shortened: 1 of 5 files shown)


src/rpc/Counters.cpp
```diff
@@ -76,10 +79,11 @@ Counters::report()
         counters["forwarded"] = std::to_string(info.forwarded);
         counters["duration_us"] = std::to_string(info.duration);
 
-        obj[method] = std::move(counters);
+        rpc[method] = std::move(counters);
     }
+    obj["work_queue"] = workQueue_.get().report();
```
godexsoft, line 84: Do we have a JS stringy for this? i suppose same question for the lines above.

src/rpc/handlers/ServerInfo.cpp
```diff
@@ -48,7 +48,7 @@ doServerInfo(Context const& context)
     if (admin)
     {
         info[JS(counters)] = boost::json::object{};
-        info[JS(counters)].as_object()[JS(rpc)] = context.counters.report();
```
godexsoft, line 51: Is it intended that it seems to be stored in a different path?

src/rpc/handlers/ServerInfo.cpp
```diff
@@ -48,7 +48,7 @@ doServerInfo(Context const& context)
...
+        info[JS(counters)].as_object() = context.counters.report();
```
godexsoft: Maybe i'm wrong but i think you can just assign the result of reports straight into the correct path, without creating an object there first and then referencing it by as_object. worth checking

godexsoft (commented):
Please check the minor things i commented on. otherwise looks good!

src/rpc/Counters.cpp
cjcobb23: Not for this, but for some of the others. It's a macro that wraps fields defined in rippled. We should add some sort of mechanism to define fields just for clio as well.

src/rpc/handlers/ServerInfo.cpp
cjcobb23: I just moved the addition of the "rpc" object inside of `Counters::report()`, so it should ultimately still be in the same path.

src/rpc/handlers/ServerInfo.cpp
cjcobb23: You can, good call

godexsoft (approved):
Nice one. Let's think about adding our own fields in a separate issue. Ideally we wouldn't mix with stray string literals here and there and only use those constants 😎

Issue #221, opened 2022-07-21
Missing  'limit' parameter in the clio response of "account_lines", "account_channels", "account_objects", "account_offers" requests.


mounikakun:
Here is the link for sample requests I have tested on

[](https://docs.google.com/document/d/13QoYtKHHcn6-IV0o4taOMXFZEGeTImRNMEUmx8ji2AU/edit
)

mounikakun:
Sample requests I have tested:

https://docs.google.com/document/d/13QoYtKHHcn6-IV0o4taOMXFZEGeTImRNMEUmx8ji2AU/edit

Sample rippled and clio request responses:

https://ripplelabs.atlassian.net/wiki/spaces/~6244a5228678e900705a63ad/pages/edit-v2/2426503180?draftShareId=35de7fbd-f3bc-405b-9d11-594bc358bc90

officialfrancismendoza:
@ManiMounikaKunasani I was able to find the "limit" parameter for testing account_lines when sending to s2, unless I am misunderstanding the issue. Attached a screenshot below showing the limit parameter grepped from the output

The command I ran was: 
`curl -H 'Content-Type: application/json' -d '{   "method": "account_lines",   "params": [       {           "account": "rpRq4fC6kfeS7cR42YL5ZJMTNSj9noedU7",           "ledger_index": "validated",           "limit": 2       }   ]}' https://s2-clio.ripple.com:51233/`

![image](https://user-images.githubusercontent.com/47958858/192069488-6813a4f9-aa8e-4d0c-80ff-359faaba76c3.png)

officialfrancismendoza:
@ManiMounikaKunasani just to double-check, I made a request, but on the sample account found in the XRPL documentation for account_limit [here](https://xrpl.org/account_lines.html) and grepped the limit in the Clio response. As the account for the sample account_limit in the [Clio requests google doc](https://docs.google.com/document/d/13QoYtKHHcn6-IV0o4taOMXFZEGeTImRNMEUmx8ji2AU/edit) did not exist, I got fuller output this time, but they both still returned limit

The command I ran was:
`curl -H 'Content-Type: application/json' -d'{    "method": "account_lines",    "params": [        {            "account": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",            "ledger_index": "validated",            "limit":2        }    ]}' https://s2-clio.ripple.com:51233/`

[...]

PR #328, opened 2022-09-26
Return limit for assorted account methods

officialfrancismendoza:
**Issue** (#221): Limit parameter is not returned for account_lines, account_channels, account_objects, account_offers checks
**Fix**: Add limit parameter within response object for each respective file

### DIFF
--- a/src/rpc/handlers/AccountChannels.cpp
+++ b/src/rpc/handlers/AccountChannels.cpp
@@ -85,6 +85,7 @@ doAccountChannels(Context const& context)
 
     response[JS(account)] = ripple::to_string(accountID);
     response[JS(channels)] = boost::json::value(boost::json::array_kind);
+    response[JS(limit)] = limit;
     boost::json::array& jsonChannels = response.at(JS(channels)).as_array();
 
     auto const addToResponse = [&](ripple::SLE&& sle) {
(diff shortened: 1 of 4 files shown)


test_scripts/clio_223.js
```diff
@@ -0,0 +1,38 @@
+/*Script provided by Nathan Nichols*/
```
godexsoft: Imho, not a great name for the script - `clio_223`. i suppose it is referring to the issue 223 but i would try call it something meaningful instead like `requestAccountChannels` for example. 

Another thing is - i'd probably not put these two scripts in the codebase at all. At the very least not until everyone agree we want to have those in the codebase and they should be in javascript. We already have one uberscript written in python called `test.py`. 

@cjcobb23 what do you think about tests scripts in the repo?

cjcobb23: Yes let's not add new scripts to the repo. Pretty sure what you want to do is already handled in test.py. And I feel that it's questionable whether test.py should even be in the repo either. You can still use this script to test @officialfrancismendoza , but just don't add it to the repo.

cjcobb23:
Please give the PR a descriptive title and link to the issue in the description.

godexsoft (dismissed):
:+1:

cjcobb23 (dismissed):
This needs to be squashed to one commit, with a good commit message, and then its good to go.

officialfrancismendoza:
@godexsoft squashed to one commit. Can you review? If looks good, please approve & merge. Thank you! :)

godexsoft (dismissed):
LGTM

godexsoft (dismissed):

godexsoft (approved):

Issue #268, opened 2022-08-23
“book_offers” with invalid taker_gets currency field in the request do not return "dstAmtMalformed" error.


mounikakun:
Expected response: `dstAmtMalformed`
Actual Response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "zz"
           },
           "taker_pays": {
               "currency": "JOD",
               "issuer": "rfqNt3Q7jDiWzt9v23ef7F8Fr4KJu1obMb"
           },
           "limit":20
       }
   ]
}`

PR #333, opened 2022-09-28
Return dstAmtMalformed in book offers

officialfrancismendoza:
**Issue** (#268): book_offers API request incorrectly returns invalidParams instead of dstAmtMalformed.
**Fix**: Add rpcDST_AMT_MALFORMED in get_currency.

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1380,12 +1380,12 @@ parseBook(boost::json::object const& request)
     ripple::Currency pay_currency;
     if (!ripple::to_currency(
             pay_currency, taker_pays.at("currency").as_string().c_str()))
-        return Status{Error::rpcINVALID_PARAMS, "badTakerPaysCurrency"};
+        return Status{Error::rpcSRC_CUR_MALFORMED, "badTakerPaysCurrency"};
 
     ripple::Currency get_currency;
     if (!ripple::to_currency(
             get_currency, taker_gets["currency"].as_string().c_str()))
-        return Status{Error::rpcINVALID_PARAMS, "badTakerGetsCurrency"};
+        return Status{Error::rpcDST_AMT_MALFORMED, "badTakerGetsCurrency"};
 
     ripple::AccountID pay_issuer;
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1383,7 +1383,7 @@ parseBook(boost::json::object const& request)
     ripple::Currency get_currency;
     if (!ripple::to_currency(
             get_currency, taker_gets["currency"].as_string().c_str()))
-        return Status{Error::rpcINVALID_PARAMS, "badTakerGetsCurrency"};
+        return Status{Error::rpcDST_AMT_MALFORMED, "badTakerGetsCurrency"};
```
godexsoft, line 1388: There is a block above that does a very similar thing for `takerPaysCurrency`. Can you please verify with @ManiMounikaKunasani whether it also needs to change or should stay as `rpcINVALID_PARAMS`. I feel like it would have to change the same way.

cjcobb23: We definitely need to make this same change for `takerPaysCurrency`

src/rpc/RPCHelpers.cpp
```diff
@@ -1380,12 +1378,12 @@ parseBook(boost::json::object const& request)
     ripple::Currency pay_currency;
     if (!ripple::to_currency(
             pay_currency, taker_pays.at("currency").as_string().c_str()))
-        return Status{Error::rpcINVALID_PARAMS, "badTakerPaysCurrency"};
+        return Status{Error::rpcDST_AMT_MALFORMED, "badTakerPaysCurrency"};
```
cjcobb23: This is actually supposed to be `rpcSRC_CUR_MALFORMED`. Here's the analogous line in rippled: https://github.com/XRPLF/rippled/blob/fe05b8c4feb815e1aa6afd7ff8c85dbc92bcd651/src/ripple/rpc/handlers/BookOffers.cpp#L83

cjcobb23 (changes requested):

cjcobb23 (approved):
LGTM

Issue #264, opened 2022-08-23
“account_tx” with the invalid ledger_index_max returns "invalidParams" error instead of "lgrIdxMalformed" error.


mounikakun:
“account_tx” with the ledger specified by the ledger_index_max does not exist, or if it does exist but the server does not have it.
Expected response: `lgrIdxMalformed`
Actual response: `invalidParams`
Suggested fix: return `lgrIdxsInvalid`

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": 31089994,
           "ledger_index_min": -1
       }
   ]
}`

mounikakun:
“account_tx” with the ledger specified by the ledger_index_min does not exist, or if it does exist but the server does not have it.
Expected response: `lgrIdxMalformed`
Actual response: `invalidParams`
Suggested fix: return `lgrIdxsInvalid`

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": -1,
           "ledger_index_min": 19499032
       }
   ]
}`

PR #336, opened 2022-09-29
Missing lgrIdxMalformed error message for account_tx API call

officialfrancismendoza:
**Issue** (#264): account_tx API call mistakenly returns invalidParams error message when the ledger specified by the ledger_index_min does not exist, or if it does exist but the server does not have it.
**Fix**: Add rpcLGR_IDX_MALFORMED error messages to ledger sequence min and max out of range conditionals (#264)


cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):

PR #337, opened 2022-09-30
Limit concurrent requests

cjcobb23:
This is a follow up to https://github.com/XRPLF/clio/pull/187

I changed the logic of that PR to just clear the whole request counting map, instead of deleting as requests finish.

Defaults were also changed. Now, the byte based rate limiter limits to 10 MB/s, and the request based limiter limits to 1 concurrent request. Sweep period defaults to 10 seconds. 10 MB/s is roughly the throughput of calling ledger_data repeatedly in a loop from localhost when the cache is full.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/337)

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -10,29 +10,37 @@
 class DOSGuard
 {
     boost::asio::io_context& ctx_;
-    std::shared_mutex ftMtx_;   // protects ipFetchCount_
-    std::shared_mutex reqMtx_;  // protects ipRequestCount_
-    std::unordered_map<std::string, std::atomic_uint32_t> ipFetchCount_;
-    std::unordered_map<std::string, std::atomic_uint32_t> ipRequestCount_;
+    std::shared_mutex fetchMtx_;  // protects ipFetchCount_
+    std::shared_mutex reqMtx_;    // protects ipRequestCount_
+    std::unordered_map<std::string, uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, uint32_t> ipRequestCount_;
     std::unordered_set<std::string> const whitelist_;
-    std::uint32_t const maxFetches_;
-    std::uint32_t const sweepInterval_;
-    std::uint32_t const maxConcurrentRequests_;
+    std::optional<uint32_t> const maxFetches_;
+    std::optional<uint32_t> const maxConcurrentRequests_;
+    std::optional<uint32_t> const sweepInterval_;
+    // We track the time we swept so that way a request that started before the
+    // sweep but finished after doesn't decrement the request count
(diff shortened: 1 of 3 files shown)


src/webserver/DOSGuard.h
```diff
@@ -69,20 +109,79 @@ class DOSGuard
...
+    {
+        if (!maxConcurrentRequests_)
+            return;
+        std::unique_lock lck{reqMtx_};
+        auto it = ipRequestCount_.find(ip);
+        if (it != ipRequestCount_.end() && it->second != 0)
+            --it->second;
+    }
+    bool
+    isOk(std::string const& ip, std::uint32_t requestCount)
```
godexsoft: consider marking the member function as `const`

src/webserver/DOSGuard.h
```diff
@@ -69,20 +109,79 @@ class DOSGuard
         }
     }
 
+    std::uint32_t
+    getFetches(std::string const& ip)
```
godexsoft: consider marking the member function as `const`

src/webserver/DOSGuard.h
```diff
@@ -69,20 +109,79 @@ class DOSGuard
...
+        std::shared_lock lk{fetchMtx_};
+        auto it = ipFetchCount_.find(ip);
+        if (it == ipFetchCount_.end())
+            return 0;
+        else
+            return it->second;
+    }
+
+    std::uint32_t
+    getRequests(std::string const& ip)
```
godexsoft: consider marking `const` and using `auto const it =` in the body

src/webserver/DOSGuard.h
```diff
@@ -104,38 +203,53 @@ class DOSGuard
     {
         if (whitelist_.contains(ip))
             return true;
+        bool fetchesOk = !maxFetches_ || getFetches(ip) < maxFetches_;
```
godexsoft: this looks very similar to `isOk`. maybe `isOk` can be reused here?

src/webserver/DOSGuard.h
```diff
@@ -104,38 +203,53 @@ class DOSGuard
...
     {
-        if (whitelist_.contains(ip))
+        if (whitelist_.contains(ip) || !maxFetches_)
             return true;
-
         {
-            std::unique_lock lck(mtx_);
+            std::unique_lock lck{fetchMtx_};
             auto it = ipFetchCount_.find(ip);
+            auto count = 0;
```
godexsoft: this is an `int`. consider using `uint32_t` here to have the same type we are fetching a few lines below

godexsoft (changes requested):

Issue #261, opened 2022-08-23
“account_tx” with unfunded account do not return actMalformed error.


mounikakun:
The address specified in the account field of the request does not correspond to an account in the ledger.
Expected output: `actMalformed` error
Actual output: No error
Suggested fix: return account not found and update documentation.

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz",
           "binary": false,
           "forward": false,
           "ledger_index_max": -1,
           "ledger_index_min": -1
       }
   ]
}
`

godexsoft:
@ManiMounikaKunasani can't reproduce with current `develop`.
For the account in the sample request both `rippled` and `clio` return success with empty transactions.
If i change a letter in the account to make it invalid - i get actMalformed from both `rippled` and `clio`.

legleux:
Getting same response as of 29f1f860d860433913278784a5361379e8cc69a4
```
curl rippled:5005 -d'{ "method": "account_tx", "params": [ { "account": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz", "binary": false, "forward": false, "ledger_index
_max": -1, "ledger_index_min": -1 } ] } '
{"result":{"account":"rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz","ledger_index_max":33861998,"ledger_index_min":33861311,"limit":0,"status":"success","transactions":[],"validated":true}}
```
Clio:
```
curl clio:8080 -d'{ "method": "account_tx", "params": [ { "account": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz", "binary": false, "forward": false, "ledger_index
_max": -1, "ledger_index_min": -1 } ] } '
{"result":{"ledger_index_min":33833664,"ledger_index_max":33862003,"transactions":[],"account":"rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz","validated":true,"status":"success"},"warnings":[{"id":2001,"message":"This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"}]}
```

PR #338, opened 2022-10-01
Return account not found error for unfunded account in account_tx

officialfrancismendoza:
**Issue** (#261): account_tx API call with unfunded account does not return any error message
**Fix**: Return rpcACT_NOT_FOUND and implement new conditional for specific error message

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1,8 +1,6 @@
 #include <boost/algorithm/string.hpp>
 #include <backend/BackendInterface.h>
 #include <rpc/RPCHelpers.h>
-#include <webserver/WsBase.h>
-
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -220,6 +218,11 @@ getAccount(
         return {};
     }
 
+    if (!request.contains(account))
```
cjcobb23, line 221: This line doesn't build. Also, this is not actually what we want to do. We want to check if the account exists on the ledger.

cjcobb23 (changes requested):

godexsoft (changes requested):
This is still not fixed. See @cjcobb23 comment for details.

Issue #263, opened 2022-08-23
“account_tx” with invalid ledger_index_max returns "invalidParams" instead of "lgrIdxsInvalid"


mounikakun:
The request specifies a ledger_index_max that is before the ledger_index_min.
Expected response:`lgrIdxsInvalid`
Actual response: `invalidParams`

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": 21085176,
           "ledger_index_min": -1
       }
   ]
}`

mounikakun:
“account_tx” with Either the request specifies a ledger_index_max that is before the ledger_index_min, and the server does not have a validated ledger range because it is not synced with the network.
Expected response: `lgrIdxsInvalid`
Actual response: `invalidParams`

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": 19399132,
           "ledger_index_min": -1
       }
   ]
}`

legleux:
as of 29f1f860d860433913278784a5361379e8cc69a4
```
"method": "account_tx", "params": [{"account": "rhYtqWeTaAiMv8qTnke98HsTxy9XZBdp5X", "binary": false, "forward": false, "ledger_index_ma
x": 21085176, "ledger_index_min": -1, "Fee": "20"}]}'
```
returns `lgrIdxMalformed` and the `error_message`s are different:
rippled: `"Ledger indexes invalid."`
clio: `ledgerSeqMaxOutOfRange`
```
{"result":{"error":"lgrIdxMalformed","error_code":58,"error_message":"ledgerSeqMaxOutOfRange","status":"error","type":"response","request":{"method":"account_tx","params":[{"account":"rhYtqWeTaAiMv8qTnke98HsTxy9XZBdp5X","binary":false,"forward":false,"ledger_index_max":21085176,"ledger_index_min":-1,"Fee":"20"}]}},"warnings":[{"id":2001,"message":"This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"}]}
```

godexsoft:
[...]

PR #339, opened 2022-10-02
Return lgrIdxsInvalid error for ledger_max_index less than ledger_min_index

officialfrancismendoza:
**Issue** (#263): Request with ledger_max_index before ledger_min_index specifies "invalidParams" as wrong error message
**Fix**: Add new conditional and rpcLGR_IDXS_INVALID error message

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1539,47 +1539,54 @@ traverseTransactions(
     }
 
     auto minIndex = context.range.minSequence;
+    auto maxIndex = context.range.maxSequence;
+    std::optional<int64_t> min;
+    std::optional<int64_t> max;
+
     if (request.contains(JS(ledger_index_min)))
     {
-        auto& min = request.at(JS(ledger_index_min));
-
-        if (!min.is_int64())
+        if (!request.at(JS(ledger_index_min)).is_int64())
+        {
             return Status{
                 RippledError::rpcINVALID_PARAMS, "ledgerSeqMinNotNumber"};
+        }
 
-        if (min.as_int64() != -1)
+        min = request.at(JS(ledger_index_min)).as_int64();
+
+        if (*min != -1)
         {
-            if (context.range.maxSequence < min.as_int64() ||
-                context.range.minSequence > min.as_int64())
+            if (context.range.maxSequence < *min ||
+                context.range.minSequence > *min)
                 return Status{
                     RippledError::rpcLGR_IDX_MALFORMED,
                     "ledgerSeqMinOutOfRange"};
             else
-                minIndex = boost::json::value_to<std::uint32_t>(min);
+                minIndex = static_cast<uint32_t>(*min);
         }
 
         if (forward && !cursor)
             cursor = {minIndex, 0};
     }
 
-    auto maxIndex = context.range.maxSequence;
     if (request.contains(JS(ledger_index_max)))
     {
-        auto& max = request.at(JS(ledger_index_max));
-
-        if (!max.is_int64())
+        if (!request.at(JS(ledger_index_max)).is_int64())
+        {
             return Status{
                 RippledError::rpcINVALID_PARAMS, "ledgerSeqMaxNotNumber"};
+        }
 
-        if (max.as_int64() != -1)
+        max = request.at(JS(ledger_index_max)).as_int64();
+
+        if (*max != -1)
         {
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1602,6 +1602,16 @@ traverseTransactions(
...
 
+    if (request.contains(JS(ledger_index_min)) &&
+        request.contains(JS(ledger_index_max)))
+    {
+        auto& max = request.at(JS(ledger_index_max));
+        auto& min = request.at(JS(ledger_index_min));
+
+        if (max.as_int64() < min.as_int64())
+            return Status{Error::rpcLGR_IDXS_INVALID, "lgrIdxsInvalid"};
+    }
```
cjcobb23: This is correct, but inefficient, since we already parsed the min and the max from the request higher up in the function. Let's avoid parsing the same value twice, and just use some of the existing variables.

cjcobb23 (dismissed):

src/rpc/RPCHelpers.cpp
officialfrancismendoza: @cjcobb23 resolved by moving auto max and min from preexisting occurences outside by one order of precedence

src/rpc/RPCHelpers.cpp
```diff
@@ -1559,10 +1559,10 @@ traverseTransactions(
     }
 
     auto minIndex = context.range.minSequence;
+    auto& min = request.at(JS(ledger_index_min));
```
godexsoft: I think this will now crash if you don't pass `ledger_index_min`?

src/rpc/RPCHelpers.cpp
```diff
@@ -1581,10 +1581,10 @@ traverseTransactions(
     }
 
     auto maxIndex = context.range.maxSequence;
+    auto& max = request.at(JS(ledger_index_max));
```
godexsoft: Same problem here.

godexsoft (changes requested):

src/rpc/RPCHelpers.cpp
```diff
@@ -1539,48 +1539,42 @@ traverseTransactions(
...
-        auto& min = request.at(JS(ledger_index_min));
+        if (request.at(JS(ledger_index_min)).is_int64())
+            min = request.at(JS(ledger_index_min)).as_int64();
 
-        if (!min.is_int64())
+        else
             return Status{
                 RippledError::rpcINVALID_PARAMS, "ledgerSeqMinNotNumber"};
 
-        if (min.as_int64() != -1)
```
godexsoft, line 1550: Don't we still need to support this -1 case?

officialfrancismendoza: @godexsoft you're right as we pivoted away from unit64 to int64, so this check becomes necessary again

src/rpc/RPCHelpers.cpp
```diff
@@ -1589,6 +1593,17 @@ traverseTransactions(
             cursor = {maxIndex, INT32_MAX};
     }
 
+    if (request.contains(JS(ledger_index_min)) &&
```
godexsoft: I think this check is no longer needed. You should be able to just do 
```
if (min && max && max < min) 
    return error;
```

src/rpc/RPCHelpers.cpp
```diff
@@ -1539,47 +1539,51 @@ traverseTransactions(
...
 
     auto minIndex = context.range.minSequence;
+    auto maxIndex = context.range.maxSequence;
+    std::optional<int64_t> min;
+    std::optional<int64_t> max;
+
     if (request.contains(JS(ledger_index_min)))
     {
-        auto& min = request.at(JS(ledger_index_min));
+        if (request.at(JS(ledger_index_min)).is_int64())
```
godexsoft: I'd suggest to simplify this code to something like 
```
if (not req.at(index_min).is_int64()) 
    return error;
min = req.at(index_min).as_int64();
if (*min != -1) { 
    ... 
}
```
And the same transformation for `max` below.

godexsoft (commented):
A few more simplifications are possible

godexsoft (approved):
For future PRs - you can omit curly brackets for one line `if` statements, especially with 'return' (if it is still readable).
Otherwise, LGTM. Let's get it merged.

Issue #276, opened 2022-08-23
“ledger_entry” request provided an invalid combination of fields returns "invalidParams" instead of "malformedRequest"


mounikakun:
Expected response: `malformedRequest`
Actual response: `invalidParams`

Sample request:
`{
    "method": "ledger_entry",
    "params": [
        {
            "index": "7DB0788C020F02780A673DC74757F23823FA3014C1866E72CC4CD8B226CD6EF4a",
            "ledger_index": "validated"
        }
    ]
}`

cjcobb23:
@ManiMounikaKunasani I don't think the example request is actually an invalid combination of fields. I think the example request is possibly a malformed index, because it a has a trailing `a`, but that should result in invalid params as an error

mounikakun:
Yes by mistake I added only half part.
“ledger_entry” request provided an invalid combination of fields, or provided the wrong type for one or more fields.Expected:malformedRequest
--
Malformed value in `index ` should return `malformedRequest` but Clio is returning `invalidParams`

Sample rippled response for same request:
`{
    "result": {
        "error": "malformedRequest",
        "ledger_hash": "ACEFACCFABD4777066108B607963C08CF74025453257B735179A8B15615ED403",
        "ledger_index": 22304628,
        "request": {
            "command": "ledger_entry",
            "index": "7DB0788C020F02780A673DC74757F23823FA3014C1866E72CC4CD8B226CD6EF4a",
            "ledger_index": "validated"
        },
        "status": "error",
        "validated": true
    }
}`

Ref:[(https://xrpl.org/ledger_entry.html#ledger_entry)]

cjcobb23:
that just doesn't line up with what the docs say: https://xrpl.org/ledger_entry.html#possible-errors

cjcobb23:
but it does seem like thats what rippled does: https://github.com/XRPLF/rippled/blob/fe05b8c4feb815e1aa6afd7ff8c85dbc92bcd651/src/ripple/rpc/handlers/LedgerEntry.cpp#L58

mounikakun:
Yes it should be either "malformedRequest" or "invalidParams" but both rippled and Clio should have same response.

PR #340, opened 2022-10-04
Return malformedRequest error code for ledger_entry

officialfrancismendoza:
**Issue** (#276): ledger_entry API request returns invalidParams instead of malformedRequest
**Fix**: Add conditional to account for both ledger and ledger_index params and malformedRequest error message (#276)

Note that rippled ErrorCodes.h does not possess a dedicated error code for "malformedRequest", so left as a text return for now.

### DIFF
--- a/src/rpc/handlers/LedgerEntry.cpp
+++ b/src/rpc/handlers/LedgerEntry.cpp
@@ -40,7 +40,13 @@ doLedgerEntry(Context const& context)
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{RippledError::rpcINVALID_PARAMS, "malformedIndex"};
     }
-    else if (request.contains(JS(account_root)))
+    if (request.contains(JS(ledger_index)))
+    {
+        if (!request.at(JS(ledger_index)).is_string() &&
+            !request.at(JS(ledger_index)).is_int64())
+            return Status{RippledError::rpcLGR_IDX_MALFORMED};
+    }
+    if (request.contains(JS(account_root)))
     {
         if (!request.at(JS(account_root)).is_string())
             return Status{
@@ -53,7 +59,7 @@ doLedgerEntry(Context const& context)
         else
             key = ripple::keylet::account(*account).key;
     }
-    else if (request.contains(JS(check)))
+    if (request.contains(JS(check)))
     {
         if (!request.at(JS(check)).is_string())
             return Status{RippledError::rpcINVALID_PARAMS, "checkNotString"};
@@ -63,7 +69,7 @@ doLedgerEntry(Context const& context)
             return Status{RippledError::rpcINVALID_PARAMS, "checkMalformed"};
         }
     }
-    else if (request.contains(JS(deposit_preauth)))
+    if (request.contains(JS(deposit_preauth)))
     {
         if (!request.at(JS(deposit_preauth)).is_object())
         {
@@ -118,7 +124,7 @@ doLedgerEntry(Context const& context)
                 key = ripple::keylet::depositPreauth(*owner, *authorized).key;
         }
     }
-    else if (request.contains(JS(directory)))
+    if (request.contains(JS(directory)))
     {
         if (!request.at(JS(directory)).is_object())
         {
@@ -191,7 +197,7 @@ doLedgerEntry(Context const& context)
             }
         }
     }
-    else if (request.contains(JS(escrow)))
+    if (request.contains(JS(escrow)))
     {
         if (!request.at(JS(escrow)).is_object())
         {
@@ -230,7 +236,7 @@ doLedgerEntry(Context const& context)
             }
         }
     }
-    else if (request.contains(JS(offer)))
+    if (request.contains(JS(offer)))
     {
         if (!request.at(JS(offer)).is_object())
         {
@@ -266,7 +272,7 @@ doLedgerEntry(Context const& context)
             }
         }
     }
-    else if (request.contains(JS(payment_channel)))
+    if (request.contains(JS(payment_channel)))
     {
         if (!request.at(JS(payment_channel)).is_string())
             return Status{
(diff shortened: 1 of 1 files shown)


src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -29,14 +29,30 @@ doLedgerEntry(Context const& context)
...
     {
         if (!request.at(JS(index)).is_string())
             return Status{Error::rpcINVALID_PARAMS, "indexNotString"};
 
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{Error::rpcINVALID_PARAMS, "malformedIndex"};
     }
+    else if (request.contains(JS(ledger_index)))
+    {
+        if (!request.at(JS(ledger_index)).is_string())
```
cjcobb23: ledger_index does not need to be string, it can also be a number

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -29,14 +29,30 @@ doLedgerEntry(Context const& context)
...
+        if (!request.at(JS(index)).is_string())
+            return Status{Error::rpcINVALID_PARAMS, "malformedRequest"};
+
+        if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
+            return Status{Error::rpcINVALID_PARAMS, "malformedRequest"};
+
+        if (!request.at(JS(ledger_index)).is_string())
+            return Status{Error::rpcINVALID_PARAMS, "malformedRequest"};
+    }
+    else if (request.contains(JS(index)))
```
cjcobb23: What exactly are these additions doing? We were already checking that `index` was a string below, why are we checking it again now?

cjcobb23 (changes requested):

cjcobb23:
I actually suspect the linked issue does not correctly describe the problem

cjcobb23:
@officialfrancismendoza the bug here is that when `index` is not parseable for whatever reason, return malformed request instead of invalid params

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -37,6 +37,12 @@ doLedgerEntry(Context const& context)
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{Error::rpcINVALID_PARAMS, "malformedIndex"};
     }
+    else if (request.contains(JS(ledger_index)))
```
godexsoft: I'm not sure if your code does what was planned or not but i noticed that the whole block is using if-else. this means that if request contains `index` it will never even have a chance to run your code, even if everything was fine with the `index` param. And the same for all other checks down the line. I think this is a bug, wdyt @cjcobb23 ?

officialfrancismendoza: @godexsoft that's what I suspected as well. That's initially why I had that block of code atop parsing index and ledger_index concurrently to handle the edge case of both fields being present but malformed some other way, whereas the single conditionals could be part of the general case of just that parameter being available

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -37,7 +37,15 @@ doLedgerEntry(Context const& context)
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{Error::rpcINVALID_PARAMS, "malformedIndex"};
     }
-    else if (request.contains(JS(account_root)))
+    if (request.contains(JS(ledger_index)))
+    {
+        if (!request.at(JS(ledger_index)).is_string())
```
godexsoft: Seems like the logic is incorrect here. Now if someone passes a `ledger_index` but it's not a string it will straight return an error, not even trying to check whether it's an integer or not. You have to check for both and return an error only if index is neither string nor integer.

godexsoft (changes requested):

godexsoft (changes requested):
The final `else' needs to be removed entirely, otherwise we will always have an error if "ticket" is not passed.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -37,7 +37,13 @@ doLedgerEntry(Context const& context)
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{Error::rpcINVALID_PARAMS, "malformedIndex"};
     }
-    else if (request.contains(JS(account_root)))
+    if (request.contains(JS(ledger_index)))
+    {
+        if (!request.at(JS(ledger_index)).is_string() ||
```
godexsoft: As it is now, if i send true as ledger_index it will crash the application.
If you would fix as_int64() to the intended is_int64() then the condition will always resolve to true as 'ledger_index' can't be both a string and int64 at the same time.

godexsoft (changes requested):

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -37,7 +37,13 @@ doLedgerEntry(Context const& context)
         if (!key.parseHex(request.at(JS(index)).as_string().c_str()))
             return Status{RippledError::rpcINVALID_PARAMS, "malformedIndex"};
     }
-    else if (request.contains(JS(account_root)))
+    if (request.contains(JS(ledger_index)))
+    {
+        if (!request.at(JS(ledger_index)).is_string() ||
```
godexsoft: This is always true - ledger_index can't be both `string` and `int64`. 
Did you run this and passed it `ledger_index` of different types to see what happens?

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -37,7 +37,13 @@ doLedgerEntry(Context const& context)
...
+            !request.at(JS(ledger_index)).is_int64())
+            return Status{Error::rpcLGR_IDX_MALFORMED, "malformedIndex"};
```
godexsoft: No need to pass second param.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -319,7 +325,7 @@ doLedgerEntry(Context const& context)
 
         key = ripple::keylet::line(*id1, *id2, currency).key;
     }
-    else if (request.contains(JS(ticket)))
+    if (request.contains(JS(ticket)))
```
godexsoft, line 328: The final `else` is still there - need to remove (I could swear you did that before?).

godexsoft (changes requested):

PR #342, opened 2022-10-04
Fix nft_sell_offers/nft_buy_offers limit and marker

### LINKED ISSUE #335: "nft_sell_offers" and "nft_buy_offers" do not have "marker" parameter in the response for any limit value.
Sample request:
`{
  "method": "nft_sell_offers",
  "params": [{
    "nft_id": "000000004C17E5B8E33245D98126F4A6DF4EA57FF71C027B0000099B00000000",
    "ledger_index": "validated",
    "limit": 1
  }]
}`

Response:
`{
    "result": {
        "nft_id": "000000004C17E5B8E33245D98126F4A6DF4EA57FF71C027B0000099B00000000",
        "offers": [
            {
                "index": "0C4FF88C889CCB9B37C401C3C9F27A8098E4FE440C06D14DA27548CC9207CE14",
                "flags": 1,
                "owner": "rfALyz8SX9UN444d7iQnpjnWLhtr8Cm2f5",
                "amount": "2400000"
            }
        ],
        "validated": true,
        "status": "success"
    },
    "warnings": [
        {
            "id": 2001,
        },
        {
            "id": 2002,
        }
    ]
}`

Expected response:
`{
    "result": {
        "limit": 1,
        "marker": "19FFA2DAA50D050F220B719DAE42D04CC1BAF1AB10FFCE7FAFE3F49261BEAE7B",
        "nft_id": "000000004C17E5B8E33245D98126F4A6DF4EA57FF71C027B0000099B00000000",
        "offers": [
            {
                "amount": "2400000",
                "flags": 1,
                "nft_offer_index": "0C4FF88C889CCB9B37C401C3C9F27A8098E4FE440C06D14DA27548CC9207CE14",
                "owner": "rfALyz8SX9UN444d7iQnpjnWLhtr8Cm2f5"
            }
        ],
        "status": "success"
    }
}`

godexsoft:
Fixes #335 

While fixing the logic part a random segfault started happening so there is some extra fixing to avoid UB.


src/rpc/handlers/NFTOffers.cpp
```diff
@@ -100,13 +97,13 @@ enumerateNFTOffers(
...
         lgrInfo.seq,
-        limit,
+        reserve,
         {},
         context.yield,
         [&offers](ripple::SLE const& offer) {
             if (offer.getType() == ripple::ltNFTOKEN_OFFER)
             {
-                offers.emplace_back(offer);
+                offers.push_back(offer);
```
cjcobb23: can we actually move this offer object?

src/rpc/handlers/NFTOffers.cpp
```diff
@@ -126,6 +123,7 @@ enumerateNFTOffers(
     for (auto const& offer : offers)
         appendNftOfferJson(offer, jsonOffers);
 
+    response.insert_or_assign(JS(offers), std::move(jsonOffers));
```
cjcobb23, line 141: Why do we need this syntax versus `response[JS(offers)] = std::move(jsonOffers);`

src/rpc/handlers/NFTOffers.cpp
```diff
@@ -126,6 +123,7 @@ enumerateNFTOffers(
     for (auto const& offer : offers)
         appendNftOfferJson(offer, jsonOffers);
```
cjcobb23: Why don't we just return the json offer object from this function call, and push the value into the array? I feel like passing `jsonOffers` into the function is an unecessary amount of indirection and side effects, with no real benefit

src/rpc/handlers/NFTOffers.cpp
godexsoft: operator[] returns a reference to the element inside of the response whereas `insert_or_assign` accepts an rvalue-ref. this means that `response[...] = std::move(...)` will default-construct and then move-assign. the `insert_or_assign` version will just call the move constructor.

src/rpc/handlers/NFTOffers.cpp
godexsoft: i agree!

src/rpc/handlers/NFTOffers.cpp
godexsoft: not as it was (`const ref`). now we can tho :)

src/rpc/handlers/NFTOffers.cpp
```diff
@@ -8,29 +8,38 @@
...
 
-namespace RPC {
+namespace json = boost::json;
+
+namespace ripple {
 
-static void
-appendNftOfferJson(ripple::SLE const& offer, boost::json::array& offers)
+void
+tag_invoke(json::value_from_tag, json::value& jv, SLE const& offer)
```
cjcobb23, line 16: I'd probably leave a comment here about how this tag_invoke stuff works

src/rpc/handlers/NFTOffers.cpp
```diff
@@ -8,29 +8,38 @@
 #include <algorithm>
 #include <rpc/RPCHelpers.h>
 
-namespace RPC {
+namespace json = boost::json;
+
+namespace ripple {
```
cjcobb23, line 13: Why do we need this to be in namespace `ripple`?

cjcobb23, line 13: Are we using this namespace just so we don't have to type out `ripple::` a bunch of times in `tag_invoke`? Feel like I'd just add `using namespace ripple;` to the beginning of `tag_invoke`, but I guess it probably doesn't matter, since we never call `tag_invoke` ourselves.

src/rpc/handlers/NFTOffers.cpp
```diff
@@ -121,7 +130,7 @@ enumerateNFTOffers(
     }
 
     for (auto const& offer : offers)
-        appendNftOfferJson(offer, jsonOffers);
+        jsonOffers.push_back(json::value_from(offer));
```
cjcobb23: Feel like I'd leave a comment here saying this calls `tag_invoke`

src/rpc/handlers/NFTOffers.cpp
godexsoft: no, it's because `tag_invoke` needs to be in the same namespace as the type it converts to json as per `boost::json` docs.

cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):
LGTM

Issue #274, opened 2022-08-23
“ledger_entry” with the request improperly specified the ticket.owner sub-field do not return "malformedOwner".


mounikakun:
Expected response: `malformedOwner`
Actual response: `invalidParams`

Sample request:
`{
  "method": "ledger_entry",
  "params": [{
    "ticket": {
      "owner": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
      "ticket_sequence": 23
    },
    "ledger_index": "validated"
  }]
}`

godexsoft:
rippled seems to return `malformedRequest` instead of `malformedOwner` so i'm gonna do that instead @ManiMounikaKunasani. Is that ok with you?

PR #344, opened 2022-10-05
Return malformedOwner in ticket.owner for ledger_entry

officialfrancismendoza:
**Issue** (#274): ledger_entry API call returns invalidParams instead of malformedOwner in ticket request's owner field
**Fix**: Add malformedOwner to conditionals and change account to owner in request


godexsoft (approved):
LGTM

Issue #273, opened 2022-08-23
“ledger_entry” with the request improperly specified the deposit_preauth.owner sub-field do not return "malformedOwner"


mounikakun:
Expected response: `malformedOwner`
Actual response: `invalidParams`

Sample request:
`{
  "method": "ledger_entry",
  "params": [{
    "deposit_preauth": {
      "owner": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
      "authorized": "ra5nK24KXen9AHvsdFTKHSANinZseWnPcX"
    },
    "ledger_index": "validated"
  }]
}`

officialfrancismendoza:
To expound on the recurrence: One edge case found by @ManiMounikaKunasani contained a perfectly formatted owner string that was not found. Arguably "malformedOwner" should not be the proper error message to pass here as the request is perfectly formed, but the account owner doesn't exist. 
![image](https://user-images.githubusercontent.com/47958858/204623318-415f0ae1-0f95-4c64-a185-964ce551007d.png)
![image](https://user-images.githubusercontent.com/47958858/204623370-453351c9-1eda-4afd-b44d-1b60f069bf58.png)
![image](https://user-images.githubusercontent.com/47958858/204623412-28e825f8-e393-42e0-8c13-b3ff272cc795.png)

officialfrancismendoza:
@injaelee closing this issue as per our discussion with @godexsoft as to edge case being unclear and code handling cases

Issue #272, opened 2022-08-23
“ledger_entry” with request improperly specified an Address field returns "invalidParams" instead of "malformedAddress".


mounikakun:
account_root is an invalid address in the request.
Expected response: `malformedAddress`
Actual response: `invalidParams`

Sample request:
`{
    "method": "ledger_entry",
    "params": [
        {
            "account_root": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
            "ledger_index": "validated"
        }
    ]
}`

mounikakun:
“ledger_entry”  of directory with invalid owner in the request.
Expected response: `malformedAddress`
Actual response: `invalidParams`

Sample request:
`{
    "method": "ledger_entry",
    "params": [
        {
            "directory": {
              "owner": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
              "sub_index": 0
            },
            "ledger_index": "validated"
        }
    ]
}`

mounikakun:
“ledger_entry” of offer with invalid account in the request.
Expected response: `malformedAddress`
Actual response: `invalidParams`

Sample request:
`{
  "method": "ledger_entry",
  "params": [
    {
      "offer": {
        "account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
        "seq": 359
      },
      "ledger_index": "validated"
    }
  ]
}`

mounikakun:
“ledger_entry” of ripple_state with invalid account in the request.
Expected response: `malformedAddress`
Actual response: `invalidParams`

Sample request:
`{
  "method": "ledger_entry",
  "params": [{
    "ripple_state": {
      "accounts": [
        "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpna",
        "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW"
      ],
      "currency": "USD"
    },
    "ledger_index": "validated"
  }]
}`

mounikakun:
“ledger_entry” of escrow with invalid account in the request.
Expected response: `malformedAddress`
Actual response: `invalidParams`
Suggested fix: `malformedRequest`

Sample request:
`{
  "method": "ledger_entry",
  "params": [{
    "escrow": {
      "account": "rL4fPHi2FWGwRGRQSH7gBcxkuo2b9NTjKKa",
      "seq": 126
    },
    "ledger_index": "validated"
  }]
[...]

PR #345, opened 2022-10-05
Return malformedOwner for deposit_preauth.owner in ledger_entry

officialfrancismendoza:
**Issue** (#273): ledger_entry API call returns invalidParams instead of malformedOwner in deposit_preauth.owner
**Fix**: Add malformedOwner error message 

**NOTE**: Similar issues to #272, #274, #276. malformedOwner does not have its own unique error code. Furthermore, we may not want to replace "ownerNotString" with "malformedOwner" as the former is technically a more accurate error message.


godexsoft (approved):
LGTM considering the notes in the pr description

PR #346, opened 2022-10-05
Handle client disconnect during rpc

### LINKED ISSUE #211: Implement timeout to align with client's socket read timeout of 3 seconds
Currently the websocket read timeout used by XRPL clients is 3 seconds and if server doesn't response within the 3 seconds, client will receive socket timeout, however the Clio server side doesn't have such timeout guarded and would continue its processing and return the respond even though the underline connection had already been abandoned by the client. We need a mechanism to stop this silent errors by implementing mechanism to keep track of request's connection duration to align  with client's timeout and raise exception/error when the processing exceeds the timeout value.

Alternatively we need to look into the XRPL client implementation to see if it's possible to negotiate the timeout value since if client is requesting large amount of data, there's a chance that the 3 seconds is not sufficient for server to fulfill.

godexsoft:
Fixes #211 

Initially we wanted to gracefully abort handling any `rpc` for disconnected clients on both websocket and over http. However only `ws` allows us to detect client disconnects. The way it could be done in the future is by introducing our own heartbeat system - something we don't want to do now.

This "fix" detects clients that dropped the connection while their request was in `clio`s work queue. Once detected, a warning log message is written which includes the `tag` of the session that dropped out.

### DIFF
--- a/src/etl/ReportingETL.h
+++ b/src/etl/ReportingETL.h
@@ -81,6 +81,12 @@ private:
     {
         std::string ip;
         int port;
+
+        // TODO: remove when clang catches up to gcc/msvc in regards to
+        // emplace_back with brace initialization (c++20)
+        ClioPeer(std::string ip, int port) : ip{ip}, port{port}
+        {
+        }
     };
 
     std::vector<ClioPeer> clioPeers;
--- a/src/rpc/WorkQueue.cpp
+++ b/src/rpc/WorkQueue.cpp
@@ -1,11 +1,19 @@
 #include <rpc/WorkQueue.h>
 
-WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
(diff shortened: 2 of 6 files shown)


src/etl/ReportingETL.h
```diff
@@ -81,6 +81,12 @@ class ReportingETL
     {
         std::string ip;
         int port;
+
+        // TODO: remove when clang catches up to gcc/msvc in regards to
+        // emplace_back with brace initialization (c++20)
+        ClioPeer(std::string ip, int port) : ip{ip}, port{port}
+        {
+        }
```
godexsoft, line 89: note: this can be avoided by using `push_back` at the callsite bu that would be 1 move vs. inplace creation - not very important difference considering this is parsed/created once. happy to leave this or do `push_back` instead - just shout.

godexsoft:
This is not really fixing anything and the codebase changed a lot since this PR was opened. Closing it without merging - should be easy to add those changes again if we ever wanted to.

Issue #266, opened 2022-08-23
“book_offers” with unfunded taker_gets issuer returns "invalidParams" instead of "srcIsrMalformed"


mounikakun:
The issuer field of the taker_pays field in the request is not valid.
Expected response: `srcIsrMalformed`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "INR",
               "issuer": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz"
           },
           "taker_pays": {
               "currency": "JOD",
               "issuer": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDM"
           },
           "limit":20
       }
   ]
}`

mounikakun:
“book_offers” with invalid taker_pays issuer.
Expected response: `srcIsrMalformed`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "XRP"
           },
           "taker_pays": {
               "currency": "JOD",
               "issuer": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDM"
           },
           "limit":20
       }
   ]
}`

mounikakun:
“book_offers” with taker_pays currency “XRP” and issuer.
Expected response: `srcIsrMalformed`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "XRP"
           },
           "taker_pays": {
               "currency": "XRP",
               "issuer": "rfqNt3Q7jDiWzt9v23ef7F8Fr4KJu1obMb"
           },
           "limit":20
       }
   ]
}`

mounikakun:
“book_offers” with taker_pays currency not “XRP” and no issuer.
Expected response: `srcIsrMalformed`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
[...]

PR #355, opened 2022-10-07
Return srcIsrMalformed for taker_gets issuer in book_offers

officialfrancismendoza:
**Issue** (#266): book_offers API call returns "invalidParams" instead of "srcIsrMalformed" for taker_gets issuer
**Fix**:  Add srcIsMalformed and extra conditional (#266)

**NOTE**: Request for subscribe API call in latest comment seems to be malformed, but logic enters RPCHelpers.cpp nonetheless

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1386,11 +1386,11 @@ parseBook(boost::json::object const& request)
 
         if (!ripple::to_issuer(
                 pay_issuer, taker_pays.at("issuer").as_string().c_str()))
-            return Status{Error::rpcINVALID_PARAMS, "badTakerPaysIssuer"};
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1411,16 +1411,19 @@ parseBook(boost::json::object const& request)
...
             "specification."};
 
     if (!isXRP(pay_currency) && isXRP(pay_issuer))
         return Status{
-            Error::rpcINVALID_PARAMS,
+            Error::rpcSRC_ISR_MALFORMED,
             "Invalid field 'taker_pays.issuer', expected non-XRP "
             "issuer."};
 
+    if ((!isXRP(pay_currency)) && (!taker_pays.contains("issuer")))
```
godexsoft, line 1412: Imho, no real need for the extra set of parentheses here - i'd drop them but it's up to you.

godexsoft (approved):

Issue #267, opened 2022-08-23
“book_offers” with the invalid taker_pays currency field in the request does not return "srcCurMalformed"


mounikakun:
Expected response: `srcCurMalformed`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "XRP"
           },
           "taker_pays": {
               "currency": "aa",
               "issuer": "rfqNt3Q7jDiWzt9v23ef7F8Fr4KJu1obMb"
           },
           "limit":20
       }
   ]
}`

mounikakun:
subscribe to order books with empty taker_pays also returns "invalidParams" instead of "srcCurMalformed"

Sample request:
`{
    "id": "Example subscribe to XRP/GateHub USD order book",
    "command": "subscribe",
    "books": [
        {
            "taker_pays": {
            },
            "taker_gets": {
                "currency": "USD",
                "issuer": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
            },
            "snapshot": true
        }
    ]
}`

mounikakun:
subscribe to order books with malformed taker_pays currency returns "invalidParams" instead of "srcCurMalformed"

Sample request:
`{
    "id": "Example subscribe to XRP/GateHub USD order book",
    "command": "subscribe",
    "books": [
        {
            "taker_pays": {
                "currency": "XR"
            },
            "taker_gets": {
                "currency": "USD",
                "issuer": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq"
            },
            "snapshot": true
        }
    ]
}`

mounikakun:
Issue exists for unsubscribe to books with empty taker_pays also.

mounikakun:
unsubscribe to books with taker_pays currency XRP and issuer

mounikakun:
unsubscribe to books with malformed taker_pays currency

PR #356, opened 2022-10-07
Return srcCurMalformed for taker_pays currency in book_offers

officialfrancismendoza:
**Issue** (#267): book_offers API returns "invalidParams" instead of "srcCurMalformed" for taker_pays currency field
**Fix**: Add srcCurMalformed in taker_pays currency field

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1317,7 +1317,7 @@ std::variant<Status, ripple::Book>
 parseBook(boost::json::object const& request)
 {
     if (!request.contains("taker_pays"))
-        return Status{RippledError::rpcBAD_MARKET, "missingTakerPays"};
+        return Status{RippledError::rpcSRC_CUR_MALFORMED, "missingTakerPays"};
 
     if (!request.contains("taker_gets"))
         return Status{RippledError::rpcINVALID_PARAMS, "missingTakerGets"};
@@ -1331,7 +1331,7 @@ parseBook(boost::json::object const& request)
     auto taker_pays = request.at("taker_pays").as_object();
     if (!taker_pays.contains("currency"))
         return Status{
-            RippledError::rpcINVALID_PARAMS, "missingTakerPaysCurrency"};
+            RippledError::rpcSRC_CUR_MALFORMED, "missingTakerPaysCurrency"};
 
     if (!taker_pays.at("currency").is_string())
         return Status{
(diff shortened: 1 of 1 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1421,6 +1421,12 @@ parseBook(boost::json::object const& request)
             "Invalid field 'taker_pays.issuer', expected non-XRP "
             "issuer."};
 
+    if (isXRP(pay_currency) && isXRP(pay_issuer))
+        return Status{
+            Error::rpcSRC_CUR_MALFORMED,
+            "Unneeded field 'taker_pays.issuer' for XRP currency "
```
godexsoft, line 1398: Is this the same error you would get from rippled? LGTM if it is.

officialfrancismendoza: @godexsoft this was a reused error message for a new conditional to handle the new case of XRP pay currency with an XRP specified issuer (with the latter not being necessary at all). This solves one of the cases @ManiMounikaKunasani added later

officialfrancismendoza:
@godexsoft can you review in light of the new merge commit? If everything looks good, will squash and then ask you to approve and then merge it, as I don't have privileges to merge automatically

src/rpc/RPCHelpers.cpp
```diff
@@ -661,7 +661,7 @@ traverseOwnedNodes(
     std::uint32_t limit,
     std::optional<std::string> jsonCursor,
     boost::asio::yield_context& yield,
-    std::function<void(ripple::SLE&&)> atOwnedNode)
+    std::function<void(ripple::SLE)> atOwnedNode)
```
godexsoft: Why change this?

src/rpc/RPCHelpers.cpp
```diff
@@ -836,7 +836,9 @@ traverseOwnedNodes(
     for (auto i = 0; i < objects.size(); ++i)
     {
         ripple::SerialIter it{objects[i].data(), objects[i].size()};
-        atOwnedNode(ripple::SLE{it, keys[i]});
+        ripple::SLE sle(it, keys[i]);
+
+        atOwnedNode(sle);
```
godexsoft: Does not look like you need this temporary but if you really want to have it you would have to `std::move()` it if you left the signature of above callback as it should be (`void(ripple::SLE&&)`).

godexsoft (changes requested):

src/rpc/RPCHelpers.cpp
```diff
@@ -1703,4 +1711,4 @@ traverseTransactions(
     return response;
 }
 
-}  // namespace RPC
+}  // namespace RPC
```
godexsoft: You may need to setup VSCode (or the IDE you are using) to automatically add a newline to the end of any files you change.

godexsoft (approved):
LGTM.

Issue #269, opened 2022-08-23
“book_offers” with same currency for taker_gets and taker_pays do not return "badMarket"


mounikakun:
Expected response: `badMarket`
Actual response: `invalidParams`

Sample request:
`{
   "method": "book_offers",
   "params": [
       {
           "taker": "r97U4iT9m8Q2FeYnAQ1gFKGFDoKe8E7Rh",
           "taker_gets": {
               "currency": "XRP"
           },
           "taker_pays": {
               "currency": "XRP"
           },
           "limit":20
       }
   ]
}`

mounikakun:
subscribe to order books with XRP currency in taker gets and pays.
Expected response: `badMarket`
Actual response: `invalidParams`

Sample request:
`{
    "id": "Example subscribe to XRP/GateHub USD order book",
    "command": "subscribe",
    "books": [
        {
            "taker_pays": {
                "currency": "XRP"
            },
            "taker_gets": {
                "currency": "XRP"
            },
            "snapshot": true
        }
    ]
}`

mounikakun:
subscribe to order books with same currency other than XRP and issuer in taker gets and pays.

Sample request:
`{
    "id": "Example subscribe to XRP/GateHub USD order book",
    "command": "subscribe",
    "books": [
        {
            "taker_pays": {
                "currency": "USD",
                "issuer": "r3s769ETJw8TvPUxMDjF8VZChZYMWSqgMu"
            },
            "taker_gets": {
                "currency": "USD",
                "issuer": "r3s769ETJw8TvPUxMDjF8VZChZYMWSqgMu"
            },
            "snapshot": true
        }
    ]
}`

mounikakun:
Issue exists for unsubscribe also

PR #357, opened 2022-10-07
Return badMarket for same currency in taker_gets and taker_pays in book_offers

officialfrancismendoza:
**Issue** (#269): book_offers API request returns "invalidParams" instead of "badMarket" when taker_gets and taker_pays have the same currency.
**Fix**: Add badMarket to taker_gets and taker_pays having equivalent currency conditional.


godexsoft (approved):
LGTM

PR #358, opened 2022-10-10
Abstraction for config

### LINKED ISSUE #321: Abstraction for config
Currently the config is parsed from json and is passed around as a `boost::json::object`. This leads to duplication of access patterns (implementations of lambda/functions to get a specific config parameter as optional etc.). It's also not testable and we straight up use string literals as keys everywhere in the codebase. 
It would be nice to introduce some api around this config. So that once parsed it's self-contained and exposes easy access functions for the valid config parameters found while parsing the json. 

At a high level, something like this:
``` cpp
Config cfg(jsonObject); // or filepath perhaps
auto val = cfg.value_or<std::string>(Config::optionName, "default value"); // string or "default value"
auto val2 = cfg.value<uint32_t>(Config::anotherOption); // throws if not found
auto opt = cfg.optValue<uint64_t>(Config::something); // value wrapped in `std::optional`, can be `std::nullopt` if not found
```

godexsoft:
Fixes #321

This implements a simple abstraction on top of `boost::json` that allows to uniformly manage configuration options for `clio`.
It introduces helper functions to read config options using `keys` (mapped directly to the json configuration file). These functions allow for different access patterns such as automatically falling back to a user-specified value if the option is not present, throwing custom error messages, returning `std::optional` values, etc.

### DIFF
--- a/src/backend/BackendFactory.h
+++ b/src/backend/BackendFactory.h
@@ -4,30 +4,23 @@
 #include <boost/algorithm/string.hpp>
 #include <backend/BackendInterface.h>
 #include <backend/CassandraBackend.h>
+#include <config/Config.h>
 
 namespace Backend {
 std::shared_ptr<BackendInterface>
-make_Backend(boost::asio::io_context& ioc, boost::json::object const& config)
+make_Backend(boost::asio::io_context& ioc, clio::Config const& config)
 {
     BOOST_LOG_TRIVIAL(info) << __func__ << ": Constructing BackendInterface";
 
-    boost::json::object dbConfig = config.at("database").as_object();
-
-    bool readOnly = false;
-    if (config.contains("read_only"))
-        readOnly = config.at("read_only").as_bool();
-
-    auto type = dbConfig.at("type").as_string();
-
+    auto readOnly = config.valueOr("read_only", false);
+    auto type = config.value<std::string>("database.type");
     std::shared_ptr<BackendInterface> backend = nullptr;
 
     if (boost::iequals(type, "cassandra"))
     {
-        if (config.contains("online_delete"))
-            dbConfig.at(type).as_object()["ttl"] =
-                config.at("online_delete").as_int64() * 4;
-        backend = std::make_shared<CassandraBackend>(
-            ioc, dbConfig.at(type).as_object());
+        auto cfg = config.section("database." + type);
+        auto ttl = config.valueOr<uint32_t>("online_delete", 0) * 4;
+        backend = std::make_shared<CassandraBackend>(ioc, cfg, ttl);
     }
 
     if (!backend)
--- a/src/backend/CassandraBackend.cpp
+++ b/src/backend/CassandraBackend.cpp
@@ -979,19 +979,6 @@ CassandraBackend::isTooBusy() const
 void
 CassandraBackend::open(bool readOnly)
 {
-    auto getString = [this](std::string const& field) -> std::string {
-        if (config_.contains(field))
-        {
-            auto jsonStr = config_[field].as_string();
-            return {jsonStr.c_str(), jsonStr.size()};
-        }
-        return {""};
-    };
-    auto getInt = [this](std::string const& field) -> std::optional<int> {
-        if (config_.contains(field) && config_.at(field).is_int64())
-            return config_[field].as_int64();
-        return {};
-    };
     if (open_)
     {
         assert(false);
@@ -1005,7 +992,8 @@ CassandraBackend::open(bool readOnly)
     if (!cluster)
         throw std::runtime_error("nodestore:: Failed to create CassCluster");
 
-    std::string secureConnectBundle = getString("secure_connect_bundle");
+    std::string secureConnectBundle =
+        config_.valueOr<std::string>("secure_connect_bundle", "");
 
     if (!secureConnectBundle.empty())
     {
@@ -1025,12 +1013,9 @@ CassandraBackend::open(bool readOnly)
     }
     else
     {
-        std::string contact_points = getString("contact_points");
-        if (contact_points.empty())
-        {
-            throw std::runtime_error(
-                "nodestore: Missing contact_points in Cassandra config");
-        }
+        std::string contact_points = config_.valueOrThrow<std::string>(
+            "contact_points",
+            "nodestore: Missing contact_points in Cassandra config");
         CassError rc =
             cass_cluster_set_contact_points(cluster, contact_points.c_str());
         if (rc != CASS_OK)
@@ -1043,7 +1028,7 @@ CassandraBackend::open(bool readOnly)
             throw std::runtime_error(ss.str());
         }
 
-        auto port = getInt("port");
+        auto port = config_.maybeValue<int>("port");
         if (port)
         {
             rc = cass_cluster_set_port(cluster, *port);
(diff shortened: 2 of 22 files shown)


src/config/Config.h
```diff
@@ -0,0 +1,407 @@
...
+     * @return Result Value of desired type
+     * @throws std::runtime_error Thrown if conversion to Result is not possible
+     * or value does not exist under key
+     */
+    template <typename Result>
+    [[nodiscard]] Result
+    value(key_type key) const
+    {
+        return maybe_value<Result>(key).value();
+    }
```
cjcobb23: Do we actually need this member? Seems like it would be preferable to always use one of the other members.

src/config/Config.h
```diff
@@ -0,0 +1,407 @@
...
+            if (not store->get().is_object())
+                throw std::runtime_error(
+                    "Can't write to '" + key + "': not an object");
+            store->get().as_object().insert_or_assign(
+                final_key, std::forward<Type>(value));
+        }
+        else
+            throw std::runtime_error(
+                "Can't write into a null store under '" + key + "'");
+    }
```
cjcobb23: We use this in one place right? I think something to do with online delete. We can probably refactor that, and remove the need for this function, as well as the functions it requires. What do you think?

unittests/config.cpp
```diff
@@ -0,0 +1,241 @@
...
+                "test",
+                json::object({
+                    {"str", "hello"},
+                    {"int", 9042},
+                    {"bool", true},
+                }),
+            }}),
+        },
+        {"top", 420},
+    })};
```
cjcobb23: Why don't we parse an actual config here, such as the example config or some such?

src/backend/CassandraBackend.cpp
```diff
@@ -1025,7 +1013,8 @@ CassandraBackend::open(bool readOnly)
     }
     else
     {
-        std::string contact_points = getString("contact_points");
+        std::string contact_points =
+            config_.value_or<std::string>("contact_points", "");
```
cjcobb23: do you want to use `value_or_throw` here?

src/backend/CassandraBackend.cpp
```diff
@@ -1087,14 +1077,18 @@ CassandraBackend::open(bool readOnly)
...
+    if (auto val = config_.maybe_value<int>("max_write_requests_outstanding");
+        val)
+        maxWriteRequestsOutstanding = *val;
+
+    if (auto val = config_.maybe_value<int>("max_read_requests_outstanding");
+        val)
+        maxReadRequestsOutstanding = *val;
+
+    if (auto val = config_.maybe_value<int>("sync_interval"); val)
+        syncInterval_ = *val;
```
cjcobb23: Why this syntax versus:
```
maxWriteRequestsOutstanding = config_.value_or<int>("max_write_requests_outstanding",maxWriteRequestsOutstanding);
```
I believe this is the syntax you used elsewhere.

src/backend/CassandraBackend.cpp
```diff
@@ -1153,25 +1146,24 @@ CassandraBackend::open(bool readOnly)
         cass_ssl_free(context);
     }
 
-    std::string keyspace = getString("keyspace");
+    auto keyspace = config_.value_or<std::string>("keyspace", "");
```
cjcobb23: Why not use `value_or_throw`?

src/backend/CassandraBackend.cpp
```diff
@@ -1153,25 +1146,24 @@ CassandraBackend::open(bool readOnly)
...
         BOOST_LOG_TRIVIAL(warning)
             << "No keyspace specified. Using keyspace clio";
         keyspace = "clio";
     }
 
-    int rf = getInt("replication_factor") ? *getInt("replication_factor") : 3;
-
-    std::string tablePrefix = getString("table_prefix");
+    auto rf = config_.value_or<int>("replication_factor", 3);
+    auto tablePrefix = config_.value_or<std::string>("table_prefix", "");
```
cjcobb23: I think we should remove this config option. Though maybe not necessarily as part of this PR.

src/config/Config.h
```diff
@@ -0,0 +1,407 @@
...
+     *
+     * @tparam Result The desired return type
+     * @param key The key to check
+     * @param fallback The fallback value
+     * @return Result Value of desired type
+     * @throws std::runtime_error Thrown if conversion to Result is not possible
+     */
+    template <typename Result>
+    [[nodiscard]] Result
+    value_or(key_type key, Result fallback) const
```
cjcobb23: I'm curious, why snake case for all of these functions? We use camel case for most things in the codebase.

src/backend/BackendFactory.h
```diff
@@ -4,30 +4,24 @@
...
 
     if (boost::iequals(type, "cassandra"))
     {
+        auto cfg = config.section("database." + type);
         if (config.contains("online_delete"))
-            dbConfig.at(type).as_object()["ttl"] =
-                config.at("online_delete").as_int64() * 4;
-        backend = std::make_shared<CassandraBackend>(
-            ioc, dbConfig.at(type).as_object());
+            cfg.set_value("ttl", config.value<int64_t>("online_delete") * 4);
```
cjcobb23: So this online_delete functionality is not actually currently supported. The whole thing is a WIP and needs more testing. It seems this is the only reason we need `set_value`. I'd rather refactor this and remove `set_value`. We can pass the `ttl` in as a constructor param, or we can just require `ttl` be specified in the cassandra config.

cjcobb23 (changes requested):
See my comments about `set_value`. I'd like to refactor to see that removed. Otherwise all of my other comments are non-blocking and up to you.

unittests/config.cpp
godexsoft: It would be nice to pass an actual config but we can't guarantee that the file will be present in the expected location. There are two solutions: don't use a file in tests or specify exactly where the file is expected relative to CWD when running the tests. We can certainly think about setting some expectation. wdyt?

src/backend/CassandraBackend.cpp
godexsoft: one less assignment. it's not all that important tbh in this context - i can change to the other option

src/backend/CassandraBackend.cpp
godexsoft: i was going for the same behaviour as we already had but i agree we could use the throwing version in many places instead

src/config/Config.h
godexsoft: it's my personal preference to write code in the style of stl and boost. i think i'm tok used to this style so it comes out by default 😅
i simply got carried away and forgot we use camel case 😂 i will change it to camel 👍

src/backend/BackendFactory.h
godexsoft: great. i will happily remove 'set_value' and this param

src/config/Config.h
godexsoft: i think we could use it for most places where we can get away with a graceful exit on exception with a meaningful default message about what config option was set to a wrong value. we can use 'value_or_throw' if we want a custom error message. i think using 'value_or' is a good option when we don't require the user to set a value. they all got their own uses imo

unittests/config.cpp
godexsoft: Although we can't have a file directly, i am able to make a temporary file, write a JSON string there and use that instead. So we will have it tested as well.

cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):

Issue #278, opened 2022-08-23
“tx” with min_ledger and max_ledger fields of the request are more than 1000 apart returns no error.


mounikakun:
Expected response: `excessiveLgrRange`
Actual response: No error 
Suggested fix: change the docs for clio to accept infinitely large range

Sample request:
`{
   "method": "tx",
   "params": [
       {
           "transaction": "823C3BB277B19ED2AA7ED20CC2AD62B064CC9A20B59B7C6C5F3821A6A53676A4",
           "binary": false,
           "min_ledger":"19499130",
           "max_ledger": "21007767"
       }
   ]
}`

godexsoft:
Reopening so @ManiMounikaKunasani can close when verified.

legleux:
This issue is resolved but now unless you pass the range as a string you get:
```
{'error': 'invalidParams', 'error_code': 31, 'error_message': 'Invalid field min_ledger, not uint.', 'request': {'meth...', 'min_ledger': '33866512', 'transaction': '95376492DE6B2B3192EAC759A9868F2F85546B9C7251F65FF1DD56D9BD73755A'}]}
```

PR #359, opened 2022-10-11
Return excessiveLgrRange for tx

officialfrancismendoza:
**Issue** (#278): tx API call returns no error when min_ledger and max_ledger are more than 1000 apart
**Fix**: Add excessiveLgrRange and conditional (#278)

### DIFF
--- a/src/rpc/handlers/Tx.cpp
+++ b/src/rpc/handlers/Tx.cpp
@@ -36,6 +36,32 @@ doTx(Context const& context)
     if (!range)
         return Status{RippledError::rpcNOT_READY};
 
+    if (request.contains(JS(min_ledger)) && request.contains(JS(max_ledger)))
+    {
+        if (request.at(JS(min_ledger)).is_int64() &&
+            request.at(JS(max_ledger)).is_int64())
+        {
+            const auto& minLedger = request.at(JS(min_ledger)).as_int64();
+            const auto& maxLedger = request.at(JS(max_ledger)).as_int64();
+
+            if (minLedger >= 0 && maxLedger >= 0 && maxLedger > minLedger)
+            {
+                if (maxLedger - minLedger > 1000)
+                    return Status{
+                        RippledError::rpcEXCESSIVE_LGR_RANGE,
+                        "excessiveLgrRange"};
+            }
+
+            else
+                return Status{
+                    RippledError::rpcLGR_IDXS_INVALID, "lgrIdxsInvalid"};
+        }
+
+        else
+            return Status{
+                RippledError::rpcLGR_IDX_MALFORMED, "lgrIdxMalformed"};
+    }
+
     auto dbResponse = context.backend->fetchTransaction(hash, context.yield);
     if (!dbResponse)
         return Status{RippledError::rpcTXN_NOT_FOUND};


src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,16 @@ doTx(Context const& context)
     if (!range)
         return Status{Error::rpcNOT_READY};
 
+    if ((request.contains(JS(min_ledger))) &&
```
godexsoft: Consider removing extra parens - there is no need for them in this case

src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,16 @@ doTx(Context const& context)
...
+        (request.contains(JS(max_ledger))))
+    {
+        auto& minLedger = request.at(JS(min_ledger));
```
godexsoft: Please check if `auto const&` works for these two and if it does - consider enforcing `const` where possible

godexsoft (changes requested):

src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,15 @@ doTx(Context const& context)
     if (!range)
         return Status{Error::rpcNOT_READY};
 
+    if (request.contains(JS(min_ledger)) && request.contains(JS(max_ledger)))
+    {
+        const auto& minLedger = request.at(JS(min_ledger));
+        const auto& maxLedger = request.at(JS(max_ledger));
+
+        if ((maxLedger.as_int64() - minLedger.as_int64()) > 1000)
```
godexsoft: Consider adding a check for `max > min` as well now that we have this check.

src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,15 @@ doTx(Context const& context)
     if (!range)
         return Status{Error::rpcNOT_READY};
 
+    if (request.contains(JS(min_ledger)) && request.contains(JS(max_ledger)))
+    {
+        const auto& minLedger = request.at(JS(min_ledger));
```
godexsoft: Please add a check for `is_int64()` first and then consider grabbing `as_int64()` right here - you also will not need it to be a reference that way.

godexsoft (changes requested):

src/rpc/handlers/Tx.cpp
godexsoft: Btw. we are checking these fields but are they even used anywhere? I could not find any mentions of `min_ledger` in the entire codebase.

officialfrancismendoza: @godexsoft min_ledger and max_ledger are parseable fields in the request that Mounika outlined here 
image

src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,18 @@ doTx(Context const& context)
     if (!range)
         return Status{RippledError::rpcNOT_READY};
 
+    if (request.contains(JS(min_ledger)) && request.contains(JS(max_ledger)) &&
+        request.at(JS(min_ledger)).is_int64() &&
```
godexsoft: I'd add these `is_int64()` checks inside instead and return some error code if they are passed but not of the correct type.

src/rpc/handlers/Tx.cpp
```diff
@@ -36,6 +36,18 @@ doTx(Context const& context)
...
+        request.at(JS(max_ledger)).is_int64())
+    {
+        const auto& minLedger = request.at(JS(min_ledger)).as_int64();
+        const auto& maxLedger = request.at(JS(max_ledger)).as_int64();
+
+        if (maxLedger - minLedger > 1000 && maxLedger > minLedger)
```
godexsoft: Looks like I'm allowed to pass negative ledgers.

godexsoft: And most importantly - this code still does a check for nothing - we don't use the parameters we check. This means you have to go to rippled codebase and see how this stuff works exactly and if we can/should port it to clio.

godexsoft (changes requested):

src/rpc/handlers/Tx.cpp
officialfrancismendoza: @godexsoft added a new ClioError code, rpcMALFORMED_PARAMS, as a counterpart to the rippled INVALID_PARAMS. Functionally in rippled, invalidParams was passed when unknown parameters were input. In contrast, rippled possesses error codes for specific parameters of paramount importance, but no universal malformed param error code previously existed

Seeing as min and max ledger are used locally, I considered this merited a more universal error code, with the hardcoded error string a more specific explanation as to what non-major parameter was malformed. Major parameters like accounts, currency types, etc. should use the more specific error codes rippled passes by default 
image

officialfrancismendoza: Update to the above @godexsoft: found a specific error code for malformed ledger indexes in rippled, but kept rpcMALFORMED_PARAMS in Errors.h for future use. Lmk if you want me to delete it otherwise

src/rpc/handlers/Tx.cpp
officialfrancismendoza: @godexsoft for context as outlined in the original issue [here](https://github.com/XRPLF/clio/issues/278), there was no previous check in Rippled to determine if there was a difference of 1000 or more between the optional min_ledger and max_ledger parameters in the doc [here](https://xrpl.org/tx.html). 

According to the documentation, specifying min_ledger and max_ledger would return the set of transactions between those two. In rippled's Tx.cpp, (relative path: build/_deps/rippled-src/src/ripple/rpc/handlers/Tx.cpp) the min_ledger and max_ledger would be formed into pairs and then passed as a separate call. 

Should we do the same for Clio, or pass in the tx set differently?
image

officialfrancismendoza: @godexsoft this issue is getting taken over by CJ due to shared concerns about missing min_ledger and max_ledger handling functionality in Clio that exists in Rippled

cjcobb23:
Fixing this in https://github.com/XRPLF/clio/pull/407

Issue #280, opened 2022-08-24
“ledger” if you specified full  as true, but not connected to the server as an admin do not return "noPermission" error.


mounikakun:
Expected response: `noPermission`
Actual response: No error
Suggested fix: document on clio part for ledger request.

Sample request:
`{
   "method": "ledger",
   "params": [
       {
           "ledger_index": "validated",
           "accounts": false,
           "full": true,
           "transactions": false,
           "expand": false,
           "owner_funds": false
       }
   ]
}`

mounikakun:
“ledger” if you specified  accounts as true, but are not connected to the server as an admin.
Expected response: `noPermission`
Actual response: No error
Suggested fix: document on clio part for ledger request.

Sample request:
`{
   "method": "ledger",
   "params": [
       {
           "ledger_index": "validated",
           "accounts": true,
           "full": false,
           "transactions": false,
           "expand": false,
           "owner_funds": false
       }
   ]
}`

godexsoft:
@ManiMounikaKunasani please note that clio does not support anything like "being connected as an admin" so instead we decided to return NOT_SUPPORTED error in case the user is attempting to use these parameters ("accounts" and/or "full").

mounikakun:
@godexsoft this issue was raised to fix the documentation part of `ledger` specific to Clio.

godexsoft:
@cjcobb23 please escalate with relevant people in order to get the docs updated :+1:

cjcobb23:
https://github.com/XRPLF/xrpl-dev-portal/issues/1618

cindyyan317:
Close this one since the document of "full" has been updated.

PR #360, opened 2022-10-11
Return noPermission in ledger admin call

officialfrancismendoza:
**Issue** (#280): ledger API call does not return an error message at all instead of returning noPermission error message when using admin methods.
**Fix**: Add noPermission error and conditional.

### DIFF
--- a/src/rpc/handlers/Ledger.cpp
+++ b/src/rpc/handlers/Ledger.cpp
@@ -46,6 +46,12 @@ doLedger(Context const& context)
         diff = params.at("diff").as_bool();
     }
 
+    if (params.contains(JS(full)))
+        return Status{RippledError::rpcNOT_SUPPORTED};
+
+    if (params.contains(JS(accounts)))
+        return Status{RippledError::rpcNOT_SUPPORTED};
+
     auto v = ledgerInfoFromRequest(context);
     if (auto status = std::get_if<Status>(&v))
         return *status;


src/rpc/handlers/Ledger.cpp
```diff
@@ -45,6 +45,13 @@ doLedger(Context const& context)
         diff = params.at("diff").as_bool();
     }
 
+    if (params.contains(JS(full)) && params.contains(JS(accounts)))
+    {
+        if ((params.at(JS(full)).as_bool() &&
```
godexsoft: I wonder if this will crash if i send a `full` or `accounts` that is not a `bool`?

cjcobb23:
We actually don't support these flags at all, admin or not. The ledger is far too big to return in one RPC. Is there any sort of not supported error we can return?

src/rpc/handlers/Ledger.cpp
```diff
@@ -45,6 +45,13 @@ doLedger(Context const& context)
...
+             params.at(JS(accounts)).as_bool()) == false)
+            return Status{Error::rpcNO_PERMISSION, "noPermission"};
```
godexsoft: We can use `rpcNOT_SUPPORTED` here or once we merge #396 we will be able to add and return a custom `clio` error for this if desired.

src/rpc/handlers/Ledger.cpp
godexsoft: I'd say let's remove this inner `if` checks completely (avoiding the crash) and just return `rpcNOT_SUPPORTED` when either `full` or `accounts` is present in the request.

godexsoft:
> We actually don't support these flags at all, admin or not. The ledger is far too big to return in one RPC. Is there any sort of not supported error we can return?

@officialfrancismendoza as @cjcobb23 suggested above - let's use `rpcNOT_SUPPORTED` in case any of these fields are passed be it a bool or not. This will simplify the fix a lot and we won't be parsing useless flags we aren't using.

godexsoft (changes requested):
Let's use `rpcNOT_SUPPORTED` - see comment.

src/rpc/handlers/Ledger.cpp
```diff
@@ -46,6 +46,18 @@ doLedger(Context const& context)
         diff = params.at("diff").as_bool();
     }
 
+    if (params.contains(JS(full)) && params.contains(JS(accounts)))
```
godexsoft: Split this into two separate `if`s each checking its own field. If any of the fields is present (we don't care the type anymore) just return the `rpcNOT_SUPPORTED` code.

src/rpc/handlers/Ledger.cpp
```diff
@@ -46,6 +46,18 @@ doLedger(Context const& context)
...
+    {
+        if (!(params.at(JS(full)).is_bool()) &&
+            !(params.at(JS(accounts)).is_bool()))
+            break;
+
+        else if (
+            (params.at(JS(full)).as_bool() &&
```
godexsoft: Checking for bool is irrelevant if we don't support the field at all.

src/rpc/handlers/Ledger.cpp
```diff
@@ -46,6 +46,18 @@ doLedger(Context const& context)
...
+             params.at(JS(accounts)).as_bool()) == false)
+            return Status{RippledError::rpcNOT_SUPPORTED, "notSupported"};
```
godexsoft: You don't need to pass "notSupported" as second argument - it should auto-fill the message from rippled codebase.

godexsoft (changes requested):

godexsoft (approved):
Looks like this should do it.

Issue #279, opened 2022-08-24
“ledger” with one or more fields are specified incorrectly do not return "invalidParams".


mounikakun:
Expected response: `invalidParams`
Actual response: `lgrNotFound`

Sample request:
`{
   "method": "ledger",
   "params": [
       {
           "ledger_index": "ghgj"
       }
   ]
}`

officialfrancismendoza:
@ManiMounikaKunasani I don't think we should downgrade from LGR_NOT_FOUND to invalidParams (as it's less accurate), but have pushed the PR anyway

PR #365, opened 2022-10-17
Add rpcINVALID_PARAMS error code

officialfrancismendoza:
**Issue (#279)**: Ledger API call with one or more fields specified incorrectly does not return invalidParams
**FIx**: Add rpcINVALID_PARAMS error code 

**Note**: Upon testing, hit ledgerIndexMalformed error instead of ledgerNotFound. Don't think it's a good idea to downgrade from lgrNotFound to invalidParams (as the error message is less specific), but pushed the PR anyway

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -592,7 +592,7 @@ ledgerInfoFromRequest(Context const& ctx)
     }
 
     if (!ledgerSequence)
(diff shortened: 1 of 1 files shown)


godexsoft (dismissed):
LGTM

godexsoft (dismissed):
LGTM

godexsoft (approved):

PR #370, opened 2022-10-25
fix bug on cache download from peer when ledger not found

cjcobb23:
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/370)


godexsoft (approved):
I'd prefer some sort of utility to take care of this delay and max-attempts deal but it's ok as is - can refactor later.

PR #371, opened 2022-10-25
Implement logging abstraction

### LINKED ISSUE #290: Explore and improve logging system
`Boost.Log` is currently used as the logging facility.
While useful, it's quite possible that it's not the fastest and not the most convenient way to go about logging.

Explore alternatives, especially in regards to using `std::format` which allows for a faster printer and better syntax.
It would also be good to have a wrapper on top of whatever backend we end up choosing:
```
// Now:
BOOST_LOG_TRIVIAL(info) << "Log some value " << value;

// After (if using a macro):
LOG_INFO("Log some value {}", value);

// Or better:
clio::logi("info log line {1}: {0}", first, second);
```
> Note: Initial implementation can use `Boost.Log` under the hood.

Another direction to look into is allowing to configure different levels for different components. For example while working on a new `RPC` call it may be useful to see verbose logging from the `RPC` component but not the 100s of lines per second of verbose messages from `ETLSource`. 
It could look something like this:
```
clio::logi("generic info log line");
clio::etl_logv("verbose log line for ETL stuff");
clio::rpc_logd("debug log line for RPC stuff");
```

godexsoft:
Fixes #290 

This change does the following few things:
- Introduces an abstraction for logging within `clio` (helps with separation of concerns, testability, DI, etc.)
- Implements a logger for tracking `Performance` specific things - to be used by Sophia in performance testing
- Each `Logger` object will use their own `mutex` so that logs in one part of the app don't interlock with all others.
- Adds a globally available `LogService` that can be used to raise Alerts and push logs into the General channel.
- Adds unit tests for logging primitives

The intention is to help performance and battle the 40% performance decrease we have seen lately due to many new log lines (multiple on each rpc).

### DIFF
--- a/src/log/Logger.cpp
+++ b/src/log/Logger.cpp
@@ -1,0 +1,191 @@
+#include <config/Config.h>
+#include <log/Logger.h>
+
+#include <algorithm>
+#include <array>
+#include <filesystem>
+
+namespace clio {
+
+Logger LogService::general_log_ = Logger{"General"};
+Logger LogService::alert_log_ = Logger{"Alert"};
+
+std::ostream&
+operator<<(std::ostream& stream, Severity sev)
+{
+    static constexpr std::array<const char*, 6> labels = {
+        "TRC",
+        "DBG",
+        "NFO",
+        "WRN",
+        "ERR",
+        "FTL",
+    };
+
+    return stream << labels.at(static_cast<int>(sev));
+}
+
+Severity
+tag_invoke(boost::json::value_to_tag<Severity>, boost::json::value const& value)
+{
+    if (not value.is_string())
+        throw std::runtime_error("`log_level` must be a string");
+    auto const& logLevel = value.as_string();
+
+    if (boost::iequals(logLevel, "trace"))
+        return Severity::TRACE;
+    else if (boost::iequals(logLevel, "debug"))
+        return Severity::DEBUG;
+    else if (boost::iequals(logLevel, "info"))
+        return Severity::INFO;
+    else if (
+        boost::iequals(logLevel, "warning") || boost::iequals(logLevel, "warn"))
+        return Severity::WARNING;
+    else if (boost::iequals(logLevel, "error"))
+        return Severity::ERROR;
+    else if (boost::iequals(logLevel, "fatal"))
+        return Severity::FATAL;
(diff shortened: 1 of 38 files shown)


src/log/Logger.cpp
```diff
@@ -0,0 +1,191 @@
...
+    {
+        boost::log::add_console_log(std::cout, keywords::format = format);
+    }
+
+    auto logDir = config.maybeValue<std::string>("log_directory");
+    if (logDir)
+    {
+        boost::filesystem::path dirPath{logDir.value()};
+        if (!boost::filesystem::exists(dirPath))
+            boost::filesystem::create_directories(dirPath);
```
cjcobb23, line 79: Is there a race condition here, where two `LogService`s try to create the same directory?

src/webserver/WsBase.h
```diff
@@ -114,8 +118,7 @@ class WsSession : public WsBase,
         if (!ec_ && ec != boost::asio::error::operation_aborted)
         {
             ec_ = ec;
-            BOOST_LOG_TRIVIAL(info)
-                << tag() << __func__ << ": " << what << ": " << ec.message();
+            perfLog_.info() << tag() << ": " << what << ": " << ec.message();
```
cjcobb23, line 121: Should these be at info or debug? This seems like its used to trace the execution of an individual request at a very fine grained level, so seems more like debug.

cjcobb23 (dismissed):
Left a few comments, but this LGTM

src/webserver/WsBase.h
godexsoft: It used to be `info` in a common log but now it's info in the 'Performance` channel which we can also tweak separately or turn off completely. We could ask Sophia what she wants this to be but i think it's not too important as it will be most likely turned off on any production system outside of performance testing.

src/log/Logger.cpp
godexsoft: This static `init` function is called once for the entire application (in main.cpp). `LogService` is like a static wrapper for the `boost::log::core`.
However, i forgot to disable construction of `LogService` instances so that's something to add for safety and to make it obvious that you can't create another `LogService` within one app.

cjcobb23 (approved):

PR #373, opened 2022-10-26
Change index name for nft buy/sell offers api

shawnxie999:
The NFT offer index is name "nft_offer_index" in rippled. We want to conform to the API design in clio


godexsoft (approved):
LGTM

Issue #275, opened 2022-08-23
“ledger_entry” with request improperly specified a Currency Code field do not return "malformedCurrency"


mounikakun:
Expected response: `malformedCurrency`
Actual response: `invalidParams`

Sample request:
`{
  "method": "ledger_entry",
  "params": [{
    "ripple_state": {
      "accounts": [
        "rEgAvY1qyEfZxwSaDg5UdLtNQzyvSPEXtV",
        "rhyxmFfxujVTfxBjug37kouzms6YzWpAUy"
      ],
      "currency": "aa"
    },
    "ledger_index": "validated"
  }]
}`

officialfrancismendoza:
@ManiMounikaKunasani there is no dedicated error code for malformedCurrency within ErrorCodes.h. Closing this issue as the plaintext error already returns "malformedCurrency'" and needs no further modification

officialfrancismendoza:
@ManiMounikaKunasani changed the "malformedCurrency" text error to "invalidParams" but don't think this is advisable as we are downgrading from a more specific error message. A custom error code for malformedCurrency does not within rippled's ErrorCodes.h

mounikakun:
@officialfrancismendoza I think there's some misunderstanding here. Expected change from this issue is the `error` in Clio response should be `malformedCurrency` instead of `invalidParams`.

Sample rippled response:
`{
    "result": {
        "error": "malformedCurrency",
        "ledger_hash": "2490B57E90FC49F9F930A1C0E7C31DF253B61F28444A1A0171A515F33E946D60",
        "ledger_index": 23070262,
        "request": {
            "command": "ledger_entry",
            "ledger_index": "validated",
            "ripple_state": {
                "accounts": [
                    "rEgAvY1qyEfZxwSaDg5UdLtNQzyvSPEXtV",
                    "rhyxmFfxujVTfxBjug37kouzms6YzWpAUy"
                ],
                "currency": "aa"
            }
        },
        "status": "error",
        "validated": true
    }
}`

Sample Clio response:
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "malformedCurrency",
        "status": "error",
        "type": "response",
        "request": {
            "method": "ledger_entry",
            "params": [
                {
                    "ripple_state": {
                        "accounts": [
                            "rEgAvY1qyEfZxwSaDg5UdLtNQzyvSPEXtV",
                            "rhyxmFfxujVTfxBjug37kouzms6YzWpAUy"
[...]

PR #378, opened 2022-10-31
ledger_entry does not return malformedCurrency

officialfrancismendoza:
**Issue (#275):** Ledger entry API call with improperly specified currency code field does not return "malformedCurrency"
**Fix**: Create custom rpcMALFORMED_CURRENCY error code and revert back to malformedCurrency hardcoded string

**Note**: Based on Mounika's feedback, I created a custom error code within Rippled's ErrorCodes.h to match with the original malformedCurrency hardcoded string, as the former did not exist before. Was told before to not change Rippled files and keep changes within Clio, but this appears to be an exception. Reviewers- please determine if this change to rippled is safe to add

### DIFF
--- a/build/_deps/rippled-src/src/ripple/protocol/ErrorCodes.h
+++ b/build/_deps/rippled-src/src/ripple/protocol/ErrorCodes.h
@@ -1,0 +1,343 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012 - 2019 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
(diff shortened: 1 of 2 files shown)


build/_deps/rippled-src/src/ripple/protocol/ErrorCodes.h
```diff
@@ -0,0 +1,343 @@
+//------------------------------------------------------------------------------
```
godexsoft, line 1: This is a file from `rippled` that is part of the `build` directory and will be replaced on a clean build of `clio`.
Nothing in the `build` directory should be ever added to the repo so unfortunately this is not a viable solution.
If we were to add this error code we would have to add it in the `rippled` repo.

Perhaps it's time for `clio` to rethink how we handle custom error codes.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -273,7 +273,7 @@ doLedgerEntry(Context const& context)
         if (!state.contains(JS(currency)) ||
             !state.at(JS(currency)).is_string())
         {
-            return Status{Error::rpcINVALID_PARAMS, "malformedCurrency"};
+            return Status{Error::rpcMALFORMED_CURRENCY, "malformedCurrency"};
```
godexsoft, line 276: You added the error codes header from `build` but forgot to add `rpcMALFORMED_CURRENCY`? Or perhaps it got overwritten by a rebuild of `clio`. Either way this does not work.

godexsoft (changes requested):
I'm thinking me or @cjcobb23 will have to think how we add custom error codes and tackle this issue ourselves.

build/_deps/rippled-src/src/ripple/protocol/ErrorCodes.h
officialfrancismendoza: @godexsoft so current way that Clio handles error codes: it directly inherits the hardcoded error string requirement that is required within Rippled. However, Clio now also utilizes the rpc error codes from ErrorCodes.h in addition to the hardcoded string

According to @cjcobb23, there is a default error string associated with every rpc error code. If there is no hardcoded error string mentioned, it reverts to whatever hardcoded message the rpc error had by default.

Seeing as we shouldn't change the Rippled build of ErrorCodes.h (which I mentioned as inadvisable before on the [original issue thread](https://github.com/XRPLF/clio/issues/275)), perhaps we can make a helper ErrorCodesClio.h file as the dedicated way to add any new error codes in the future. I tracked down which files include Rippled's ErrorCodes.h in which we can trivially add the new helper file in an attached picture below. What do you think of this approach?

In contrast, I think reverting back to hardcoded error strings is inadvisable, as there's useful data that is part of the rpc error codes that could go a long way in debugging as the system grows. 
image

src/rpc/handlers/LedgerEntry.cpp
officialfrancismendoza: @godexsoft let's talk further on how to rethink Clio custom error codes as this is the only viable solution forward

Issue #270, opened 2022-08-23
“gateway_balances” with unfunded account do not return "actNotFound" error.


mounikakun:
Expected response: `actNotFound`
Actual response:No error

Sample request:
`{
   "method": "gateway_balances",
   "params": [
       {
           "account": "rhKKk5KB8LuxmjvTSA7r3iYmJXGqnno7Mf",
           "ledger_index": "validated"
       }
   ]
}
`

cjcobb23:
fixed in https://github.com/XRPLF/clio/pull/382

PR #382, opened 2022-11-03
Return actNotFound for non-existent account in account iteration methods

### LINKED ISSUE #260: “account_objects” with unfunded account do not return actNotFound error.
The address specified in the account field of the request does not correspond to an account in the ledger.
Expected output: `actNotFound` error
Actual output: No error

Sample request:
`{
  "method": "account_objects",
  "params": [
      {
          "account": "rB7LiohHNiAnP4t1QK2RNL4YGat57sCDMz"
      }
  ]
}
`

cjcobb23:
Closes https://github.com/XRPLF/clio/issues/260 and https://github.com/XRPLF/clio/issues/270

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/382)

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -663,6 +663,10 @@ traverseOwnedNodes(
     boost::asio::yield_context& yield,
     std::function<void(ripple::SLE&&)> atOwnedNode)
 {
+    if (!backend.fetchLedgerObject(
+            ripple::keylet::account(accountID).key, sequence, yield))
+        return Status{Error::rpcACT_NOT_FOUND};
+
     auto parsedCursor =
         parseAccountCursor(backend, sequence, jsonCursor, accountID, yield);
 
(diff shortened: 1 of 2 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -663,6 +663,10 @@ traverseOwnedNodes(
     boost::asio::yield_context& yield,
     std::function<void(ripple::SLE&&)> atOwnedNode)
 {
+    if (!backend.fetchLedgerObject(
+            ripple::keylet::account(accountID).key, sequence, yield))
+        return Status{Error::rpcACT_NOT_FOUND};
```
officialfrancismendoza, line 668: @cjcobb23 since this is part of Clio, do we not have a hardcoded error string in addition to the rpc error code (rippled ErrorCodes.h)?

officialfrancismendoza (approved):

src/rpc/RPCHelpers.cpp
cjcobb23: the string is added by default, based on the code

godexsoft (approved):
LGTM

Issue #354, opened 2022-10-06
For successful transaction of subscribe and unsubscribe, result parameter has extra parameters in Clio response.


mounikakun:
`"validated": true` 
for subscribe to accounts, accounts_proposed, order books, streams

`"status": "success",
    "validated": true`
for unsubscribe to accounts, accounts_proposed, order books, streams

cjcobb23:
In general, extra fields are not a problem. However, it can't hurt to remove them, and `accounts_proposed` is not returning validated data, so `"validated":true` there is misleading

godexsoft:
This `validated` flag is added globally for all RPCs that return a meaningful JSON object.
To _remove_ it from only those methods means we need to add it to each method that _should_ return this flag.
I'm not sure it's worth the effort. Let me know what you think @cjcobb23 .

PR #383, opened 2022-11-03
Fix account ownership check

### LINKED ISSUE #222: "account_objects" request with parameter 'marker' returns "invalidParams" error.
Here is the sample request and response:
`{
  "method": "account_objects",
  "params": [
      {
          "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
          "deletion_blockers_only": false,
          "limit": 2,
          "validated": true,
          "marker": "77D55F149C997B7A15F683F9C568A2087DCEAD913962689F9E0A7A56E5F423C5,0"
      }
  ]
}`
 
 `{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "Malformed cursor",
        "status": "error",
        "type": "response",
        "request": {
            "method": "account_objects",
            "params": [
                {
                    "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
                    "deletion_blockers_only": false,
                    "limit": 2,
                    "validated": true,
                    "marker": "77D55F149C997B7A15F683F9C568A2087DCEAD913962689F9E0A7A56E5F423C5,0"
                }
            ]
        },
        "warnings": [
            "This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"
        ]
    },
    "warnings": [
        "This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request",
        "Too many requests"
    ]
}`

godexsoft:
Fixes #222 

When recreating the cursor from a marker in `account_objects` we call out to 'isOwnedByAccount' which seems to be a safety net of sorts. 
If the marker happens to land on an entry where the `accountID` we check against is the `destination` the check will fail.
The fix just adds a check for this case. 

On a second thought tho, maybe we don't need this `isOwnedByAccount` check at all?

### DIFF
--- a/src/rpc/Errors.h
+++ b/src/rpc/Errors.h
@@ -81,6 +81,34 @@ struct Status
             return *err != RippledError::rpcSUCCESS;
         return true;
     }
+
+    /**
+     * @brief Returns true if the Status contains the desired @ref RippledError
+     *
+     * @param other The RippledError to match
+     * @return bool true if status matches given error; false otherwise
+     */
+    bool
+    operator==(RippledError other) const
+    {
+        if (auto err = std::get_if<RippledError>(&code))
+            return *err == other;
+        return false;
+    }
+
+    /**
+     * @brief Returns true if the Status contains the desired @ref ClioError
+     *
+     * @param other The RippledError to match
+     * @return bool true if status matches given error; false otherwise
+     */
+    bool
+    operator==(ClioError other) const
+    {
+        if (auto err = std::get_if<ClioError>(&code))
+            return *err == other;
+        return false;
+    }
 };
 
 /**
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -81,35 +81,11 @@ getRequiredUInt(boost::json::object const& request, std::string const& field)
         throw InvalidParamsError("Missing field " + field);
 }
 
-bool
-isOwnedByAccount(ripple::SLE const& sle, ripple::AccountID const& accountID)
-{
-    if (sle.getType() == ripple::ltRIPPLE_STATE)
-    {
-        return (sle.getFieldAmount(ripple::sfLowLimit).getIssuer() ==
-                accountID) ||
-            (sle.getFieldAmount(ripple::sfHighLimit).getIssuer() == accountID);
-    }
-    else if (sle.isFieldPresent(ripple::sfAccount))
-    {
-        return sle.getAccountID(ripple::sfAccount) == accountID;
-    }
-    else if (sle.getType() == ripple::ltSIGNER_LIST)
-    {
-        ripple::Keylet const accountSignerList =
-            ripple::keylet::signers(accountID);
-        return sle.key() == accountSignerList.key;
-    }
-
-    return false;
-}
-
 std::optional<AccountCursor>
 parseAccountCursor(
     BackendInterface const& backend,
     std::uint32_t seq,
     std::optional<std::string> jsonCursor,
-    ripple::AccountID const& accountID,
     boost::asio::yield_context& yield)
 {
     ripple::uint256 cursorIndex = beast::zero;
@@ -139,19 +115,6 @@ parseAccountCursor(
     {
         return {};
     }
-
-    // We then must check if the object pointed to by the marker is actually
-    // owned by the account in the request.
-    auto const ownedNode = backend.fetchLedgerObject(cursorIndex, seq, yield);
-
-    if (!ownedNode)
-        return {};
-
-    ripple::SerialIter it{ownedNode->data(), ownedNode->size()};
-    ripple::SLE sle{it, cursorIndex};
-
-    if (!isOwnedByAccount(sle, accountID))
-        return {};
 
     return AccountCursor({cursorIndex, startHint});
 }
@@ -663,13 +626,11 @@ traverseOwnedNodes(
             ripple::keylet::account(accountID).key, sequence, yield))
         return Status{RippledError::rpcACT_NOT_FOUND};
 
-    auto parsedCursor =
-        parseAccountCursor(backend, sequence, jsonCursor, accountID, yield);
-
-    if (!parsedCursor)
+    auto maybeCursor = parseAccountCursor(backend, sequence, jsonCursor, yield);
+    if (!maybeCursor)
         return Status(ripple::rpcINVALID_PARAMS, "Malformed cursor");
 
-    auto [hexCursor, startHint] = *parsedCursor;
+    auto [hexCursor, startHint] = *maybeCursor;
 
     return traverseOwnedNodes(
         backend,
@@ -715,22 +676,21 @@ traverseOwnedNodes(
         auto hintDir =
             backend.fetchLedgerObject(hintIndex.key, sequence, yield);
 
-        if (hintDir)
-        {
-            ripple::SerialIter it{hintDir->data(), hintDir->size()};
-            ripple::SLE sle{it, hintIndex.key};
+        if (!hintDir)
+            return Status(ripple::rpcINVALID_PARAMS, "Invalid marker");
 
-            for (auto const& key : sle.getFieldV256(ripple::sfIndexes))
-            {
-                if (key == hexMarker)
-                {
-                    // We found the hint, we can start here
-                    currentIndex = hintIndex;
-                    break;
-                }
-            }
+        ripple::SerialIter it{hintDir->data(), hintDir->size()};
+        ripple::SLE sle{it, hintIndex.key};
+
+        if (auto const& indexes = sle.getFieldV256(ripple::sfIndexes);
+            std::find(std::begin(indexes), std::end(indexes), hexMarker) ==
+            std::end(indexes))
+        {
+            // result in empty dataset
+            return AccountCursor({beast::zero, 0});
         }
 
+        currentIndex = hintIndex;
         bool found = false;
         for (;;)
         {
--- a/src/rpc/handlers/Subscribe.cpp
+++ b/src/rpc/handlers/Subscribe.cpp
@@ -442,8 +442,7 @@ doUnsubscribe(Context const& context)
     if (request.contains("books"))
         unsubscribeToBooks(books, context.session, *context.subscriptions);
 
-    boost::json::object response = {{"status", "success"}};
-    return response;
+    return boost::json::object{};
 }
 
 }  // namespace RPC
--- a/src/rpc/RPC.cpp
+++ b/src/rpc/RPC.cpp
@@ -352,7 +352,7 @@ buildResponse(Context const& ctx)
         if (auto object = get_if<boost::json::object>(&v);
             object && not shouldSuppressValidatedFlag(ctx))
         {
-            (*object)["validated"] = true;
+            (*object)[JS(validated)] = true;
         }
 
         return v;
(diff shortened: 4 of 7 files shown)


cjcobb23:
So I think we can probably remove this ownership check. All we are doing is trying to make sure the marker is a valid point to start iteration. I think we can do that check in here: https://github.com/XRPLF/clio/blob/8cc2de5643004fc53bafab328257b2341b73cfd7/src/rpc/RPCHelpers.cpp#L712
and it would be much less likely to be buggy. Specifically, if `hintDir` is not found, or if it is but `hexMarker` is not found in `sfIndexes`, return an error.

godexsoft:
> Specifically, if `hintDir` is not found, or if it is but `hexMarker` is not found in `sfIndexes`, return an error.

Take a look if it's what you wanted. I was not sure what to put as the error strings so if you have better suggestions - let me know.

cjcobb23 (commented):
What is rippled's behavior when you pass an invalid marker?

godexsoft:
> What is rippled's behavior when you pass an invalid marker?

@cjcobb23 This is what i got sending an invalid marker:
```
{
  "result": {
    "error": "invalidParams",
    "error_code": 31,
    "error_message": "Invalid field 'marker'.",
    "request": {
      "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
      "command": "account_objects",
      "deletion_blockers_only": false,
      "limit": 10,
      "marker": "1111",
      "validated": true
    },
    "status": "error"
  }
}
```

If i use a valid marker but change one character in it i get this:
```
{
  "result": {
    "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
    "account_objects": [],
    "ledger_current_index": 23293951,
    "status": "success",
    "validated": false
  }
}
```
For the same type of change, with `clio` i get this:
```
{
  "result": {
    "error": "invalidParams",
    "error_code": 31,
    "error_message": "Malformed cursor",
    "status": "error",
    "type": "response",
    "request": {
      "method": "account_objects",
      "params": [
        {
          "marker": "17D55F149C997B7A15F683F9C568A2087DCEAD913962689F9E0A7A56E5F423C5,0",
          "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
          "deletion_blockers_only": false,
          "limit": 10,
          "validated": true
        }
      ]
    }
  },
  "warnings": [
    {
      "id": 2001,
    }
  ]
}
```
I also get exactly the same error for any other completely malformed marker.

So i guess the takeaway is that we need to distinguish between invalid format of a marker and a marker that does not yield a cursor, right?

cjcobb23:
> > What is rippled's behavior when you pass an invalid marker?
> 
> @cjcobb23 This is what i got sending an invalid marker:
> 
> ```
> {
>   "result": {
>     "error": "invalidParams",
>     "error_code": 31,
>     "error_message": "Invalid field 'marker'.",
>     "request": {
>       "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
>       "command": "account_objects",
>       "deletion_blockers_only": false,
>       "limit": 10,
>       "marker": "1111",
>       "validated": true
>     },
>     "status": "error"
>   }
> }
> ```
> 
> If i use a valid marker but change one character in it i get this:
> 
> ```
> {
>   "result": {
>     "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
>     "account_objects": [],
>     "ledger_current_index": 23293951,
>     "status": "success",
>     "validated": false
>   }
> }
> ```
> 
> For the same type of change, with `clio` i get this:
> 
> ```
> {
>   "result": {
>     "error": "invalidParams",
>     "error_code": 31,
>     "error_message": "Malformed cursor",
>     "status": "error",
>     "type": "response",
>     "request": {
>       "method": "account_objects",
>       "params": [
>         {
>           "marker": "17D55F149C997B7A15F683F9C568A2087DCEAD913962689F9E0A7A56E5F423C5,0",
>           "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
>           "deletion_blockers_only": false,
>           "limit": 10,
>           "validated": true
>         }
>       ]
>     }
>   },
>   "warnings": [
>     {
>       "id": 2001,
>       "message": "This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"
>     }
>   ]
> }
> ```
> 
> I also get exactly the same error for any other completely malformed marker.
> 
> So i guess the takeaway is that we need to distinguish between invalid format of a marker and a marker that does not yield a cursor, right?

Yeah one is malformed, and one is sort of invalid but well formed at least. It looks like rippled is not returning an error for the invalid case, just returning nothing. So we should just copy that.

src/rpc/Errors.h
```diff
@@ -17,6 +17,7 @@ namespace RPC {
  */
 enum class ClioError {
     rpcMALFORMED_CURRENCY = 5000,
+    rpcMARKER_NOT_FOUND = 5001,
```
godexsoft: I'm now thinking whether we should use some rippled code instead of adding our own for this hack.. maybe `rpcOBJECT_NOT_FOUND`?

src/rpc/RPCHelpers.cpp
```diff
@@ -686,7 +680,7 @@ traverseOwnedNodes(
             std::find(std::begin(indexes), std::end(indexes), hexMarker) ==
             std::end(indexes))
         {
-            return Status(ripple::rpcINVALID_PARAMS, "Marker not found");
+            return Status(ClioError::rpcMARKER_NOT_FOUND);
```
cjcobb23: so rippled just returns an empty list in this case right? If so, I'd just return `AccountCursor({beast::zero,0})`. In the handlers, if we return a 0 cursor, we don't return anything to the client

cjcobb23 (changes requested):

src/rpc/RPCHelpers.cpp
godexsoft: Ah, that might be a really good way if it works. Let me try that

godexsoft: The only thing is that i'm not sure we can handle `validated=false` that way.

cjcobb23: what is `validated=false`?

godexsoft: If i use a valid marker but change one character in it i get this from rippled:

```
{
  "result": {
    "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
    "account_objects": [],
    "ledger_current_index": 23293951,
    "status": "success",
    "validated": false
  }
}
```

with your suggestion clio now gives:
```
{
  "result": {
    "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rX",
    "account_objects": [],
    "ledger_hash": "5CF222090AFB48C88FB27FD0BB9660404C39E2BFE90B47D8D2E4C5EA7D49598C",
    "ledger_current_index": 23294159,
    "validated": true,
    "status": "success"
  },
  "warnings": [
    {
      "id": 2001,
    },
    {
      "id": 2002,
    }
  ]
}
```

as you can see, rippled sets `validated` to false while `clio` does not. 
If you look at my code changes in this pr you will see that i have added a way for any rpc to set `validated` and not have it be reset to `true` which is literally hardcoded atm. :) 
So to sum up - we need to handle `validated` and set it to `false` in this case.. somehow. With your idea however i can't distinguish empty results vs. this marker that was not found situation.

godexsoft: I could do something like this:

```
    if (auto status = std::get_if<RPC::Status>(&next))
        return *status;

    auto const& nextMarker = std::get<RPC::AccountCursor>(next);
    if (nextMarker.isNonZero())
        response[JS(marker)] = nextMarker.toString();
    else
        response[JS(validated)] = false;

    return response;
```

but this leads to the final paged portion being returned with `validated=false` because apparently the `zero` nextMarker is used to signify end of paged data.

cjcobb23: oh don't worry about that, thats an intended discrepancy between clio and rippled. clio by default returns validated data, rippled by default returns not yet validated data. If you set `"ledger_index":"validated"` in your request to rippled, this discrepancy won't exist.

godexsoft: I'm only worried about it because i bet that the next bug report will be about this field differ from rippled's output..
Also even for `clio` i think it makes sense that for a bogus marker we get non-validated data (if not an error like before). But i can't see an easier solution to my custom error code approach. Maybe we can ignore it for the time being to move on with the bugs and later deal with the next bug that will be reported for this difference.

src/rpc/handlers/Subscribe.cpp
```diff
@@ -442,8 +442,7 @@ doUnsubscribe(Context const& context)
     if (request.contains("books"))
         unsubscribeToBooks(books, context.session, *context.subscriptions);
 
-    boost::json::object response = {{"status", "success"}};
-    return response;
+    return boost::json::object{};
```
godexsoft, line 445: This is for issue #354. We did not remove this `status` back then but it seems like that's what is needed to match `clio`'s `subscribe` behaviour and `rippled` as well.

src/rpc/RPCHelpers.cpp
cjcobb23: its ok, everything coming back from clio is always validated data. It can't not be, because clio only stores validated data. The only way to get not validated data from clio is if clio forwards the request to rippled. When you send a request to rippled, you should always set `"ledger_index":"validated"` before comparing to clio's response. Or set it to a number. In your example request, you set `"ledger_current_index"` instead of `"ledger_index"`

cjcobb23: We have been very clear to everyone from the beginning that clio only serves validated data. That won't change. It's a known and intended discrepancy. It's bad UX to serve not yet validated data

src/rpc/RPCHelpers.cpp
```diff
@@ -709,22 +670,21 @@ traverseOwnedNodes(
         auto hintDir =
             backend.fetchLedgerObject(hintIndex.key, sequence, yield);
 
-        if (hintDir)
-        {
-            ripple::SerialIter it{hintDir->data(), hintDir->size()};
-            ripple::SLE sle{it, hintIndex.key};
+        if (!hintDir)
+            return Status(ripple::rpcINVALID_PARAMS, "Invalid marker");
```
cjcobb23, line 680: Is this rippled's behavior? I think this corresponds to passing a marker that has a bogus hint, which is the single integer at the end of the marker.

src/rpc/RPCHelpers.cpp
godexsoft: Oh, that's useful to know about `ledger_current_index` not being the correct field to send here. @ManiMounikaKunasani you may have to update your tests with the correct parameter, can you check please?

src/rpc/RPCHelpers.cpp
godexsoft: I think so, yes.

src/rpc/RPCHelpers.cpp
godexsoft: nvm. re above - i must have added the wrong parameter by mistake while adjusting some unrelated curl request i had in history. I'll remember the difference, thanks.

cjcobb23 (dismissed):
LGTM

cjcobb23 (dismissed):

cjcobb23 (approved):

PR #384, opened 2022-11-03
handle invalidHotWallet in gateway_balances

### LINKED ISSUE #271: “gateway_balances” with invalid account in hot wallet do not return error.
One or more of the addresses specified in the hotwallet field is not the address of an account holding currency issued by the account from the request.
Expected response: `invalidHotWallet`
Actual response: No error

Sample request:
`{
    "method": "gateway_balances",
    "params": [
        {
            "account": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q",
            "hotwallet": [
                "rhKKk5KB8LuxmjvTSA7r3iYmJXGqnno7M"
            ],
            "ledger_index": "validated",
            "strict": true
        }
    ]
}`

cjcobb23:
fixes https://github.com/XRPLF/clio/issues/271

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/384)

### DIFF
--- a/src/rpc/handlers/GatewayBalances.cpp
+++ b/src/rpc/handlers/GatewayBalances.cpp
@@ -180,6 +180,14 @@ doGatewayBalances(Context const& context)
             return obj;
         };
 
+    for (auto& hw : hotWallets)
+    {
+        if (!hotBalances.contains(hw))
+        {
+            return Status{Error::rpcINVALID_PARAMS, "invalidHotWallet"};
+        }
+    }
+
     if (auto balances = toJson(hotBalances); balances.size())
         response[JS(balances)] = balances;
     if (auto balances = toJson(frozenBalances); balances.size())
--- a/src/rpc/handlers/GatewayBalances.cpp
+++ b/src/rpc/handlers/GatewayBalances.cpp
@@ -180,10 +180,11 @@ doGatewayBalances(Context const& context)
             return obj;
         };
 
+    auto containsHotWallet = [&](auto const& hw) {
+        return hotBalances.contains(hw);
+    };
     if (not std::all_of(
-            hotWallets.begin(), hotWallets.end(), [&](auto const& hw) {
-                return hotBalances.contains(hw);
-            }))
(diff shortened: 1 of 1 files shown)


src/rpc/handlers/GatewayBalances.cpp
```diff
@@ -180,6 +180,14 @@ doGatewayBalances(Context const& context)
             return obj;
         };
 
+    for (auto& hw : hotWallets)
```
godexsoft: For better readability this could be replaced with something like:
```
auto isHotBalance = [&](auto const& hw) { return hotBalances.contains(hw); };
if (not all_of(begin(hotWallets), end(hotWallets), isHotBalance))
    return Status{Error::rpcINVALID_PARAMS, "invalidHotWallet"}; 
```
It's up to you which you prefer. Otherwise consider to use `auto const&` in the `for`.

godexsoft (dismissed):

godexsoft (dismissed):
The way you collapsed is not more readable than the for loop you had, isn't it?
The reason i had separated the lambda into a separate temporary is to make the `all_of` line read as plain English. Having it collapsed like this sort-of defeats the purpose.
You can also rest assured that both versions, with or without temporary will yield exactly the same assembly anyways. Compiler's know how to optimize those out so it's not like it's harmful to have a temporary.

I'll leave it up to you how you want to have it in the end. I don't think this is all that important honestly :)

cjcobb23:
> The way you collapsed is not more readable than the for loop you had, isn't it? The reason i had separated the lambda into a separate temporary is to make the `all_of` line read as plain English. Having it collapsed like this sort-of defeats the purpose. You can also rest assured that both versions, with or without temporary will yield exactly the same assembly anyways. Compiler's know how to optimize those out so it's not like it's harmful to have a temporary.
> 
> I'll leave it up to you how you want to have it in the end. I don't think this is all that important honestly :)

you're right lol, I added the temporary. More readable that way.

godexsoft (approved):
Perfect :+1:

PR #391, opened 2022-11-05
Port ignore_default support for account_lines rpc

### LINKED ISSUE #375: "account_lines" with "ignore_default" parameter is not returning empty lines.
Work around:
- create a `TrustSet` transaction.
- Get `account_lines` for `issuer` account in the `TrustSet` transaction with `ignore_default:true`

Sample request:
`{
    "method": "account_lines",
    "params": [
        {
            "account": "r4s4fjL4HdrWmyUSeJYd4GFAXCBN3n18ad",
            "ignore_default": true
        }
    ]
}`

Rippled response:
`{
    "result": {
        "account": "r4s4fjL4HdrWmyUSeJYd4GFAXCBN3n18ad",
        "ledger_current_index": 22929992,
        "lines": [],
        "status": "success",
        "validated": false
    }
}`

Clio response:
`{
    "result": {
        "account": "r4s4fjL4HdrWmyUSeJYd4GFAXCBN3n18ad",
        "ledger_hash": "3694E5F6F640EBDF86967C7F683B85D7AFB7FC2A9932CF42620D0D74C208F587",
        "ledger_index": 22929927,
        "lines": [
            {
                "account": "rnpAyYtbn4zo8utuyYLNz5G41hNvbVk28o",
                "balance": "-10000",
                "currency": "USD",
                "limit": "0",
                "limit_peer": "10000",
                "quality_in": 0,
                "quality_out": 0,
                "no_ripple": true,
                "no_ripple_peer": false
            }
        ],
        "validated": true,
        "status": "success"
    },
    "warnings": [
        {
            "id": 2001,
        }
    ]
}`

FYI: There's no mention of `ignore_default` parameter in `account_lines` documentation so not sure of expected response ([https://xrpl.org/account_lines.html#account_lines](url)
Here's the link of `ignore_default` code in rippled
[https://github.com/nbougalis/rippled/commit/daccb5b4c0e86dd1d6217fe33f42f487211b2fb1](url)

godexsoft:
Fixes #375 

@cjcobb23 please take a careful look whether this implementation makes sense to you. It's ported from `rippled` and adapted to `clio`'s way of doing `account_lines`. 
I have tested it and it seems to do the right thing, matching rippled from what i could tell. But my tests sample size is not that big so I can't guarantee it is 100% correct especially because i don't really understand what it needs to do exactly.

### DIFF
--- a/src/rpc/handlers/AccountLines.cpp
+++ b/src/rpc/handlers/AccountLines.cpp
@@ -121,10 +121,19 @@ doAccountLines(Context const& context)
     std::optional<std::string> marker = {};
     if (request.contains(JS(marker)))
     {
-        if (!request.at(JS(marker)).is_string())
+        if (not request.at(JS(marker)).is_string())
             return Status{Error::rpcINVALID_PARAMS, "markerNotString"};
 
         marker = request.at(JS(marker)).as_string().c_str();
+    }
+
+    auto ignoreDefault = false;
+    if (request.contains(JS(ignore_default)))
+    {
+        if (not request.at(JS(ignore_default)).is_bool())
+            return Status{Error::rpcINVALID_PARAMS, "ignoreDefaultNotBool"};
+
+        ignoreDefault = request.at(JS(ignore_default)).as_bool();
     }
 
     response[JS(account)] = ripple::to_string(accountID);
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):
LGTM. Though apparently this API flag is not even documented. Not sure how I feel about mirroring undocumented API flags.

godexsoft:
> LGTM. Though apparently this API flag is not even documented. Not sure how I feel about mirroring undocumented API flags.

I don't like this either but i think in this particular case it's best to update the docs and support this the same way `rippled` does.

PR #393, opened 2022-11-07
Implement a simple check to suppress 'validated' flag output

### LINKED ISSUE #354: For successful transaction of subscribe and unsubscribe, result parameter has extra parameters in Clio response.

godexsoft:
Fixes #354 

We decided to leave `status` output as is. 
However, `validated` flag is now suppressed for both `subscribe` and `unsubscribe` commands.

### DIFF
--- a/src/rpc/RPC.cpp
+++ b/src/rpc/RPC.cpp
@@ -305,6 +305,13 @@ isClioOnly(std::string const& method)
     return handlerTable.isClioOnly(method);
 }
 
+bool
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):
LGTM

PR #395, opened 2022-11-09
Add clang-format git hook

### LINKED ISSUE #392: Setup pre-commit hook for clang-format
Currently we have a "lint" step thru github actions which means that code commited to the repo will be checked with clang-format (not really a linter, but ok). 
However, it happens too often that code that is not formatted correctly gets pushed and this leads to `commit --amend` and `git push -f` which would be nice to avoid. 
We can easily battle it by adding a git pre-commit hook with this command: 
```
find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format-11 -i
```

godexsoft:
Fixes #392 

- Add a pre-commit hook which uses `clang-format` to correct any style issues and aborts the ongoing commit in case some files had to be changed during the pre-commit hook. 
- Add git hooks installation howto to `CONTRIBUTING.md`

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+exec 1>&2
+
+# format all relevant sources
+find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format -i
+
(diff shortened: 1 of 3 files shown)


legleux (approved):
LGTM!

PR #396, opened 2022-11-10
Introduce support for custom error codes

### LINKED ISSUE #275: “ledger_entry” with request improperly specified a Currency Code field do not return "malformedCurrency"

godexsoft:
Fixes #275 

- Introduce a CombinedError that combines rippled errors with clio's custom ones
- Refactor a bit to make it more readable + add doxygen comments
- Add unit tests for everything RPC errors/warnings
- Add custom code for `malformedCurrency` and use it in `ledger_entry`

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -6,7 +6,7 @@ exec 1>&2
 find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format -i
 
 # check how many lines differ
-lines=$(git diff | wc -l)
+lines=$(git diff src unittests | wc -l)
 
 # check if there is any updated files
 if [ "$lines" != "0" ]; then
--- a/src/rpc/Errors.h
+++ b/src/rpc/Errors.h
@@ -1,0 +1,206 @@
+#ifndef REPORTING_RPC_ERRORS_H_INCLUDED
+#define REPORTING_RPC_ERRORS_H_INCLUDED
+
+#include <ripple/protocol/ErrorCodes.h>
+
+#include <boost/json.hpp>
+
+#include <optional>
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace RPC {
+
+/**
+ * @brief Custom clio RPC Errors.
+ */
+enum class ClioError {
+    rpcMALFORMED_CURRENCY = 5000,
+};
+
+/**
+ * @brief Holds info about a particular @ref ClioError.
+ */
+struct ClioErrorInfo
+{
+    ClioError const code;
+    std::string_view const error;
+    std::string_view const message;
+};
+
+/**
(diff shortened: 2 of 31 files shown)


.githooks/pre-commit
```diff
@@ -6,7 +6,7 @@ exec 1>&2
 find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format -i
 
 # check how many lines differ
-lines=$(git diff | wc -l)
+lines=$(git diff src unittests | wc -l)
```
godexsoft, line 9: This change was not planned but i realised my mistake from yesterday and decided to fix it here and now

src/rpc/Errors.h
```diff
@@ -0,0 +1,206 @@
...
+public:
+    explicit AccountNotFoundError(std::string const& acct) : account(acct)
+    {
+    }
+    const char*
+    what() const throw() override
+    {
+        return account.c_str();
+    }
+};
```
cjcobb23, line 146: Do we actually use these? I know they were here before.

godexsoft: Yes, seems like we still do use them. We could possibly get rid of them eventually but i'd keep for now to make the changes in this pr minimal (already quite a lot).

src/rpc/Errors.h
```diff
@@ -0,0 +1,206 @@
...
+{
+    ClioError const code;
+    std::string_view const error;
+    std::string_view const message;
+};
+
+/**
+ * @brief Clio uses compatible Rippled error codes for most RPC errors.
+ */
+using Error = ripple::error_code_i;
```
cjcobb23: Should we call this `RippledError`?

godexsoft: We can. I left it as `Error` because it's less changes on call sites - we use this `Error` quite a bit. Let me try to rename it

cjcobb23: oh thats true, lots of changes, we could make a separate issue/PR and do that later

godexsoft: In the same branch, after changing `Error` to `RippledError` i get `28 files changed, 289 insertions(+), 226 deletions(-)
`. WDYT should we just do it as part of this PR then?

cjcobb23: thats not too bad, I'm ok with that if you are, but don't feel strongly

godexsoft: Let's just get it done :+1:

cjcobb23 (approved):
LGTM

PR #398, opened 2022-11-11
Fix pre-commit to only check files changed by clang-format

### LINKED ISSUE #397: Clang-format git hook allow partial commit
Currently if we try to commit only some of the changed files it will trick pre-commit git hook to think clang-format did something. 
Need to calculate changed files (to be commited) and only check those as pre-commit runs.

godexsoft:
Fixes #397

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -2,14 +2,16 @@
 
 exec 1>&2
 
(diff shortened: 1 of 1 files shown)


legleux (approved):
:+1: :+1:

Issue #330, opened 2022-09-28
"account_nfts" with incorrect account returns "invalidParams" instead of "actNotFound" error


mounikakun:
Sample request:
`{
  "method": "account_nfts",
  "params": [{
    "account": "rhuEDQ2Q9UeTfu1LBWq7ZzxgrfYV6P5",
    "ledger_index": "validated"
  }]
}`
Response:
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "accountMalformed",
        "status": "error",
        "type": "response",
        "request": {
            "method": "account_nfts",
            "params": [
                {
                    "account": "fgdhj",
                    "ledger_index": "validated"
                }
            ]
        }
    },
    "warnings": [
        {
            "id": 2001,
            "message": "This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"
        }
    ]
}`

mounikakun:
invalidParams - One or more fields are specified incorrectly, or one or more required fields are missing.

actNotFound - The Address specified in the account field of the request does not correspond to an account in the ledger.
[(https://xrpl.org/account_nfts.html#account_nfts)]

"account" in sample response is malformed account so is this rippled or Clio bug?

cjcobb23:
The account in the request is actually malformed, as opposed to a proper account. I don't think actNotFound is the proper error here. What does rippled return?

mounikakun:
yes the account in the request is malformed one but rippled is returning `actNotFound` error. It might be an issue with rippled.

cjcobb23:
its treating the string as a seed. This is bad practice, but I guess in the vein of mirroring rippled, we should do the same

cjcobb23:
We decided to not treat arbitrary strings as account seeds in the API. This should be returning account malformed now.

mounikakun:
Clio response is returning `actMalformed` now and as it was decided to not mirror rippled response I am closing this issue.

PR #399, opened 2022-11-14
remove accountFromSeed

cjcobb23:
Addresses https://github.com/XRPLF/clio/issues/330

This results in no change in API behavior, but is rather a decision to intentionally deviate from the rippled API. Allowing clients to send seeds in place of accounts is a security risk, because then someone's secret seed is being transmitted over the network as well as being written to the logs. This is very bad practice from a security standpoint.

Now, it is expected that passing an arbitrary string in place of an account in the API will return `actMalformed` as opposed to `actNotFound`

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/399)

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -348,20 +348,6 @@ canHaveDeliveredAmount(
 }
 
 std::optional<ripple::AccountID>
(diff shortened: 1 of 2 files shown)


godexsoft (approved):
LGTM

PR #402, opened 2022-11-15
helper function for subscribe to ensure cleanup

cjcobb23:
There was a memory leak where some subscriptions were not being cleaned up. This commit fixes that, and makes it easier to ensure this mistake doesn't happen again in the future

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/402)

### DIFF
--- a/src/subscriptions/SubscriptionManager.cpp
+++ b/src/subscriptions/SubscriptionManager.cpp
@@ -155,7 +155,9 @@ SubscriptionManager::subLedger(
     boost::asio::yield_context& yield,
     std::shared_ptr<WsBase> session)
 {
-    ledgerSubscribers_.subscribe(session);
+    subscribeHelper(session, ledgerSubscribers_, [this](session_ptr session) {
+        unsubLedger(session);
+    });
 
     auto ledgerRange = backend_->fetchLedgerRange();
     assert(ledgerRange);
@@ -185,7 +187,9 @@ SubscriptionManager::unsubLedger(std::shared_ptr<WsBase> session)
 void
 SubscriptionManager::subTransactions(std::shared_ptr<WsBase> session)
 {
-    txSubscribers_.subscribe(session);
+    subscribeHelper(session, txSubscribers_, [this](session_ptr session) {
+        unsubTransactions(session);
+    });
 }
 
 void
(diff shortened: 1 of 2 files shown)


src/subscriptions/SubscriptionManager.cpp
```diff
@@ -446,14 +460,39 @@ SubscriptionManager::unsubProposedAccount(
...
 }
+void
+SubscriptionManager::subscribeHelper(
+    std::shared_ptr<WsBase>& session,
+    Subscription& subs,
+    CleanupFunction&& func)
+{
+    subs.subscribe(session);
+    std::unique_lock lk(cleanupMtx_);
+    cleanupFuncs_[session].emplace_back(func);
```
godexsoft: Here 'func' is an lvalue within the body of the function (even tho it's passed as &&, it has a name so it's an lvalue) so you are actually copying 'func'. Also `emplace_back` does not actually _emplace_ here so i suggest you do `push_back(std::move(func))` instead.

godexsoft (changes requested):
While this approach works.. perhaps this is a good opportunity to rethink the design here?
Why not use a simple wrapper class that holds onto both the `session` and the function to call on removal from the subscribers list.. you can call your lambda in the destructor and have a strong guarantee that it will be called at all times even if we throw an exception somewhere between adding it to the map and registering the cleanup function. I'd try to use a single vector/map here instead of two separate containers. WDYT?

cjcobb23:
> While this approach works.. perhaps this is a good opportunity to rethink the design here?
> Why not use a simple wrapper class that holds onto both the `session` and the function to call on removal from the subscribers list.. you can call your lambda in the destructor and have a strong guarantee that it will be called at all times even if we throw an exception somewhere between adding it to the map and registering the cleanup function. I'd try to use a single vector/map here instead of two separate containers. WDYT?

Should be possible. We'd still need a collection inside the wrapper class to track which cleanup functions to call for that session.

godexsoft:
> Should be possible. We'd still need a collection inside the wrapper class to track which cleanup functions to call for that session.

Yea, i see. Up to you if you want to refactor it now or we redesign it later.
Please at least still fix the `emplace_back`.

cjcobb23:
I think since this is a memory leak, i'd rather get the fix in now and redesign later

godexsoft (dismissed):
LGTM.

godexsoft (approved):
LGTM.

PR #407, opened 2022-11-17
include searched_all in error response of tx

### LINKED ISSUE #278: “tx” with min_ledger and max_ledger fields of the request are more than 1000 apart returns no error.

cjcobb23:
This PR fixes a couple of bugs. Fixes https://github.com/XRPLF/clio/issues/278. Also performs other checks on the ledger range supplied by the user in order to align with rippled. Also actually implements the `searched_all` functionality: https://xrpl.org/tx.html#not-found-response
Supersedes: https://github.com/XRPLF/clio/pull/359

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/407)

### DIFF
--- a/src/rpc/Errors.cpp
+++ b/src/rpc/Errors.cpp
@@ -132,10 +132,9 @@ makeError(Status const& status)
         status.code);
     if (status.extraInfo)
     {
-        for (auto it = status.extraInfo->begin(); it != status.extraInfo->end();
-             ++it)
+        for (auto& [key, value] : status.extraInfo.value())
         {
-            res[it->key()] = it->value();
+            res[key] = value;
         }
     }
     return res;
--- a/src/rpc/Errors.cpp
+++ b/src/rpc/Errors.cpp
@@ -106,7 +106,7 @@ makeError(Status const& status)
         return str.empty() ? nullopt : make_optional(str);
     };
 
-    return visit(
+    auto res = visit(
         overloadSet{
             [&status, &wrapOptional](RippledError err) {
                 if (err == ripple::rpcUNKNOWN)
@@ -130,6 +130,15 @@ makeError(Status const& status)
             },
         },
         status.code);
+    if (status.extraInfo)
+    {
+        for (auto it = status.extraInfo->begin(); it != status.extraInfo->end();
+             ++it)
+        {
+            res[it->key()] = it->value();
+        }
+    }
+    return res;
 }
 
 }  // namespace RPC
--- a/src/rpc/Errors.h
+++ b/src/rpc/Errors.h
@@ -50,9 +50,12 @@ struct Status
     CombinedError code = RippledError::rpcSUCCESS;
     std::string error = "";
     std::string message = "";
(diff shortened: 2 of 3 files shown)


src/rpc/Errors.cpp
```diff
@@ -130,6 +130,15 @@ makeError(Status const& status)
             },
         },
         status.code);
+    if (status.extraInfo)
+    {
+        for (auto it = status.extraInfo->begin(); it != status.extraInfo->end();
```
godexsoft: Wouldn't `for(auto const& [key, value] : status.extraInfo)` or something similar work here?

src/rpc/handlers/Tx.cpp
```diff
@@ -31,14 +31,35 @@ doTx(Context const& context)
...
     auto dbResponse = context.backend->fetchTransaction(hash, context.yield);
     if (!dbResponse)
+    {
+        if (rangeSupplied)
+        {
+            bool searchedAll = range->maxSequence >= maxLedger &&
+                range->minSequence <= minLedger;
+            boost::json::object extra;
+            extra["searched_all"] = searchedAll;
+            return Status{RippledError::rpcTXN_NOT_FOUND, extra};
```
godexsoft: Not a very expensive copy of `extra` but can be made into a `move` if you `std::move(extra)` and change/add a constructor of `Status` to support move. Depending on how often this is hit it can be important.

src/rpc/Errors.cpp
cjcobb23: good call, changed

src/rpc/handlers/Tx.cpp
cjcobb23: changed the constructor to support move

src/rpc/Errors.h
```diff
@@ -50,9 +50,12 @@ struct Status
     CombinedError code = RippledError::rpcSUCCESS;
     std::string error = "";
     std::string message = "";
+    std::optional<boost::json::object> extraInfo;
 
     Status() = default;
     /* implicit */ Status(CombinedError code) : code(code){};
+    Status(CombinedError code, boost::json::object&& extraInfo)
+        : code(code), extraInfo(extraInfo){};
```
godexsoft: I think you still need to `std::move` here as `extraInfo` is an lvalue here too.

godexsoft (changes requested):

godexsoft (dismissed):
LGTM.

godexsoft (approved):

PR #413, opened 2022-11-21
Return srcCurMalformed on invalid taker_pays in book_offers

### LINKED ISSUE #267: “book_offers” with the invalid taker_pays currency field in the request does not return "srcCurMalformed"

godexsoft:
Fixes #267

### DIFF
--- a/CMake/settings.cmake
+++ b/CMake/settings.cmake
@@ -1,1 +1,6 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-narrowing -Wno-dangling-else -Wno-deprecated-declarations")
+target_compile_options(clio  
+  PUBLIC -Wall
+         -Werror 
+         -Wno-narrowing
+         -Wno-deprecated-declarations 
+         -Wno-dangling-else)
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1317,46 +1317,48 @@ std::variant<Status, ripple::Book>
 parseBook(boost::json::object const& request)
 {
     if (!request.contains("taker_pays"))
-        return Status{RippledError::rpcBAD_MARKET, "missingTakerPays"};
+        return Status{
+            RippledError::rpcINVALID_PARAMS, "Missing field 'taker_pays'"};
 
     if (!request.contains("taker_gets"))
-        return Status{RippledError::rpcINVALID_PARAMS, "missingTakerGets"};
+        return Status{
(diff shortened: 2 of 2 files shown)


CMake/settings.cmake
```diff
@@ -1 +1,10 @@
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-narrowing -Wno-dangling-else -Wno-deprecated-declarations")
+# set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-narrowing -Wno-dangling-else -Wno-deprecated-declarations")
+
+target_compile_options(clio
```
godexsoft, line 1: This is a better way of setting flags for `clio` and derivatives (`clio_server`, `clio_tests`, etc.) only.

legleux: This might work?
```
 PUBLIC -Wall
        -Werror 
        -Wno-narrowing
        -Wno-deprecated-declarations 
        -Wno-unused-local-typedefs 
        -Wno-unknown-pragmas 
        -Wno-dangling-else)

File: src/rpc/RPCHelpers.cpp [hunk 2]
```diff
@@ -1331,7 +1331,7 @@ parseBook(boost::json::object const& request)
     auto taker_pays = request.at("taker_pays").as_object();
     if (!taker_pays.contains("currency"))
         return Status{
-            RippledError::rpcINVALID_PARAMS, "missingTakerPaysCurrency"};
+            RippledError::rpcSRC_CUR_MALFORMED, "missingTakerPaysCurrency"};
```
cjcobb23: Is this the same error message rippled returns? I think rippled returns `"Missing field 'taker_pays.currency'."` But I could be reading the code wrong.

godexsoft: Actually it returns 
``` 
"error": "srcCurMalformed",
"error_code": 69,
"error_message": "Source currency is malformed.",
```
I have tried to match `rippled` as close as possible. Will push the changes.

cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):

PR #415, opened 2022-11-21
Change returned error to accountNotFound

### LINKED ISSUE #277: “no ripple check” with invalid account returns "invalidParams" instead of "actNotFound"
Expected response: `actNotFound`

Sample request:
`{
   "method": "noripple_check",
   "params": [
       {
           "account": "raeR1nAGkMm18hHpUz1ZP6wGKLA4WCh8rXa",
           "limit": 2,
           "role": "gateway",
           "transactions": true
       }
   ]
}
`

godexsoft:
Fixes #277

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -193,8 +193,7 @@ getAccount(
         return {};
     }
 
-    return Status{
-        RippledError::rpcACT_MALFORMED, field.to_string() + "Malformed"};
+    return Status{RippledError::rpcACT_NOT_FOUND};
 }
 
 Status


src/rpc/RPCHelpers.cpp
```diff
@@ -193,8 +193,7 @@ getAccount(
         return {};
     }
 
-    return Status{
-        RippledError::rpcACT_MALFORMED, field.to_string() + "Malformed"};
+    return Status{RippledError::rpcACT_NOT_FOUND};
```
cjcobb23, line 196: I don't think we should make this change. The account is malformed, as opposed to well formed but not found. This is the issue we were talking about before, where rippled interprets malformed accounts as just strings and treats them as a seed. Given that we decided to not accept seeds as accounts, I believe we are already returning the right error here, and there is no change that needs to be made to the code as is.

cjcobb23 (changes requested):

src/rpc/RPCHelpers.cpp
godexsoft: @ManiMounikaKunasani please see above and let me know if you have objections.

godexsoft:
Was decided not to allow seeds. Closing

PR #418, opened 2022-11-22
Add limit to response only if next marker is available

### LINKED ISSUE #221: Missing  'limit' parameter in the clio response of "account_lines", "account_channels", "account_objects", "account_offers" requests.

godexsoft:
Fixes #221 

Fixing issues from the latest comments on #221:
- Add 'limit' to response only if a `marker` is also available (meaning there is more data) for
  - Account channels
  - Account lines
  - Account offers

### DIFF
--- a/src/rpc/handlers/AccountChannels.cpp
+++ b/src/rpc/handlers/AccountChannels.cpp
@@ -85,7 +85,6 @@ doAccountChannels(Context const& context)
 
     response[JS(account)] = ripple::to_string(accountID);
     response[JS(channels)] = boost::json::value(boost::json::array_kind);
-    response[JS(limit)] = limit;
     boost::json::array& jsonChannels = response.at(JS(channels)).as_array();
 
     auto const addToResponse = [&](ripple::SLE&& sle) {
@@ -118,7 +117,10 @@ doAccountChannels(Context const& context)
     auto nextMarker = std::get<RPC::AccountCursor>(next);
 
     if (nextMarker.isNonZero())
+    {
         response[JS(marker)] = nextMarker.toString();
+        response[JS(limit)] = limit;
+    }
 
     return response;
 }
--- a/src/rpc/handlers/AccountLines.cpp
+++ b/src/rpc/handlers/AccountLines.cpp
@@ -140,7 +140,6 @@ doAccountLines(Context const& context)
     response[JS(account)] = ripple::to_string(accountID);
     response[JS(ledger_hash)] = ripple::strHex(lgrInfo.hash);
     response[JS(ledger_index)] = lgrInfo.seq;
-    response[JS(limit)] = limit;
     response[JS(lines)] = boost::json::value(boost::json::array_kind);
     boost::json::array& jsonLines = response.at(JS(lines)).as_array();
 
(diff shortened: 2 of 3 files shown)


cjcobb23:
I'm a bit confused why we need this. First of all, there's no harm in returning the `"limit"` field if rippled does not return the field. Extra info is ok, won't break anyone's script. Unless there's some special logic around when the field should or shouldn't be returned.

Second, limit is not documented at all for `account_lines`, and for `account_objects` and `account_channels`, it's listed as "May be omitted". I'm confused why we are going through the trouble to mirror rippled's behavior for a field that is marked as may be omitted.

Third, the docs definitely don't say "Omitted if next marker is not available." This is some totally undocumented runtime behavior of rippled. This I do not think we need to emulate. The docs say "The limit that was used in this request, if any." I proposed for all of these requests, we just do this. Return the limit that was specified in the request.

godexsoft:
> I'm a bit confused why we need this. First of all, there's no harm in returning the `"limit"` field if rippled does not return the field. Extra info is ok, won't break anyone's script. Unless there's some special logic around when the field should or shouldn't be returned.
> 
> Second, limit is not documented at all for `account_lines`, and for `account_objects` and `account_channels`, it's listed as "May be omitted". I'm confused why we are going through the trouble to mirror rippled's behavior for a field that is marked as may be omitted.
> 
> Third, the docs definitely don't say "Omitted if next marker is not available." This is some totally undocumented runtime behavior of rippled. This I do not think we need to emulate. The docs say "The limit that was used in this request, if any." I proposed for all of these requests, we just do this. Return the limit that was specified in the request.

The limit used can differ from the limit specified by the client as we have min/max/default limits for each rpc.
I think if we are to always display the limit we should at least report back the real limit used, not the one sent to us in the request. WDYT?

cjcobb23:
> > I'm a bit confused why we need this. First of all, there's no harm in returning the `"limit"` field if rippled does not return the field. Extra info is ok, won't break anyone's script. Unless there's some special logic around when the field should or shouldn't be returned.
> > Second, limit is not documented at all for `account_lines`, and for `account_objects` and `account_channels`, it's listed as "May be omitted". I'm confused why we are going through the trouble to mirror rippled's behavior for a field that is marked as may be omitted.
> > Third, the docs definitely don't say "Omitted if next marker is not available." This is some totally undocumented runtime behavior of rippled. This I do not think we need to emulate. The docs say "The limit that was used in this request, if any." I proposed for all of these requests, we just do this. Return the limit that was specified in the request.
> 
> The limit used can differ from the limit specified by the client as we have min/max/default limits for each rpc. I think if we are to always display the limit we should at least report back the real limit used, not the one sent to us in the request. WDYT?

I'm fine with just always reporting back the real limit used. As long as it's something consistent, easy to code and easy to understand, I'm ok with it.

godexsoft:
> I'm fine with just always reporting back the real limit used. As long as it's something consistent, easy to code and easy to understand, I'm ok with it.

We can just close this issue then as it should already be doing exactly that after @officialfrancismendoza added the `limit` lines.

@ManiMounikaKunasani fyi, see the reasoning and decision above.

PR #417, opened 2022-11-21
Add custom error for malformed owner and request

godexsoft:
Fixes #274 

- Add two custom errors
- Try to match rippled errors

PR #419, opened 2022-11-22
Use custom malformedAddress error in ledger_entry

### LINKED ISSUE #272: “ledger_entry” with request improperly specified an Address field returns "invalidParams" instead of "malformedAddress".

godexsoft:
Fixes #272 
- Add a custom error code for malformedAddress
- Use the new error in relevant places (except for `escrow` case where `account` should not be supported according to docs)

### DIFF
--- a/src/rpc/Errors.h
+++ b/src/rpc/Errors.h
@@ -19,6 +19,7 @@ enum class ClioError {
     rpcMALFORMED_CURRENCY = 5000,
     rpcMALFORMED_REQUEST = 5001,
     rpcMALFORMED_OWNER = 5002,
+    rpcMALFORMED_ADDRESS = 5003,
 };
 
 /**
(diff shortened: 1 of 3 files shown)


src/rpc/Errors.h
```diff
@@ -17,6 +17,7 @@ namespace RPC {
  */
 enum class ClioError {
     rpcMALFORMED_CURRENCY = 5000,
+    rpcMALFORMED_ADDRESS = 5003,
```
godexsoft, line 22: Note that we skip to 5003. This is because in PR #417 we add 5001 and 5002 and this will be merged after #417

cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):

Issue #411, opened 2022-11-18
Add copyright info to each file


godexsoft:
We should add and maintain copyright comments similar to what `rippled` does:
```
//------------------------------------------------------------------------------
/*
    This file is part of clio: https://github.com/XRPLF/clio
    Copyright (c) 2022 Ripple Labs Inc.
    Permission to use, copy, modify, and/or distribute this software for any
    purpose  with  or without fee is hereby granted, provided that the above
    copyright notice and this permission notice appear in all copies.
    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
*/
//==============================================================================
```

j0ker70:
Hello. Sorry for asking but I am trying to contribute to my first open source project. Can you be more specific to which files should I actually add the copyright files? I have checked the `rippled` project. As it seems, in that project not all files have copyright statements included.

godexsoft:
Hi @j0ker70,
Only `.h` and `.cpp` source files should have this copyright at the top. 
If in `rippled` they missed this in some files it's probably a mistake.
If you would like to contribute to `clio` as your first opensource project - you are very welcome to take this issue or any other issue with `good first issue` label on it.

j0ker70:
Thanks for replying @godexsoft. I am a little stuck now. As told in `CONTRIBUTING.md`, I have completed these commands.

```
# Create a backup of your branch
git branch <your feature branch>_bk

# Rebase and squash commits into one
git checkout develop
git pull origin develop
git checkout <your feature branch>
git rebase -i develop
```

But after that changes are made in the feature-branch, not in develop branch. Please, can you tell me how can I push my changes?

[...]

PR #422, opened 2022-11-23
Update headers to use #pragma once

### LINKED ISSUE #286: Use `#pragma once` instead of include guards
Currently we have ifdefs/guards everywhere. 
It's better to use `#pragma once` instead as the compiler caches those and it's generally more efficient than relying on the pre-processor. 
As part of this issue please check whether `#pragma once` is supported on all our target compilers. It's worth noting that as we target **c++20** it's pretty safe to assume that modern major compilers are targeted and they should have support for `#pragma once`.

godexsoft:
Fixes #286

### DIFF
--- a/src/webserver/Ssl.h
+++ b/src/webserver/Ssl.h
@@ -1,24 +1,4 @@
-//------------------------------------------------------------------------------
-/*
-    This file is part of rippled: https://github.com/ripple/rippled
-    Copyright (c) 2021 Ripple Labs Inc.
-
-    Permission to use, copy, modify, and/or distribute this software for any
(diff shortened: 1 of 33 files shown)


src/webserver/Ssl.h
```diff
@@ -1,24 +1,4 @@
-//------------------------------------------------------------------------------
```
godexsoft, line 1: Removing this as this is clearly wrong anyways. Will be addressed in #411

officialfrancismendoza (approved):
Nice job @godexsoft. LGTM

legleux (approved):
Beautiful

PR #431, opened 2022-12-01
Add workaround for source_location for AppleClang 

### LINKED ISSUE #428: Fix source_location build error on MacOSX
On MacOSX, when using AppleClang (default, comes with Xcode), neither `source_location` nor `experimental::source_location` is available.

We need to
1) Check whether we have source_location during configuration/compilation
2) include the correct header from std; or
3) use our own implementation for when it's not available

godexsoft:
Fixes #428 

Introduces a small custom `source_location` clone that is used for compilers without direct support for `std::source_location` nor `std::experimental::source_location`.

### DIFF
--- a/CMake/deps/SourceLocation.cmake
+++ b/CMake/deps/SourceLocation.cmake
@@ -1,0 +1,11 @@
+include(CheckIncludeFileCXX)
+
+check_include_file_cxx("source_location" SOURCE_LOCATION_AVAILABLE)
+if(SOURCE_LOCATION_AVAILABLE) 
+    target_compile_definitions(clio PUBLIC "HAS_SOURCE_LOCATION")
+endif()
+
+check_include_file_cxx("experimental/source_location" EXPERIMENTAL_SOURCE_LOCATION_AVAILABLE)
+if(EXPERIMENTAL_SOURCE_LOCATION_AVAILABLE)     
+    target_compile_definitions(clio PUBLIC "HAS_EXPERIMENTAL_SOURCE_LOCATION")
+endif()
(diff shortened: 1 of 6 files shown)


CMake/deps/SourceLocation.cmake
```diff
@@ -0,0 +1,11 @@
+include(CheckIncludeFileCXX)
+
+check_include_file_cxx("source_location" SOURCE_LOCATION_AVAILABLE)
```
godexsoft, line 3: on mac this somehow finds `boost/assert/source_location.hpp` and so in Logger.h we check that builtin is available on top of this.

src/log/Logger.h
```diff
@@ -16,17 +16,58 @@
...
+    }
+    std::size_t
+    line() const
+    {
+        return line_;
+    }
+};
+using source_location_t = SourceLocation;
+#define CURRENT_SRC_LOCATION \
+    source_location_t(__builtin_FILE(), __builtin_LINE())
```
godexsoft, line 69: Since we require c++20, we assume here that these builtins should exist.

src/log/Logger.h
```diff
@@ -16,28 +16,69 @@
...
  */
 enum class Severity {
-    TRACE,
-    DEBUG,
-    INFO,
-    WARNING,
-    ERROR,
-    FATAL,
+    TRC,
+    DBG,
```
godexsoft, line 77: Typical preprocessor bug. "DEBUG" was replaced with "1" by the preprocessor when building from Debug configuration.

cjcobb23 (dismissed):
LGTM

cjcobb23 (approved):

Issue #427, opened 2022-11-30
Populate git commit hash into clio version string


godexsoft:
Currently `clio` is encoding git commit hash for debug builds only.
Having the commit hash for release does not hurt anyone but has an added benefit - it allows us to know which exact code is used while going thru `performance testing` and/or `running integration tests`. Multiple times already we had a `custom build` installed on a server for perf/integration testing and we had no way to prove which code it's build from other than hoping that we did not copy/install the wrong thing.

Plan of attack:
- Talk to @legleux to find out if there are objections
- See `src/main/impl/Build.cpp` 
- Version must always include the git commit hash. for both release and debug builds
- Make sure `server_info` shows the version string with the commit hash for debug and release.

PR #432, opened 2022-12-02
Add GIT_COMMIT_HASH to Clio Version String For Release

officialfrancismendoza:
**Issue (#427)**: GIT_COMMIT_HASH only populates for debug and not for release
**Fix**: Add additional macros so GIT_COMMIT_HASH populates for debug and release

### DIFF
--- a/src/main/impl/Build.cpp
+++ b/src/main/impl/Build.cpp
@@ -15,7 +15,7 @@ namespace Build {
 char const* const versionString = "1.0.3"
 // clang-format on
 
-#if defined(DEBUG) || defined(SANITIZER)
+#if defined(DEBUG) || defined(SANITIZER) || defined(RELEASE)
     "+"
 #ifdef CLIO_GIT_COMMIT_HASH
     CLIO_GIT_COMMIT_HASH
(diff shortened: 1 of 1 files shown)


src/main/impl/Build.cpp
```diff
@@ -36,6 +36,9 @@ char const* const versionString = "1.0.3"
         "-release"
 #endif
 
+#define VAL(CLIO_GIT_COMMIT_HASH) #CLIO_GIT_COMMIT_HASH
+#define TOSTRING(CLIO_GIT_COMMIT_HASH) VAL(CLIO_GIT_COMMIT_HASH)
```
cjcobb23, line 40: Why did we need to add these? Why couldn't we follow the same pattern we were following for debug builds?

src/main/impl/Build.cpp
```diff
@@ -44,10 +47,12 @@ getClioVersionString()
 {
     static std::string const value = [] {
         std::string const s = versionString;
+        std::string const t = std::to_string(CLIO_GIT_COMMIT_HASH);
         beast::SemanticVersion v;
         if (!v.parse(s) || v.print() != s)
             throw std::runtime_error(s + ": Bad server version string");
-        return s;
+        std::string const result = s + ", Git Commit Hash: " + t;
```
cjcobb23, line 54: Does this change the format of the string returned?

cjcobb23 (changes requested):

PR #433, opened 2022-12-02
add connection counting

cjcobb23:
Limit the number of concurrent connections the server allows from a specific ip. Defaults to 1. This will apply to open subscriptions as well. Also raises the default max fetches value.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/433)

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -13,16 +13,19 @@ class DOSGuard
     boost::asio::io_context& ctx_;
     std::mutex mtx_;  // protects ipFetchCount_
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::uint32_t> ipConnCount_;
     std::unordered_set<std::string> const whitelist_;
     std::uint32_t const maxFetches_;
     std::uint32_t const sweepInterval_;
+    std::uint32_t const maxConnCount_;
 
 public:
     DOSGuard(clio::Config const& config, boost::asio::io_context& ctx)
         : ctx_{ctx}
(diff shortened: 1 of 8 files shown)


src/webserver/DOSGuard.h
```diff
@@ -13,16 +13,19 @@ class DOSGuard
     boost::asio::io_context& ctx_;
     std::mutex mtx_;  // protects ipFetchCount_
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    std::unordered_map<std::string, std::uint32_t> ipConnCount_;
```
godexsoft, line 16: Consider one `unordered_map` + struct instead of separate maps.

src/webserver/DOSGuard.h
```diff
@@ -54,11 +57,44 @@ class DOSGuard
             return true;
 
         std::unique_lock lck(mtx_);
-        auto it = ipFetchCount_.find(ip);
-        if (it == ipFetchCount_.end())
-            return true;
+        bool fetchesOk = true;
+        bool connsOk = true;
+        {
+            auto it = ipFetchCount_.find(ip);
```
godexsoft, line 63: Consider immediately invoked lambda or a helper function for this. 
If you only have one map it could return a pair:
```
auto [fetchesOk, connsOk] = [this] { 
    auto it = talley_.find(ip); 
    return ...; 
}();
```

godexsoft (approved):
LGTM. Adding a few suggestions but overall feel free to merge as is.

cjcobb23:
Suggestions are good, but I'm going to merge this as is in the interest of time. May circle back later and refactor.

PR #436, opened 2022-12-06
Implement cli parsing using boost::po

### LINKED ISSUE #367: Rewrite CLI parser to use `boost::po`
Currently we are parsing `argv`/`argc` manually. We can do better by using `boost::po` like `rippled` does.

godexsoft:
Fixes #367 

Use `boost::program_options` to parse command line options.
Does not change the way `clio` is to be ran (with positional argument for config file path) but also adds:
- `-h` and `--help` to print help message
- `-v` and `--version` to print version and exit
- `-c` and `--conf` OR 1st positional argument to specify the configuration file path
- defaults config file path to `/etc/clio.json` (similarly to `rippled`). final default path is TBD.

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -45,7 +45,7 @@ namespace po = boost::program_options;
 std::string
 parseCli(int argc, char* argv[])
 {
-    static constexpr char defaultConfigPath[] = "/etc/clio.json";
+    static constexpr char defaultConfigPath[] = "/etc/opt/clio/config.json";
 
     // clang-format off
     po::options_description description("Options");
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -16,7 +16,9 @@
 #include <boost/asio/strand.hpp>
 #include <boost/beast/websocket.hpp>
 #include <boost/date_time/posix_time/posix_time_types.hpp>
+#include <boost/filesystem/path.hpp>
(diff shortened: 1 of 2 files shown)


src/main/main.cpp
```diff
@@ -31,9 +33,65 @@
...
+ * @brief Parse command line and return path to configuration file
+ *
+ * @param argc
+ * @param argv
+ * @return std::string Path to configuration file
+ */
+std::string
+parseCli(int argc, char* argv[])
+{
+    static constexpr char defaultConfigPath[] = "/etc/clio.json";
```
cjcobb23: Is this reasonable @legleux ?

legleux: That's funny, I did mention that to him. :eyes:

godexsoft: @cjcobb23 @legleux Can you both help come up with the best default? maybe there is a known location we already tend to store `clio`s config on our test/prod servers?

legleux: @godexsoft 
If this is the default place clio is going to look for a config, it should be `/etc/opt/clio/config.json` which is a link to `/opt/clio/etc/config.json`.
I can change the filename to whatever we want.

godexsoft: Ok, changing it.

cjcobb23 (approved):
LGTM

Issue #250, opened 2022-08-15
Add `start_sequence` to the example config


Mwni:
The `start_sequence` parameter currently is not documented. It should at least be included in the example configuration files. But then the issue is, which example value for the sequence should be picked?

cjcobb23:
Good question. I see 3 options here:

1. Allow start_sequence to be `-1`, in which case its ignored
2. Figure out how to add comments to json
3. Document all of the config options in the README somewhere.

What do you think?

Mwni:
Since options 1 and 2 are hacks: option 3. I would welcome documentation for the config in general.
Is a .toml based config feasible?

cjcobb23:
I agree that 3 is best. .toml is probably feasible, haven't looked into it.

PR #438, opened 2022-12-09
Update documentation and config with start_sequence 

officialfrancismendoza:
**Issue (#250)**: No documentation regarding start_sequence in Clio config file
**Fix**: Update documentation regarding optional start_sequence parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,23 @@ server is running
 to the IP of your Clio server. This entry can take the form of a comma-separated list if
 you are running multiple Clio nodes.
 
-In addition, the parameters `ssl_cert_file` and `ssl_key_file` can also be added to the top level of precedence of our Clio config. `ssl_cert_file` specifies the filepath for your SSL cert while `ssl_key_file` specifies the filepath for your SSL key. It is up to you how to change ownership of these folders for your designated Clio user. Your options include:
+
+In addition, the parameter `start_sequence` can be included and configured within the top level of the config file. This parameter specifies the sequence of first ledger to extract if the database is empty. Note that ETL extracts ledgers in order and that no backfilling functionality currently exists, meaning Clio will not retroactively learn ledgers older than the one you specify. Choosing to specify this or not will yield the following behavior:
+- If this setting is absent and the database is empty, ETL will start with the next ledger validated by the network. 
+- If this setting is present and the database is not empty, an exception is thrown.
+
+In addition, the optional parameter `finish_sequence` can be added to the json file as well, specifying where the ledger can stop.
+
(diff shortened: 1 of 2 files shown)


godexsoft (commented):
This text looks good.

I think we should also add it to `example-config.json` and maybe even add a brief comment explaining the same thing in the example config.

In `clio` configs, you can use `//` or `/* */` style comments in JSON.

officialfrancismendoza:
@godexsoft left as additional comments under README.md since it's not possible to have pure comments in JSON files (data only). Could do a representation as a .toml file instead, but figured putting them in the documentation is a cleaner solution

README.md
```diff
@@ -72,6 +72,21 @@ server is running
...
+- If this setting is present and the database is not empty, an exception is thrown.
+
+In addition, the optional parameter `finish_sequence` can be added to the json file as well, specifying where the ledger can stop.
+
+To add `start_sequence` and/or `finish_sequence` to the config.json file appropriately, they will be on the same top level of precedence as other parameters (such as database, etl_sources, read_only, etc.) and be specified with an integer. Here is an example below, added at the end of the config.json:
+
+```sh
+    ...
+    "read_only": false,
+    // "start_sequence": [integer] the ledger index to start from
```
godexsoft: In this case, when you are adding it to the readme like this, why keep the `//` for comments here?

README.md
```diff
@@ -72,6 +72,21 @@ server is running
...
 
+In addition, the parameter `start_sequence` can be included and configured within the top level of the config file. This parameter specifies the sequence of first ledger to extract if the database is empty. Note that ETL extracts ledgers in order and that no backfilling functionality currently exists, meaning Clio will not retroactively learn ledgers older than the one you specify. Choosing to specify this or not will yield the following behavior:
+- If this setting is absent and the database is empty, ETL will start with the next ledger validated by the network. 
+- If this setting is present and the database is not empty, an exception is thrown.
+
+In addition, the optional parameter `finish_sequence` can be added to the json file as well, specifying where the ledger can stop.
+
+To add `start_sequence` and/or `finish_sequence` to the config.json file appropriately, they will be on the same top level of precedence as other parameters (such as database, etl_sources, read_only, etc.) and be specified with an integer. Here is an example below, added at the end of the config.json:
+
+```sh
```
godexsoft: Is there no `json` variant for syntax highlighting?

godexsoft (changes requested):
Leaving a few questions. 
Also, while the JSON spec does not allow comments, clio's parser of JSON config certainly allows comments. So you can add the comments to the example config too (as comments).

README.md
officialfrancismendoza: @godexsoft not that I'm aware of for plaintext, but there is for code if encapsulated in ```sh ```to give the highlighting you would find on a CLI
image

godexsoft: Please try both `json` and then `javascript` and see if anything looks like actual json and not `shell` after syntax highlighting on github.

godexsoft (approved):

PR #441, opened 2022-12-09
Fix gateway balances to match rippled output

### LINKED ISSUE #271: “gateway_balances” with invalid account in hot wallet do not return error.

godexsoft:
Fixes #271 

Looks like `gateway_balances` had a few mistakes. 
@cjcobb23 please carefully check the logic changes with regards to negating the output. 
As it is right now it looks like it matches the output rippled produces for the same addresses. Before the change it was showing flipped negated amounts.

### DIFF
--- a/src/rpc/handlers/GatewayBalances.cpp
+++ b/src/rpc/handlers/GatewayBalances.cpp
@@ -25,7 +25,7 @@ doGatewayBalances(Context const& context)
     std::map<ripple::AccountID, std::vector<ripple::STAmount>> frozenBalances;
     std::set<ripple::AccountID> hotWallets;
 
-    if (request.contains("hot_wallet"))
+    if (request.contains(JS(hotwallet)))
     {
         auto getAccountID =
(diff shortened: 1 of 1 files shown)


godexsoft:
@cjcobb23 @legleux this implementation seems to produce the same output as `rippled` does. At least for successful response with balances. Before the change it would output negated number compared to `rippled`s output. 
Let's carefully test this with integration tests and compare to `rippled` in all possible ways.

cjcobb23 (approved):
LGTM

PR #449, opened 2022-12-13
Increase file descriptor limit

legleux:
System ran out of file descriptors. 
Default limit is 1024 and can be hit with many connections open.


godexsoft (approved):
Beautiful :)

Issue #435, opened 2022-12-06
Add extra compilation options for clio


godexsoft:
Current set of options and warnings when building `clio` is quite narrow (see `CMake/settings.cmake`).
Let's implement more safety nets. 

Add at least these recommended/standard options: 
```
-Wextra 
-Wshadow
-Wnon-virtual-dtor
-pedantic
```

This will break compilation of `clio` which will need to be fixed for all supported compilers.

godexsoft:
Let's do this in the future. Not now. Codebase is not ready yet.

PR #450, opened 2022-12-13
Add extra compilation options

officialfrancismendoza:
**Issue** (#435): No additional compilation options for Clio are supported
**Fix**: Configure Clio to support additional compilation options

### DIFF
--- a/src/backend/BackendInterface.cpp
+++ b/src/backend/BackendInterface.cpp
@@ -173,7 +173,6 @@ BackendInterface::fetchBookOffers(
     ripple::uint256 const& book,
     std::uint32_t const ledgerSequence,
     std::uint32_t const limit,
-    std::optional<ripple::uint256> const& cursor,
     boost::asio::yield_context& yield) const
 {
     // TODO try to speed this up. This can take a few seconds. The goal is
--- a/src/backend/BackendInterface.h
+++ b/src/backend/BackendInterface.h
@@ -23,7 +23,6 @@
 #include <backend/DBHelpers.h>
 #include <backend/SimpleCache.h>
 #include <backend/Types.h>
-#include <config/Config.h>
 #include <log/Logger.h>
 
 #include <boost/asio/spawn.hpp>
@@ -182,7 +181,7 @@ protected:
      */
 
 public:
-    BackendInterface(clio::Config const& config)
+    BackendInterface()
     {
     }
     virtual ~BackendInterface()
@@ -479,7 +478,6 @@ public:
         ripple::uint256 const& book,
         std::uint32_t const ledgerSequence,
         std::uint32_t const limit,
-        std::optional<ripple::uint256> const& cursor,
         boost::asio::yield_context& yield) const;
 
     /**
--- a/src/backend/CassandraBackend.cpp
+++ b/src/backend/CassandraBackend.cpp
@@ -59,10 +59,11 @@ processAsyncWriteResponse(T& requestParams, CassFuture* fut, F func)
             std::make_shared<boost::asio::steady_timer>(
                 backend.getIOContext(),
                 std::chrono::steady_clock::now() + wait);
-        timer->async_wait([timer, &requestParams, func](
-                              const boost::system::error_code& error) {
-            func(requestParams, true);
-        });
+        timer->async_wait(
+            [timer, &requestParams, func](
+                [[maybe_unused]] const boost::system::error_code& error) {
+                func(requestParams, true);
+            });
     }
     else
     {
@@ -210,11 +211,11 @@ CassandraBackend::doWriteLedgerObject(
             this,
             std::make_tuple(seq, key),
             [this](auto& params) {
-                auto& [sequence, key] = params.data;
+                auto& [sequence, k] = params.data;
 
                 CassandraStatement statement{insertDiff_};
                 statement.bindNextInt(sequence);
-                statement.bindNextBytes(key);
+                statement.bindNextBytes(k);
                 return statement;
             },
             "ledger_diff");
@@ -222,12 +223,12 @@ CassandraBackend::doWriteLedgerObject(
         this,
         std::make_tuple(std::move(key), seq, std::move(blob)),
         [this](auto& params) {
-            auto& [key, sequence, blob] = params.data;
+            auto& [k2, sequence, b] = params.data;
 
             CassandraStatement statement{insertObject_};
-            statement.bindNextBytes(key);
+            statement.bindNextBytes(k2);
             statement.bindNextInt(sequence);
-            statement.bindNextBytes(blob);
+            statement.bindNextBytes(b);
             return statement;
         },
         "ledger_object");
@@ -236,24 +237,24 @@ CassandraBackend::doWriteLedgerObject(
 void
 CassandraBackend::writeSuccessor(
     std::string&& key,
-    std::uint32_t const seq,
+    std::uint32_t const sequence,
     std::string&& successor)
 {
     log_.trace() << "Writing successor. key = " << key.size() << " bytes. "
-                 << " seq = " << std::to_string(seq)
+                 << " seq = " << std::to_string(sequence)
                  << " successor = " << successor.size() << " bytes.";
     assert(key.size() != 0);
     assert(successor.size() != 0);
     makeAndExecuteAsyncWrite(
         this,
-        std::make_tuple(std::move(key), seq, std::move(successor)),
+        std::make_tuple(std::move(key), sequence, std::move(successor)),
         [this](auto& params) {
-            auto& [key, sequence, successor] = params.data;
+            auto& [k, seq, suc] = params.data;
 
             CassandraStatement statement{insertSuccessor_};
-            statement.bindNextBytes(key);
-            statement.bindNextInt(sequence);
-            statement.bindNextBytes(successor);
+            statement.bindNextBytes(k);
+            statement.bindNextInt(seq);
+            statement.bindNextBytes(suc);
             return statement;
         },
         "successor");
@@ -267,10 +268,10 @@ CassandraBackend::writeLedger(
         this,
         std::make_tuple(ledgerInfo.seq, std::move(header)),
         [this](auto& params) {
-            auto& [sequence, header] = params.data;
+            auto& [sec, hdr] = params.data;
             CassandraStatement statement{insertLedgerHeader_};
-            statement.bindNextInt(sequence);
-            statement.bindNextBytes(header);
+            statement.bindNextInt(sec);
+            statement.bindNextBytes(hdr);
             return statement;
         },
         "ledger");
@@ -305,8 +306,8 @@ CassandraBackend::writeAccountTransactions(
                     record.txHash),
                 [this](auto& params) {
                     CassandraStatement statement(insertAccountTx_);
-                    auto& [account, lgrSeq, txnIdx, hash] = params.data;
-                    statement.bindNextBytes(account);
+                    auto& [acc, lgrSeq, txnIdx, hash] = params.data;
+                    statement.bindNextBytes(acc);
                     statement.bindNextIntTuple(lgrSeq, txnIdx);
                     statement.bindNextBytes(hash);
                     return statement;
@@ -343,7 +344,7 @@ CassandraBackend::writeNFTTransactions(std::vector<NFTTransactionsData>&& data)
 void
 CassandraBackend::writeTransaction(
     std::string&& hash,
-    std::uint32_t const seq,
+    std::uint32_t const sequence,
     std::uint32_t const date,
     std::string&& transaction,
     std::string&& metadata)
@@ -353,7 +354,7 @@ CassandraBackend::writeTransaction(
 
     makeAndExecuteAsyncWrite(
         this,
-        std::make_pair(seq, hash),
+        std::make_pair(sequence, hash),
         [this](auto& params) {
             CassandraStatement statement{insertLedgerTransaction_};
             statement.bindNextInt(params.data.first);
@@ -365,18 +366,18 @@ CassandraBackend::writeTransaction(
         this,
         std::make_tuple(
             std::move(hash),
-            seq,
+            sequence,
             date,
             std::move(transaction),
             std::move(metadata)),
         [this](auto& params) {
             CassandraStatement statement{insertTransaction_};
-            auto& [hash, sequence, date, transaction, metadata] = params.data;
-            statement.bindNextBytes(hash);
-            statement.bindNextInt(sequence);
-            statement.bindNextInt(date);
-            statement.bindNextBytes(transaction);
-            statement.bindNextBytes(metadata);
+            auto& [hsh, seq, d, trans, meta] = params.data;
+            statement.bindNextBytes(hsh);
+            statement.bindNextInt(seq);
+            statement.bindNextInt(d);
+            statement.bindNextBytes(trans);
+            statement.bindNextBytes(meta);
             return statement;
         },
         "transaction");
@@ -431,17 +432,17 @@ CassandraBackend::hardFetchLedgerRange(boost::asio::yield_context& yield) const
         log_.error() << "No rows";
         return {};
     }
-    LedgerRange range;
-    range.maxSequence = range.minSequence = result.getUInt32();
+    LedgerRange rng;
+    rng.maxSequence = rng.minSequence = result.getUInt32();
     if (result.nextRow())
     {
-        range.maxSequence = result.getUInt32();
+        rng.maxSequence = result.getUInt32();
     }
-    if (range.minSequence > range.maxSequence)
+    if (rng.minSequence > rng.maxSequence)
     {
-        std::swap(range.minSequence, range.maxSequence);
+        std::swap(rng.minSequence, rng.maxSequence);
     }
-    return range;
+    return rng;
 }
 
 std::vector<TransactionAndMetadata>
@@ -464,10 +465,10 @@ struct ReadCallbackData
 
     std::atomic_bool errored = false;
     ReadCallbackData(
-        std::atomic_int& numOutstanding,
-        handler_type& handler,
-        std::function<void(CassandraResult&)> onSuccess)
-        : numOutstanding(numOutstanding), handler(handler), onSuccess(onSuccess)
+        std::atomic_int& numOutstan,
+        handler_type& hndlr,
+        std::function<void(CassandraResult&)> onScs)
+        : numOutstanding(numOutstan), handler(hndlr), onSuccess(onScs)
     {
     }
 
@@ -525,6 +526,7 @@ CassandraBackend::fetchTransactions(
     std::vector<TransactionAndMetadata> results{numHashes};
     std::vector<std::shared_ptr<ReadCallbackData<result_type>>> cbs;
     cbs.reserve(numHashes);
+
     auto timeDiff = util::timed([&]() {
         for (std::size_t i = 0; i < hashes.size(); ++i)
         {
@@ -532,13 +534,13 @@ CassandraBackend::fetchTransactions(
             statement.bindNextBytes(hashes[i]);
 
             cbs.push_back(std::make_shared<ReadCallbackData<result_type>>(
-                numOutstanding, handler, [i, &results](auto& result) {
-                    if (result.hasResult())
+                numOutstanding, handler, [i, &results](auto& rslt2) {
+                    if (rslt2.hasResult())
                         results[i] = {
-                            result.getBytes(),
-                            result.getBytes(),
-                            result.getUInt32(),
-                            result.getUInt32()};
+                            rslt2.getBytes(),
+                            rslt2.getBytes(),
+                            rslt2.getUInt32(),
+                            rslt2.getUInt32()};
                 }));
 
             executeAsyncRead(statement, processAsyncRead, *cbs[i]);
@@ -849,9 +851,9 @@ CassandraBackend::doFetchLedgerObjects(
     for (std::size_t i = 0; i < keys.size(); ++i)
     {
         cbs.push_back(std::make_shared<ReadCallbackData<result_type>>(
-            numOutstanding, handler, [i, &results](auto& result) {
-                if (result.hasResult())
-                    results[i] = result.getBytes();
+            numOutstanding, handler, [i, &results](auto& rslt2) {
+                if (rslt2.hasResult())
+                    results[i] = rslt2.getBytes();
             }));
         CassandraStatement statement{selectObject_};
         statement.bindNextBytes(keys[i]);
@@ -995,7 +997,7 @@ CassandraBackend::isTooBusy() const
 }
 
 void
-CassandraBackend::open(bool readOnly)
+CassandraBackend::open([[maybe_unused]] bool readOnly)
 {
     if (open_)
     {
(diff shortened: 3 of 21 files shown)


src/backend/CassandraBackend.cpp
```diff
@@ -209,24 +209,24 @@ CassandraBackend::doWriteLedgerObject(
             this,
             std::make_tuple(seq, key),
             [this](auto& params) {
-                auto& [sequence, key] = params.data;
+                auto& [sequence, _key] = params.data;
```
godexsoft: Let's not use `_` suffix and/or prefix for this purpose as `_` usually expresses a special group of variables, like member variables in a class/struct.

I suggest we use the following scheme instead.
For every shadowed variable, rename the inner/later one to a shortened name:
`session` -> `ses`, `key` -> `k`, `blob` -> `b`, etc.

If the variables both were one character (e.g. `k`) let's do the opposite and change the shadowed/top variable with the longer name that makes sense for it (e.g. `key`).

src/backend/CassandraBackend.cpp
```diff
@@ -247,12 +247,12 @@ CassandraBackend::writeSuccessor(
         this,
         std::make_tuple(std::move(key), seq, std::move(successor)),
         [this](auto& params) {
-            auto& [key, sequence, successor] = params.data;
+            auto& [_key, sequence, _successor] = params.data;
```
godexsoft: Let's call these `k`, `seq` and `suc`. Changing all three to be consistent.

src/backend/CassandraBackend.cpp
```diff
@@ -266,10 +266,10 @@ CassandraBackend::writeLedger(
         this,
         std::make_tuple(ledgerInfo.seq, std::move(header)),
         [this](auto& params) {
-            auto& [sequence, header] = params.data;
+            auto& [sequence, _header] = params.data;
```
godexsoft: `sec`, `hdr`

src/backend/CassandraBackend.cpp
```diff
@@ -304,8 +304,8 @@ CassandraBackend::writeAccountTransactions(
                     record.txHash),
                 [this](auto& params) {
                     CassandraStatement statement(insertAccountTx_);
-                    auto& [account, lgrSeq, txnIdx, hash] = params.data;
-                    statement.bindNextBytes(account);
+                    auto& [_account, lgrSeq, txnIdx, hash] = params.data;
```
godexsoft: `acc`

src/backend/CassandraBackend.cpp
```diff
@@ -370,12 +370,12 @@ CassandraBackend::writeTransaction(
             std::move(metadata)),
         [this](auto& params) {
             CassandraStatement statement{insertTransaction_};
-            auto& [hash, sequence, date, transaction, metadata] = params.data;
-            statement.bindNextBytes(hash);
+            auto& [_hash, sequence, _date, _transaction, _metadata] = params.data;
```
godexsoft: `hsh`, `seq`, `d`, `trans`, `meta` or something like that I guess.

src/backend/CassandraBackend.cpp
```diff
@@ -1182,14 +1182,14 @@ CassandraBackend::open(bool readOnly)
     auto executeSimpleStatement = [this](std::string const& query) {
         CassStatement* statement = makeStatement(query.c_str(), 0);
         CassFuture* fut = cass_session_execute(session_.get(), statement);
-        CassError rc = cass_future_error_code(fut);
+        CassError _rc = cass_future_error_code(fut);
```
godexsoft: this could be renamed to say `err` or `ec`

src/backend/CassandraBackend.cpp
```diff
@@ -1221,10 +1221,10 @@ CassandraBackend::open(bool readOnly)
             cass_future_free(fut);
             if (rc != CASS_OK)
             {
-                std::stringstream ss;
-                ss << "nodestore: Error connecting Cassandra session at all: "
+                std::stringstream _ss;
```
godexsoft: stringstream should probably be removed entirely later but for now, how about we reuse the other `ss`? we can just `ss.str("")` here or even assign `ss = std::stringstream{};` and that should do.

src/rpc/RPCHelpers.cpp
```diff
@@ -720,8 +720,8 @@ traverseOwnedNodes(
                 return Status(
                     ripple::rpcINVALID_PARAMS, "Owner directory not found");
 
-            ripple::SerialIter it{ownerDir->data(), ownerDir->size()};
-            ripple::SLE sle{it, currentIndex.key};
+            ripple::SerialIter _it{ownerDir->data(), ownerDir->size()};
+            ripple::SLE _sle{it, currentIndex.key};
 
             for (auto const& key : sle.getFieldV256(ripple::sfIndexes))
```
godexsoft: You need to be very very careful. This looks like it's now using the wrong `sle`... and you really gotta check all the code you changed in this PR so that it's not suddenly using the wrong variable. This is exactly one of the dangers of shadowing and why we try to fix this in the first place

src/rpc/handlers/Subscribe.cpp
```diff
@@ -260,14 +260,14 @@ validateAndGetBooks(
...
             auto getOrderBook = [&snapshot, &backend, &rng, &takerID](
-                                    auto book,
-                                    boost::asio::yield_context& yield) {
-                auto bookBase = getBookBase(book);
+                                    auto _book,
+                                    boost::asio::yield_context& _yield) {
+                auto bookBase = getBookBase(_book);
                 auto [offers, retMarker] = backend->fetchBookOffers(
-                    bookBase, rng->maxSequence, 200, {}, yield);
+                    bookBase, rng->maxSequence, 200, {}, _yield);
```
godexsoft: i guess `yld` could work

src/webserver/WsBase.h
```diff
@@ -440,10 +440,10 @@ class WsSession : public WsBase,
             send(std::move(responseStr));
         };
 
-        boost::json::value raw = [](std::string const&& msg) {
+        boost::json::value raw = [](std::string const&& _msg) {
```
godexsoft: Let's call the outer one `message` and this one `msg`

godexsoft (changes requested):
It's a good start. Let's not use `_` for this though. See my other comments inline :+1:

officialfrancismendoza:
@godexsoft this push is the most recent batch for the explicit _ prepends mentioned in your comments, will modify the rest in a separate push (have to go through the commit history). In the same future commit, will add prepends to eliminate all warning messages regarding unused params

Update: Made the latest push eliminating unused param errors for all non-build files

src/backend/BackendInterface.cpp
```diff
@@ -173,7 +173,7 @@ BackendInterface::fetchBookOffers(
     ripple::uint256 const& book,
     std::uint32_t const ledgerSequence,
     std::uint32_t const limit,
-    std::optional<ripple::uint256> const& cursor,
+    [[maybe_unused]] std::optional<ripple::uint256> const& cursor,
```
cindyyan317: why not just remove this parameter . it is not used anyway.

src/backend/BackendInterface.h
```diff
@@ -182,7 +182,7 @@ class BackendInterface
      */
 
 public:
-    BackendInterface(clio::Config const& config)
+    BackendInterface([[maybe_unused]] clio::Config const& config)
```
cindyyan317: remove

src/etl/ETLSource.cpp
```diff
@@ -462,7 +464,7 @@ ETLSourceImpl<Derived>::onRead(boost::beast::error_code ec, size_t size)
 
         log_.trace() << "calling async_read - " << toString();
         derived().ws().async_read(
-            readBuffer_, [this](auto ec, size_t size) { onRead(ec, size); });
+            readBuffer_, [this](auto e, size_t _size) { onRead(e, _size); });
```
cindyyan317: since the parameter size is not used, 
we can 
`ETLSourceImpl<Derived>::onRead(
    boost::beast::error_code ec,
    [[maybe_unused]] size_t _)`

then no need to amend lambda

src/etl/ReportingETL.cpp
```diff
@@ -826,7 +826,7 @@ ReportingETL::monitor()
                  << "Starting monitor loop. sequence = " << nextSequence;
     while (true)
     {
-        if (auto rng = backend_->hardFetchLedgerRangeNoThrow();
+        if (auto rg = backend_->hardFetchLedgerRangeNoThrow();
```
cindyyan317, line 829: Alert: define rg , but not use it? it may introduce an issue, are we using the previous rng?

src/rpc/handlers/GatewayBalances.cpp
```diff
@@ -171,9 +171,9 @@ doGatewayBalances(Context const& context)
     if (!sums.empty())
     {
         boost::json::object obj;
-        for (auto const& [k, v] : sums)
+        for (auto const& [k, v1] : sums)
```
cindyyan317: instead of here, i feel we can make the original v more meaningful?
see line 35. 
`    auto v = ledgerInfoFromRequest(context);`

src/rpc/handlers/Ledger.cpp
```diff
@@ -153,9 +153,8 @@ doLedger(Context const& context)
     {
         header["diff"] = boost::json::value(boost::json::array_kind);
         boost::json::array& jsonDiff = header.at("diff").as_array();
-        auto diff =
-            context.backend->fetchLedgerDiff(lgrInfo.seq, context.yield);
-        for (auto const& obj : diff)
+        auto dff = context.backend->fetchLedgerDiff(lgrInfo.seq, context.yield);
```
cindyyan317: maybe diffLedgerObject more accurate here

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -245,8 +243,8 @@ SubscriptionManager::subBook(
     std::shared_ptr<WsBase> session)
 {
     subscribeHelper(
-        session, book, bookSubscribers_, [this, book](session_ptr session) {
-            unsubBook(book, session);
+        session, book, bookSubscribers_, [this, book](session_ptr sesh) {
```
cindyyan317, line 246: wonder after u remove the lambda parameter , the function parameter session is not used. does it throw warning?

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -174,8 +174,8 @@ SubscriptionManager::subLedger(
     boost::asio::yield_context& yield,
     std::shared_ptr<WsBase> session)
 {
-    subscribeHelper(session, ledgerSubscribers_, [this](session_ptr session) {
-        unsubLedger(session);
+    subscribeHelper(session, ledgerSubscribers_, [this](session_ptr sesh) {
```
cindyyan317, line 177: if session at 175 not used , can we do 
`session_ptr _)`
or just comment out

src/webserver/WsBase.h
```diff
@@ -374,8 +374,8 @@ class WsSession : public WsBase,
                 // if the result is forwarded - just use it as is
                 // but keep all default fields in the response too.
                 if (isForwarded)
-                    for (auto const& [k, v] : result)
-                        response.insert_or_assign(k, v);
+                    for (auto const& [k, v2] : result)
```
cindyyan317: The original v can be more meaningful
`auto [response, timeDiff] =
                util::timed([&]() { return RPC::buildResponse(*context); });`

src/etl/ReportingETL.cpp
```diff
@@ -908,7 +908,7 @@ ReportingETL::loadCacheFromClioPeer(
         std::optional<boost::json::value> marker;
 
         log_.trace() << "Sending request";
-        auto getRequest = [&](auto marker) {
+        auto getRequest = [&]([[maybe_unused]] auto _marker) {
```
cindyyan317: just use _ 
or just comment the parameter out 
like:`    auto f = [](auto /**/){
        std::cout<<456<<std::endl;
    };
    f(2);`

cindyyan317 (commented):
Nice to see we are fixing so many bombs

src/etl/ETLSource.cpp
godexsoft: or maybe even `ETLSourceImpl<Derived>::onRead( boost::beast::error_code ec, size_t)`

src/backend/BackendInterface.cpp
officialfrancismendoza: Removal of cursor causes the following error localized to BackendInterface.cpp upon rebuild
image

src/backend/BackendInterface.h
officialfrancismendoza: This parameter is referenced within CassandraBackend.h, where it is fed in as input for config_. config_ faces same issue, but could unknowingly create further bugs in the rest of the codebase

image
image

officialfrancismendoza: Update: Deletion did in fact cause a bug chain to appear. config_ from CassandraBackend.h was necessary in CassandraBackend.cpp. Reverting for now. 
image

src/subscriptions/SubscriptionManager.cpp
officialfrancismendoza: @cindyyan317 It will throw compilation error unless I fundamentally modfiy subscribeHelper in SubscriptionManager.h, as it does not match the format.
image

Upon modifying subscribeHelper to get rid of this, however, introduced bug chains as this parameter was utilized elsewhere within SubscriptionManager.cpp. As a general rule, I think it is better to keep these parameters, despite being unused, so that bug chains are not unknowingly created in the future + these parameters can be utilized in the future need be 
image

src/subscriptions/SubscriptionManager.cpp
officialfrancismendoza: See prior comment above regarding removing session and assorted consequences. Regarding session_ptr sesh, it is a necessary parameter for subscription and unsubscription methods within SubscriptionManager.cpp

src/etl/ETLSource.cpp
officialfrancismendoza: Got primary expression errors when attempting to implement either of your suggestions, so will investigate further into this today

image

officialfrancismendoza: Attempting either fix leaves invalid conversion and expected primary expressions within parameters due to explicit type definition. Was able to make a partial modification with specifying only ETLSourceImpl<Derived>, but could not modify params for onRead so far

image

Experimented with a few variations: just size_t, size_t _, and _, but all gave compilation errors. 
image
image
image

src/backend/BackendInterface.h
godexsoft: > This parameter is referenced within CassandraBackend.h, where it is fed in as input for config_. config_ faces same issue, but could unknowingly create further bugs in the rest of the codebase
> 
> image image

All you gotta do, is remove the config parameter from `BackendInterface.h`, along with the `#include` for config.
You could actually completely remove the constructor of `BackendInterface`, because it will become default once config is removed.

The config in `CassandreBackend` is of course there to stay. What you will have to do however is remove the call to `BackendInterface` constructor now that it's a `default constructor` again:
```
 CassandraBackend(boost::asio::io_context& ioc, util::Config const& config, uint32_t ttl)
        : config_(config)
        , ttl_(ttl)
    {
        work_.emplace(ioContext_);
        ioThread_ = std::thread([this]() { ioContext_.run(); });
    }
```

src/backend/BackendInterface.cpp
godexsoft: Huh? you just need to remove it from both `BackendInterface.h` and `BackendInterface.cpp`, then find where it is used (literally 2 places) and remove the 4th parameter that's passed to the member function (marker).

src/etl/ETLSource.cpp
godexsoft: It's because of two things:
1) you seem to have removed the type of the first argument
2) `_` seems to be a special keyword in on of the boost libs included. should not be a problem if you make the code valid

What i mean is 
```
template <class Derived>
void
ETLSourceImpl<Derived>::onRead(boost::beast::error_code ec, size_t)
```

Don't forget that you of course need to fix the same thing in the header:
```
    /// Callback
    void
    onRead(boost::beast::error_code ec, size_t);
```

You gotta distinguish between calling functions and defining/declaring them.

src/etl/ReportingETL.cpp
godexsoft: if we had uses of `rng` after this line (in the scope of `if`) then it should now be `rg` everywhere in there. 
You seem to have made a bug. You are getting the new `rg` but checking the old one. I fear that there is going to be many of those in other fixes you did - please go check all your changes super thoroughly, otherwise we might end up with multiple bugs added to the codebase.

```
if (auto rg = backend_->hardFetchLedgerRangeNoThrow(); rg && rg->maxSequence >= nextSequence) {
    publishLedger(nextSequence, {});
    ++nextSequence;
}
```

godexsoft:
> Nice to see we are fixing so many bombs

Sadly we don't have unit-tests to find all the bugs that are likely introduced by the changes..
I think there are going to be many places where only the variable was renamed but all the uses of that variable stayed with the old name which of course means we now use an entirely different variable for the logic.

src/backend/BackendInterface.cpp
cindyyan317: Remove the declaration and definition and usages.

src/etl/ReportingETL.cpp
officialfrancismendoza: Functionally it makes no difference as to whether rng (by removing auto to reference the variable that already exists within method scope) or rg (shadowed param solution) is used because rng in the original code is not referenced further in the while loop aside the if condition

The functional difference of the while loop is to monitor the ledger sequence. The difference is that with rg, you create a new copy for the sequence data whereas with rng you're overwriting the old variable used to store it, which wasn't utilized in there anyway

src/rpc/Errors.h
```diff
@@ -74,25 +74,24 @@ struct Status
     std::optional<boost::json::object> extraInfo;
 
     Status() = default;
-    /* implicit */ Status(CombinedError code) : code(code){};
-    Status(CombinedError code, boost::json::object&& extraInfo)
-        : code(code), extraInfo(std::move(extraInfo)){};
+    /* implicit */ Status(CombinedError cde) : code(cde){};
+    Status(CombinedError cde, boost::json::object&& _extraInfo)
+        : code(cde), extraInfo(std::move(_extraInfo)){};
```
cindyyan317: miss this _?

cindyyan317:
LGTM. Most of comments have been addressed. Need to remove some  local variable underscore

cindyyan317 (dismissed):
The changes look good to me. Alex had run integration tests against this change, let us double confirm with him before submitting.

src/backend/CassandraBackend.h
```diff
@@ -703,10 +703,10 @@ class CassandraBackend : public BackendInterface
 
 public:
     CassandraBackend(
-        boost::asio::io_context& ioc,
+        [[maybe_unused]] boost::asio::io_context& ioc,
```
godexsoft, line 706: This makes no sense.. we are not using `ioc` at all for some reason. We don't even need to pass it at all in this case. Let's keep it as is for now but also keep this in mind for future - we should either remove it or use it.

src/backend/DBHelpers.h
```diff
@@ -41,12 +41,12 @@ struct AccountTransactionsData
 
     AccountTransactionsData(
         ripple::TxMeta& meta,
-        ripple::uint256 const& txHash,
-        beast::Journal& j)
+        ripple::uint256 const& txHsh,
+        [[maybe_unused]] beast::Journal& j)
```
godexsoft, line 45: This should have been removed instead. Let's keep for now and remove later for the sake of finishing with this PR.

src/rpc/RPCHelpers.cpp
```diff
@@ -103,10 +103,10 @@ getRequiredUInt(boost::json::object const& request, std::string const& field)
 
 std::optional<AccountCursor>
 parseAccountCursor(
-    BackendInterface const& backend,
-    std::uint32_t seq,
+    [[maybe_unused]] BackendInterface const& backend,
```
godexsoft, line 106: If we don't use those, why keep them? Let's remove another time. Keep for now for the sake of finishing this PR.

src/rpc/RPCHelpers.cpp
```diff
@@ -699,8 +699,8 @@ traverseOwnedNodes(
         if (!hintDir)
             return Status(ripple::rpcINVALID_PARAMS, "Invalid marker");
 
-        ripple::SerialIter it{hintDir->data(), hintDir->size()};
-        ripple::SLE sle{it, hintIndex.key};
+        ripple::SerialIter itHint{hintDir->data(), hintDir->size()};
+        ripple::SLE sle{itHint, hintIndex.key};
```
godexsoft: I'd rename this to `hintSle` - this would have revealed the problems you introduced below.

src/rpc/RPCHelpers.cpp
```diff
@@ -721,8 +721,8 @@ traverseOwnedNodes(
                 return Status(
                     ripple::rpcINVALID_PARAMS, "Owner directory not found");
 
-            ripple::SerialIter it{ownerDir->data(), ownerDir->size()};
-            ripple::SLE sle{it, currentIndex.key};
+            ripple::SerialIter itOwner{ownerDir->data(), ownerDir->size()};
+            ripple::SLE sleOwner{itOwner, currentIndex.key};
```
godexsoft, line 725: Obvious bug - `sle` not renamed below. And also on line 745.

src/rpc/RPCHelpers.cpp
godexsoft: I have warned you before. Seems like you did not carefully check still?

src/rpc/handlers/Ledger.cpp
```diff
@@ -153,9 +153,9 @@ doLedger(Context const& context)
     {
         header["diff"] = boost::json::value(boost::json::array_kind);
         boost::json::array& jsonDiff = header.at("diff").as_array();
-        auto diff =
+        auto diffLedgerObject =
```
godexsoft: In this context it makes much more sense to have removed `bool diff` to `bool wantDiff` and leave this one as it was before.

src/webserver/WsBase.h
```diff
@@ -344,13 +344,13 @@ class WsSession : public WsBase,
 
             response = getDefaultWsResponse(id);
 
-            auto [v, timeDiff] =
+            auto [additionalResponse, timeDiff] =
```
godexsoft: Probably not the best name for it? I'd say `builtResponse` or `response` and would have renamed the above `response` to `defaultResponse` instead.

src/webserver/WsBase.h
```diff
@@ -449,7 +450,7 @@ class WsSession : public WsBase,
             {
                 return boost::json::value{nullptr};
             }
-        }(std::move(msg));
+        }(std::move(message));
```
godexsoft, line 453: Not your bug but: the lambda is taking `std::string const&&` which obviously can't ever support the `std::move` attempted on this line.

unittests/Backend.cpp
```diff
@@ -2210,92 +2213,93 @@ TEST_F(BackendTest, cacheIntegration)
...
+                    return _lgrInfo;
                 };
-                auto writeLedger = [&](auto lgrInfo, auto objs, auto state) {
+                auto writeLedger = [&](auto _lgrInfo, auto objs, auto state) {
                     backend->startWrites();
 
                     backend->writeLedger(
-                        lgrInfo, std::move(ledgerInfoToBinaryString(lgrInfo)));
+                        _lgrInfo,
+                        std::move(ledgerInfoToBinaryString(_lgrInfo)));
```
godexsoft, line 2239: Not your bug: useless std::move.

godexsoft (changes requested):
There is at least one serious bug in the PR. The rest is good to fix but not critical.

src/rpc/RPCHelpers.cpp
cindyyan317: Interesting, how compiler pass?

godexsoft:
Let's not merge it for now. We can come back to it in the future when the codebase is better prepared (covered with unit-tests etc.) for these changes.

PR #451, opened 2022-12-14
Document dos_guard in example config. Log when client surpasses rate limit

cjcobb23:
I also added a max size for the work queue in the example config.

- - -
This change is [Reviewable](https://reviewable.io/reviews/xrplf/clio/451)


godexsoft (approved):
Looks good!

PR #453, opened 2022-12-15
Unit-tests for slightly modified DOSGuard

### LINKED ISSUE #452: Implement unit-tests for DOSGuard
DOSGuard can be covered with unit-tests without any/much modifications. Let's get it covered.

godexsoft:
Fixes #452 

This is one way to approach unit-testing something that requires `boost::asio::io_context` without trying to mock or proxy it somehow. Instead we just inject a strategy (handler) that will do whatever it needs to and call out to the DOSGuard itself when needed (e.g. to reset some counter via `clear`).

I'm open to other ideas how to test this.
I have also implemented a fixture that wraps and runs a real `boost::asio::io_context` in a `std::jthread`. We can make use of it when we absolutely have no choice but to test with a real context. For the `DOSGuard` however doing that would mean to actually wait for a timer and sync that with the test thread. This obviously introduces longer running tests - not ideal.

I'm now also thinking that we need to test the strategy itself. For this we must call `setup` with a mock/fake somehow so the strategy might also have to become a template.

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -20,57 +20,80 @@
 #pragma once
 
 #include <boost/asio.hpp>
+#include <boost/iterator/transform_iterator.hpp>
+
+#include <chrono>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
 
 #include <config/Config.h>
 
-class DOSGuard
+namespace clio {
+
+class BaseDOSGuard
 {
-    boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
+public:
+    virtual ~BaseDOSGuard() = default;
+
+    virtual void
+    clear() noexcept = 0;
+};
+
+/**
+ * @brief A simple denial of service guard used for rate limiting.
+ *
+ * @tparam SweepHandler Type of the sweep handler
+ */
+template <typename SweepHandler>
+class BasicDOSGuard : public BaseDOSGuard
+{
+    mutable std::mutex mtx_;  // protects ipFetchCount_
     std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
     std::unordered_map<std::string, std::uint32_t> ipConnCount_;
     std::unordered_set<std::string> const whitelist_;
+
     std::uint32_t const maxFetches_;
-    std::uint32_t const sweepInterval_;
     std::uint32_t const maxConnCount_;
     clio::Logger log_{"RPC"};
 
 public:
-    DOSGuard(clio::Config const& config, boost::asio::io_context& ctx)
-        : ctx_{ctx}
-        , whitelist_{getWhitelist(config)}
+    /**
+     * @brief Constructs a new DOS guard.
+     *
+     * @param config Clio config
+     * @param sweepHandler Sweep handler that implements the sweeping behaviour
+     */
+    BasicDOSGuard(clio::Config const& config, SweepHandler& sweepHandler)
+        : whitelist_{getWhitelist(config)}
         , maxFetches_{config.valueOr("dos_guard.max_fetches", 100000000u)}
-        , sweepInterval_{config.valueOr("dos_guard.sweep_interval", 10u)}
         , maxConnCount_{config.valueOr("dos_guard.max_connections", 1u)}
     {
-        createTimer();
+        sweepHandler.setup(this);
     }
 
-    void
-    createTimer()
-    {
-        auto wait = std::chrono::seconds(sweepInterval_);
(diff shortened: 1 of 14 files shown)


cindyyan317:
Hi @godexsoft  I think for DosGuard, we can config the timer duration, waiting can be short. So I'd like to see if it is possible to run real io_context. Could you please elaborate the strategy u mentioned here? Love to learn more

src/webserver/DOSGuard.h
```diff
@@ -20,56 +20,71 @@
...
+ * @brief A simple denial of service guard used for rate limiting.
+ *
+ * @tparam SweepHandler Type of the sweep handler
+ */
+template <typename SweepHandler>
+class BasicDOSGuard
 {
-    boost::asio::io_context& ctx_;
-    std::mutex mtx_;  // protects ipFetchCount_
+    std::reference_wrapper<SweepHandler> sweepHandler_;
```
cjcobb23: Why does this need to be a reference?

godexsoft: What else would you implement it as? 
I'm open for suggestions!

My current understanding:
1) it can't be unset, we need some strategy for sweeping so a pointer is not a very good option
2) it can't be an object copied/moved in because the whole point of it is to be able to change it from outside of DOSGuard itself (for testing, etc.)
3) it can be a `SweepHandler &` instead but we already had that discussion a few months ago :D

cjcobb23: I was thinking copied/moved. Why exactly do we need to change it outside of the DOSGuard once we construct the DOSGuard? I thought the point was to just be able to pass in different kinds of sweepers, but we can do that at construction right? Maybe I'm missing something.

godexsoft: How can we then test it from a unit test without going thru a real timer in a real `io_context` on a real thread?
The whole point is to be able to pretend something happened (by faking/forcing a behaviour) to see how the DOSGuard reacts (e.g. we can fake a timer calling out to `clear` by avoiding using a timer and just call the `clear` directly from a fake sweep handler). 

Now the real question is how do we test `IntervalSweepHandler` itself. We can of course set it up on a real `io_context` and then we just wait inside the unit test until timer fires and the relevant data is cleared which we can validate by checking `isOk(IP)`. Since here we are not testing the `DOSGuard` anymore but testing only the `IntervalSweepHandler` it seems strange to use a read `DOSGuard` instead of a mock. 

Ideally in such a test we create the **real** `SweepHandler` but set it up with a mocked `DOSGuard`. Then, as the timer fires, the real `SweepHandler` will call `guardMock->clear()` inside and we can check it by just monitoring that `clear` was called. In terms of a mock it will look something like this:

```cpp
template <typename SweepHandler>
struct BasicDOSGuardMock
{
    BasicDOSGuardMock(SweepHandler& handler)
    {
        handler.setup(this); // real sweep handler is given a mocked DOSGuard
    }

   // generates a `clear` member function that we can set expectations on
    MOCK_METHOD(void, clear, (), (noexcept));
};

class IntervalSweepHandlerTest : public AsioContextTest
{
protected:
    Config cfg{json::parse(JSONData)}; // real config and real sweep handler
    IntervalSweepHandler sweepHandler{cfg, ctx}; // ctx is a real io_context, comes from AsioContextTest
    BasicDOSGuardMock<IntervalSweepHandler> guard{sweepHandler}; // mocked guard given real sweep handler
};

TEST_F(IntervalSweepHandlerTest, SweepAfterInterval)
{
    EXPECT_CALL(guard, clear()).Times(AtLeast(1)); // we expect a call to `clear`
    std::this_thread::sleep_for(std::chrono::milliseconds(1500)); // wait for the call 
} // expectation should be fulfilled here
```

cindyyan317 (dismissed):
LGTM

src/webserver/Listener.h
```diff
@@ -51,7 +51,7 @@ class Detector
     std::shared_ptr<ETLLoadBalancer> balancer_;
     std::shared_ptr<ReportingETL const> etl_;
     util::TagDecoratorFactory const& tagFactory_;
-    DOSGuard& dosGuard_;
+    clio::DOSGuard& dosGuard_;
```
cindyyan317, line 54: it is kinda weird , we need to use clio:: to use something in the same folder. If I understand correctly, you also plan to put all these class under clio:: , right?

godexsoft: We can reconsider this.. i don't mind having to type `clio::` in headers tho. and for implementation files you can always do `using namespace clio;`.
Ideally it would eventually all be in one namespace `clio`. When that's done you will not need to type `clio::` anymore as we will be in the same namespace already.

PR #455, opened 2022-12-16
add time measurement profiler

cindyyan317:
1 Create a profiler function to measure the time consuming of functions
2 Add unittest

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ from which data can be extracted. The rippled node does not need to be running o
 
 ## Building
 
-Clio is built with CMake. Clio requires at least GCC-11 (C++20), and Boost 1.75.0 or later.
+Clio is built with CMake. Clio requires at least GCC-11/clang-14.0.0 (C++20), and Boost 1.75.0.
 
 Use these instructions to build a Clio executable from the source. These instructions were tested on Ubuntu 20.04 LTS.
 
@@ -79,13 +79,11 @@ In addition, the parameter `start_sequence` can be included and configured withi
 
 In addition, the optional parameter `finish_sequence` can be added to the json file as well, specifying where the ledger can stop.
 
-To add `start_sequence` and/or `finish_sequence` to the config.json file appropriately, they will be on the same top level of precedence as other parameters (such as database, etl_sources, read_only, etc.) and be specified with an integer. Here is an example below, added at the end of the config.json:
+To add `start_sequence` and/or `finish_sequence` to the config.json file appropriately, they will be on the same top level of precedence as other parameters (such as `database`, `etl_sources`, `read_only`, etc.) and be specified with an integer. Here is an example snippet from the config file:
 
 ```json
-    ...
-    "read_only": false,
-    "start_sequence": [integer] the ledger index to start from
-    "finish_sequence": [integer] the ledger index to finish at
+"start_sequence": 12345,
+"finish_sequence": 54321
 ```
 
 The parameters `ssl_cert_file` and `ssl_key_file` can also be added to the top level of precedence of our Clio config. `ssl_cert_file` specifies the filepath for your SSL cert while `ssl_key_file` specifies the filepath for your SSL key. It is up to you how to change ownership of these folders for your designated Clio user. Your options include:
@@ -93,16 +91,15 @@ The parameters `ssl_cert_file` and `ssl_key_file` can also be added to the top l
 - Changing the permissions directly so it's readable by your Clio user
 - Running Clio as root (strongly discouraged)
 
-An example of how to include `ssl_cert_file` and `ssl_key_file` in the Clio config are below:
+An example of how to specify `ssl_cert_file` and `ssl_key_file` in the config:
+
 ```json
-{
-    "server":{
-        "ip": "0.0.0.0",
-        "port": 51233
-    },
-    "ssl_cert_file" : "/full/path/to/cert.file",
-    "ssl_key_file" : "/full/path/to/key.file"
-}
+"server":{
+    "ip": "0.0.0.0",
+    "port": 51233
+},
+"ssl_cert_file" : "/full/path/to/cert.file",
+"ssl_key_file" : "/full/path/to/key.file"
 ```
 
 Once your config files are ready, start rippled and Clio. It doesn't matter which you
@@ -185,8 +182,32 @@ You must:
 ## Logging
 Clio provides several logging options, all are configurable via the config file and are detailed below.
 
-`log_level`: The minimum level of severity at which the log message will be outputted.
+`log_level`: The minimum level of severity at which the log message will be outputted by default.
 Severity options are `trace`, `debug`, `info`, `warning`, `error`, `fatal`. Defaults to `info`.
+
+`log_format`: The format of log lines produced by clio. Defaults to `"%TimeStamp% (%SourceLocation%) [%ThreadID%] %Channel%:%Severity% %Message%"`.
+Each of the variables expands like so
+- `TimeStamp`: The full date and time of the log entry
+- `SourceLocation`: A partial path to the c++ file and the line number in said file (`source/file/path:linenumber`)  
+- `ThreadID`: The ID of the thread the log entry is written from
+- `Channel`: The channel that this log entry was sent to
+- `Severity`: The severity (aka log level) the entry was sent at
+- `Message`: The actual log message
+
+`log_channels`: An array of json objects, each overriding properties for a logging `channel`. 
+At the moment of writing, only `log_level` can be overriden using this mechanism.
+
+Each object is of this format:
+```json
+{
+    "channel": "Backend",
+    "log_level": "fatal"
+}
+``` 
+If no override is present for a given channel, that channel will log at the severity specified by the global `log_level`.
+Overridable log channels: `Backend`, `WebServer`, `Subscriptions`, `RPC`, `ETL` and `Performance`.
+
+> **Note:** See `example-config.json` for more details. 
 
 `log_to_console`: Enable/disable log output to console. Options are `true`/`false`. Defaults to true.
 
@@ -202,6 +223,11 @@ rotate the current log file. Defaults to 12 hours.
 
 Note, time-based log rotation occurs dependently on size-based log rotation, where if a
 size-based log rotation occurs, the timer for the time-based rotation will reset.
+
+`log_tag_style`: Tag implementation to use. Must be one of:
+- `uint`: Lock free and threadsafe but outputs just a simple unsigned integer 
+- `uuid`: Threadsafe and outputs a UUID tag
+- `none`: Don't use tagging at all
 
 ## Cassandra / Scylla Administration
 


src/util/Profiler.h
```diff
@@ -0,0 +1,36 @@
...
+#include <chrono>
+#include <type_traits>
+#include <utility>
+
+namespace util::detail {
+
+/// @brief Profiler function to measure the time consuming
+/// @param func function object, can be a lamdba or function wrapper
+/// @return return a pair if function wrapper has return value: result of
+/// function wrapper and the elapsed time during executing the given function
```
cjcobb23: What are the units of time?

cindyyan317: ha. i think it is nanoseconds, will add it in comment~

src/util/Profiler.h
```diff
@@ -0,0 +1,36 @@
+
+#include <chrono>
+#include <type_traits>
+#include <utility>
+
+namespace util::detail {
```
godexsoft: Let's expose the new utility in `namespace util` without `detail`. 
`detail` is used for implementation details which ideally are not to be used outside of the feature you are making.

So to sum up - exposed APIs should go into just `util` but any implementation details can go into `util::detail` :+1:

src/util/Profiler.h
```diff
@@ -0,0 +1,36 @@
+
```
godexsoft: 1. our top level copyright message missing (copy from any other header) - i should probably add a githook for this
2. `#pragma once` missing

src/util/Profiler.h
```diff
@@ -0,0 +1,36 @@
...
+/// @return return a pair if function wrapper has return value: result of
+/// function wrapper and the elapsed time during executing the given function
+/// only return the elapsed time if function wrapper does not have return value
+template <typename F>
+auto
+timed(F func)
+{
+    auto start = std::chrono::system_clock::now();
+
+    if constexpr (std::is_same<typename std::result_of_t<F()>, void>{})
```
godexsoft: You can use `if constexpr (std::is_same_v<decltype(func()), void>)` here as a shorter syntax i believe

src/util/Profiler.h
```diff
@@ -0,0 +1,36 @@
...
+namespace util::detail {
+
+/// @brief Profiler function to measure the time consuming
+/// @param func function object, can be a lamdba or function wrapper
+/// @return return a pair if function wrapper has return value: result of
+/// function wrapper and the elapsed time during executing the given function
+/// only return the elapsed time if function wrapper does not have return value
+template <typename F>
+auto
+timed(F func)
```
godexsoft: Maybe we can avoid copying here (in some cases) by using `F&& func`?

unittests/ProfilerTest.cpp
```diff
@@ -0,0 +1,41 @@
+#include "util/Profiler.h"
+#include <gtest/gtest.h>
+#include <thread>
+
+namespace util::detail {
```
godexsoft: The tests should not be in a namespace themselves. Let's just `using namespace util;` etc. 
Take a look at other unit-tests we have for an example :+1:

unittests/ProfilerTest.cpp
```diff
@@ -0,0 +1,41 @@
+#include "util/Profiler.h"
```
godexsoft: Copyright header missing - copy from any other source file please

godexsoft: I think we always use `#include <headerfile>` style everywhere (except two places that should probably be fixed too).

unittests/ProfilerTest.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <gtest/gtest.h>
+#include <thread>
+
+namespace util::detail {
+TEST(ProfilerTest, HasReturnValue)
+{
+    auto [ret, time] = timed([]() {
+        std::this_thread::sleep_for(std::chrono::seconds(1));
```
godexsoft: Let's use `std::chrono::milliseconds(5)` to waste less time running tests :+1:

unittests/ProfilerTest.cpp
```diff
@@ -0,0 +1,41 @@
...
+};
+
+TEST(ProfilerTest, Functor)
+{
+    auto time = timed(FunctorTest());
+
+    ASSERT_NE(time, 0);
+}
+
+}  // namespace util::detail
```
godexsoft: Github does not like it when we don't add the final `\n` to files. You can configure your IDE to automatically insert newline at end of files and it will usually work fine that way.

godexsoft (changes requested):
Great job @cindyyan317! I added a few comments here and there, mostly small things that we do a certain way already etc.

You also still need to use your new utility in all/most places where it can be used in the codebase :1st_place_medal:

src/util/Profiler.h
cindyyan317: yes , it is ! :)

src/util/Profiler.h
cjcobb23: I think would be better to return milliseconds. I think everywhere we log, we convert to milliseconds in one way or another.

cindyyan317: Good point.

src/util/Profiler.h
```diff
@@ -0,0 +1,57 @@
...
+namespace util {
+
+/// @brief Profiler function to measure the time consuming
+/// @param func function object, can be a lamdba or function wrapper
+/// @return return a pair if function wrapper has return value: result of
+/// function wrapper and the elapsed time(ms) during executing the given
+/// function only return the elapsed time if function wrapper does not have
+/// return value
+template <typename F>
+auto
```
godexsoft: Let's add `[[nodiscard]]` here

src/util/Profiler.h
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma once
+
+#include <chrono>
+#include <type_traits>
+#include <utility>
+
+namespace util {
+
+/// @brief Profiler function to measure the time consuming
```
godexsoft: While some code in the codebase is using `///` style comments for doxygen, we prefer `/** */` style in new code - please check Config or Logging abstraction and how doxy comments look there :+1:

src/util/Profiler.h
```diff
@@ -0,0 +1,57 @@
...
+template <typename F>
+auto
+timed(F&& func)
+{
+    auto start = std::chrono::system_clock::now();
+
+    if constexpr (std::is_same_v<decltype(func()), void>)
+    {
+        func();
+        return std::chrono::duration_cast<std::chrono::milliseconds>(
```
godexsoft: this std::chrono::milliseconds could be a template parameter `DurationType` defaulted to `milliseconds`. since we already return `auto` it should be easy:
```cpp
auto elapsedMillis = util::timed([]() { ... }); // default to millis
auto elapsedNanos = util::timed<std::chrono::nanoseconds>([]() { ... }); // custom
```

godexsoft: And also use it on line 50 same way

unittests/ProfilerTest.cpp
```diff
@@ -0,0 +1,70 @@
...
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <gtest/gtest.h>
+#include <thread>
+#include <util/Profiler.h>
+
+using namespace util;
+TEST(ProfilerTest, HasReturnValue)
```
godexsoft: If the test suite is `ProfilerTest` should the `timed` util perhaps be called `profile`?

src/util/Profiler.h
godexsoft: And `noexcept` but only if we can guarantee `duration_cast` will not throw in this situation. Might not be feasible if we allow customization thru tparam `DurationType` tho.

godexsoft (changes requested):
Adding a few more small things that could improve this pr further.
Also, we still need to use this new profiler in all places where we currently hardcode this by hand.

src/util/Profiler.h
cindyyan317: [nodiscard] added. I did not add "noexcept", because if the func throws exception, the whole process will be terminated.

src/etl/ReportingETL.cpp
```diff
@@ -620,12 +620,14 @@ ReportingETL::runETLPipeline(uint32_t startSequence, int numExtractors)
...
-                std::optional<org::xrpl::rpc::v1::GetLedgerResponse>
-                    fetchResponse{fetchLedgerDataAndDiff(currentSequence)};
-                auto end = std::chrono::system_clock::now();
-
-                auto time = ((end - start).count()) / 1000000000.0;
+                auto [fetchResponse, timeDiff] =
+                    util::timed<std::chrono::nanoseconds>([&]() {
+                        std::optional<org::xrpl::rpc::v1::GetLedgerResponse>
+                            fetchResponse{
+                                fetchLedgerDataAndDiff(currentSequence)};
```
godexsoft: I think we can safely do 
```cpp
return fetchLedgerDataAndDiff(currentSequence);
```
instead of all this.
WDYT?

src/rpc/RPCHelpers.cpp
```diff
@@ -1652,60 +1649,58 @@ traverseTransactions(
...
-        cursorJson[JS(ledger)] = retCursor->ledgerSequence;
-        cursorJson[JS(seq)] = retCursor->transactionIndex;
-        response[JS(marker)] = cursorJson;
-    }
-
-    for (auto const& txnPlusMeta : blobs)
-    {
-        if ((txnPlusMeta.ledgerSequence < minIndex && !forward) ||
-            (txnPlusMeta.ledgerSequence > maxIndex && forward))
+    auto timeDiff = util::timed([&, &retCursor = retCursor, &blobs = blobs]() {
```
godexsoft: Doesn't `&` already cover `retCursor` and `blobs`?

cindyyan317: it should be use explicitly, otherwise the compiler will complain the retCursor is not used :(.

src/etl/ReportingETL.cpp
cindyyan317: ha. yes. we can do that

src/rpc/RPCHelpers.cpp
godexsoft: Ha! in this case you can either
```cpp
[[maybe_unused]] auto [blobs, retCursor] = ...;
```
i can see how this can be undesired. so maybe we can do this as a compromise:
```cpp
util::timed([&, &retCursor, &blobs]() { ... });
```

src/webserver/WsBase.h
```diff
@@ -343,11 +344,10 @@ class WsSession : public WsBase,
...
 
-            auto start = std::chrono::system_clock::now();
-            auto v = RPC::buildResponse(*context);
-            auto end = std::chrono::system_clock::now();
-            auto us = std::chrono::duration_cast<std::chrono::microseconds>(
-                end - start);
+            auto [v, timeDiff] =
+                util::timed([&]() { return RPC::buildResponse(*context); });
+
+            auto us = std::chrono::duration<int, std::milli>(timeDiff);
```
godexsoft: Looks like it changed from microseconds to milliseconds here?

src/webserver/HttpBase.h
```diff
@@ -437,11 +438,10 @@ handle_request(
...
         boost::json::object response;
-        auto start = std::chrono::system_clock::now();
-        auto v = RPC::buildResponse(*context);
-        auto end = std::chrono::system_clock::now();
-        auto us =
-            std::chrono::duration_cast<std::chrono::microseconds>(end - start);
+        auto [v, timeDiff] =
+            util::timed([&]() { return RPC::buildResponse(*context); });
+
+        auto us = std::chrono::duration<int, std::milli>(timeDiff);
```
godexsoft: Looks like it changed from microseconds to milliseconds here?

src/rpc/handlers/NFTHistory.cpp
```diff
@@ -38,28 +39,25 @@ doNFTHistory(Context const& context)
...
+        context,
+        [&tokenID, &outerFuncName](
+            std::shared_ptr<Backend::BackendInterface const> const& backend,
+            std::uint32_t const limit,
+            bool const forward,
+            std::optional<Backend::TransactionsCursor> const& cursorIn,
+            boost::asio::yield_context& yield)
+            -> Backend::TransactionsAndCursor {
+            auto const [txnsAndCursor, timeDiff] =
+                util::timed([&, &tokenID = tokenID]() {
```
godexsoft: `&tokenID` should be sufficient? Same as above in the outer lambda capture

src/webserver/HttpBase.h
godexsoft: I think you can also just do 
```cpp
auto [v, us] = util::timed<std::micro>([&]() { ... });
```

src/rpc/RPCHelpers.cpp
```diff
@@ -798,14 +799,10 @@ traverseOwnedNodes(
                         .count()
```
godexsoft: Looks like this is also a case we can fix to use the new utility

godexsoft: Search for `end - start` and also for `duration_cast<` to see if you can find even more places :+1:

godexsoft (changes requested):
Leaving a few more comments. Other than that I think this is great and we should merge it after fixing the small things i noticed :1st_place_medal:

src/webserver/WsBase.h
cindyyan317: The output of timed is milli , so the duration unit is milli.

src/webserver/HttpBase.h
cindyyan317: :(. i could not . The us has to be duration. the output of timed is absolute duration count in a certain unit.
I think the milli is okay here, the timeDiff is milli, so duration should be milli. The logDuration will cast it to other unit.

src/rpc/RPCHelpers.cpp
cindyyan317: This one is tricky. We could not put the code into a lambda . The code return the whole function in the middle. I suggest we leave this tricky one when we have unittest covered , or at least we know how to test it case by case.
We have a few places have not changed:
1 The return value is not copyable, could not be wrapped by timed. Like CassandraBackend.cpp
2 Multiple stage duration 
3 WorkQueue.h : start time and end time could not put in one function. End time is in a coroutine.

src/rpc/RPCHelpers.cpp
cindyyan317: Screenshot 2022-12-20 at 09 27 12

src/rpc/handlers/NFTHistory.cpp
cindyyan317: Got "["](error: '&' cannot precede a capture when the capture default is '&') .

godexsoft: Ah, ok. C++...

src/webserver/WsBase.h
godexsoft: But `us` was in microseconds before, no? And also, why do we need to wrap `timeDiff` at all? it's already `millis` as it is returned from `timed` so i don't see why not just pass `timeDiff` instead of `us` if it's the same? 
In general, as long as the end result prints in the same unit as before - i don't mind how it's implemented too much :+1:

PR #456, opened 2022-12-20
Build Clio on macOS and Ubuntu 22.04

legleux:
We should be building/testing on the preferred development environment and the next Linux version we're going to support.


godexsoft (approved):
LGTM.

PR #459, opened 2022-12-21
Populate namespaces

godexsoft:
This is a draft but at least this builds and runs both the tests and `clio` itself.
I'm not a fan of the `subscription` namespace, perhaps we should think of a better name.


src/backend/BackendInterface.cpp
```diff
@@ -20,16 +20,17 @@
...
 using namespace clio;
 
 // local to compilation unit loggers
 namespace {
-clio::Logger gLog{"Backend"};
+util::Logger gLog{"Backend"};
 }  // namespace
 
-namespace Backend {
+namespace clio::backend {
```
cindyyan317: instead of backend, how do u think we call it clio::data?
backend is kinda large scope

godexsoft: Would you also rename `backend` directory to `data`? currently the namespaces match their respective source directories. 
I think long term `data` is a good name - we will likely introduce some abstraction/layer for accessing data on top of what is now known as `BackendInterface`.. eventually.

What about `subscription`? Any better name ideas there? :)

src/subscriptions/SubscriptionManager.h
```diff
@@ -20,18 +20,24 @@
...
 
+namespace clio {
+
+namespace web {
 class WsBase;
+}  // namespace web
+
+namespace subscription {
 
 class Subscription
```
cindyyan317, line 37: I am okay with the namespace name. 
But this class is a little bit confusing . it has subscribe function, as well as publish function. In other place, the instances of it are called subscribers. It not make sense seeing something "subscribers.publish(something)"

godexsoft: Yea, i agree and we can rename it :+1: 
This pr however is mostly about namespaces - we better keep diffs as small as possible (and perform more refactorings in other dedicated prs). This diff is already pretty big as is..

src/backend/BackendInterface.cpp
godexsoft: Ended up renaming `backend` to `data` and `subscriptions` to `feed`. 
Also renamed the directories same way.

src/subscriptions/SubscriptionManager.h
godexsoft: Let's leave the internals of these confusing classes for a later refactoring otherwise the diff will be even bigger than it already is.

godexsoft:
This is so behind now. One day we will get back to this.. maybe easier to do from scratch again tho.

godexsoft:
This was implemented in #817 and #820

PR #462, opened 2023-01-06
Feature/dosguard : add request limit 

cindyyan317:
Test: https://github.com/cindyyan317/my_test/tree/main/dosguard

1 Add request limit to Dosguard
2 Change dosguard invoke places

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -48,13 +48,24 @@ public:
 template <typename SweepHandler>
 class BasicDOSGuard : public BaseDOSGuard
 {
-    mutable std::mutex mtx_;  // protects ipFetchCount_
-    std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    // Accumulated state per IP, state will be reset accordingly
+    struct ClientState
+    {
+        // accumulated transfered byte
+        std::uint32_t transferedByte = 0;
+        // accumulated served requests count
+        std::uint32_t requestsCount = 0;
+    };
+
+    mutable std::mutex mtx_;
+    // accumulated states map
+    std::unordered_map<std::string, ClientState> ipState_;
     std::unordered_map<std::string, std::uint32_t> ipConnCount_;
     std::unordered_set<std::string> const whitelist_;
 
     std::uint32_t const maxFetches_;
     std::uint32_t const maxConnCount_;
+    std::uint32_t const maxRequestCount_;
     clio::Logger log_{"RPC"};
 
 public:
@@ -68,6 +79,7 @@ public:
         : whitelist_{getWhitelist(config)}
         , maxFetches_{config.valueOr("dos_guard.max_fetches", 100000000u)}
         , maxConnCount_{config.valueOr("dos_guard.max_connections", 1u)}
+        , maxRequestCount_{config.valueOr("dos_guard.max_requests", 10u)}
     {
         sweepHandler.setup(this);
     }
@@ -98,25 +110,33 @@ public:
         if (whitelist_.contains(ip))
             return true;
 
-        std::unique_lock lck(mtx_);
-        bool fetchesOk = true;
-        bool connsOk = true;
         {
-            auto it = ipFetchCount_.find(ip);
-            if (it != ipFetchCount_.end())
-                fetchesOk = it->second <= maxFetches_;
-        }
-        {
+            std::unique_lock lck(mtx_);
+            if (ipState_.find(ip) != ipState_.end())
+            {
+                auto [transferedByte, requests] = ipState_.at(ip);
+                if (transferedByte > maxFetches_ || requests > maxRequestCount_)
+                {
+                    log_.warn()
+                        << "Dosguard:Client surpassed the rate limit. ip = "
+                        << ip << " Transfered Byte:" << transferedByte
+                        << " Requests:" << requests;
+                    return false;
+                }
+            }
             auto it = ipConnCount_.find(ip);
             if (it != ipConnCount_.end())
             {
-                connsOk = it->second <= maxConnCount_;
+                if (it->second > maxConnCount_)
+                {
+                    log_.warn()
+                        << "Dosguard:Client surpassed the rate limit. ip = "
+                        << ip << " Concurrent connection:" << it->second;
+                    return false;
+                }
             }
         }
-        if (!fetchesOk || !connsOk)
-            log_.warn() << "Client surpassed the rate limit. ip = " << ip;
-
-        return fetchesOk && connsOk;
+        return true;
     }
 
     /**
@@ -170,11 +190,32 @@ public:
 
         {
             std::unique_lock lck(mtx_);
-            auto it = ipFetchCount_.find(ip);
-            if (it == ipFetchCount_.end())
-                ipFetchCount_[ip] = numObjects;
-            else
-                it->second += numObjects;
+            ipState_[ip].transferedByte += numObjects;
+        }
+
+        return isOk(ip);
+    }
+
+    /**
+     * @brief Adds one request for the given ip address.
+     *
+     * If the total sums up to a value equal or larger than maxRequestCount_
+     * the operation is no longer allowed and false is returned; true is
+     * returned otherwise.
+     *
+     * @param ip
+     * @return true
+     * @return false
+     */
+    [[maybe_unused]] bool
+    request(std::string const& ip) noexcept
+    {
+        if (whitelist_.contains(ip))
+            return true;
+
+        {
+            std::unique_lock lck(mtx_);
+            ipState_[ip].requestsCount++;
         }
 
         return isOk(ip);
@@ -188,7 +229,7 @@ public:
     clear() noexcept override
     {
         std::unique_lock lck(mtx_);
-        ipFetchCount_.clear();
+        ipState_.clear();
     }
 
 private:
--- a/src/webserver/HttpBase.h
+++ b/src/webserver/HttpBase.h
@@ -238,6 +238,27 @@ public:
         if (ec)
             return httpFail(ec, "read");
 
+        auto ip = derived().ip();
+
+        if (!ip)
+        {
+            return;
+        }
+
+        auto const httpResponse = [&](http::status status,
+                                      std::string content_type,
+                                      std::string message) {
+            http::response<http::string_body> res{status, req_.version()};
+            res.set(
+                http::field::server,
+                "clio-server-" + Build::getClioVersionString());
+            res.set(http::field::content_type, content_type);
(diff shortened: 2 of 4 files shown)


src/webserver/DOSGuard.h
```diff
@@ -68,6 +79,7 @@ class BasicDOSGuard : public BaseDOSGuard
         : whitelist_{getWhitelist(config)}
         , maxFetches_{config.valueOr("dos_guard.max_fetches", 100000000u)}
         , maxConnCount_{config.valueOr("dos_guard.max_connections", 1u)}
+        , maxRequestCount_{config.valueOr("dos_guard.max_requests", 10u)}
```
cindyyan317, line 82: we can discuss the default value here , maybe 100

src/webserver/DOSGuard.h
```diff
@@ -98,25 +110,33 @@ class BasicDOSGuard : public BaseDOSGuard
         if (whitelist_.contains(ip))
             return true;
 
-        std::unique_lock lck(mtx_);
-        bool fetchesOk = true;
-        bool connsOk = true;
-        {
```
cindyyan317, line 104: merge these two scopes... idk why they are here. rewrite a bit, remove fetchesOK and consOK. not need a variable

src/webserver/DOSGuard.h
```diff
@@ -98,25 +110,33 @@ class BasicDOSGuard : public BaseDOSGuard
...
-        }
         {
+            std::unique_lock lck(mtx_);
+            if (ipState_.find(ip) != ipState_.end())
+            {
+                auto [transferedByte, requests] = ipState_.at(ip);
+                if (transferedByte > maxFetches_ || requests > maxRequestCount_)
+                {
+                    log_.warn()
+                        << "Dosguard:Client surpassed the rate limit. ip = "
```
cindyyan317, line 121: add log for better investigating the issue when limit is over

src/webserver/HttpBase.h
```diff
@@ -260,10 +281,16 @@ class HttpBase : public util::Taggable
...
         }
 
-        auto ip = derived().ip();
-
-        if (!ip)
-            return;
+        // to avoid overwhelm work queue, the request limit check should be
+        // before posting to queue the web socket creation will be guarded via
+        // connection limit
+        if (!dosGuard_.request(ip.value()))
```
cindyyan317, line 287: don't post to work queue if dosguard fail

src/webserver/HttpBase.h
```diff
@@ -380,12 +401,6 @@ handle_request(
         return send(httpResponse(
             http::status::bad_request, "text/html", "Expected a POST request"));
 
-    if (!dosGuard.isOk(ip))
```
cindyyan317, line 383: remove dosGuard in handle_request. the check should happen outside the business logic.

src/webserver/WsBase.h
```diff
@@ -452,19 +452,22 @@ class WsSession : public WsBase,
         }(std::move(msg));
 
         boost::json::object request;
-        if (!raw.is_object())
```
cindyyan317, line 455: We shall protect server from any requests, even invalid parameter requests. Bad user can send tons of invalid request to overwhelm server. And also, invalid parameter request should not end the session, after senderror , we shall continue async read.

src/webserver/WsBase.h
```diff
@@ -452,19 +452,22 @@ class WsSession : public WsBase,
...
-
-        auto id = request.contains("id") ? request.at("id") : nullptr;
-
-        if (!dosGuard_.isOk(*ip))
+        // dosGuard served request++ and check ip address
+        // dosGuard should check before any request, even invalid request
+        if (!dosGuard_.request(*ip))
         {
-            sendError(RPC::RippledError::rpcSLOW_DOWN, id, request);
+            sendError(RPC::RippledError::rpcSLOW_DOWN, nullptr, request);
```
cindyyan317, line 459: not return here, because even the same connection , we allow service resumes , when requests slow down.

src/webserver/DOSGuard.h
godexsoft: I think 10 is actually a good number, maybe more than enough. Ultimately we should encourage clients to use WS and do most of their requests thru one open connection. So the remaining allowed connections are basically for subscriptions/feeds.

src/webserver/DOSGuard.h
```diff
@@ -48,13 +48,24 @@ class BaseDOSGuard
 template <typename SweepHandler>
 class BasicDOSGuard : public BaseDOSGuard
 {
-    mutable std::mutex mtx_;  // protects ipFetchCount_
-    std::unordered_map<std::string, std::uint32_t> ipFetchCount_;
+    // Accumulated state per IP, state will be reset accordingly
+    struct ClientState
+    {
+        // accumulated transfered byte
+        std::uint32_t TransferedByte = 0;
```
godexsoft: The rest of the codebase seems to (mostly) have lowercase camel for variable names and members like these. Let's try to keep it that way?

src/webserver/DOSGuard.h
cindyyan317: The requests via ws also limit by this number.

src/webserver/DOSGuard.h
```diff
@@ -170,11 +190,49 @@ class BasicDOSGuard : public BaseDOSGuard
 
         {
             std::unique_lock lck(mtx_);
-            auto it = ipFetchCount_.find(ip);
-            if (it == ipFetchCount_.end())
-                ipFetchCount_[ip] = numObjects;
+            auto it = ipState_.find(ip);
+            if (it == ipState_.end())
+            {
+                ipState_.emplace(ip, ClientState{.TransferedByte = numObjects});
```
godexsoft: I have a suspicion that this code will not compile on clang. Parenthesized initialization was not supported a month ago so chances are it still is not.. I'm not sure about AppleClang but i guess same story?

godexsoft: Also, can we not rely on `operator[]` default constructing a zeroed ClinetState here?
So like simplify all this to `ipState_[ip].bytesTransferred += numObjects;`. Unless i'm misunderstanding something. WDYT?

src/webserver/DOSGuard.h
```diff
@@ -170,11 +190,49 @@ class BasicDOSGuard : public BaseDOSGuard
...
+
+    /**
+     * @brief Adds one request for the given ip address.
+     *
+     * If the total sums up to a value equal or larger than maxRequestCount_
+     * the operation is no longer allowed and false is returned; true is
+     * returned otherwise.
+     *
+     * @param ip
+     * @param
```
godexsoft: some stray @param here

src/webserver/DOSGuard.h
```diff
@@ -170,11 +190,49 @@ class BasicDOSGuard : public BaseDOSGuard
...
+    {
+        if (whitelist_.contains(ip))
+            return true;
+
+        {
+            std::unique_lock lck(mtx_);
+            auto it = ipState_.find(ip);
+            if (it == ipState_.end())
+            {
+                ipState_.emplace(ip, ClientState{.RequestsCount = 1});
```
godexsoft: Seem like again we can simplify all this to one line using `operator[]`?

src/webserver/HttpBase.h
godexsoft: If `ip` is nullopt for whatever reason (probably never?) `.value()` will crash the app entirely. I think Maybe we should leave that check with early return in here?

src/webserver/WsBase.h
```diff
@@ -452,19 +452,22 @@ class WsSession : public WsBase,
...
+            sendError(RPC::RippledError::rpcINVALID_PARAMS, nullptr, request);
         }
         else
         {
+            request = raw.as_object();
+
+            auto id = request.contains("id") ? request.at("id") : nullptr;
             perfLog_.debug() << tag() << "Adding to work queue";
 
             if (!queue_.postCoro(
```
godexsoft, line 473: would we still need to pass `dosGuard_.isWhiteListed(*ip)` to postCoro now that we check everything beforehand?

godexsoft (commented):
Good stuff! I left a few minor comments to think about and check. Overall it looks much better than before :+1:

src/webserver/DOSGuard.h
godexsoft: Ah, right, this is requestCount, not connectionCount :+1:
We will have to think what a good number is, you are right.

src/webserver/DOSGuard.h
cindyyan317: you mean rely on the default constructing?

src/webserver/DOSGuard.h
cindyyan317: do we distinguish public / private variable by name?

src/webserver/WsBase.h
cindyyan317: yes , yes. i did not see there is a call to whitelist .

src/webserver/HttpBase.h
cindyyan317: IP valid check is done at line 243

src/webserver/DOSGuard.h
cindyyan317: Great comment. fix it.

src/webserver/DOSGuard.h
godexsoft: as discussed in slack, private members (e.g. in a class) get a `_` at the end of the name while public members (e.g. struct) get no such marker.

src/webserver/HttpBase.h
```diff
@@ -238,6 +238,27 @@ class HttpBase : public util::Taggable
         if (ec)
             return httpFail(ec, "read");
 
+        auto ip = derived().ip();
+
+        if (!ip)
+        {
+            return;
```
godexsoft, line 245: you were asking me if we prefer curly brackets or no brackets for one line ifs. in this case i'd drop them as it's just a 'return' statement and not some long line.

godexsoft (dismissed):
LGTM! still left one comment but that can be left as is too, no problem.

src/webserver/HttpBase.h
officialfrancismendoza: Same sentiments as Alex here. Just be sure that the one-liner still fits formatting-wise when you run the following command to ensure linting compatibility: `find src unittests -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format-11 -i`

officialfrancismendoza (dismissed):

godexsoft (dismissed):

godexsoft (approved):

PR #463, opened 2023-01-06
Fix LGR_IDXS_INVALID error code mismatch

officialfrancismendoza:
**Problem**: Upon Q&A, error code for https://github.com/XRPLF/clio/issues/263 had mismatch with  rippled error message (misconstrued for lgr_idxs_malformed)
**Fix**: Add correct error message hotfix to old account_tx ledger index max out-of-bounds PR


godexsoft (approved):
This should do it.

PR #467, opened 2023-01-10
replace unique_lock with scoped_lock

cindyyan317:


cindyyan317:
TEST: Build pass
Found we use unique_lock a lot. Replace some unique_lock with scoped_lock to avoid overhead.

godexsoft (approved):
LGTM! Let's have a separate project to replace all '()' with '{}' everywhere in the codebase

PR #471, opened 2023-01-12
Demote confusing errors that are not really errors

### LINKED ISSUE #468: Downgrade log level for couple of messages in ETLSource
Currently we have a log line that is marked as `warning` for "Error getting ledger" and `error` for "Error executing function" which is not really an error at all.
This happens when `clio` needs a ledger seq that is not present in the `rippled` that is accessed thru the ETLSource. Clio will try all sources and if all sources failed to provide the ledger required `clio` will wait for a couple seconds before trying again. Eventually `rippled` will backfill and the ledger will become available to `clio`. 

This log line should probably be something like "INF Ledger xxx is not yet available from this ETL source" and when all sources failed it could say "INF Ledger xxx is not yet available from any of the configured sources, waiting xxx seconds and retrying" or similar.

godexsoft:
Fixes #468

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -993,7 +993,7 @@ ETLLoadBalancer::fetchLedger(
             }
             else
             {
-                log.warn() << "Error getting ledger = " << ledgerSequence
+                log.warn() << "Could not fetch ledger " << ledgerSequence
                            << ", Reply: " << response.DebugString()
                            << ", error_code: " << status.error_code()
                            << ", error_msg: " << status.error_message()
@@ -1186,9 +1186,9 @@ ETLLoadBalancer::execute(Func f, uint32_t ledgerSequence)
         numAttempts++;
         if (numAttempts % sources_.size() == 0)
         {
-            log_.error() << "Error executing function "
-                         << " - ledger sequence = " << ledgerSequence
-                         << " - Tried all sources. Sleeping and trying again";
+            log_.info() << "Ledger sequence " << ledgerSequence
+                        << " is not yet available from any configured sources. "
+                        << "Sleeping and trying again";
             std::this_thread::sleep_for(std::chrono::seconds(2));
         }
     }


legleux:
If you're seeing this message are you guaranteed to be properly connected to an ETL source?

If the ETL source is configured such that the ledger will never be available will it get out of this loop?

godexsoft:
@legleux The original implementation is already flawed so that the loop does not exits. While this is a potential problem and what you noted is true - this "issue" is not about fixing those behaviours but just demoting the severity of the messages so that people who see them don't panic. Those "errors" are something `clio` should be able to recover from if ETL sources are configured correctly. On the other hand "error" is not "fatal" either. But people still find it scary so I rather have it as a warn/info for this case.

Before trying to go into the loop there is a check (and wait) for the required seq to be validated by the network. This does not mean however that the required seq is available for fetching - it only checks that the latest seq validated is at least the one we ask for or newer. So in a way, yes, we should not be going into this loop until we know that ETL source connection is ok.
However, the indefinite loop itself will be fixed when we get to redesign and update the ETL subsystem.

cindyyan317:
If I understand correctly, `clio` will continue trying and it might succeed later. Indeed , it is not an error.

cindyyan317 (approved):

legleux:
This comment was more about the error message itself. I'm familiar with this message and get that it's not really an "error".
I meant to get some info to maybe help to make the message more helpful as when I saw it, I couldn't tell if the problem was with Clio or rippled config or if it would resolve itself once rippled was synced or...

godexsoft:
> This comment was more about the error message itself. I'm familiar with this message and get that it's not really an "error". I meant to get some info to maybe help to make the message more helpful as when I saw it, I couldn't tell if the problem was with Clio or rippled config or if it would resolve itself once rippled was synced or...

If you have some better wording - please let me know and i'll apply it. I don't mind what wording we use, i just don't want it to say "error". 
Thanks

PR #473, opened 2023-01-16
Fix bug with ClioVersion to prevent crash at runtime

godexsoft:
Remove semver check at runtime preventing crash and fix for tagless fork.
@legleux if you want to enforce semver check please do it either in CMake or at compile time but definitely not at runtime 👍


godexsoft:
@legleux this is a quick workaround for us to be able to build on our cloned repos again without crashing at application start.  There is still something strange going on as the version on my machine comes out with the branch but on CI it did not populate the branch as one'd expect resulting in `2023xxxxx--gitrev` instead of `2023xxxxx-branch-gitrev`. It does work on my local machine though. I hope you can fix this a second time later on and make it work everywhere - no rush. As long as i did not break it for release/master.

cindyyan317 (approved):
LGTM :)

PR #477, opened 2023-01-18
add code coverage job

cindyyan317:
This PR adds 
1 code coverage for gcc compiler
2 enable code coverage on CI  
     *show code coverage report markdown in log
  like:
                Package | Line Rate | Branch Rate | Complexity | Health
                -------- | --------- | ----------- | ---------- | ------
                src.backend | 0% | 0% | 0 | ❌
                src.config | 95% | 59% | 0 | ✔
                src.etl | 0% | 0% | 0 | ❌
                src.log | 90% | 32% | 0 | ✔
                src.rpc | 78% | 33% | 0 | ✔
                src.subscriptions | 0% | 0% | 0 | ❌
                src.util | 100% | 0% | 0 | ✔
                src.webserver | 100% | 55% | 0 | ✔
                **Summary** | **37%** (164 / 447) | **15%** (174 / 1173) | **0** | ❌
    *upload report to CI
Ideally, we can add coverage summary in PR comments. But github does not allow GITHUB_TOKEN to write PR https://docs.github.com/en/actions/security-guides/automatic-token-authentication. To add the coverage summary to comment, we need another workflow triggered by "build.yml". see https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,6 @@ jobs:
           cmake --build clio-packages/build --parallel $(nproc)
           cp ./clio-packages/build/clio-prefix/src/clio-build/clio_tests .
           mv ./clio-packages/build/*.${{ matrix.type.suffix }} .
-
       - name: Artifact packages
         uses: actions/upload-artifact@v3
         with:
@@ -78,8 +77,6 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - name: ubuntu-22.04
-            experimental: true
           - name: macos-11
             experimental: true
           - name: macos-12
@@ -110,7 +107,6 @@ jobs:
           mac_flags='cxxflags="-std=c++14"'
         fi
         ./b2 ${mac_flags}
-
     - name: install deps
       run: |
         if [[ ${{ matrix.os.name }} =~ mac ]];then
@@ -118,7 +114,6 @@ jobs:
         elif [[ ${{matrix.os.name }} =~ ubuntu ]];then
             sudo apt-get -y install git pkg-config protobuf-compiler libprotobuf-dev libssl-dev wget build-essential doxygen bison flex autoconf clang-format
         fi
-
     - name: Build clio
       run: |
         export BOOST_ROOT=$(pwd)/boost
@@ -127,7 +122,7 @@ jobs:
         if ! cmake --build build -j$(nproc); then
           echo '# 🔥${{ matrix.os.name }}🔥 failed!💥' >> $GITHUB_STEP_SUMMARY
         fi
-
+        
   test_clio:
     name: Test Clio
     runs-on: [self-hosted, Linux]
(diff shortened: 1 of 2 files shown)


CMake/coverage.cmake
```diff
@@ -35,8 +35,46 @@ function(add_converage module)
         add_custom_command(TARGET ${module}-ccov POST_BUILD
             COMMENT "Open ${module}-llvm-cov/index.html in your browser to view the coverage report."
         )
+    elseif("${CMAKE_C_COMPILER_ID}" MATCHES "GNU" OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
+        message("[Coverage] Building with Gcc Code Coverage Tools")
+
+        find_program(GCOV_PATH gcov)
+        IF(NOT GCOV_PATH)
```
godexsoft: While not very important, consider not mixing `ALL_CAPS` and `all_lower` styles in one file?

.github/workflows/build.yml
```diff
@@ -127,7 +122,7 @@ jobs:
         if ! cmake --build build -j$(nproc); then
           echo '# 🔥${{ matrix.os.name }}🔥 failed!💥' >> $GITHUB_STEP_SUMMARY
         fi
-
+
```
godexsoft, line 125: Could remove this space and just leave the newline. Up to you though.

CMake/coverage.cmake
```diff
@@ -35,8 +35,46 @@ function(add_converage module)
...
+	       MESSAGE(FATAL_ERROR "gcov not found! Aborting...")
+        ENDIF() # NOT GCOV_PATH
+        find_program(GCOVR_PATH gcovr)
+        IF(NOT GCOVR_PATH)
+	       MESSAGE(FATAL_ERROR "gcovr not found! Aborting...")
```
godexsoft: Some formatting could be nice in the file. The tab spaces are not equal on different lines like this one.

CMake/coverage.cmake
```diff
@@ -35,8 +35,46 @@ function(add_converage module)
...
+        ADD_CUSTOM_TARGET(${module}-ccov-report
+        COMMAND ${module}  ${TEST_PARAMETER}
+            COMMAND rm -rf ${COV_OUTPUT_PATH}
+            COMMAND mkdir ${COV_OUTPUT_PATH}
+            COMMAND gcovr -r ${CMAKE_SOURCE_DIR} --object-directory=${PROJECT_BINARY_DIR}  --html-details ${COV_OUTPUT_PATH}/index.html --exclude='${CMAKE_SOURCE_DIR}/unittests/' --exclude='${PROJECT_BINARY_DIR}/'
+            WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+            COMMENT "Running gcovr to produce Cobertura code coverage report."
+        )
+        add_custom_command(TARGET ${module}-ccov-report POST_BUILD
+        COMMENT "Open ${COV_OUTPUT_PATH}/index.html in your browser to view the coverage report."
```
godexsoft: Could use a tab or some spaces to make it look more in line with the other formatting of this file. While this happens automatically for C++ files with `clang-format`, unfortunately, we have to maintain it ourselves in other files :-(

godexsoft (dismissed):
LGTM! Left a few comments in case you want to make it prettier but all optional.

legleux:
I think the code coverage should be a completely different step rather than using the experimental Ubuntu 22.04 build to do it.

This PR is also adding a bunch of stuff that isn't really necessary and produced quite different results from what I tested last month. Does CI need to install/run multiple code coverage implementations? (go with gcc or clang)
Then again, I can't really tell how the results are being generated from `coverage.cmake` :grimacing:

`src.backend` says 0%? That's the most extensively covered part of Clio?
edit:
just got some of my questions answered fro @godexsoft

cindyyan317:
@legleux  I believe @godexsoft has answered some of your questions.
Just keep a note here:
1 Why Ubuntu 22.04 ? 
   I am not a CI expert. As you can see from the build.yml, I just move Ubuntu 22.04 from build_dev jobs to code_coverage, adding more steps related with code coverage. Please feel free to correct it.
2 The test result:
   Because we have to run code coverage with available gcda/gcno files, we need to run it on building machine where db is not available. So I filter out backend testcases. Github action provides services container to run the service in runner. I have tried Cassandra service, however it fails to start. We have planed to remove db dependency from unittests. I think it is okay to disable backend unittests for now. The main purpose of unittest code coverage is to make sure our new added code is covered sufficiently. 
3 We do support both clang and gcc. But for CI usage, I think it is okay to run only one build tool to generate the report. I believe the results from two compilers are identical.

CMake/coverage.cmake
cindyyan317: will update these format issue. :). sometimes i just paste what i get from reference page, that's why u see a lot of format mismatch~

godexsoft (dismissed):
Awesome 🥳

godexsoft (approved):
Beautiful 🥇

PR #481, opened 2023-01-19
Update gitignore and readme

cindyyan317:
sometimes we need two sets of makefiles on the same machine, 
we shall ignore build_clang / build_gcc/ build_ccov_enable such.

update readme , adding gcovr dep


godexsoft (approved):
LGTM

Issue #298, opened 2022-09-07
Leftover bypass should be removed


godexsoft:
See 8aaa365e.
Two bypasses were added, the first one is already gone but the second one is still in the codebase.
This is likely not desired. Let's test without the bypass and get rid of it.

PR #484, opened 2023-01-20
Eliminate extraneous bypass in ETLSource.cpp

officialfrancismendoza:
**Problem (#298):** Persistent leftover bypass in ETLSource.cpp that automatically disables proper check if source has proper ledgerSequence
**Fix:** Remove bypass

### DIFF
--- a/src/etl/ETLSource.cpp
+++ b/src/etl/ETLSource.cpp
@@ -1159,7 +1159,11 @@ ETLLoadBalancer::execute(Func f, uint32_t ledgerSequence)
 
         log_.debug() << "Attempting to execute func. ledger sequence = "
                      << ledgerSequence << " - source = " << source->toString();
-        if (source->hasLedger(ledgerSequence) || true)
+        // Originally, it was (source->hasLedger(ledgerSequence) || true)
+        /* Sometimes rippled has ledger but doesn't actually know. However,
+        but this does NOT happen in the normal case and is safe to remove
+        This || true is only needed when loading full history standalone */
+        if (source->hasLedger(ledgerSequence))
         {
             bool res = f(source);
             if (res)


cindyyan317:
I think there is something tricky happening here.
We can't remove this line until we refactor the function "loadInitialLedger" .

src/etl/ETLSource.cpp
```diff
@@ -1159,7 +1159,7 @@ ETLLoadBalancer::execute(Func f, uint32_t ledgerSequence)
 
         log_.debug() << "Attempting to execute func. ledger sequence = "
                      << ledgerSequence << " - source = " << source->toString();
-        if (source->hasLedger(ledgerSequence) || true)
+        if (source->hasLedger(ledgerSequence))
```
cindyyan317, line 1166: I am happy to see this line got remove !!!
but here is something we need to figure out first.
`
ETLLoadBalancer::loadInitialLedger(uint32_t sequence, bool cacheOnly)
{
    execute(
        [this, &sequence, cacheOnly](auto& source) {
            bool res =
                source->loadInitialLedger(sequence, downloadRanges_, cacheOnly);
            if (!res)
            {
                log_.error() << "Failed to download initial ledger."
                             << " Sequence = " << sequence
                             << " source = " << source->toString();
            }
            return res;
        },
        sequence);
}
`
The execute func is used to init the ledger. So "sequence" from "**loadIntialLedger**" is not in the range definitely. It might be root cause why we have this poison here. 
To fix this line, we need to refactor the larger scope of code.

cjcobb23: This line is just here because sometimes rippled has the ledger but doesn't actually know about it. This doesn't happen in the normal case and should be safe to remove. I believe I only needed this when loading full history in standalone mode.

cindyyan317: @officialfrancismendoza  To confirm it is okay to call loadInitialLedger in execute function with the actual check enable, can we have a test on emtpy db ?

cindyyan317: Thanks @cjcobb23  for clarification. 
@officialfrancismendoza  Can we also add CJ's comment on the code to note there is the rare case we might need it.

cindyyan317 (dismissed):

src/etl/ETLSource.cpp
officialfrancismendoza: Added relevant note and rebased @cindyyan317

godexsoft: I'm slightly worried to remove this just based on CJ's knowledge and without unit-tests to support any assumptions but i'm willing to take the risk. Hopefully CJ is right and we will not hit this case in normal operation.

godexsoft (approved):

PR #485, opened 2023-01-25
Fix clio-server link issue

cindyyan317:
When enable CODE_COVERAGE, we build and link clio lib and clio_tests with coverage dependency. Building clio_server will get a linker issue. We dont really need clio_server run with coverage dependency, but just avoid the error message,
I am changing the link option from private->public .


godexsoft (approved):
LGTM

PR #487, opened 2023-01-26
Implement basics of RPC subsystem

godexsoft:
Implements #494

This adds some basic helpers for the nextgen rpc system which is testable and separates concerns:
* Parsing of JSON requests is done in a common and testable way
* Handlers don't know about JSON at all and operate on handler-defined `Input` and `Output` POD structs
* Results and errors are handled using `ripple::Expected` (will be replaced with `std::expected` eventually)
* Handlers expose a `Spec` that defines how the JSON is to be validated
* AnyHandler is a type-erased wrapper that helps convert `JSON->Input` and `Output->JSON` (should probably use a strategy to specify how this should happen)

While not fully ready to be integrated into the main app yet, this should be sufficient to start implementing/migrating handlers and adding unit-tests for them before we make the switch eventually.


src/rpc/common/Validators.h
```diff
@@ -0,0 +1,247 @@
...
+ * @brief Check that the type is the same as what was expected
+ *
+ * @tparam Expected The expected type that value should be convertible to
+ * @param value The json value to check the type of
+ * @return true if convertible; false otherwise
+ */
+template <typename Expected>
+[[nodiscard]] bool static checkType(boost::json::value const& value)
+{
+    auto has_error = false;
```
cindyyan317: I saw we are using camel case for local variable in some places. Do we have a style rule?

src/rpc/common/Validators.h
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    [[nodiscard]] MaybeError
+    verify(boost::json::value const& value, std::string_view key) const;
+};
+
+/**
+ * @brief A validator that simply requires a field to be present
+ */
+struct required final
```
cindyyan317: "required" is a keyword in C++20. maybe we shall avoid using it as struct name. https://en.cppreference.com/w/cpp/keyword/requires

src/rpc/common/Validators.cpp
```diff
@@ -0,0 +1,89 @@
...
+    return {};
+}
+
+[[nodiscard]] MaybeError
+required::verify(boost::json::value const& value, std::string_view key) const
+{
+    if (not value.is_object() or not value.as_object().contains(key.data()))
+        return Error{RPC::Status{
+            RPC::RippledError::rpcINVALID_PARAMS,
+            "Required field '" + std::string{key} + "' missing"}};
```
cindyyan317, line 50: I like user friendly error info like this. So the caller can log it or print it out. Can we have more detailed info elsewhere

src/rpc/common/Validators.h
godexsoft: Good catch! my personal preference is snake case and i still sometimes forget that `clio` uses camel 👍

src/rpc/common/Validators.h
godexsoft: "required" is not a keyword, "requires" is. But you got a point there - they are pretty similar. I was thinking whether we should stick to our convention of `CamelCase` for structs/classes in this case too. What do you think? it would make the spec code look a little different tho.
Another thing I had in mind is, instead of `verify` we could just use `operator()()` and that should allow us to instantly use a lambda or any callable as a `Requirement`. I think. But we can also leave that for the next iteration - clearly this code will still be evolving in the near future.

src/rpc/common/Validators.cpp
godexsoft: That was my intention. I removed it for brevity for the time being and only left it in this one place as an example of what is now possible. We should definitely be adding custom messages like this one in all validators in the future but that means we need to agree with integration tests that the message is not going to match `rippled` in most cases.

src/util/Expected.h
```diff
@@ -0,0 +1,264 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+    Copyright (c) 2021 Ripple Labs Inc.
```
cindyyan317, line 5: delete this line?

godexsoft: That would violate `rippled`s requirement to keep this notice when copying the code. I added this line on purpose :)

src/util/Expected.h
```diff
@@ -0,0 +1,264 @@
...
+    E val_;
+};
+
+// Unexpected deduction guide that converts array to const*.
+template <typename E, std::size_t N>
+Unexpected(E (&)[N]) -> Unexpected<E const*>;
+
+// Definition of Expected.  All of the machinery comes from boost::result.
+template <class T, class E>
+class [[nodiscard]] Expected
```
cindyyan317, line 141: This is a nodiscard class. Can we omit the extra [[nodiscard]] at usage side?

src/rpc/common/Validators.h
cindyyan317: `verify` is more readable

CMakeLists.txt
```diff
@@ -97,12 +100,16 @@ target_link_libraries(clio_server PUBLIC clio)
 if(BUILD_TESTS)
   set(TEST_TARGET clio_tests)
   add_executable(${TEST_TARGET}
-  unittests/RPCErrors.cpp
-  unittests/Backend.cpp
-  unittests/Logger.cpp
-  unittests/Config.cpp
-  unittests/ProfilerTest.cpp
-  unittests/DOSGuard.cpp)
+    unittests/Playground.cpp
```
cindyyan317, line 103: don't forget this one

cindyyan317:
Look forward to using the framework!
LGTM
Will approve after removing the playground.cpp

CMakeLists.txt
godexsoft: i did not forget it - playground.cpp as it is now should be in the project (otherwise you would have to write your own each time). the point is not to commit the tmp stuff into it later on :)

src/util/Expected.h
godexsoft: Hm, good point. I'll check where we can drop it in user code.

src/rpc/common/Validators.h
godexsoft: that is true. let's leave as is for now at least.

cindyyan317 (approved):

cindyyan317 (dismissed):

PR #488, opened 2023-01-26
Add unittests for subscription

cindyyan317:
1 Update subscription , make it to be testable
2 Add test
3 Also add some util functions to create test objects.

Unittest coverage -> 100%
Screenshot 2023-01-27 at 13 50 32

### DIFF
--- a/src/subscriptions/SubscriptionManager.cpp
+++ b/src/subscriptions/SubscriptionManager.cpp
@@ -21,57 +21,6 @@
 #include <subscriptions/SubscriptionManager.h>
 #include <webserver/WsBase.h>
 
-template <class T>
-inline void
-sendToSubscribers(
-    std::shared_ptr<Message> const& message,
-    T& subscribers,
-    std::atomic_uint64_t& counter)
-{
-    for (auto it = subscribers.begin(); it != subscribers.end();)
-    {
-        auto& session = *it;
-        if (session->dead())
-        {
-            it = subscribers.erase(it);
-            --counter;
-        }
-        else
-        {
-            session->send(message);
-            ++it;
-        }
-    }
-}
-
-template <class T>
-inline void
-addSession(
-    std::shared_ptr<WsBase> session,
-    T& subscribers,
-    std::atomic_uint64_t& counter)
-{
-    if (!subscribers.contains(session))
-    {
-        subscribers.insert(session);
-        ++counter;
-    }
-}
-
-template <class T>
-inline void
-removeSession(
-    std::shared_ptr<WsBase> session,
-    T& subscribers,
-    std::atomic_uint64_t& counter)
-{
-    if (subscribers.contains(session))
-    {
-        subscribers.erase(session);
-        --counter;
-    }
-}
-
 void
 Subscription::subscribe(std::shared_ptr<WsBase> const& session)
 {
@@ -93,55 +42,6 @@ Subscription::publish(std::shared_ptr<Message> const& message)
 {
     boost::asio::post(strand_, [this, message]() {
         sendToSubscribers(message, subscribers_, subCount_);
-    });
-}
-
-template <class Key>
-void
-SubscriptionMap<Key>::subscribe(
-    std::shared_ptr<WsBase> const& session,
-    Key const& account)
-{
-    boost::asio::post(strand_, [this, session, account]() {
-        addSession(session, subscribers_[account], subCount_);
-    });
-}
-
-template <class Key>
-void
-SubscriptionMap<Key>::unsubscribe(
-    std::shared_ptr<WsBase> const& session,
-    Key const& account)
-{
-    boost::asio::post(strand_, [this, account, session]() {
-        if (!subscribers_.contains(account))
-            return;
-
-        if (!subscribers_[account].contains(session))
-            return;
-
-        --subCount_;
-
-        subscribers_[account].erase(session);
-
-        if (subscribers_[account].size() == 0)
-        {
-            subscribers_.erase(account);
-        }
-    });
-}
-
-template <class Key>
-void
-SubscriptionMap<Key>::publish(
-    std::shared_ptr<Message>& message,
-    Key const& account)
-{
-    boost::asio::post(strand_, [this, account, message]() {
-        if (!subscribers_.contains(account))
-            return;
-
-        sendToSubscribers(message, subscribers_[account], subCount_);
     });
 }
 
@@ -345,8 +245,6 @@ SubscriptionManager::pubTransaction(
 
     for (auto const& node : meta->getNodes())
     {
-        if (!node.isFieldPresent(ripple::sfLedgerEntryType))
-            assert(false);
         if (node.getFieldU16(ripple::sfLedgerEntryType) == ripple::ltOFFER)
         {
             ripple::SField const* field = nullptr;
@@ -388,9 +286,6 @@ SubscriptionManager::pubBookChanges(
     ripple::LedgerInfo const& lgrInfo,
     std::vector<Backend::TransactionAndMetadata> const& transactions)
 {
-    if (bookChangesSubscribers_.empty())
-        return;
-
     auto const json = RPC::computeBookChanges(lgrInfo, transactions);
     auto const bookChangesMsg =
         std::make_shared<Message>(boost::json::serialize(json));
--- a/unittests/SubscriptionManagerTest.cpp
+++ b/unittests/SubscriptionManagerTest.cpp
@@ -1,0 +1,880 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <subscriptions/SubscriptionManager.h>
+#include <util/Fixtures.h>
+#include <util/MockBackend.h>
+#include <util/MockWsBase.h>
+#include <util/TestObject.h>
+#include <webserver/WsBase.h>
+
+#include <boost/json/parse.hpp>
+#include <gmock/gmock.h>
+
+#include <chrono>
+
+using namespace std::chrono_literals;
+namespace json = boost::json;
+using namespace Backend;
+using ::testing::Return;
+
+// common const
+constexpr static auto CURRENCY = "0158415500000000C1F76FF6ECB0BAC600000000";
+constexpr static auto ISSUER = "rK9DrarGKnVEo2nYp5MfVRXRYf5yRX3mwD";
+constexpr static auto ACCOUNT1 = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn";
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+constexpr static auto LEDGERHASH =
+    "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+constexpr static auto LEDGERHASH2 =
+    "1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC";
+constexpr static auto TXNID =
+    "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC321";
+
+/*
+ *  test subscription factory method and report function
+ */
+TEST(SubscriptionManagerTest, InitAndReport)
+{
+    constexpr static auto ReportReturn = R"({
+        "ledger":0,
+        "transactions":0,
+        "transactions_proposed":0,
+        "manifests":0,
+        "validations":0,
+        "account":0,
+        "accounts_proposed":0,
+        "books":0,
+        "book_changes":0
+    })";
+    clio::Config cfg;
+    auto backend = std::make_shared<MockBackend>(cfg);
+    auto subManager =
+        SubscriptionManager::make_SubscriptionManager(cfg, backend);
+    EXPECT_EQ(subManager->report(), json::parse(ReportReturn));
+}
+
+void
+CheckSubscriberMessage(
+    std::string out,
+    std::shared_ptr<WsBase> session,
+    int retry = 10)
+{
+    auto sessionPtr = static_cast<MockSession*>(session.get());
+    while (retry-- != 0)
+    {
+        std::this_thread::sleep_for(20ms);
+        if ((!sessionPtr->message.empty()) &&
+            json::parse(sessionPtr->message) == json::parse(out))
+        {
+            return;
+        }
+    }
+    EXPECT_TRUE(false) << "Could not wait the subscriber message, expect:"
+                       << out << " Get:" << sessionPtr->message;
+}
+
+// Fixture contains test target and mock backend
+class SubscriptionManagerSimpleBackendTest : public MockBackendTest
+{
+protected:
+    clio::Config cfg;
+    std::shared_ptr<SubscriptionManager> subManagerPtr;
+    util::TagDecoratorFactory tagDecoratorFactory{cfg};
+    std::shared_ptr<WsBase> session;
+    void
+    SetUp() override
+    {
+        MockBackendTest::SetUp();
+        subManagerPtr =
+            SubscriptionManager::make_SubscriptionManager(cfg, mockBackendPtr);
+        session = std::make_shared<MockSession>(tagDecoratorFactory);
+    }
+    void
+    TearDown() override
+    {
+        MockBackendTest::TearDown();
+        subManagerPtr.reset();
+    }
+};
+
+/*
+ * test report function and unsub functions
+ */
+TEST_F(SubscriptionManagerSimpleBackendTest, ReportCurrentSubscriber)
+{
+    constexpr static auto ReportReturn = R"({
+        "ledger":0,
+        "transactions":2,
+        "transactions_proposed":2,
+        "manifests":2,
+        "validations":2,
+        "account":2,
+        "accounts_proposed":2,
+        "books":2,
+        "book_changes":2
+    })";
+    std::shared_ptr<WsBase> session1 =
+        std::make_shared<MockSession>(tagDecoratorFactory);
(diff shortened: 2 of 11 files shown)


src/subscriptions/SubscriptionManager.cpp
```diff
@@ -21,57 +21,6 @@
 #include <subscriptions/SubscriptionManager.h>
 #include <webserver/WsBase.h>
 
-template <class T>
```
cindyyan317, line 24: These functions are moved to header file to be tested.

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -345,8 +245,6 @@ SubscriptionManager::pubTransaction(
 
     for (auto const& node : meta->getNodes())
     {
-        if (!node.isFieldPresent(ripple::sfLedgerEntryType))
```
cindyyan317, line 348: line 294 deserializeTxPlusMeta already check this. This line could not be reached

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -388,9 +286,6 @@ SubscriptionManager::pubBookChanges(
     ripple::LedgerInfo const& lgrInfo,
     std::vector<Backend::TransactionAndMetadata> const& transactions)
 {
-    if (bookChangesSubscribers_.empty())
```
cindyyan317, line 391: We should not check the subscribers here. The sub method run asynchronously. If a message is published just after subscribing, the session will be ignored.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
godexsoft: minor - it's 2023 already :) here and everywhere else where this header was added.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <boost/json/parse.hpp>
```
godexsoft: minor - we try to group and order the includes:
1st group - local files (in this case anything in unittest/*)
2nd group - 3rd party libraries like boost/gmock etc.
3rd group - std stuff like 'chrono'

here and everywhere else your changes affect :)

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+constexpr static auto LEDGERHASH =
+    "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+constexpr static auto LEDGERHASH2 =
+    "1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC";
+constexpr static auto TXNID =
+    "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC321";
+
+/**
+ *  test subscription factory method and report function
```
godexsoft, line 50: anything starting with `/**` ends up in generated documentation (doxygen). i'd avoid documenting unit tests (consider using regular comments instead `/*`)

for brevity i'm not going to repeat this comment for every instance - please fix wherever you added this and you recognize it should not be there 🥇

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+ * */
+TEST(SubscriptionManagerTest, InitAndReport)
+{
+    constexpr static auto ReportReturn = R"JSON(
+            {
```
godexsoft: minor - formatting could be nicer here. i'd put `{` and `}` next to the opening and closing parens but it's up to you.

here and everywhere else (all formatting of json seems a bit random).

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+            "transactions_proposed":0,
+            "manifests":0,
+            "validations":0,
+            "account":0,
+            "accounts_proposed":0,
+            "books":0,
+            "book_changes":0
+            }
+    )JSON";
+    clio::Config cfg{json::parse(R"JSON({})JSON")};
```
godexsoft: minor - here there is not much point in using `R"JSON( )JSON"` since you are not actually multilining nor using `"` within the JSON. 
other than that, it's possible to use just `R"({ "some" : "json" })"` and the result would be the same.
again, up to you.

godexsoft: and for default config you could just do `clio::Config cfg;`

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+                       << out << " Get:" << session->message;
+}
+
+// Fixture contains test target and mock backend
+class SubscriptionManagerSimpleBackendTest : public NoLoggerFixture
+{
+protected:
+    clio::Config cfg{json::parse(R"JSON({})JSON")};
+    std::shared_ptr<SubscriptionManager> subManagerPtr;
+    std::shared_ptr<BackendInterface> mockBackendPtr;
```
godexsoft: It may be worth extracting `MockBackendFixture` and inherit from it instead. I feel like this is going to be useful in a whole lot of places soon! wdyt? I'd put it with other fixtures we have.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    auto account = GetAccountIDWithString(ACCOUNT1);
+    subManagerPtr->subAccount(account, session1);
+    subManagerPtr->subAccount(account, session2);
+    subManagerPtr->subProposedAccount(account, session1);
+    subManagerPtr->subProposedAccount(account, session2);
+    auto issue1 = GetIssue(CURRENCY, ISSUER);
+    ripple::Book book{ripple::xrpIssue(), issue1};
+    subManagerPtr->subBook(book, session1);
+    subManagerPtr->subBook(book, session2);
+    std::this_thread::sleep_for(20ms);
```
godexsoft, line 155: ideally we should not do anything with threads if that's possible. if it's too much refactoring to support that then let's leave it for now and get back to this later.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    subManagerPtr->unsubManifest(session1);
+    subManagerPtr->unsubProposedTransactions(session1);
+    subManagerPtr->unsubTransactions(session1);
+    subManagerPtr->unsubValidation(session1);
+    subManagerPtr->unsubAccount(account, session1);
+    subManagerPtr->unsubProposedAccount(account, session1);
+    subManagerPtr->unsubBook(book, session1);
+    std::this_thread::sleep_for(20ms);
+    auto report = subManagerPtr->report();
+    EXPECT_EQ(report["book_changes"], 1);
```
godexsoft: this block of checks is the same as below and really the same as `ReportReturn`. consider making a helper function/lambda that you give one `int` and it spits out a `json::object` that you can just compare to `report` here with one call.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    EXPECT_EQ(report["transactions"], 0);
+    EXPECT_EQ(report["manifests"], 0);
+    EXPECT_EQ(report["accounts_proposed"], 0);
+    EXPECT_EQ(report["account"], 0);
+    EXPECT_EQ(report["books"], 0);
+}
+
+TEST_F(SubscriptionManagerSimpleBackendTest, SubscriptionManagerLedgerUnSub)
+{
+    MockBackend* rawBackendPtr = (MockBackend*)mockBackendPtr.get();
```
godexsoft: i bet there is a way to avoid both raw pointer and C-style casting or at least the latter. check `EXPECT_CALL` and `ON_CALL` docs please 🥇 

there is many instances of the same issue in the proposed changes. ideally we should get rid of all of them. 
if raw pointer is unavoidable, let's at least not use c-style casts.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    std::shared_ptr<WsBase> session =
+        std::make_shared<MockSession>(tagDecoratorFactory);
+    const uint32_t rangeMin = 10;
```
godexsoft: consider removing if we are not reusing these constants - can just use the values straight away (10u, 30u), possibly with comment like `// min` and `// max`

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    const uint32_t rangeMax = 30;
+    mockBackendPtr->updateRange(rangeMin);
+    mockBackendPtr->updateRange(rangeMax);
+    boost::asio::io_context ctx;
+    auto ledgerinfo = CreateLedgerInfo(LEDGERHASH, 30);
+    ripple::LedgerInfo();
```
godexsoft: what is this? do we need it?
also same thing in a few more places.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+    // mock fetchLedgerBySequence return this ledger
+    ON_CALL(*rawBackendPtr, fetchLedgerBySequence)
+        .WillByDefault(Return(ledgerinfo));
+    EXPECT_CALL(*rawBackendPtr, fetchLedgerBySequence).Times(1);
+    // mock doFetchLedgerObject return fee setting ledger object
+    auto feeBlob = CreateFeeSettingBlob(1, 2, 3, 4, 0);
+    ON_CALL(*rawBackendPtr, doFetchLedgerObject).WillByDefault(Return(feeBlob));
+    EXPECT_CALL(*rawBackendPtr, doFetchLedgerObject).Times(1);
+    boost::asio::spawn(ctx, [this, &session](boost::asio::yield_context yield) {
+        subManagerPtr->subLedger(yield, session);
```
godexsoft, line 200: oof.. we should really refactor this to get rid of yield in the api of submanager.. but i understand it's a lot of work rn so let's leave it for later :(

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+ * we need to construct a valid account in the transaction
+ * this test subscribe the proposed transaction for two accounts
+ * but only forward a transaction with one of them
+ * check the correct session is called
+ */
+TEST_F(
+    SubscriptionManagerSimpleBackendTest,
+    SubscriptionManagerAccountProposedTransaction)
+{
+    std::shared_ptr<WsBase> session =
```
godexsoft: seems like we have a bunch of tests that start like this - consider refactoring into a separate fixture and suite.

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+ * test ledger stream
+ * check 1 subscribe response, 2 publish message
+ * mock backend to return fee ledger object
+ */
+TEST_F(SubscriptionManagerSimpleBackendTest, SubscriptionManagerLedger)
+{
+    MockBackend* rawBackendPtr = (MockBackend*)mockBackendPtr.get();
+    std::shared_ptr<WsBase> session =
+        std::make_shared<MockSession>(tagDecoratorFactory);
+    const uint32_t rangeMin = 10;
```
godexsoft: same as before

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+        std::make_shared<MockSession>(tagDecoratorFactory);
+    auto account = GetAccountIDWithString(ACCOUNT1);
+    subManagerPtr->subProposedAccount(account, session);
+
+    std::shared_ptr<WsBase> sessionIdle =
+        std::make_shared<MockSession>(tagDecoratorFactory);
+    auto accountIdle = GetAccountIDWithString(ACCOUNT2);
+    subManagerPtr->subProposedAccount(accountIdle, sessionIdle);
+
+    auto dummyTransaction =
```
godexsoft: i liked the constexpr static. consider adding for consistency :)

unittests/SubscriptionTest.cpp
```diff
@@ -0,0 +1,203 @@
...
+protected:
+    clio::Config cfg{json::parse(R"JSON({})JSON")};
+    util::TagDecoratorFactory tagDecoratorFactory{cfg};
+};
+
+// subscribe/unsubscribe the same session would not change the count
+TEST_F(SubscriptionTest, SubscriptionCount)
+{
+    Subscription sub(ctx);
+    std::shared_ptr<WsBase> session1(new MockSession(tagDecoratorFactory));
```
godexsoft: let's use std::make_shared here and all other instances of `new` :)

unittests/SubscriptionTest.cpp
```diff
@@ -0,0 +1,203 @@
...
+    ctx.run();
+    EXPECT_EQ(sub.count(), 0);
+}
+
+TEST_F(SubscriptionTest, SubscriptionMapCount)
+{
+    std::shared_ptr<WsBase> session1(new MockSession(tagDecoratorFactory));
+    std::shared_ptr<WsBase> session2(new MockSession(tagDecoratorFactory));
+    std::shared_ptr<WsBase> session3(new MockSession(tagDecoratorFactory));
+    SubscriptionMap<std::string> subMap(ctx);
```
godexsoft, line 136: consider refactoring to a separate suite, `SubscriptionMapTest`

unittests/util/TestObject.h
```diff
@@ -0,0 +1,135 @@
...
+ * Create AccountID object with string
+ */
+[[nodiscard]] ripple::AccountID
+GetAccountIDWithString(std::string id);
+
+/**
+ * Create a simple ledgerInfo object with only hash and seq
+ */
+[[nodiscard]] ripple::LedgerInfo
+CreateLedgerInfo(const char* ledgerHash, ripple::LedgerIndex seq);
```
godexsoft: if using `string_view` is an option, we should prefer it over `const char*`. please consider here and everywhere else 👍

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,961 @@
...
+/**
+ * test Manifest
+ * Subscription Manager forward the manifest message to subscribers
+ */
+TEST_F(SubscriptionManagerSimpleBackendTest, SubscriptionManagerManifestTest)
+{
+    std::shared_ptr<WsBase> session =
+        std::make_shared<MockSession>(tagDecoratorFactory);
+    subManagerPtr->subManifest(session);
+    constexpr static auto dummyManifest = R"JSON(
```
godexsoft: for simple json like this, consider using one liners to save space.

unittests/util/MockWsBase.h
```diff
@@ -0,0 +1,50 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <webserver/WsBase.h>
+
+class MockSession : public WsBase
```
godexsoft: i guess this can be a struct if everything is to stay public? same below.

godexsoft (changes requested):
Great job getting so many tests in!
Let's try to fix some simple issues that i noted. 
Some of them may be harder than others and can be left for next iteration/sprint instead.

unittests/SubscriptionManagerTest.cpp
cindyyan317: i think it is done automatically in my vs code~ i will try to switch it off

cindyyan317: we have "SortIncludes:    true" cfg in our clang-format setting. let us turn it off in other PR

unittests/SubscriptionManagerTest.cpp
cindyyan317: yea, my plan is getting it tested first, then remove this part.

unittests/SubscriptionManagerTest.cpp
cindyyan317: use rangeMin, rangeMax is more expressive to set the range. the API we call is the same one "updaterange"

unittests/SubscriptionManagerTest.cpp
cindyyan317: forgot to delete it :(

unittests/SubscriptionManagerTest.cpp
cindyyan317: no , expect_call and on_call receive obj http://google.github.io/googletest/reference/mocking.html

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,880 @@
...
+    },
+    "type":"transaction",
+    "validated":true,
+    "status":"closed",
+    "ledger_index":33,
+    "ledger_hash":"1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC",
+    "engine_result_code":0,
+    "engine_result":"tesSUCCESS",
+    "engine_result_message":"The transaction was applied. Only final in a validated ledger."
+    })";
```
godexsoft: minor: this tab could be removed

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,880 @@
...
+        ISSUER, ripple::lsfGlobalFreeze, 1, 10, 2, TXNID, 3);
+    ON_CALL(*rawBackendPtr, doFetchLedgerObject(kk, testing::_, testing::_))
+        .WillByDefault(Return(accountRoot.getSerializer().peekData()));
+    subManagerPtr->pubTransaction(trans1, ledgerinfo);
+    CheckSubscriberMessage(TransactionForOwnerFundFrozen, session);
+}
+
+/*
+ * test subscribe account
+ * */
```
godexsoft: minor: looks like a different format of comments to what we use normally (compared to your other comments in this file)

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,880 @@
...
+    "validated":true,
+    "status":"closed",
+    "ledger_index":33,
+    "ledger_hash":"1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC",
+    "engine_result_code":0,
+    "engine_result":"tesSUCCESS",
+    "engine_result_message":"The transaction was applied. Only final in a validated ledger."
+    })";
+
+/**
```
godexsoft: minor: still a doxygen comment here

unittests/SubscriptionManagerTest.cpp
```diff
@@ -0,0 +1,880 @@
...
+        "engine_result":"tesSUCCESS",
+        "engine_result_message":"The transaction was applied. Only final in a validated ledger."
+    })";
+    CheckSubscriberMessage(AccountPublish, session);
+}
+
+/*
+ * test subscribe order book
+ * Create/Delete/Update offer node will trigger publish
+ * */
```
godexsoft: minor: strange comment format again

godexsoft (commented):
Found just a few minor things that you missed i think. LGTM in general :)

godexsoft (dismissed):
LGTM! 🚀

godexsoft (approved):

Issue #244, opened 2022-08-11
support ip ranges in whitelist


cjcobb23:

godexsoft:
It would be useful to provide a description what actually needs to be done here.
From what we can tell it's probably something like this:

```
"dos_guard":
{
    "whitelist":["192.168.0.*", "192.168.1.[1-5]"]
}
```

I imagine this can only be really useful for some local network where the router assigns IP addresses using DHCP (e.g. allow all users on local network to access `clio` running on some machine in the network).
@cjcobb23 if you have a minute, please elaborate on what kind of format you had in mind for this. Thanks

injaelee:
Let's support CIDR notations for the whitelist. There are many references to whitelisting IP addresses in many frameworks / platforms. Here's one reference. (cc: @godexsoft @officialfrancismendoza)
- [EC2 Network Access](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/authorizing-access-to-an-instance.html)

officialfrancismendoza:
@injaelee I was assuming classless IPv4 + CIDR based on your comment in the slack. That said, in the process of making additional unit tests to handle the following cases: [reserve IPv4 Addresses](https://www.tutorialspoint.com/ipv4/ipv4_reserved_addresses.htm), [no continuous 0's and 1's for host addresses](https://community.cisco.com/t5/switching/valid-host-address/td-p/1413517), etc.
![image](https://user-images.githubusercontent.com/47958858/216129719-09120caf-1aa4-4181-84e7-20a46ddc2de2.png)

PR #489, opened 2023-01-27
Add support for IP Ranges in DOSGuard Whitelist

officialfrancismendoza:
**Problem (#244):** Clio DOSGuard only takes in explicit IP addresses in whitelist, does not support ranges
**Fix**: Add parsing functionality to generate IP addresses for ranges delimited by [number1-number2] in any byte

### DIFF
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -204,7 +204,6 @@ public:
         std::string const& ip,
         std::string const& subnetIPwCIDR) const
     {
-        // [!!!] TODO: Check and see if restoring condition helps
         // ip being checked does not have CIDR
         if (subnetIPwCIDR.find("/") != std::string::npos &&
             checkParsedIPv4Address(ip) == true)
@@ -252,7 +251,6 @@ public:
         std::string const& ip,
         std::string const& subnetIPwCIDR) const
     {
-        // [!!!] TODO: Check and see if restoring condition helps
         // ip being checked does not have CIDR
         if (subnetIPwCIDR.find("/") != std::string::npos &&
             checkParsedIPv6Address(ip) == true)
@@ -379,18 +377,19 @@ public:
     checkIfInSubnetMixed(std::string const& ip, std::string const& subnetIP)
         const
     {
-        // [!!!] TODO: Restore extra checks for subnetIP
         // Check that both addresses are IPv4
-        if (ip.find(".") != std::string::npos)
+        if (ip.find(".") != std::string::npos &&
+            subnetIP.find(".") != std::string::npos)
         {
             return isIPv4AddressInSubnet(ip, subnetIP);
         }
 
-        // [!!!] TODO: Restore extra checks for subnetIP
         // Check that both addresses are IPv6
         else if (
-            ip.find(":") != std::string::npos ||
-            ip.find("::") != std::string::npos)
+            (ip.find(":") != std::string::npos ||
+             ip.find("::") != std::string::npos) &&
+            (subnetIP.find(":") != std::string::npos ||
+             subnetIP.find("::") != std::string::npos))
         {
             return isIPv6AddressInSubnet(ip, subnetIP);
         }
@@ -417,24 +416,6 @@ public:
     }
 
     /**
-     * @brief Checks if IP address is malformed
-     *
-     */
-    [[nodiscard]] bool
-    isValidIPMixed(std::string const& ip) const
-    {
-        try
-        {
-            boost::asio::ip::address::from_string(ip);
-            return true;
-        }
-        catch (const std::exception& e)
-        {
-            return false;
-        }
-    }
-
-    /**
      * @brief Check whether an ip address is in the whitelist or not
      *
      * @param ip The ip address to check (raw IP, NO CIDR)
@@ -451,8 +432,14 @@ public:
         }
 
         // CASE #2: Check if IP has malformed input.
-        if (!isValidIPMixed(ip))
+        boost::system::error_code error_code;
+        boost::asio::ip::address addr =
+            boost::asio::ip::make_address(ip, error_code);
+
+        // Neither a valid IPv4 nor a valid IPv6 address
+        if (error_code)
         {
+            // throw std::invalid_argument(error_code.message());
             return false;
         }
 
--- a/src/webserver/DOSGuard.h
+++ b/src/webserver/DOSGuard.h
@@ -29,6 +29,10 @@
 
 #include <config/Config.h>
 
+#include <iostream>
+#include <string>
+#include <vector>
+
 namespace clio {
 
 class BaseDOSGuard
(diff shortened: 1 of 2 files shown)


officialfrancismendoza:
@cindyyan317 @godexsoft don't review this just yet, just generated this so I can atomically roll back + track rebase history as this is developed, as there will be fair amount of code churn from upcoming dev sprint

officialfrancismendoza:
**Note**: Additional details are required to determine how far we would like to go with IP subnet functionality. Do we also generate subnet masks, network masks, broadcast IP, etc? 

Additional unit tests and checks will be appended to handle the following cases: [reserve IPv4 Addresses](https://www.tutorialspoint.com/ipv4/ipv4_reserved_addresses.htm), [no continuous 0's and 1's for host addresses](https://community.cisco.com/t5/switching/valid-host-address/td-p/1413517), etc.

cindyyan317:
> **Note**: Additional details are required to determine how far we would like to go with IP subnet functionality. Do we also generate subnet masks, network masks, broadcast IP, etc?
> 
> Additional unit tests and checks will be appended to handle the following cases: [reserve IPv4 Addresses](https://www.tutorialspoint.com/ipv4/ipv4_reserved_addresses.htm), [no continuous 0's and 1's for host addresses](https://community.cisco.com/t5/switching/valid-host-address/td-p/1413517), etc.
@injaelee @godexsoft

officialfrancismendoza:
> **Note**: Injae requested support for IPv6 be added as well, with @godexsoft specifying shared whitelist for IPv4 and IPv6

officialfrancismendoza:
@injaelee @godexsoft @cindyyan317  as part of new SOPs, issues with associated notes and accounting for churn on origin/develop will be tracked as a **draft** first

officialfrancismendoza:
**Note**: IPV4 CIDR check incorrect so far. The remainder gets applied to ALL BYTES if less than 8 whereas the rest should be "0" after the first one. Meaning, in our previous results, misled with IPv4 and IPv6 test cases. Currently, failing only for IPv4. Heavy utilization of boost::asio::ip library, but currently no function exists for whether valid CIDR for given IP address

**Two critical bugs caught:**
- [ X] Malformed generation of network address (solution: pre=populate with "zero" address to eliminate possibility of population by garbage values)
- [ X] Malformed generation of subnet mask (solution: additional else condition to ensure bitwise shift only occurs once when remaining CIDR bits less than 8 from counter)

Questions
- [ ] Verify assumption below of determining if IP address in a subnet
- [ ] Verify if correct in assuming IP address w. CIDR split is evaluated as the following: see if IP address fits into network address with the same CIDR (ie: Does 10.3.255.254 fit in 10.0.0.0/24)

Operation to determine if in subnet: determine if bitwise_and(extracted_ip, subnet_mask) == bitwise_and(network_address, subnet_mask). 

UPDATE: Bugs resolved and fully functioning as of 13FEB2023

Screenshots of malformed subnet mask and network address generation (now resolved)
image
image

officialfrancismendoza:
Test

unittests/DOSGuard.cpp
```diff
@@ -81,14 +93,75 @@ class DOSGuardTest : public NoLoggerFixture
...
 };
 
 TEST_F(DOSGuardTest, Whitelisting)
 {
+    // Test cases for individual IPv4 address
     EXPECT_TRUE(guard.isWhiteListed("127.0.0.1"));
     EXPECT_FALSE(guard.isWhiteListed(IP));
+
+    // Test cases for list of IPv4 addresses
+    EXPECT_TRUE(guard_mixed.isWhiteListed("127.0.0.1/24"));
```
godexsoft: Looks like you did not fully understand the purpose of the whitelist and what it means to check whether an IP is within a subnet or not.

Here is a test that should be passing and is currently failing:

```cpp
// assume "whitelist": ["192.168.0.0/24"]
EXPECT_TRUE(guard_mixed.isWhiteListed("192.168.0.1"));
EXPECT_FALSE(guard_mixed.isWhiteListed("192.168.1.1"));
```

I believe I told you before that `isWhiteListed` can never get a mask, only a real IP, hence this test makes no sense.
No need to review the rest of the code for now because it fundamentally does not do what it should.

godexsoft (changes requested):

unittests/DOSGuard.cpp
officialfrancismendoza: @godexsoft I went off an incorrect assumption regarding intended scope of isWhitelisted. Functionality for checking subnets in different methods and assumed those to be public, but will integrate that functionality + testing live into the isWhitelisted (now clarified as the **only** public method to users) and getWhitelist methods in the next iteration. Standby

officialfrancismendoza:
@cindyyan317 the core logic in the latest push of the PR is ready for review

src/webserver/DOSGuard.h
```diff
@@ -84,17 +415,65 @@ class BasicDOSGuard : public BaseDOSGuard
...
      * @return true
-     * @return false
+     * @return falseconvertedIPsToStrings
      */
     [[nodiscard]] bool
     isWhiteListed(std::string const& ip) const noexcept
     {
-        return whitelist_.contains(ip);
+        // CASE #1: Check if IP is CIDR. Treat as malformed input
+        if (ip.find("/") != std::string::npos)
```
cindyyan317, line 448: A bit redundant in logic. The IP is from the connection, we can be sure the IP is valid, right? 
The two if in for loop can be merged to one, try to use std::any_of.

src/webserver/DOSGuard.h
```diff
@@ -84,17 +415,65 @@ class BasicDOSGuard : public BaseDOSGuard
         sweepHandler.setup(this);
     }
 
+    /**
+     * @brief Checks if IP address is malformed
+     *
+     * @param ip IP address to check
+     */
+    [[nodiscard]] bool
+    isValidIPMixed(std::string const& ip) const
```
cindyyan317, line 424: This function can be removed. We don't need to call it in isWhiteListed

src/webserver/DOSGuard.h
```diff
@@ -69,6 +73,333 @@ class BasicDOSGuard : public BaseDOSGuard
...
+    checkValidityOfWhitelist(std::string const& ip) const
+    {
+        if (ip.find(".") != std::string::npos)
+        {
+            return isValidIPv4(ip);
+        }
+
+        else if (
+            ip.find(":") != std::string::npos ||
+            ip.find("::") != std::string::npos)
```
cindyyan317, line 358: If it can not find one ":", can it find two ":"?

src/webserver/DOSGuard.h
```diff
@@ -69,6 +73,333 @@ class BasicDOSGuard : public BaseDOSGuard
...
+     * @brief Checks if raw IPv4 address is valid
+     *
+     * @param ip Raw IPv4 address to check
+     */
+    [[nodiscard]] bool
+    checkParsedIPv4Address(std::string const& ip) const
+    {
+        try
+        {
+            boost::asio::ip::address_v4::from_string(ip);
```
cindyyan317, line 86: 1 .from_string has an override which can give you error_code instead of exception. 
2. Does this function need to be public? Same question is worthy to ask yourself for other functions.

src/webserver/DOSGuard.h
```diff
@@ -69,6 +73,333 @@ class BasicDOSGuard : public BaseDOSGuard
...
+        }
+    }
+
+    /**
+     * @brief Checks if IPv4 address has a valid CIDR notation
+     *
+     * @param ip The IP address (IPv4)
+     */
+    [[nodiscard]] bool
+    checkCIDRIPv4(std::string const& ip) const
```
cindyyan317, line 122: to validate a valid subnet , i think we can call "make_network_v4". then you can hold a subnet or subnet hosts in memory. When you need to check, you don't need to call make_network again. Imagine, when we receive a request, we want to do it timely.

src/webserver/DOSGuard.h
```diff
@@ -236,14 +615,75 @@ class BasicDOSGuard : public BaseDOSGuard
     [[nodiscard]] std::unordered_set<std::string> const
     getWhitelist(clio::Config const& config) const
     {
-        using T = std::unordered_set<std::string> const;
-        auto whitelist = config.arrayOr("dos_guard.whitelist", {});
+        // Data structures: string cast elements, complement, anwswer
```
cindyyan317, line 618: Since you have the function to check if it is a valid ip6/ip4 raw IP , if it is a valid ip6/ip4 subnet ip. I think you can use filter or copy_if to simplify the loop logic. 
`

unittests/DOSGuard.cpp
```diff
@@ -81,14 +97,99 @@ class DOSGuardTest : public NoLoggerFixture
...
+        "21DA:00D4:0000:2F4C:02BC:00FF:FE18:4C5A/64"));
+    EXPECT_TRUE(guard_mixed.checkValidityOfWhitelist(
+        "21DA:00D4::2F4C:02BC:00FF:FE18:4C5A/64"));
+    EXPECT_FALSE(
+        guard_mixed.checkValidityOfWhitelist("2001::eab:dead::a0:abcd:-1"));
+    EXPECT_FALSE(guard_mixed.checkValidityOfWhitelist(
+        "21DA:00D4:0000:2F4C:02BC:00FF:FE18:-1/64"));
+}
+
+TEST_F(DOSGuardTest, checkIfInSubnet)
```
cindyyan317, line 176: This test can be covered by testing isWhiteListed. If the internal functions are called by the other functions, we can just test the public functions.

unittests/DOSGuard.cpp
```diff
@@ -81,14 +97,99 @@ class DOSGuardTest : public NoLoggerFixture
...
+        guard_mixed.isWhiteListed("0000:0000:0EAB:DEAD:FFFF:FFFF:FFFF:FFFF"));
+
+    // Check that CIDR notation is not allowed
+    EXPECT_FALSE(guard_mixed.isWhiteListed("10.3.255.254/14"));
+    EXPECT_FALSE(guard_mixed.isWhiteListed("10.3.255.254/16"));
+    EXPECT_FALSE(guard_mixed.isWhiteListed("127.0.0.1/24"));
+    EXPECT_FALSE(guard_mixed.isWhiteListed("2001:0:eab:DEAD:0:A0:ABCD:4E/64"));
+}
+
+TEST_F(DOSGuardTest, checkIfIPNotMalformed)
```
cindyyan317, line 153: You can test against the getWhitelist function. Once your decide which functions are public , which functions are private, you will have more clear idea about how to test interface.

unittests/DOSGuard.cpp
```diff
@@ -54,6 +54,22 @@ constexpr static auto JSONData2 = R"JSON(
     }
 )JSON";
 
+// Note: Can have a mixed bag whitelist like the below
+// Whitelist contains raw IP not in any subnet, raw IP in a subnet, a subnet
+// viable with no raw IPs in whitelist, a subnet viable with a raw IP in a
+// whitelist
+constexpr static auto JSONData3 = R"JSON(
```
cindyyan317, line 61: You can reuse old config and existing guard. No need to create a new one.

cindyyan317:
I left a few of comments regarding the logic and structure. I bet @godexsoft will have more insight on the coding. His comments have been always useful. Let us wait for his comments and you can address the issues together.

src/webserver/DOSGuard.h
officialfrancismendoza: @cindyyan317 I had that to protect against the malformed input like ":::", which shouldn't exist and would pass iff I only left the former ":" condition. Unless the input ip can be guaranteed not to be malformed (which is true for isWhitelisted but not for loading the initial config in getWhitelist), an IPv6 address can only have ":" (canonical) or "::" (compressed)

src/webserver/DOSGuard.h
officialfrancismendoza: @cindyyan317 removing because of your assertion that input ip from isWhitelisted will never be malformed

godexsoft:
> I left a few of comments regarding the logic and structure. I bet @godexsoft will have more insight on the coding. His comments have been always useful. Let us wait for his comments and you can address the issues together.

Oh, just noticed this message. I'd like to see @officialfrancismendoza address your comments first and then I can do another pass.

PR #490, opened 2023-01-31
Run test on mac

cindyyan317:
Run unittest on mac .
Different behavior between gcc and clang could cause unittest result differs.
Also we remove macos11 node which has been failing for long.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,9 @@
 name: Build Clio
 on:
   push:
-    branches:  [master, release/*, develop, develop-next]
+    branches: [master, release/*, develop, develop-next]
   pull_request:
-    branches:  [master, release/*, develop, develop-next]
+    branches: [master, release/*, develop, develop-next]
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -150,57 +139,57 @@ jobs:
...
-        curl -s -OJL "https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/boost_1_77_0.tar.gz"
-        tar zxf boost_1_77_0.tar.gz
-        mv boost_1_77_0 boost
-        cd boost
-        ./bootstrap.sh
-        ./b2
-
-    - name: install deps
-      run: |
+      - uses: actions/checkout@v3
```
cindyyan317, line 142: format is done by yaml formatter

.github/workflows/build.yml
```diff
@@ -70,59 +70,48 @@ jobs:
...
+
+      - name: Build boost
+        if: steps.boost.outputs.cache-hit != 'true'
+        run: |
+          curl -s -OJL "https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/boost_1_77_0.tar.gz"
+          tar zxf boost_1_77_0.tar.gz
+          mv boost_1_77_0 boost
+          cd boost
+          ./bootstrap.sh
+          ./b2 cxxflags="-std=c++14"
```
godexsoft, line 98: I wonder, why not c++20 here? provided we are c++20 enabled in `clio`? idk if it makes a difference or not in the end but i imagine boost could have some extra code for newer standards.

.github/workflows/build.yml
godexsoft: yea, this is good. probably more correct format to start with.

godexsoft (approved):
LGTM! could experiment with c++20 for boost compilation but up to you.

.github/workflows/build.yml
cindyyan317: idk this. Let us confirm with @legleux

cindyyan317:
Fixes #490

PR #502, opened 2023-02-06
Fix backend test fail

cindyyan317:
We found cacheIntegration fails occasionally. Like this one https://github.com/XRPLF/clio/actions/runs/4076284584/jobs/7024125391
The case fails when comparing the result of fetchLedgerBySequence and fetchLedgerByHash. fetchLedgerByHash is to search the 'ledger_hashes' by hash to get sequence, then call fetchLedgerBySequence inside.
The problem is we could get wrong seq from ledger_hashes table. Because the generateNextLedger could generate the same hash for different ledger seq. generateNextLedger shuffles the same hash multiple times to get hash for 20 test targets.
From the error message, we also can tell the mismatch happens at the sequence number. We saw the parentHash is the same as well, that can be explained: because the previous hash is same, so the next parentHash is the same.
I did an experiment (mimic our unittest steps, the hash is also from unittest)to show the probability. 
```
int main()
{
  std::set<std::string> x;
  const char *const delim = "";
  std::vector<std::string> v = {"DB", "65", "EF", "4C", "7C", "FD", "F8", "FA", "6A", "82", "D9", "80", "13", "CE", "CD", "A2", "5D", "96", "6E", "52", "AC", "48", "58", "37", "91", "41", "C2", "59", "5A", "FA", "E2", "4E"};
  unsigned seed =
      std::chrono::system_clock::now().time_since_epoch().count();
  int times = 1000;
  while (times-- != 0)
  {
    std::srand(std::time(nullptr));
    std::shuffle(v.begin(), v.end(), std::default_random_engine(seed));
    std::ostringstream s;
    std::copy(v.begin(), v.end(), std::ostream_iterator<std::string>(s, delim));
    x.insert(s.str());
    std::cout << s.str() << std::endl;
  }
  std::cout << x.size() << std::endl;
}
```

shuffle is called 1000 times, but only generate 50-300 different result.
This diff adds a function to generate unique hash. I also add more log, in case it fails again, we have more information.

### DIFF
--- a/unittests/Backend.cpp
+++ b/unittests/Backend.cpp
@@ -802,19 +802,17 @@ TEST_F(BackendTest, Basic)
                 auto generateNextLedger = [seed](auto lgrInfo) {
                     ++lgrInfo.seq;
                     lgrInfo.parentHash = lgrInfo.hash;
-                    std::srand(std::time(nullptr));
+                    static auto randomEngine = std::default_random_engine(seed);
                     std::shuffle(
                         lgrInfo.txHash.begin(),
                         lgrInfo.txHash.end(),
-                        std::default_random_engine(seed));
+                        randomEngine);
                     std::shuffle(
                         lgrInfo.accountHash.begin(),
                         lgrInfo.accountHash.end(),
-                        std::default_random_engine(seed));
+                        randomEngine);
                     std::shuffle(
-                        lgrInfo.hash.begin(),
-                        lgrInfo.hash.end(),
-                        std::default_random_engine(seed));
+                        lgrInfo.hash.begin(), lgrInfo.hash.end(), randomEngine);
                     return lgrInfo;
                 };
                 auto writeLedger = [&](auto lgrInfo,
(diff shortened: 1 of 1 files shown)


unittests/Backend.cpp
```diff
@@ -2214,18 +2214,23 @@ TEST_F(BackendTest, cacheIntegration)
...
+                    // make sure the hash is unique
+                    auto uniqueHashGenerator =
+                        [seed](ripple::uint256& u256Hash) mutable {
+                            static std::unordered_set<std::string> hashSet;
+                            do
+                            {
+                                std::shuffle(
+                                    u256Hash.begin(),
+                                    u256Hash.end(),
+                                    std::default_random_engine(seed));
```
godexsoft: Let's not recreate the generator with the same seed everytime - this yields same sequence everytime. We should create it once and reuse.

unittests/Backend.cpp
```diff
@@ -2214,18 +2214,23 @@ TEST_F(BackendTest, cacheIntegration)
                     ++lgrInfo.seq;
                     lgrInfo.parentHash = lgrInfo.hash;
                     std::srand(std::time(nullptr));
```
godexsoft: This line can be removed - it's doing nothing at all when using random_engine with preset seed.

godexsoft (dismissed):
It's up to you if you want to spend another 1min to make those two lambdas into one function at the top of the file and reuse it :) not too important as this whole file is going to be deleted one day.

godexsoft (approved):
LGTM.

PR #503, opened 2023-02-06
Rpc: support rpc call without parameter

cindyyan317:
Some RPC call can have zero parameter.

### DIFF
--- a/src/rpc/common/Concepts.h
+++ b/src/rpc/common/Concepts.h
@@ -49,12 +49,21 @@ concept Requirement = requires(T a) {
  */
 // clang-format off
 template <typename T>
-concept Handler = requires(T a, typename T::Input in, typename T::Output out) {
+concept HandlerWithInput = requires(T a, typename T::Input in, typename T::Output out) {
     { a.spec() } -> std::same_as<RpcSpecConstRef>;
-    { a.process(in) } -> std::same_as<HandlerReturnType<decltype(out)>>;
-} 
-&& boost::json::has_value_from<typename T::Output>::value
-&& boost::json::has_value_to<typename T::Input>::value;
+    { a.process(in) } -> std::same_as<HandlerReturnType<decltype(out)>>;}
+    && boost::json::has_value_to<typename T::Input>::value;
+
+template <typename T>
+concept HandlerWithoutInput = requires(T a, typename T::Output out) {
+    { a.process() } -> std::same_as<HandlerReturnType<decltype(out)>>;};
+
+template <typename T>
+concept Handler = 
+(HandlerWithInput<T>
+||
+HandlerWithoutInput<T>)
+&& boost::json::has_value_from<typename T::Output>::value;
 // clang-format on
 
 }  // namespace RPCng
--- a/src/rpc/common/impl/Processors.h
+++ b/src/rpc/common/impl/Processors.h
@@ -24,6 +24,9 @@
 
 namespace RPCng::detail {
 
+template <typename>
+static constexpr bool unsupported_handler_v = false;
+
 template <Handler HandlerType>
 struct DefaultProcessor final
 {
@@ -33,19 +36,35 @@ struct DefaultProcessor final
     {
         using boost::json::value_from;
         using boost::json::value_to;
+        if constexpr (HandlerWithInput<HandlerType>)
+        {
+            // first we run validation
+            auto const spec = handler.spec();
+            if (auto const ret = spec.validate(value); not ret)
+                return Error{ret.error()};  // forward Status
 
-        // first we run validation
-        auto const spec = handler.spec();
-        if (auto const ret = spec.validate(value); not ret)
-            return Error{ret.error()};  // forward Status
+            auto const inData = value_to<typename HandlerType::Input>(value);
 
-        auto const inData = value_to<typename HandlerType::Input>(value);
-
-        // real handler is given expected Input, not json
-        if (auto const ret = handler.process(inData); not ret)
-            return Error{ret.error()};  // forward Status
+            // real handler is given expected Input, not json
+            if (auto const ret = handler.process(inData); not ret)
+                return Error{ret.error()};  // forward Status
+            else
+                return value_from(ret.value());
+        }
+        else if constexpr (HandlerWithoutInput<HandlerType>)
+        {
+            // no input to pass, ignore the value
+            if (auto const ret = handler.process(); not ret)
+                return Error{ret.error()};  // forward Status
+            else
+                return value_from(ret.value());
+        }
         else
-            return value_from(ret.value());
+        {
+            // when concept HandlerWithInput and HandlerWithoutInput not cover
+            // all Handler case
+            static_assert(unsupported_handler_v<HandlerType>);
+        }
     }
 };
 
--- a/src/rpc/ngHandlers/Ping.h
+++ b/src/rpc/ngHandlers/Ping.h
@@ -19,48 +19,20 @@
 
 #pragma once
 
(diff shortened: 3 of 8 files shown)


src/rpc/common/Concepts.h
```diff
@@ -49,12 +49,17 @@ concept Requirement = requires(T a) {
  */
 // clang-format off
 template <typename T>
-concept Handler = requires(T a, typename T::Input in, typename T::Output out) {
+concept Handler = 
+((requires(T a, typename T::Input in, typename T::Output out) {
```
cindyyan317: check input jsonable when input present.

src/rpc/common/Concepts.h
```diff
@@ -49,12 +49,17 @@ concept Requirement = requires(T a) {
...
     { a.spec() } -> std::same_as<RpcSpecConstRef>;
```
godexsoft, line 53: i think we need the spec in both cases, right?
also, i'd split into separate smaller concepts and then combine into one big concept using smaller ones for readability :)

src/rpc/common/impl/Processors.h
```diff
@@ -31,21 +31,61 @@ struct DefaultProcessor final
...
         const
     {
-        using boost::json::value_from;
-        using boost::json::value_to;
+        // can not do the check in one if statement
+        // first check the nested type exist
+        // then check has a process function called by this type
+        constexpr bool hasNonVoidInputField =
+            requires(const HandlerType& t) { typename HandlerType::Input; };
+        if constexpr (hasNonVoidInputField)
```
godexsoft: If you split into smaller concepts in Concepts.h you can probably leverage it here?

src/rpc/common/impl/Processors.h
```diff
@@ -31,21 +31,61 @@ struct DefaultProcessor final
...
+        {
+            constexpr bool hasProcessWithInput =
```
godexsoft: and here too.. seems like unnecessary code duplication.

src/rpc/common/impl/Processors.h
```diff
@@ -31,21 +31,61 @@ struct DefaultProcessor final
...
+                if (auto const ret = handler.process(inData); not ret)
+                    return Error{ret.error()};  // forward Status
+                else
+                    return value_from(ret.value());
+            }
+        }
         else
-            return value_from(ret.value());
+        {
+            return Error{RPC::Status{"Wrong call"}};
```
godexsoft: You can use the static_assert trick here to make this fail to compile instead of a runtime error:

```cpp
template <typename>
static constexpr bool unsupported_handler_v = false;

// down the line ...
else { static_assert(unsupported_handler_v<HandlerType>); }
```

src/rpc/common/impl/Processors.h
```diff
@@ -31,21 +31,61 @@ struct DefaultProcessor final
     operator()(HandlerType const& handler, boost::json::value const& value)
         const
     {
-        using boost::json::value_from;
-        using boost::json::value_to;
+        // can not do the check in one if statement
+        // first check the nested type exist
+        // then check has a process function called by this type
+        constexpr bool hasNonVoidInputField =
```
godexsoft: you probably want `static constexpr`, almost always.

src/rpc/ngHandlers/Ping.h
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+
+namespace RPCng {
+class PingHandler
+{
+public:
+    using Output = std::string;
+    using Result = RPCng::HandlerReturnType<Output>;
+
+    Result
```
godexsoft, line 32: This does look very clean indeed 🥇

src/rpc/common/Concepts.h
godexsoft: Actually, nvm. you are right, we need it only for when we actually have some input to validate.
Please try to split into smaller concepts and built on composition - overall, looks like a good solution to me 👍

src/rpc/common/impl/Processors.h
cindyyan317: I can not put them in one if . seems it does not support short circuit

src/rpc/common/impl/Processors.h
cindyyan317: I think we can't have compile error here. Because I add the process() in Model class and AnyHandler to support no input process. So DefaultProcessor need valid method definition for both process(input) and proces() . If we wanna constrain it in compile time, we have to more changes on AnyHandler / Model

src/rpc/common/impl/Processors.h
```diff
@@ -33,19 +36,35 @@ struct DefaultProcessor final
...
+        }
+        else if constexpr (HandlerWithoutInput<HandlerType>)
+        {
+            // no input to pass, ignore the value
+            if (auto const ret = handler.process(); not ret)
+                return Error{ret.error()};  // forward Status
+            else
+                return value_from(ret.value());
+        }
         else
```
godexsoft, line 62: This code is unreachable, right? we already say at the very top that we expect a `Handler` which is either with input or without input. Do you add this just in case for future?

src/rpc/ngHandlers/Ping.h
```diff
@@ -0,0 +1,37 @@
...
+
+#pragma once
+
+#include <rpc/common/Types.h>
+
+namespace RPCng {
+class PingHandler
+{
+public:
+    using Output = boost::json::object;
```
godexsoft: Could this be `void`?

src/rpc/ngHandlers/Ping.h
```diff
@@ -0,0 +1,37 @@
...
+    using Result = RPCng::HandlerReturnType<Output>;
+
+    Result
+    process() const
+    {
+        return boost::json::object{};
```
godexsoft: This could be `return {};`

godexsoft (commented):
In general, LGTM. I still have a few questions tho, see inline.

src/rpc/common/impl/Processors.h
cindyyan317: yea. it is for the case, handler != input + without input. in case we update the concepts wrongly.

src/rpc/ngHandlers/Ping.h
cindyyan317: Add a dedicated type for this , as we discussed

src/rpc/common/Types.h
```diff
@@ -53,4 +53,14 @@ struct FieldSpec;
 
 using RpcSpecConstRef = RpcSpec const&;
 
+struct VoidOutput
+{
+};
+
+inline void
+tag_invoke(boost::json::value_from_tag, boost::json::value& jv, VoidOutput o)
```
godexsoft: No need for a name for the VoidOutput parameter. This would have been a warning turned error when we introduce flags again.. i'm sure it would work without the name :)

src/rpc/ngHandlers/Ping.h
```diff
@@ -0,0 +1,38 @@
...
+class PingHandler
+{
+public:
+    using Output = VoidOutput;
+    using Result = HandlerReturnType<Output>;
+
+    Result
+    process() const
+    {
+        return VoidOutput{};
```
godexsoft: `return Output{};` since we specify it above already ;)

godexsoft (changes requested):
Let's get the last few tiny changes in - getting this as good as possible is important right now because this will be the source of copy-paste later on 🚀

godexsoft (dismissed):
Looks great 🚀

godexsoft (approved):

PR #507, opened 2023-02-14
Exit 1 on failed experimental builds so build step fails

legleux:
GitHub actions is interpreting the successful echoing of the message that the step failed as a successful step result which allows the next step to run which is guaranteed to fail since the build failed.

This didn't matter when we were ignoring the experimental failure but now we want to really fail the workflow if any OS fails to build.

This change fails the build step that _actually_ failed

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
 
   build_clio:
     name: Build Clio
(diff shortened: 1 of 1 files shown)


cindyyan317:
LGTM

cindyyan317 (approved):

godexsoft (approved):
LGTM!

PR #509, opened 2023-02-14
Fix mac build failure

cindyyan317:
homebrew installs the cassandra to /usr/local/lib and /usr/local/include. (only for some version of os) CMake omit it when we include the path. 
see some discussion -> https://stackoverflow.com/questions/58541131/why-is-cmake-removing-some-include-directories-from-include-directories-calls
The fix is to include the real cassandra path


godexsoft (dismissed):
LGTM, thanks for fixing this 🚀

godexsoft (approved):

PR #510, opened 2023-02-14
Add common validator

cindyyan317:
Add ledger index validator and ledger hash validator
Put this change to a separate PR to reduce the noise of handler PR.

### DIFF
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -23,8 +23,6 @@
 #include <rpc/common/Specs.h>
 #include <rpc/common/Types.h>
 
-#include <charconv>
-
 namespace RPCng::validation {
 
 /**
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -19,8 +19,6 @@
 
 #pragma once
 
-#include <ripple/basics/base_uint.h>
-
 #include <rpc/common/Concepts.h>
 #include <rpc/common/Specs.h>
 #include <rpc/common/Types.h>
@@ -359,7 +357,8 @@ extern CustomValidator LedgerIndexValidator;
 /**
  * @brief Provide a common used validator for ledger index
  * LedgerIndex must be a string or int
- * If it is string, it must be "validated" or "int"
+ * If the specified LedgerIndex is a string, it's value must be either
+ * "validated" or a valid integer value represented as a string.
  */
 extern CustomValidator LedgerHashValidator;
 
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -19,9 +19,13 @@
 
 #pragma once
 
+#include <ripple/basics/base_uint.h>
+
 #include <rpc/common/Concepts.h>
 #include <rpc/common/Specs.h>
 #include <rpc/common/Types.h>
+
+#include <charconv>
 
 namespace RPCng::validation {
 
@@ -339,5 +343,24 @@ public:
     [[nodiscard]] MaybeError
     verify(boost::json::value const& value, std::string_view key) const;
 };
+
+/**
+ * @brief Helper function to check if sv is an uint32 number or not
+ */
+[[nodiscard]] bool
+checkIsU32Numeric(std::string_view sv);
+
+/**
+ * @brief Provide a common used validator for ledger hash
+ * LedgerHash must be a string and hex
+ */
+extern CustomValidator LedgerIndexValidator;
+
+/**
+ * @brief Provide a common used validator for ledger index
+ * LedgerIndex must be a string or int
+ * If it is string, it must be "validated" or "int"
+ */
+extern CustomValidator LedgerHashValidator;
 
 }  // namespace RPCng::validation
--- a/src/rpc/common/Validators.cpp
+++ b/src/rpc/common/Validators.cpp
@@ -22,6 +22,7 @@
 
 #include <boost/json/value.hpp>
 
+#include <charconv>
 #include <string_view>
 
 namespace RPCng::validation {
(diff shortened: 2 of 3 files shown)


src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
     verify(boost::json::value const& value, std::string_view key) const;
 };
 
+// clang-format off
+/**
+ * @brief Provide a common used validator for ledger hash
+ * LedgerHash must be a string and hex
+ */
+inline CustomValidator LedgerHashValidator = validation::CustomValidator{
```
godexsoft: Do we need `validation::CustomValidator` here? i think we are already in the `validation` namespace so could probably avoid it?

godexsoft: I'm not sure how i feel about these global objects tbh. I would also prefer to have it in the `.cpp` file if that's possible. You may be able to declare them in the header and define in the `.cpp` file, wdyt?

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
     verify(boost::json::value const& value, std::string_view key) const;
 };
 
+// clang-format off
```
godexsoft: Is there any way we can avoid disabling formatting for such a big block of code? I understand that clang-format messes up the formatting for the lambda so maybe let's try to disable formatting only for the relevant lines?

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
...
+/**
+ * @brief Provide a common used validator for ledger index
+ * LedgerIndex must be a string or int
+ * If it is string, it must be "validated" or "int"
+ */
+inline CustomValidator LedgerIndexValidator = validation::CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        auto err = Error{RPC::Status{
+                    RPC::RippledError::rpcINVALID_PARAMS, "ledgerIndexMalformed"}};
+        if(!value.is_string() && !(value.is_uint64() || value.is_int64())){
```
godexsoft: `){` - this is why i don't like to disable formatting for a big block of code..

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
...
+            return err;
+        }
+        if(value.is_string() && value.as_string()!="validated"){
+            auto str = value.as_string();
+            uint32_t unused;
+            auto [_, ec] { std::from_chars(str.data(), str.data() + str.size(), unused) };
```
godexsoft: This looks a bit confusing. Can we use `auto [_, ec] = std::from_chars(...);` instead here?

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
...
+ * If it is string, it must be "validated" or "int"
+ */
+inline CustomValidator LedgerIndexValidator = validation::CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        auto err = Error{RPC::Status{
+                    RPC::RippledError::rpcINVALID_PARAMS, "ledgerIndexMalformed"}};
+        if(!value.is_string() && !(value.is_uint64() || value.is_int64())){
+            return err;
+        }
+        if(value.is_string() && value.as_string()!="validated"){
```
godexsoft: `()!="` all collapsed. as well as `){`.

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,49 @@ class CustomValidator final
...
+            auto str = value.as_string();
+            uint32_t unused;
+            auto [_, ec] { std::from_chars(str.data(), str.data() + str.size(), unused) };
+            if (ec != std::errc())
+            {
```
godexsoft: I'd remove these curly braces - they don't help the code readability in this case and are ok to omit imo.

src/rpc/common/Validators.h
godexsoft: Does the string have to be lower case? what if the user sends "Validated" or "VALIDATED" is that considered invalid?
We could use `boost::iequals` here since we already use boost - search the codebase, we already use it in many places.

src/rpc/common/Validators.h
godexsoft: I'd also extract this to `checkIsNumeric(std::string_view)` or similar - we may want to adjust the implementation for performance later and/or reuse it in other validators.

godexsoft (changes requested):
Left a few suggestions 👍

src/rpc/common/Validators.h
cindyyan317: To keep the logic align with the previous , it seems previous implementation only check against "validated"

src/rpc/common/Validators.h
```diff
@@ -19,10 +19,14 @@
 
 #pragma once
 
+#include <ripple/basics/base_uint.h>
```
godexsoft: I think this and the other include are only needed in the cpp file now?

src/rpc/common/Validators.h
```diff
@@ -340,4 +344,23 @@ class CustomValidator final
...
+/**
+ * @brief Provide a common used validator for ledger hash
+ * LedgerHash must be a string and hex
+ */
+extern CustomValidator LedgerIndexValidator;
+
+/**
+ * @brief Provide a common used validator for ledger index
+ * LedgerIndex must be a string or int
+ * If it is string, it must be "validated" or "int"
```
godexsoft: The 'or "int"' part makes me think it's a literal string `"int"`. 
We should say something like:
```
If the specified LedgerIndex is a string, it's value must be either "validated" or a valid integer value represented as a string.
```

godexsoft (commented):
In general, all good but i'd update the doc and move the includes.

godexsoft (approved):
LGTM 🚀

PR #518, opened 2023-02-21
Fix marker issue

cindyyan317:
We used the next page as part of marker. When we request the next part of response with this marker, the code 
turn empty result. We should use the current page as marker.

Test pass after fix
Screenshot 2023-02-21 at 08 54 13

### DIFF
--- a/unittests/rpc/RPCHelpersTest.cpp
+++ b/unittests/rpc/RPCHelpersTest.cpp
@@ -1,0 +1,420 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <rpc/RPCHelpers.h>
+#include <util/Fixtures.h>
+#include <util/TestObject.h>
+
+#include <fmt/core.h>
+
+#include <variant>
+
+using namespace RPC;
+using namespace testing;
+
+constexpr static auto ACCOUNT = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn";
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+constexpr static auto INDEX1 =
+    "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC321";
+constexpr static auto INDEX2 =
+    "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC322";
+constexpr static auto TXNID =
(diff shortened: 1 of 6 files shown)


unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
godexsoft: 2023 :D

godexsoft: maybe we should update everywhere to "2021-present".. to avoid this issue in future. not in this pr ofcourse.

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
...
+        MockBackendTest::SetUp();
+    }
+    void
+    TearDown() override
+    {
+        MockBackendTest::TearDown();
+    }
+
+protected:
+    boost::asio::io_context ctx;
```
godexsoft: Wouldn't it be possible to derive from both `MockBackendTest` and one of our `*AsioTest` fixtures?

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
...
+        static_cast<MockBackend*>(mockBackendPtr.get());
+    // fetch account object return something
+    auto fake = Blob{'f', 'a', 'k', 'e'};
+    ON_CALL(*rawBackendPtr, doFetchLedgerObject).WillByDefault(Return(fake));
+    EXPECT_CALL(*rawBackendPtr, doFetchLedgerObject).Times(1);
+
+    boost::asio::spawn(ctx, [this](boost::asio::yield_context yield) {
+        auto account = GetAccountIDWithString(ACCOUNT);
+        auto ret = traverseOwnedNodes(
+            *mockBackendPtr, account, 9, 10, "nothex,abc", yield, [](auto sle) {
```
godexsoft: `auto sle` will give us unused variable error later on. let's remove `sle` here and in other instances to prevent it erroring out in the future?

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
...
+        EXPECT_EQ(
+            cursor->toString(),
+            "0000000000000000000000000000000000000000000000000000000000000000,"
+            "0");
+    });
+    ctx.run();
+}
+
+// limit = 10, return 10 objects and marker
+TEST_F(RPCHelpersTest, TraverseOwnedNodesNoMarker2)
```
godexsoft: Can you think of a more descriptive name for this test? :)

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
...
+        auto cursor = std::get_if<AccountCursor>(&ret);
+        EXPECT_TRUE(cursor != nullptr);
+        EXPECT_EQ(count, 10);
+        EXPECT_EQ(cursor->toString(), fmt::format("{},0", INDEX1));
+    });
+    ctx.run();
+}
+
+// 10 objects per page, limit is 15, return the second page as marker
+TEST_F(RPCHelpersTest, TraverseOwnedNodesNoMarker3)
```
godexsoft: Maybe a more descriptive name for the test :)

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -0,0 +1,420 @@
...
+        EXPECT_TRUE(cursor != nullptr);
+        EXPECT_EQ(count, limit);
+        EXPECT_EQ(cursor->toString(), fmt::format("{},{}", INDEX1, pageNum));
+    });
+    ctx.run();
+}
+
+// Send a valid marker, but marker contain an unexisting index
+// return empty
+TEST_F(RPCHelpersTest, TraverseOwnedNodesWithMarker2)
```
godexsoft: Here too please :)

unittests/util/TestObject.h
```diff
@@ -135,3 +135,24 @@ CreateMetaDataForCancelOffer(
     uint32_t transactionIndex,
     int finalTakerGets,
     int finalTakerPays);
+
+/*
+ * Create a owner dir ledger object
```
godexsoft, line 140: I'm not sure these comments help anything as they are the same as the function name.
If you wanted to make those into doxygen documentation then consider using `/** */` comment instead, otherwise i think they are safe to remove everywhere in this file, wdyt?

godexsoft (commented):
This is really awesome! thanks for fixing the bug and adding unit-tests. 
I am leaving a few small suggestions but overall this is great 🥇

unittests/util/TestObject.h
cindyyan317: I think we intentional not use doxygen comment because it is for unittest. I will remove repeated comments

godexsoft: yes, you are right. we don't need doxy for this nor we really need these comments at all :) it's ok for them to stay for now - let's remove later tho.

godexsoft (approved):
Awesome 🥇

PR #519, opened 2023-02-22
Account channel

cindyyan317:
Implement account channel via new rpc.
Extract AccountValidator and MarkerValidator.
Adapt RPCHelpers functions.

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -613,7 +613,7 @@ getLedgerInfoFromHashOrSeq(
 
         return *lgrInfo;
     }
-    const auto ledgerSequence = ledgerIndex.value_or(maxSeq);
+    auto const ledgerSequence = ledgerIndex.value_or(maxSeq);
     // return without check db
     if (ledgerSequence > maxSeq)
         return err;
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -703,9 +703,9 @@ ngTraverseOwnedNodes(
     boost::asio::yield_context& yield,
     std::function<void(ripple::SLE&&)> atOwnedNode)
 {
-    auto maybeCursor = parseAccountCursor(jsonCursor);
+    auto const maybeCursor = parseAccountCursor(jsonCursor);
     // the format is checked in RPC framework level
-    auto [hexCursor, startHint] = *maybeCursor;
+    auto const [hexCursor, startHint] = *maybeCursor;
 
     return traverseOwnedNodes(
         backend,
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -102,11 +102,7 @@ getRequiredUInt(boost::json::object const& request, std::string const& field)
 }
 
 std::optional<AccountCursor>
-parseAccountCursor(
-    BackendInterface const& backend,
-    std::uint32_t seq,
-    std::optional<std::string> jsonCursor,
-    boost::asio::yield_context& yield)
+parseAccountCursor(std::optional<std::string> jsonCursor)
 {
     ripple::uint256 cursorIndex = beast::zero;
     std::uint64_t startHint = 0;
@@ -596,6 +592,39 @@ ledgerInfoFromRequest(Context const& ctx)
     return *lgrInfo;
 }
 
+// extract ledgerInfoFromRequest's parameter from context
+std::variant<Status, ripple::LedgerInfo>
+getLedgerInfoFromHashOrSeq(
+    BackendInterface const& backend,
+    boost::asio::yield_context& yield,
+    std::optional<std::string> ledgerHash,
+    std::optional<uint32_t> ledgerIndex,
+    uint32_t maxSeq)
+{
+    std::optional<ripple::LedgerInfo> lgrInfo;
+    auto const err =
+        RPC::Status{RPC::RippledError::rpcLGR_NOT_FOUND, "ledgerNotFound"};
+    if (ledgerHash)
+    {
+        lgrInfo =
+            backend.fetchLedgerByHash(ripple::uint256{*ledgerHash}, yield);
+        if (!lgrInfo || lgrInfo->seq > maxSeq)
+            return err;
+
+        return *lgrInfo;
+    }
+    const auto ledgerSequence = ledgerIndex.value_or(maxSeq);
+    // return without check db
+    if (ledgerSequence > maxSeq)
+        return err;
+
+    lgrInfo = backend.fetchLedgerBySequence(ledgerSequence, yield);
+    if (!lgrInfo)
+        return err;
+
+    return *lgrInfo;
+}
+
 std::vector<unsigned char>
 ledgerInfoToBlob(ripple::LedgerInfo const& info, bool includeHash)
 {
@@ -646,10 +675,36 @@ traverseOwnedNodes(
             ripple::keylet::account(accountID).key, sequence, yield))
         return Status{RippledError::rpcACT_NOT_FOUND};
 
-    auto maybeCursor = parseAccountCursor(backend, sequence, jsonCursor, yield);
+    auto const maybeCursor = parseAccountCursor(jsonCursor);
     if (!maybeCursor)
         return Status(ripple::rpcINVALID_PARAMS, "Malformed cursor");
 
+    auto [hexCursor, startHint] = *maybeCursor;
+
+    return traverseOwnedNodes(
+        backend,
+        ripple::keylet::ownerDir(accountID),
+        hexCursor,
+        startHint,
+        sequence,
+        limit,
+        jsonCursor,
+        yield,
+        atOwnedNode);
+}
+
+std::variant<Status, AccountCursor>
+ngTraverseOwnedNodes(
+    BackendInterface const& backend,
+    ripple::AccountID const& accountID,
+    std::uint32_t sequence,
+    std::uint32_t limit,
+    std::optional<std::string> jsonCursor,
+    boost::asio::yield_context& yield,
+    std::function<void(ripple::SLE&&)> atOwnedNode)
+{
+    auto maybeCursor = parseAccountCursor(jsonCursor);
+    // the format is checked in RPC framework level
     auto [hexCursor, startHint] = *maybeCursor;
 
     return traverseOwnedNodes(
--- a/src/rpc/common/Concepts.h
+++ b/src/rpc/common/Concepts.h
@@ -50,7 +50,7 @@ concept Requirement = requires(T a) {
  */
 // clang-format off
 template <typename T>
-concept CoroutineProcess = requires(T a, typename T::Input in, typename T::Output out, boost::asio::yield_context* y) {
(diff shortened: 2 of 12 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -596,6 +592,39 @@ ledgerInfoFromRequest(Context const& ctx)
...
+std::variant<Status, ripple::LedgerInfo>
+getLedgerInfoFromHashOrSeq(
+    BackendInterface const& backend,
+    boost::asio::yield_context& yield,
+    std::optional<std::string> ledgerHash,
+    std::optional<uint32_t> ledgerIndex,
+    uint32_t maxSeq)
+{
+    std::optional<ripple::LedgerInfo> lgrInfo;
+    auto err =
```
godexsoft: minor: Let's `auto const` anything you are not changing as a rule of thumb :)

src/rpc/RPCHelpers.cpp
```diff
@@ -596,6 +592,39 @@ ledgerInfoFromRequest(Context const& ctx)
...
+    if (ledgerHash)
+    {
+        lgrInfo =
+            backend.fetchLedgerByHash(ripple::uint256{*ledgerHash}, yield);
+        if (!lgrInfo || lgrInfo->seq > maxSeq)
+            return err;
+
+        return *lgrInfo;
+    }
+    auto ledgerSequence = ledgerIndex.value_or(maxSeq);
```
godexsoft: minor: `auto const` here would be nice.

src/rpc/RPCHelpers.cpp
```diff
@@ -646,7 +675,7 @@ traverseOwnedNodes(
             ripple::keylet::account(accountID).key, sequence, yield))
         return Status{RippledError::rpcACT_NOT_FOUND};
 
-    auto maybeCursor = parseAccountCursor(backend, sequence, jsonCursor, yield);
+    auto maybeCursor = parseAccountCursor(jsonCursor);
```
godexsoft: minor: And here too

src/rpc/RPCHelpers.cpp
```diff
@@ -664,6 +693,32 @@ traverseOwnedNodes(
...
+ngTraverseOwnedNodes(
+    BackendInterface const& backend,
+    ripple::AccountID const& accountID,
+    std::uint32_t sequence,
+    std::uint32_t limit,
+    std::optional<std::string> jsonCursor,
+    boost::asio::yield_context& yield,
+    std::function<void(ripple::SLE&&)> atOwnedNode)
+{
+    auto maybeCursor = parseAccountCursor(jsonCursor);
```
godexsoft: minor: `auto const` here too

src/rpc/common/Validators.cpp
```diff
@@ -126,4 +127,42 @@ CustomValidator LedgerIndexValidator = CustomValidator{
...
 
+CustomValidator AccountValidator = CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_string())
+        {
+            return Error{RPC::Status{
+                RPC::RippledError::rpcINVALID_PARAMS,
+                std::string(key) + "NotString"}};
+        }
+        // we are using accountFromStringStrict from RPCHelpers, after we remove
```
godexsoft: It's always useful to add `TODO:` in front of comments like this one so that we can see it easier later and not forget to change this. Consider adding :)

src/rpc/common/Validators.cpp
```diff
@@ -126,4 +127,42 @@ CustomValidator LedgerIndexValidator = CustomValidator{
...
+
+CustomValidator MarkerValidator = CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_string())
+        {
+            return Error{RPC::Status{
+                RPC::RippledError::rpcINVALID_PARAMS,
+                std::string(key) + "NotString"}};
+        }
+        // we are using parseAccountCursor from RPCHelpers, after we remove
```
godexsoft: Same thing with `TODO:`.

src/rpc/common/Validators.h
```diff
@@ -360,4 +360,17 @@ extern CustomValidator LedgerIndexValidator;
...
 
+/**
+ * @brief Provide a common used validator for account
+ * Account must be a string and the converted public key is valid
+ */
+extern CustomValidator AccountValidator;
+
+/**
+ * @brief Provide a common used validator for marker
+ *  Markder is composed of a comma separated index and start hint. The
```
godexsoft: Looks like a typo. `Markder` -> `Marker`.

src/rpc/ngHandlers/AccountChannels.cpp
```diff
@@ -0,0 +1,208 @@
...
+    if (auto const& v = channelSle[~ripple::sfDestinationTag])
+        channel.destinationTag = *v;
+
+    jsonChannels.push_back(channel);
+}
+
+AccountChannelsHandler::Result
+AccountChannelsHandler::process(AccountChannelsHandler::Input input) const
+{
+    auto range = sharedPtrBackend_->fetchLedgerRange();
```
godexsoft: minor: Let's try to make everything that we can into `auto const` in this entire member function🥇

src/rpc/ngHandlers/AccountChannels.cpp
```diff
@@ -0,0 +1,208 @@
...
+    obj = {
+        {"account", output.account},
+        {"ledger_hash", output.ledgerHash},
+        {"ledger_index", output.ledgerIndex},
+        {"validated", output.validated},
+        {"limit", output.limit},
+        {"channels", boost::json::value_from(output.channels)}};
+    if (output.marker)
+        obj["marker"] = output.marker.value();
+    jv.emplace_object() = obj;
```
godexsoft: is this better than `jv = obj` or does it not work without it?

src/rpc/ngHandlers/AccountChannels.h
```diff
@@ -0,0 +1,130 @@
...
+    {
+        std::string account;
+        std::optional<std::string> destinationAccount;
+        std::optional<std::string> ledgerHash;
+        std::optional<uint32_t> ledgerIndex;
+        uint32_t limit = 50;
+        std::optional<std::string> marker;
+    };
+
+    using Input = HandlerInput;
```
godexsoft: Is there point calling it `HandlerInput` and then aliasing it here? why not just call the struct Input and Output from the start, just curious if there is a reason other than having the `using` aliases nicely together in a group?

unittests/rpc/handlers/AccountChannelsTest.cpp
```diff
@@ -0,0 +1,915 @@
...
+        MockBackendTest::TearDown();
+        SyncAsioContextTest::TearDown();
+    }
+};
+
+TEST_F(RPCAccountHandlerTest, NonHexLedgerHash)
+{
+    boost::asio::spawn(ctx, [this](boost::asio::yield_context yield) {
+        auto const handler =
+            AnyHandler{AccountChannelsHandler{yield, mockBackendPtr}};
```
godexsoft: Oops, this `yield` here is **bad**.. imagine this situation, we will have a `std::unordered_map<string, AnyHandler>` of already existing handlers and we will call `process` on the existing handler instances, not create new ones for each response. This means that `yield` will be new each time (and hopefully eventually there will not be one at all). So we can't inject it once for the entire application 😢

src/rpc/ngHandlers/AccountChannels.h
```diff
@@ -0,0 +1,130 @@
...
+
+        return rpcSpec;
+    }
+
+    Result
+    process(Input input) const;
+
+private:
+    // dependencies
+    const std::reference_wrapper<boost::asio::yield_context> yieldCtx_;
```
godexsoft: Sadly we can't have this as a dependency since it's a new context for each new rpc call. I think we can introduce some artificial `asio::spawn` inside of our `RPCHelpers` just to avoid having this dependency here. Eventually we should be able to completely remove the need for these `yield_context`s that leak everywhere throughout `clio` atm. 😢

godexsoft (changes requested):
Great progress on this! ⭐ 
I like mostly everything, it's coming along nicely. There is only one big issue - yield_context can't be a dependency. We need to find a workaround for it. Let's discuss 🤙

src/rpc/ngHandlers/AccountChannels.h
cindyyan317: I think the yield is from our work queue. Previously, context will hold the reference of it and yield when db operation happens. If we want to artifically 'spawn' something, where we should execute it ? another thread pool?

src/rpc/common/Concepts.h
```diff
@@ -50,7 +50,7 @@ concept Requirement = requires(T a) {
  */
 // clang-format off
 template <typename T>
-concept CoroutineProcess = requires(T a, typename T::Input in, typename T::Output out, boost::asio::yield_context* y) {
+concept CoroutineProcess = requires(T a, typename T::Input in, typename T::Output out, boost::asio::yield_context& y) {
```
cindyyan317, line 53: I feel reference is more friendly when we use it in handler. try to hide pointer thing inside framework

src/rpc/RPCHelpers.cpp
```diff
@@ -596,6 +592,39 @@ ledgerInfoFromRequest(Context const& ctx)
...
+    if (ledgerHash)
+    {
+        lgrInfo =
+            backend.fetchLedgerByHash(ripple::uint256{*ledgerHash}, yield);
+        if (!lgrInfo || lgrInfo->seq > maxSeq)
+            return err;
+
+        return *lgrInfo;
+    }
+    const auto ledgerSequence = ledgerIndex.value_or(maxSeq);
```
godexsoft: Let's not mix east and west const in one function (or better in the entire codebase) 😃

src/rpc/ngHandlers/AccountChannels.h
```diff
@@ -0,0 +1,124 @@
...
+
+        return rpcSpec;
+    }
+
+    Result
+    process(Input input, boost::asio::yield_context& yield) const;
+
+private:
+    // dependencies
+    const std::shared_ptr<BackendInterface> sharedPtrBackend_;
```
godexsoft: In `clio` we prefer to have the member variables at the top of the class (see other classes in the codebase). 
Although my personal preference is `private` block at the end like you have it here, we should probably try to stick with the convention we have established.

Also, west const here again.
And i'd leave the private member function here, only move the dependency member to the top for visibility.

unittests/rpc/handlers/impl/FakesAndMocks.h
```diff
@@ -122,7 +122,7 @@ class CoroutineHandlerFake
     }
 
     Result
-    process(Input input, boost::asio::yield_context* ptrYield) const
+    process(Input input, boost::asio::yield_context& ptrYield) const
```
godexsoft: `ptrYield` does not represent the reality anymore - i'd rename to `yield` like other places have it.

godexsoft (changes requested):
A few more small things! overall, we are almost ready to close this! great work🥇

godexsoft (approved):
Great work!

PR #521, opened 2023-02-23
Provide coroutine process interface for handler

cindyyan317:
Provide process coroutine interface to execute handler in a coroutine.

### DIFF
--- a/src/rpc/common/AnyHandler.h
+++ b/src/rpc/common/AnyHandler.h
@@ -78,11 +78,17 @@ public:
      * @return JSON result or @ref RPC::Status on error
      */
     [[nodiscard]] ReturnType
+    process(boost::json::value const& value) const
+    {
+        return pimpl_->process(value);
+    }
+
+    [[nodiscard]] ReturnType
     process(
         boost::json::value const& value,
-        boost::asio::yield_context* ptrYield = nullptr) const
+        boost::asio::yield_context& ptrYield) const
     {
-        return pimpl_->process(value, ptrYield);
+        return pimpl_->process(value, &ptrYield);
     }
 
 private:
--- a/src/rpc/common/AnyHandler.h
+++ b/src/rpc/common/AnyHandler.h
@@ -69,10 +69,7 @@ public:
     operator=(AnyHandler&&) = default;
 
     /**
-     * @brief Process incoming JSON by the stored handler,if ptrYield is
-     * provided, will try to call the handler process with yield_context. If
-     * handler fails to provide the process coroutine interface, it will raise
-     * compile error
+     * @brief Process incoming JSON by the stored handler
      *
      * @param value The JSON to process
      * @return JSON result or @ref RPC::Status on error
@@ -83,6 +80,13 @@ public:
         return pimpl_->process(value);
     }
 
+    /**
+     * @brief Process incoming JSON by the stored handler in a provided
+     * coroutine
+     *
+     * @param value The JSON to process
+     * @return JSON result or @ref RPC::Status on error
+     */
     [[nodiscard]] ReturnType
     process(
         boost::json::value const& value,
--- a/src/rpc/common/AnyHandler.h
+++ b/src/rpc/common/AnyHandler.h
@@ -69,15 +69,20 @@ public:
     operator=(AnyHandler&&) = default;
 
     /**
-     * @brief Process incoming JSON by the stored handler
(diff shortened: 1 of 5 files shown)


src/rpc/common/AnyHandler.h
```diff
@@ -80,13 +80,30 @@ class AnyHandler final
...
+     *
+     * @param value The JSON to process
+     * @return JSON result or @ref RPC::Status on error
+     */
+    [[nodiscard]] ReturnType
+    process(
+        boost::json::value const& value,
+        boost::asio::yield_context& ptrYield) const
+    {
+        return pimpl_->process(value, &ptrYield);
```
godexsoft, line 95: I'm ok with this because the intention is to remove `yield` later anyways but:
I would add overloads everywhere down to the concept check instead of passing a pointer down the line.
See below comment for why that is.

src/rpc/common/Concepts.h
```diff
@@ -48,10 +49,18 @@ concept Requirement = requires(T a) {
...
+template <typename T>
+concept NonCoroutineProcess = requires(T a, typename T::Input in, typename T::Output out) {
+    { a.process(in) } -> std::same_as<HandlerReturnType<decltype(out)>>;};
+
 template <typename T>
 concept HandlerWithInput = requires(T a, typename T::Input in, typename T::Output out) {
-    { a.spec() } -> std::same_as<RpcSpecConstRef>;
-    { a.process(in) } -> std::same_as<HandlerReturnType<decltype(out)>>;}
+    { a.spec() } -> std::same_as<RpcSpecConstRef>;}
+    &&(CoroutineProcess<T>||NonCoroutineProcess<T>)
```
godexsoft: Please add some spaces to keep the formatting consistent. I'd also add a space on line 62 between `;}` as well as do that on other lines above.  Also you can use `or` instead of `||` (also `and` instead of `&&`) when it makes code more readable - feel free to use it whenever you think it makes code read like english instead of machine code.

```cpp
and (IsCoroutineProcess<T> or IsNonCoroutineProcess<T>)
```

This reads much nicer IMHO. Hope you can agree 👍

src/rpc/common/Concepts.h
```diff
@@ -48,10 +49,18 @@ concept Requirement = requires(T a) {
  * as per boost::json documentation for these functions.
  */
 // clang-format off
+template <typename T>
+concept CoroutineProcess = requires(T a, typename T::Input in, typename T::Output out, boost::asio::yield_context* y) {
+    { a.process(in,y) } -> std::same_as<HandlerReturnType<decltype(out)>>;};
```
godexsoft: Add space `in,y` -> `in, y` and `;}` -> `; }` please 🙇

src/rpc/common/impl/Processors.h
```diff
@@ -44,12 +46,24 @@ struct DefaultProcessor final
...
+                // real handler is given expected Input, not json
+                if (!ret)
+                    return Error{ret.error()};  // forward Status
+                else
+                    return value_from(ret.value());
+            }
             else
-                return value_from(ret.value());
+            {
+                auto const ret = handler.process(inData, ptrYield);
```
godexsoft, line 60: Here `ptrYield` can be `nullptr`. I know this is detail code and so we should not be calling it with a `nullptr` but it's always better to make strong APIs that are hard(er) to misuse.
I'd just pass references everywhere instead and call with `yield` if the handler implements it, otherwise call without `yield`. I think it's better to rely on overload resolution in this case. If the (constexpr) logic becomes too heavy in one place you can always use private member functions to refactor to smaller bits.

src/rpc/common/AnyHandler.h
godexsoft: For this particular time we can leave it as it is to save time but for future let's consider stronger APIs ⭐

godexsoft (changes requested):

src/rpc/common/Concepts.h
cindyyan317: Will update this and previous usage

godexsoft (approved):
Looks good. Thanks for fixing the formatting.

src/rpc/common/AnyHandler.h
cindyyan317: Discussed offline. We decide not to add the override all  the way to concept for now.

PR #524, opened 2023-02-24
Handler account_currencies

cindyyan317:
Refactor account_currencies 
Add unittest
Fix ledger hash parse issue.

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -606,8 +606,10 @@ getLedgerInfoFromHashOrSeq(
         RPC::Status{RPC::RippledError::rpcLGR_NOT_FOUND, "ledgerNotFound"};
     if (ledgerHash)
     {
-        lgrInfo =
-            backend.fetchLedgerByHash(ripple::uint256{*ledgerHash}, yield);
+        // invoke uint256's constructor to parse the hex string , instead of
+        // copying buffer
+        ripple::uint256 ledgerHash256{std::string_view(*ledgerHash)};
+        lgrInfo = backend.fetchLedgerByHash(ledgerHash256, yield);
         if (!lgrInfo || lgrInfo->seq > maxSeq)
             return err;
 
--- a/src/rpc/ngHandlers/AccountCurrencies.h
+++ b/src/rpc/ngHandlers/AccountCurrencies.h
@@ -1,0 +1,93 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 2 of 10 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -606,8 +606,10 @@ getLedgerInfoFromHashOrSeq(
         RPC::Status{RPC::RippledError::rpcLGR_NOT_FOUND, "ledgerNotFound"};
     if (ledgerHash)
     {
-        lgrInfo =
-            backend.fetchLedgerByHash(ripple::uint256{*ledgerHash}, yield);
+        // invoke uint256's constructor to parse the hex string , instead of
```
cindyyan317, line 609: This function is only called in new RPC framework. We should call uint256's string_view constructor. Pass string to construct uint256 will call its ctor overload, which treat the string as container and copy the buffer.
I also add the unittest for this.

src/rpc/ngHandlers/AccountCurrencies.h
```diff
@@ -0,0 +1,87 @@
...
+    {
+        std::string ledgerHash;
+        uint32_t ledgerIndex;
+        std::set<std::string> receiveCurrencies;
+        std::set<std::string> sendCurrencies;
+        // validated should be sent via framework
+        bool validated = true;
+    };
+
+    // TODO:we did not implement the "strict" field
```
godexsoft, line 47: Do we have an issue for it? if yes - let's add a reference like `TODO [#issue]: we did not...` 👍

src/rpc/ngHandlers/AccountCurrencies.h
```diff
@@ -0,0 +1,87 @@
...
+    {
+        std::string account;
+        std::optional<std::string> ledgerHash;
+        std::optional<uint32_t> ledgerIndex;
+    };
+
+    using Result = RPCng::HandlerReturnType<Output>;
+
+    AccountCurrenciesHandler(
+        std::shared_ptr<BackendInterface>& sharedPtrBackend)
```
godexsoft: This should either be `const &` or by value (for copy-elision).

unittests/rpc/handlers/AccountCurrenciesTest.cpp
```diff
@@ -0,0 +1,308 @@
...
+constexpr static auto LEDGERHASH =
+    "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+constexpr static auto INDEX1 =
+    "1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC";
+constexpr static auto INDEX2 =
+    "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC321";
+constexpr static auto TXNID =
+    "E3FE6EA3D48F0C2B639448020EA4F03D4F4F8FFDB243A852A0F59177921B4879";
+
+class RPCAccountCurrenciesHandlerTest : public MockBackendTest,
```
godexsoft: It may be worth introducing a new fixture - `SyncAsioMockBackendTest` so that we don't have to copy this thing in every new test.

godexsoft (dismissed):
Very cool! i added a few recommendations but they are optional. ⭐

src/rpc/ngHandlers/AccountCurrencies.h
cindyyan317: We have not created issue . we have a jira for it. If we confirm it is a bug, I can create an issue

godexsoft (approved):
LGTM!

PR #526, opened 2023-02-27
Tx handler in new RPC framework

cindyyan317:
Tx handler in new RPC framework

### DIFF
--- a/src/rpc/ngHandlers/Tx.cpp
+++ b/src/rpc/ngHandlers/Tx.cpp
@@ -25,13 +25,14 @@ namespace RPCng {
 TxHandler::Result
 TxHandler::process(Input input, boost::asio::yield_context& yield) const
 {
+    constexpr static auto maxLedgerRange = 1000;
     auto const rangeSupplied = input.minLedger && input.maxLedger;
 
     if (rangeSupplied)
     {
         if (*input.minLedger > *input.maxLedger)
             return Error{RPC::Status{RPC::RippledError::rpcINVALID_LGR_RANGE}};
-        if (*input.maxLedger - *input.minLedger > 1000)
+        if (*input.maxLedger - *input.minLedger > maxLedgerRange)
             return Error{
                 RPC::Status{RPC::RippledError::rpcEXCESSIVE_LGR_RANGE}};
     }
@@ -76,7 +77,7 @@ void
 tag_invoke(
     boost::json::value_from_tag,
     boost::json::value& jv,
-    TxHandler::Output output)
+    TxHandler::Output const& output)
 {
     auto obj = boost::json::object{};
     if (output.tx)
--- a/src/rpc/ngHandlers/Tx.cpp
+++ b/src/rpc/ngHandlers/Tx.cpp
@@ -32,6 +32,7 @@ TxHandler::process(Input input, boost::asio::yield_context& yield) const
     {
         if (*input.minLedger > *input.maxLedger)
             return Error{RPC::Status{RPC::RippledError::rpcINVALID_LGR_RANGE}};
+
         if (*input.maxLedger - *input.minLedger > maxLedgerRange)
             return Error{
                 RPC::Status{RPC::RippledError::rpcEXCESSIVE_LGR_RANGE}};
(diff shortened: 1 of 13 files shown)


src/rpc/ngHandlers/Tx.cpp
```diff
@@ -0,0 +1,121 @@
...
+TxHandler::Result
+TxHandler::process(Input input, boost::asio::yield_context& yield) const
+{
+    auto const rangeSupplied = input.minLedger && input.maxLedger;
+
+    if (rangeSupplied)
+    {
+        if (*input.minLedger > *input.maxLedger)
+            return Error{RPC::Status{RPC::RippledError::rpcINVALID_LGR_RANGE}};
+        if (*input.maxLedger - *input.minLedger > 1000)
```
godexsoft: Consider moving this `1000` to some local constants static constexpr (preferably on TxHandler level, not in this function) 👍

src/rpc/ngHandlers/Tx.cpp
```diff
@@ -0,0 +1,121 @@
...
+            return Error{
+                RPC::Status{RPC::RippledError::rpcEXCESSIVE_LGR_RANGE}};
+    }
```
godexsoft, line 39: minor: This part could use a bit of spacing, maybe one line of space here would make blocks more stand out.

src/rpc/ngHandlers/Tx.cpp
```diff
@@ -0,0 +1,121 @@
...
+    output.date = dbResponse->date;
+    output.ledgerIndex = dbResponse->ledgerSequence;
+    return output;
+}
+
+void
+tag_invoke(
+    boost::json::value_from_tag,
+    boost::json::value& jv,
+    TxHandler::Output output)
```
godexsoft: Consider using `TxHandler::Output const& output` here, no need to make a copy.

src/rpc/ngHandlers/Tx.cpp
```diff
@@ -0,0 +1,121 @@
...
+{
+    auto obj = boost::json::object{};
+    if (output.tx)
+    {
+        obj = *output.tx;
+        obj["meta"] = *output.meta;
+    }
+    else
+    {
+        obj["meta"] = *output.metaStr;
```
godexsoft, line 91: Having Contracts in C++20 would have been nice.. 
Should we at least add `assert`s to verify those are not `nullopt` or do we just rely on unit-tests considering this is output, not user-input?

src/rpc/ngHandlers/Tx.h
```diff
@@ -0,0 +1,91 @@
...
+
+    Result
+    process(Input input, boost::asio::yield_context& yield) const;
+};
+
+void
+tag_invoke(
+    boost::json::value_from_tag,
+    boost::json::value& jv,
+    TxHandler::Output output);
```
godexsoft: Let's make it `const&`

godexsoft (changes requested):
Great work! Added a few comments about just a few small things to consider.

src/rpc/ngHandlers/Tx.h
cindyyan317: Probably copy from otherwhere

src/rpc/ngHandlers/Tx.cpp
cindyyan317: it is not user-input. We rely on unittests for this logic.

src/rpc/ngHandlers/Tx.cpp
cindyyan317: Good idea

src/rpc/ngHandlers/Tx.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+#include <rpc/RPCHelpers.h>
+#include <rpc/ngHandlers/Tx.h>
+
+namespace RPCng {
+
+TxHandler::Result
+TxHandler::process(Input input, boost::asio::yield_context& yield) const
+{
+    constexpr static auto maxLedgerRange = 1000;
```
godexsoft: This is now an `int` but minLedger/maxLedger are uint32.. i'd do a `1000u` here or use `uint32_t` instead of `auto` perhaps.

godexsoft (dismissed):
One small comment left but feel free to ignore (we will enable relevant warnings later and it can be fixed then). 👍

godexsoft (approved):
LGTM! 🚀

PR #530, opened 2023-03-02
Add IfType requirement to RPC framework

cindyyan317:
In some API, one field can be two type, for each type, it has the specific requirement.
For example, when the ticket is string, it fit in the uint256 hex format; when the ticket is object, it should contain certain sub field, like seq.

Add IfType to fulfil this feature.

### DIFF
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -52,16 +52,21 @@ template <typename Expected>
         if (not value.is_double())
             hasError = true;
     }
+    else if constexpr (std::is_same_v<Expected, boost::json::array>)
+    {
+        if (not value.is_array())
+            hasError = true;
+    }
+    else if constexpr (std::is_same_v<Expected, boost::json::object>)
+    {
+        if (not value.is_object())
+            hasError = true;
+    }
     else if constexpr (
         std::is_convertible_v<Expected, uint64_t> or
         std::is_convertible_v<Expected, int64_t>)
     {
         if (not value.is_int64() && not value.is_uint64())
-            hasError = true;
-    }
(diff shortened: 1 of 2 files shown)


src/rpc/common/Validators.h
```diff
@@ -308,6 +313,70 @@ class ValidateArrayAt final
     verify(boost::json::value const& value, std::string_view key) const;
 };
 
+/**
+ * @brief A meta-validator that specifies a list of requirements to run against
+ * when the type much the template parameter
```
godexsoft: Looks like a typo? 'much' -> 'match' 👍

unittests/rpc/BaseTests.cpp
```diff
@@ -220,6 +220,44 @@ TEST_F(RPCBaseTest, ArrayAtValidator)
...
+TEST_F(RPCBaseTest, IfTypeValidator)
+{
+    // clang-format off
+    auto spec = RpcSpec{
+        {"mix", Required{}, 
+                Type<std::string,json::object>{},
+                IfType<json::object>{
+                        Section{{ "limit", Required{}, Type<uint32_t>{}, Between<uint32_t>{0, 100}}, },
+                        Section{{ "limit2", Required{}, Type<uint32_t>{}, Between<uint32_t>{0, 100}}, },
+                        },
```
godexsoft, line 232: minor: Formatting looks a bit messed up and i'd remove the extra commas at the end of the list where they are not needed.

godexsoft (changes requested):
Very cool! let's fix the small typos and formatting and it looks ready to go 🚀

godexsoft (approved):
LGTM

PR #532, opened 2023-03-06
Bug fix: ledger_entry responds incorrectly when the expected type mismatch the actual type

cindyyan317:
Clio can return the entry object even when the expected type mismatch the entry object type.
Screenshot 2023-03-06 at 13 54 11
Rippled returns
Screenshot 2023-03-06 at 13 55 37

Add type check after deserialising entry.
Cliot test pass
Screenshot 2023-03-06 at 13 57 20


godexsoft (approved):
Looks good! 🚀

PR #534, opened 2023-03-07
Ledger entry in new RPC framework

cindyyan317:
1 Add currency validator 
2 Change ledger hash validator to uint256 validator which can be used for other fields
3 Add AccountBase58Validator for validate base58 account (not include public key)
4 Add test utility functions to generate ledger entries 
5 Add unittests

### DIFF
--- a/src/rpc/ngHandlers/LedgerEntry.cpp
+++ b/src/rpc/ngHandlers/LedgerEntry.cpp
@@ -120,7 +120,7 @@ LedgerEntryHandler::process(
     if (!ledgerObject || ledgerObject->size() == 0)
         return Error{RPC::Status{"entryNotFound"}};
 
-    const ripple::STLedgerEntry sle{
+    ripple::STLedgerEntry const sle{
         ripple::SerialIter{ledgerObject->data(), ledgerObject->size()}, key};
     if (input.expectedType != ripple::ltANY &&
         sle.getType() != input.expectedType)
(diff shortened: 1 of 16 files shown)


src/rpc/ngHandlers/LedgerEntry.h
```diff
@@ -0,0 +1,208 @@
...
+        std::optional<uint32_t> ledgerIndex;
+        bool binary = false;
+        // id of this ledger entry: 256 bits hex string
+        std::optional<std::string> index;
+        // index can be extracted from payment_channel, check, escrow, offer
+        // etc, expectedType is used to save the type of index
+        ripple::LedgerEntryType expectedType = ripple::ltANY;
+        // account id to address account root object
+        std::optional<std::string> accountRoot;
+        std::optional<boost::json::object> directory;
```
godexsoft, line 59: Is there a way to not leak `json::object` into rpc handler input here? Ideally we want Input to be free from json completely.

src/rpc/ngHandlers/LedgerEntry.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include <unordered_map>
+
+namespace RPCng {
+LedgerEntryHandler::Result
+LedgerEntryHandler::process(
+    LedgerEntryHandler::Input input,
+    boost::asio::yield_context& yield) const
+{
+    ripple::uint256 key;
+    if (input.index)
```
godexsoft, line 32: Maybe this whole thing should be in `tag_invoke` instead? And Input should have a `std::variant` for fields that can be represented in different ways.

godexsoft (changes requested):
Some important design questions to review. Other than that it's pretty good!

src/rpc/ngHandlers/LedgerEntry.h
cindyyan317: If we don't use json object here, we will need to create strong type for all the request types, "directory" ,"offer", "escrow" etc. Even we take effort to prepare all the type, for "directory",  we have some check like only one of "owner" and "dir_root" can be available. Current validators can not support this, we can not do it in tag_invoke either.

src/rpc/ngHandlers/LedgerEntry.cpp
cindyyan317: you mean we calculate the key index in tag_invoke?

src/rpc/ngHandlers/LedgerEntry.h
godexsoft: I see. Let's add a `TODO: extract to custom types and remove json usage from Input struct` and move on for now.

src/rpc/ngHandlers/LedgerEntry.cpp
```diff
@@ -0,0 +1,289 @@
...
+    if (auto const status = std::get_if<RPC::Status>(&lgrInfoOrStatus))
+        return Error{*status};
+
+    auto const lgrInfo = std::get<ripple::LedgerInfo>(lgrInfoOrStatus);
+    auto const ledgerObject =
+        sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, yield);
+    if (!ledgerObject || ledgerObject->size() == 0)
+        return Error{RPC::Status{"entryNotFound"}};
+
+    const ripple::STLedgerEntry sle{
```
godexsoft: West const detected 😃

src/rpc/ngHandlers/LedgerEntry.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    auto const indexFieldType = std::find_if(
+        indexFieldTypeMap.begin(),
+        indexFieldTypeMap.end(),
+        [&jsonObject](auto const& pair) {
+            auto const& [field, _] = pair;
+            return jsonObject.contains(field) &&
+                jsonObject.at(field).is_string();
+        });
+    if (indexFieldType != indexFieldTypeMap.end())
```
godexsoft, line 244: minor: I think this whole sheet of `if`s can be made into `if-elseif` with just one `return input` at the very bottom of the function. wdyt?

godexsoft (changes requested):
Few minor comments added, nothing major 🚀

godexsoft (approved):
LGTM! 🚀

Issue #464, opened 2023-01-09
Introduce the same fixes to gateway_balances as in rippled


godexsoft:
See  https://github.com/XRPLF/rippled/pull/4355
We need to make sure we are on the same page with their implementation.

cindyyan317:
Clio has the overflow issue as well. 
https://github.com/XRPLF/clio/blob/develop/src/rpc/handlers/GatewayBalances.cpp#L154

How to reproduce:
`
{
    "method": "gateway_balances",
    "params": [
        {
           "account": "raiata7DoTJL5CMHuYC8BYGE6J4psyCMcE"
        }
    ]}
`

PR #535, opened 2023-03-07
Gateway balance fix

cindyyan317:
For https://github.com/XRPLF/clio/issues/464
Refer to https://github.com/XRPLF/rippled/pull/4355
Clio has the similar issue as well. The sum overflows the STAmount range.
The fix is to send a flag to indicate there is overflow and keep the last valid sum. (same as rippled)

Before fix:
Screenshot 2023-03-07 at 14 16 11

After fix:
Screenshot 2023-03-07 at 14 15 35


godexsoft (approved):
LGTM.

PR #536, opened 2023-03-10
Gateway balance

cindyyan317:
New RPC handler gateway_balances and unittests, including overflow flag


godexsoft (dismissed):
Code looks fine. A lot of places with strange formatting of json but that's in tests so not too much of a concern. In the last file the bundle tests formatting is really bad - consider fixing it somehow 🚀

godexsoft (approved):
LGTM 👍

PR #537, opened 2023-03-10
Cassandra backend revamp

godexsoft:
- Implement RAII-based wrapper on top of Cassandra driver
- Re-implements important reading/writing routines using the wrapper
- Split into smaller logical components (Wrapper layer, Schema, SettingsProvider, ExecutionStrategy, RetryPolicy, Backend interface implementation, etc.)
- Add testability and unit tests for components that can be mocked
- Add db tests for the wrapper

Probably not complete/finished yet but let's start reviewing @cindyyan317.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -4,7 +4,7 @@ exec 1>&2
 
 # paths to check and re-format
 sources="src unittests"
-formatter="clang-format -i"
+formatter="clang-format-11 -i"
 
 first=$(git diff $sources)
 find $sources -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 $formatter
--- a/src/backend/CassandraBackendNew.h
+++ b/src/backend/CassandraBackendNew.h
@@ -1,0 +1,979 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <backend/BackendInterface.h>
+#include <backend/cassandra/Concepts.h>
+#include <backend/cassandra/Handle.h>
+#include <backend/cassandra/Schema.h>
+#include <backend/cassandra/SettingsProvider.h>
+#include <backend/cassandra/impl/ExecutionStrategy.h>
+#include <log/Logger.h>
+#include <util/Profiler.h>
+
+#include <ripple/app/tx/impl/details/NFTokenUtils.h>
+#include <boost/asio/spawn.hpp>
+
+namespace Backend::Cassandra {
+
+/**
+ * @brief Implements @ref BackendInterface for Cassandra/Scylladb
+ *
+ * Note: this is a safer and more correct rewrite of the original implementation
+ * of the backend. We deliberately did not change the interface for now so that
+ * other parts such as ETL do not have to change at all.
+ * Eventually we should change the interface so that it does not have to know
+ * about yield_context.
+ */
+template <
+    SomeSettingsProvider SettingsProviderType,
+    SomeExecutionStrategy ExecutionStrategy>
+class BasicCassandraBackend : public BackendInterface
+{
+    clio::Logger log_{"Backend"};
+
+    SettingsProviderType settingsProvider_;
+    Schema<SettingsProviderType> schema_;
+    Handle handle_;
+
+    // have to be mutable because BackendInterface constness :(
+    mutable ExecutionStrategy executor_;
+
+    std::atomic_uint32_t ledgerSequence_ = 0u;
+
+public:
+    /**
+     * @brief Create a new cassandra/scylla backend instance.
+     *
+     * @param settingsProvider
+     */
+    BasicCassandraBackend(SettingsProviderType settingsProvider)
+        : settingsProvider_{std::move(settingsProvider)}
+        , schema_{settingsProvider_}
+        , handle_{settingsProvider_.getSettings()}
+        , executor_{settingsProvider_.getSettings(), handle_}
+    {
+        if (auto const res = handle_.connect(); not res)
+            throw std::runtime_error(
+                "Could not connect to Cassandra: " + res.error());
+        if (auto const res = handle_.execute(schema_.createKeyspace); not res)
+            throw std::runtime_error(
+                "Could not create keyspace: " + res.error());
+        if (auto const res = handle_.executeEach(schema_.createSchema); not res)
+            throw std::runtime_error("Could not create schema: " + res.error());
+
+        schema_.prepareStatements(handle_);
+        log_.info() << "Created (revamped) CassandraBackend";
+    }
+
+    /*! Not used in this implementation */
+    void
+    open([[maybe_unused]] bool readOnly) override
+    {
+    }
+
+    /*! Not used in this implementation */
+    void
+    close() override
+    {
+    }
+
+    TransactionsAndCursor
+    fetchAccountTransactions(
+        ripple::AccountID const& account,
+        std::uint32_t const limit,
+        bool forward,
+        std::optional<TransactionsCursor> const& cursorIn,
+        boost::asio::yield_context& yield) const override
+    {
+        auto rng = fetchLedgerRange();
+        if (!rng)
+            return {{}, {}};
+
+        Statement statement = [this, forward, &account]() {
+            if (forward)
+                return schema_->selectAccountTxForward.bind(account);
+            else
+                return schema_->selectAccountTx.bind(account);
+        }();
+
+        auto cursor = cursorIn;
+        if (cursor)
+        {
+            statement.bindAt(1, cursor->asTuple());
+            log_.debug() << "account = " << ripple::strHex(account)
+                         << " tuple = " << cursor->ledgerSequence
+                         << cursor->transactionIndex;
+        }
+        else
+        {
+            auto const seq = forward ? rng->minSequence : rng->maxSequence;
+            auto const placeHolder =
+                forward ? 0u : std::numeric_limits<std::uint32_t>::max();
+
+            statement.bindAt(1, std::make_tuple(placeHolder, placeHolder));
+            log_.debug() << "account = " << ripple::strHex(account)
+                         << " idx = " << seq << " tuple = " << placeHolder;
+        }
+
+        // FIXME: Limit is a hack to support uint32_t properly for the time
+        // being. Should be removed later and schema updated to use proper
+        // types.
+        statement.bindAt(2, Limit{limit});
+        auto const res = executor_.read(yield, statement);
+        auto const& results = res.value();
+        if (not results.hasRows())
+        {
+            log_.debug() << "No rows returned";
+            return {};
+        }
+
+        std::vector<ripple::uint256> hashes = {};
+        auto numRows = results.numRows();
+        log_.info() << "num_rows = " << numRows;
+
+        for (auto [hash, data] :
+             extract<ripple::uint256, std::tuple<uint32_t, uint32_t>>(results))
+        {
+            hashes.push_back(hash);
+            if (--numRows == 0)
+            {
+                log_.debug() << "Setting cursor";
+                cursor = data;
+
+                // forward queries by ledger/tx sequence `>=`
+                // so we have to advance the index by one
+                if (forward)
+                    ++cursor->transactionIndex;
+            }
+        }
+
+        auto const txns = fetchTransactions(hashes, yield);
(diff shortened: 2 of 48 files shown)


src/backend/CassandraBackendNew.h
```diff
@@ -0,0 +1,943 @@
...
+        override
+    {
+        executor_.write(
+            schema_->insertLedgerHeader, ledgerInfo.seq, std::move(header));
+
+        executor_.write(
+            schema_->insertLedgerHash, ledgerInfo.hash, ledgerInfo.seq);
+
+        // FIXME: should this be atomic or protected by lock??
+        ledgerSequence_ = ledgerInfo.seq;
```
cjcobb23: pretty sure this is only accessed by a single thread at a time. IIRC it's mainly used so you don't need to pass the ledger sequence into `doFinishWrites`. But probably worth double checking

src/backend/CassandraBackendNew.h
```diff
@@ -0,0 +1,943 @@
...
+        if (cursor)
+        {
+            statement.bindAt(1, cursor->asTuple());
+            log_.debug() << "account = " << ripple::strHex(account)
+                         << " tuple = " << cursor->ledgerSequence
+                         << cursor->transactionIndex;
+        }
+        else
+        {
+            auto const seq = forward ? rng->minSequence : rng->maxSequence;
```
cindyyan317, line 127: one thing not related to this PR: seq is not really used.

src/backend/cassandra/impl/QueryExecutor.h
```diff
@@ -0,0 +1,136 @@
...
+            else
+            {
+                if (retryPolicy_.shouldRetry(res.error()))
+                    retryPolicy_.retry(
+                        [self, &handle]() { self->execute(handle); });
+                else
+                    onComplete_(std::move(res));  // report error
+            }
+        };
+        future_.emplace(handle.asyncExecute(data_, std::move(handler)));
```
cindyyan317: Do we really need this future_? seems like we don't use it after creation.

src/backend/cassandra/impl/QueryExecutor.h
```diff
@@ -0,0 +1,136 @@
...
+ * Currently it's basically a saner implementation of the previous design that
+ * was used in production without much issue but was using raw new/delete and
+ * could leak easily. This version is slightly better but the overall design is
+ * flawed and should be reworked.
+ */
+template <
+    typename StatementType,
+    typename HandleType = Handle,
+    SomeRetryPolicy RetryPolicyType = ExponentialBackoffRetryPolicy>
+class AsyncQueryExecutor
```
cindyyan317: We use it for writing request mainly. Maybe AsyncExecutor or AsyncRetryExecutor is better than AsyncQueryExecutor?

src/rpc/RPC.cpp
```diff
@@ -346,6 +346,8 @@ buildResponse(Context const& ctx)
     if (ctx.method == "ping")
         return boost::json::object{};
 
+    // FIXME: probably not a real usecase that we should support
```
cindyyan317: 👍  I don't think we should regulate traffic here too. But understand we still need to keep this interface for now. If it will return hardcode false value, we can remove maxReadRequestsOutstanding_ from new backend then.

godexsoft: Actually, this comment is a lie as it is now - i have re-added this functionality in the new backend after i had it return `false` and forgot to update this comment. While i agree we should eventually move this somewhere else, let's keep the logic as close to what it was before as possible so that other parts of clio don't fall apart.

src/backend/cassandra/impl/QueryExecutor.h
godexsoft: Good idea. Will rename.

src/backend/cassandra/impl/QueryExecutor.h
godexsoft: This future is the second component that keeps this whole thing alive - the future is holding onto the callback that captures `self` to extend lifetime. If we don't store this future then our callback would be released leading to a crash once it's called by the cassandra-driver.

src/backend/CassandraBackendNew.h
godexsoft: Thanks for letting me know. Since `tsan` is not screaming at us anymore, i think this is called from one thread indeed. I will double check tho.

src/backend/cassandra/impl/QueryExecutor.h
cindyyan317: Yea~ Driver API gets the callback's ptr but not the object.

src/backend/CassandraBackendNew.h
godexsoft: it's only used for the log line in this same block of code.. it costs nothing to have it in a constant tho vs. inlining the ternary. i'll leave it as is for now.

src/backend/CassandraBackendNew.h
```diff
@@ -0,0 +1,942 @@
...
+
+        return std::nullopt;
+    }
+
+    std::optional<ripple::LedgerInfo>
+    fetchLedgerBySequence(
+        std::uint32_t const sequence,
+        boost::asio::yield_context& yield) const override
+    {
+        log_.info() << __func__ << " call for seq " << sequence;
```
cindyyan317: this log level is different from others

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+                });
+
+            // suspend coroutine until completion handler is called
+            result.get();
+
+            numReadRequestsOutstanding_ -= statements.size();
+
+            // it's safe to call blocking get on future here as we already
+            // waited for the coroutine to resume above.
+            if (auto res = future.get(); res)
```
cindyyan317, line 261: it will either return or throw error, while loop is not used?

godexsoft: loop is used, `throwOnError` throws conditionally - only if `res.error()` is not an error we want to retry on.

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+    sync()
+    {
+        log_.debug() << "Waiting to sync all writes...";
+        std::unique_lock<std::mutex> lck(syncMutex_);
+        syncCv_.wait(lck, [this]() { return finishedAllRequests(); });
+        log_.debug() << "Sync done.";
+    }
+
+    bool
+    isTooBusy() const
```
cindyyan317, line 118: isReadingTooBusy() more precise?

src/backend/cassandra/impl/ExecutionStrategy.h
cindyyan317: I see . this part logic is as same as the previous.

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+            // it's safe to call blocking get on future here as we already
+            // waited for the coroutine to resume above.
+            if (auto res = future.get(); res)
+            {
+                return res;
+            }
+            else
+            {
+                log_.error()
+                    << "Err during batch writes (in cb): " << res.error();
```
cindyyan317: log does not reflect the real error

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+
+            // it's safe to call blocking get on future here as we already
+            // waited for the coroutine to resume above.
+            if (auto res = future.get(); res)
+            {
+                return res;
+            }
+            else
+            {
+                log_.error() << "Err during batch writes: " << res.error();
```
cindyyan317: log does not reflect the real error

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+            });
+
+        assert(futures.size() == statements.size());
+        assert(results.size() == statements.size());
+        return results;
+    }
+
+private:
+    void
+    incrementOutstandingRequestCount()
```
cindyyan317, line 402: I feel we need to update this name, it is only for writing and the name does not indicate it might get blocked.

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+        {
+            // mutex lock required to prevent race condition around spurious
+            // wakeup
+            std::lock_guard lck(syncMutex_);
+            syncCv_.notify_one();
+        }
+    }
+
+    bool
+    canAddRequest() const
```
cindyyan317: canAddWriteRequest and finishedAllWriteRequests can do better

src/backend/cassandra/impl/ExecutionStrategy.h
```diff
@@ -0,0 +1,464 @@
...
+     */
+    [[maybe_unused]] ResultOrErrorType
+    read(
+        CompletionTokenType token,
+        std::vector<StatementType> const& statements)
+    {
+        auto handler = HandlerType{token};
+        auto result = AsyncResultType{handler};
+
+        // todo: perhaps use policy instead
```
cindyyan317, line 240: Use policy +1

src/backend/cassandra/impl/Statement.h
```diff
@@ -0,0 +1,180 @@
...
+
+    template <typename Type>
+    void
+    bindAt(std::size_t const idx, Type&& value) const
+    {
+        using std::to_string;
+        auto throwErrorIfNeeded = [idx](CassError rc, std::string_view label) {
+            if (rc != CASS_OK)
+            {
+                auto const tag = '[' + std::string{label} + ']';
```
cindyyan317: can replace by fmt

src/backend/cassandra/impl/Result.h
```diff
@@ -0,0 +1,265 @@
...
+        // row managed internally by cassandra driver, hence no ManagedObject.
+        auto const* row = cass_result_first_row(*this);
+        if (row == nullptr)
+            return std::nullopt;
+
+        std::size_t idx = 0;
+        auto advanceId = [&idx]() { return idx++; };
+
+        return std::make_optional<std::tuple<RowTypes...>>(
+            {extractColumn<RowTypes>(row, advanceId())...});
```
cindyyan317, line 148: can we use idx++ directly?

godexsoft: We can and it works in practice.. however, unfortunately, the compiler complains that the order may be undefined.. and with `-Werror` it fails to compile on gcc. So i had to add this extra step - luckily it does not actually cost anything at runtime.

src/backend/cassandra/impl/ExecutionStrategy.h
godexsoft: I'll leave this as is for now - this whole strategy should be reworked later (removing yield completely).. let's leave it for then.

src/backend/cassandra/impl/ExecutionStrategy.h
godexsoft: I'll leave this for now - it mimics the naming of BackendInterface (as well as old implementation for cassandra).

cindyyan317 (dismissed):

.githooks/pre-commit
```diff
@@ -4,7 +4,7 @@ exec 1>&2
 
 # paths to check and re-format
 sources="src unittests"
-formatter="clang-format -i"
+formatter="clang-format-11 -i"
```
cindyyan317, line 7: Do we need to update the document to instruct the required version of clang-format?
Is "clang-format-11" a custom cmd?

cindyyan317 (dismissed):
LGTM

.githooks/pre-commit
godexsoft: I think it's not a custom command, it's just a specific version of clang-format. And the alias clang-format is pointing to the selected one by default is probably the newest version. We already require clang-format v11 as per documentation (as far as i remember) but we should look at moving away from using it and use the more modern version (probably latest). Out of scope for this pr tho i think.

godexsoft:
> ## [Codecov](https://codecov.io/gh/XRPLF/clio/pull/537?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) Report
> > Merging [#537](https://codecov.io/gh/XRPLF/clio/pull/537?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) ([cc21d62](https://github.com/XRPLF/clio/commit/cc21d6243e38f4f7438ef15c6bab71570136c031)) into [develop](https://codecov.io/gh/XRPLF/clio/commit/5604b37c028fd5d8b083a381b61d60130f65f681?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) ([5604b37](https://github.com/XRPLF/clio/commit/5604b37c028fd5d8b083a381b61d60130f65f681)) will **decrease** coverage by `1.73%`.
> > The diff coverage is `n/a`.
> 
> > ❗ Current head [cc21d62](https://github.com/XRPLF/clio/commit/cc21d6243e38f4f7438ef15c6bab71570136c031) differs from pull request most recent head [f0ad4c4](https://github.com/XRPLF/clio/commit/f0ad4c4b4651a4805f69d385368c4abb6d05f7c4). Consider uploading reports for the commit [f0ad4c4](https://github.com/XRPLF/clio/commit/f0ad4c4b4651a4805f69d385368c4abb6d05f7c4) to get more accurate results
> 
> ```diff
> @@             Coverage Diff             @@
> ##           develop     #537      +/-   ##
> ===========================================
> - Coverage    16.84%   15.11%   -1.73%     
> ===========================================
>   Files           99      120      +21     
>   Lines         9203    10148     +945     
>   Branches      5688     6164     +476     
> ===========================================
> - Hits          1550     1534      -16     
> - Misses        6293     7274     +981     
> + Partials      1360     1340      -20     
> ```
> 
> [see 52 files with indirect coverage changes](https://codecov.io/gh/XRPLF/clio/pull/537/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF)
> 
> Help us with your feedback. Take ten seconds to tell us [how you rate us](https://about.codecov.io/nps?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF). Have a feature suggestion? [Share it here.](https://app.codecov.io/gh/feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF)

While this is annoying, it's true that not the entire backend implementation is covered. When we delete old backend code this number will improve a lot tho. We will probably be adding more tests for the backend later on as well as possibly exclude it from clio (move it to a separate project perhaps).

cindyyan317 (approved):

PR #540, opened 2023-03-15
Custom error validator

cindyyan317:
For the same validator, the error message might be different according to the field location, eg https://xrpl.org/book_offers.html#possible-errors
Add this validator to send out a custom error when validation fails.

### DIFF
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -392,8 +392,8 @@ public:
      * @brief Constructs a validator that calls the given validator "req" and
      * return customized error "err"
      */
-    WithCustomError(Requirement const& req, RPC::Status const& err)
-        : requirement{req}, error{err}
+    WithCustomError(Requirement req, RPC::Status err)
+        : requirement{std::move(req)}, error{err}
     {
     }
 
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -378,6 +378,36 @@ private:
 };
 
 /**
+ * @brief A meta-validator that wrapp other validator to send the customized
+ * error
+ */
+template <typename Requirement>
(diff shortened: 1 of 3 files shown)


src/rpc/common/Validators.h
```diff
@@ -377,6 +377,36 @@ class IfType final
...
+{
+    Requirement requirement;
+    RPC::Status error;
+
+public:
+    /**
+     * @brief Constructs a validator that calls the given validator "req" and
+     * return customized error "err"
+     */
+    WithCustomError(Requirement const& req, RPC::Status const& err)
```
godexsoft: I think `Requirement&& req` is a better choice here in combination with `std::forward<Requirement>(req)`.
`err` could be passed by value or `rvalue ref` + `std::move` too. Value is probably better choice here for copy elision.

godexsoft (changes requested):
Small issue found, let's fix and should be good to go :)

src/rpc/common/Validators.h
cindyyan317: Requirement is a class template parameter, so Requirement&& can only reduced as rvalue .
`error: no viable constructor or deduction guide for deduction of template arguments of 'WithCustomError'
         WithCustomError{
         ^
/Users/cyan/code/clio/src/rpc/common/Validators.h:395:5: note: candidate function [with Requirement = RPCng::validation::CustomValidator] not viable: expects an rvalue for 1st argument
    WithCustomError(Requirement&& req, RPC::Status err)`

I will add a lvalue ctor

cindyyan317: Discuss offline, we will use value and move inside. Because we have some left value validators

godexsoft (approved):
LGTM :)

PR #542, opened 2023-03-16
Implement book_offers in new RPC framework

cindyyan317:
1 Implement book_offers in new RPC framework
This implementation is same as the previous book_offers. Both clio and rippled list all the offers regardless the funding status, which mismatches the document.
An open issue : https://github.com/XRPLF/xrpl-dev-portal/issues/1818
2 Remove unused parameters
3 Add IssuerValidator

### DIFF
--- a/src/rpc/common/Validators.cpp
+++ b/src/rpc/common/Validators.cpp
@@ -22,6 +22,7 @@
 #include <rpc/common/Validators.h>
 
 #include <boost/json/value.hpp>
+#include <fmt/core.h>
 
 #include <charconv>
 #include <string_view>
@@ -197,6 +198,29 @@ CustomValidator CurrencyValidator = CustomValidator{
         if (!ripple::to_currency(currency, value.as_string().c_str()))
             return Error{RPC::Status{
                 RPC::ClioError::rpcMALFORMED_CURRENCY, "malformedCurrency"}};
+        return MaybeError{};
+    }};
+
+CustomValidator IssuerValidator = CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_string())
+            return Error{RPC::Status{
+                RPC::RippledError::rpcINVALID_PARAMS,
+                std::string(key) + "NotString"}};
+        ripple::AccountID issuer;
+        if (!ripple::to_issuer(issuer, value.as_string().c_str()))
+            return Error{RPC::Status{
+                // TODO: need to align with the error
+                RPC::RippledError::rpcINVALID_PARAMS,
+                fmt::format("Invalid field '{}', bad issuer.", key)}};
+
+        if (issuer == ripple::noAccount())
+            return Error{RPC::Status{
+                RPC::RippledError::rpcINVALID_PARAMS,
+                fmt::format(
+                    "Invalid field '{}', bad issuer account "
+                    "one.",
+                    key)}};
         return MaybeError{};
     }};
 
--- a/src/rpc/handlers/BookOffers.cpp
+++ b/src/rpc/handlers/BookOffers.cpp
@@ -83,13 +83,9 @@ doBookOffers(Context const& context)
     if (auto const status = getTaker(request, takerID); status)
         return status;
 
-    ripple::uint256 marker = beast::zero;
-    if (auto const status = getHexMarker(request, marker); status)
-        return status;
-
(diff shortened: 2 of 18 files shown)


src/rpc/common/Validators.cpp
```diff
@@ -200,4 +201,27 @@ CustomValidator CurrencyValidator = CustomValidator{
...
+CustomValidator IssuerValidator = CustomValidator{
+    [](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_string())
+            return Error{RPC::Status{
+                RPC::RippledError::rpcINVALID_PARAMS,
+                std::string(key) + "NotString"}};
+        ripple::AccountID issuer;
+        if (!ripple::to_issuer(issuer, value.as_string().c_str()))
+            return Error{RPC::Status{
+                // TODO: need to align with the error
```
godexsoft, line 213: What does this mean and can this be done now or it's not possible and this is a todo for future?

src/rpc/handlers/BookOffers.cpp
```diff
@@ -83,13 +83,9 @@ doBookOffers(Context const& context)
     if (auto const status = getTaker(request, takerID); status)
         return status;
 
-    ripple::uint256 marker = beast::zero;
-    if (auto const status = getHexMarker(request, marker); status)
-        return status;
-
     auto start = std::chrono::system_clock::now();
```
godexsoft, line 86: Can we use `util::timed` here?

unittests/rpc/handlers/BookOffersTest.cpp
```diff
@@ -0,0 +1,1327 @@
...
+generateParameterBookOffersTestBundles()
+{
+    return std::vector<ParameterTestBundle>{
+        ParameterTestBundle{
+            "MissingTakerGets",
+            R"({
+                "taker_pays" : {
+                    "currency" : "USD",
+                    "issuer" : "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
+                    }
```
godexsoft: minor: Again some strange formatting here and below in other json strings. Consider fixing :)

src/rpc/ngHandlers/BookOffers.h
```diff
@@ -0,0 +1,118 @@
...
+class BookOffersHandler
+{
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Output
+    {
+        std::string ledgerHash;
+        uint32_t ledgerIndex;
+        boost::json::array offers;
```
godexsoft, line 38: Can you add a `TODO: use std::vector here instead, serialize in tag_invoke` here please? Let's try to avoid `boost::json` in handler input/output.

godexsoft (changes requested):
Adding a few comments and questions. Overall this looks good and a huge effort! congrats 🥳

src/rpc/ngHandlers/BookOffers.h
cindyyan317: We do need to use json object in output sometimes, because the output like transaction, its fields is not certain , it is hard to create a strong type to hold it.

src/rpc/handlers/BookOffers.cpp
cindyyan317: I checked this case before, it measures multiple stages.

src/rpc/common/Validators.cpp
cindyyan317: Ha. This validator is used for multiple places, they give out different error message according to the upper level field they are in. eg
"taker_pays" : {
                    "currency" : "USD",
                    "issuer" : "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
                    }
"taker_gets" : {
                    "currency" : "USD",
                    "issuer" : "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
                    }
The error message would like "taker_pays.issuer malformed". We will address all the mismatch in future. See if it is worth to change the framework to adjust this case.

godexsoft (approved):
LGTM 🚀

src/rpc/ngHandlers/BookOffers.h
godexsoft: We can use std::variant for that later. It's ok to keep it for future refinements.

PR #544, opened 2023-03-20
Fix subscription forward issue

cindyyan317:
The etl balancer holds probe etl source, when it compares with the actual etl source by address, it will always return false.
This PR change how we compare etl source.
before fix:
Screenshot 2023-03-20 at 15 13 49
after fix:
Screenshot 2023-03-20 at 15 13 03

### DIFF
--- a/src/etl/ETLSource.h
+++ b/src/etl/ETLSource.h
@@ -312,11 +312,13 @@ public:
         , subscriptions_(subscriptions)
         , balancer_(balancer)
         , forwardCache_(config, ioContext, *this)
-        , uuid_(boost::uuids::random_generator()())
         , ioc_(ioContext)
         , timer_(ioContext)
         , hooks_(hooks)
     {
+        static boost::uuids::random_generator uuidGenerator;
+        uuid_ = uuidGenerator();
+
         ip_ = config.valueOr<std::string>("ip", {});
         wsPort_ = config.valueOr<std::string>("ws_port", {});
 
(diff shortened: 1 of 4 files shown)


src/etl/ETLSource.h
```diff
@@ -323,6 +341,7 @@ class ETLSourceImpl : public ETLSource
                              << " . Remote = " << toString();
             }
         }
+        tokenStr_ = fmt::format("{}:{}:{}", ip_, wsPort_, grpcPort_);
```
godexsoft: I'm not sure we can do it this way: i think it's a legitimate usecase to have more than one of the same ETL servers configured in the config so that they just participate in load balancing on clio side. I suppose in this situation your code will get confused. 
What i think we can do is implement `operator==` properly and compare `this` pointer inside. Comparing two Probing sources would just use the underlying source's `operator==`.
Comparing by `this` is not that bad if it's hidden inside. Comparing by `shared_ptr.get()` on the other hand is really bad.

godexsoft (changes requested):
Let me know what you think of my suggestion 🚀

src/etl/ETLSource.h
cindyyan317: Even multiple ETL servers configured in the config, they can be distinguished by ip+port. 
Here the comparison is not among two probe sources, it is between one probe and one underlying source. Let us discuss offline.

godexsoft (approved):
LGTM! 🚀

PR #548, opened 2023-03-21
Implement ledger_range rpc handler

godexsoft:
Implements the nextgen rpc handler for `ledger_range` along with some unit-tests.

### DIFF
--- a/src/rpc/ngHandlers/LedgerRange.h
+++ b/src/rpc/ngHandlers/LedgerRange.h
@@ -33,7 +33,11 @@ class LedgerRangeHandler
     std::shared_ptr<BackendInterface> sharedPtrBackend_;
 
 public:
-    using Output = Backend::LedgerRange;
+    struct Output
+    {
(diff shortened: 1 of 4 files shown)


src/rpc/ngHandlers/LedgerRange.h
```diff
@@ -0,0 +1,59 @@
...
+#include <memory>
+
+namespace RPCng {
+
+class LedgerRangeHandler
+{
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Output
```
godexsoft, line 36: Reason it's not a `using Output = Backend::LedgerRange;` is that ADL would require us to have `tag_invoke` in the `Backend` namespace in this case - I did not really want to leak this implementation detail into an unrelated namespace.

cindyyan317:
LGTM 👍

cindyyan317 (approved):

PR #551, opened 2023-03-23
Implement account_lines with new RPC framework

### LINKED ISSUE #550: Implement account_lines in new RPC system
Implement nextgen handler for `account_lines`.

godexsoft:
Fixes #550

### DIFF
--- a/src/rpc/ngHandlers/AccountLines.cpp
+++ b/src/rpc/ngHandlers/AccountLines.cpp
@@ -1,0 +1,269 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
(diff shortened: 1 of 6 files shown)


src/rpc/ngHandlers/AccountLines.cpp
```diff
@@ -0,0 +1,269 @@
...
+    auto const accountID = RPC::accountFromStringStrict(input.account);
+    auto const accountLedgerObject = sharedPtrBackend_->fetchLedgerObject(
+        ripple::keylet::account(*accountID).key, lgrInfo.seq, yield);
+
+    if (not accountLedgerObject)
+        return Error{RPC::Status{
+            RPC::RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
+
+    auto const peerAccountID = input.peer
+        ? RPC::accountFromStringStrict(input.peer.value())
```
cindyyan317: *(input.peer) maybe more efficient than value()  .

src/rpc/ngHandlers/AccountLines.h
```diff
@@ -0,0 +1,135 @@
...
+    struct LineResponse
+    {
+        std::string account;
+        std::string balance;
+        std::string currency;
+        std::string limit;
+        std::string limitPeer;
+        uint32_t qualityIn;
+        uint32_t qualityOut;
+        std::optional<bool> noRipple;
```
cindyyan317: no need to be optional

src/rpc/ngHandlers/AccountLines.h
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    struct Input
+    {
+        std::string account;
+        std::optional<std::string> ledgerHash;
+        std::optional<uint32_t> ledgerIndex;
+        std::optional<std::string> peer;
+        std::optional<bool>
+            ignoreDefault;  // TODO: document
```
cindyyan317: we can just set its default value instead using optional

cindyyan317 (dismissed):
Just leave some minor things to check.

src/rpc/ngHandlers/AccountLines.cpp
godexsoft: because UB instead of exception? i don't think it's a huge performance boost but i'll change as you suggest 👍

cindyyan317 (approved):

PR #553, opened 2023-03-23
Transaction entry with new RPC framework

cindyyan317:


src/rpc/ngHandlers/TransactionEntry.cpp
```diff
@@ -0,0 +1,109 @@
...
+    // is in a different ledger than the one specified.
+    if (!dbRet || dbRet->ledgerSequence != lgrInfo.seq)
+        return Error{RPC::Status{
+            RPC::RippledError::rpcTXN_NOT_FOUND,
+            "transactionNotFound",
+            "Transaction not found."}};
+
+    auto const [txn, meta] = RPC::toExpandedJson(*dbRet);
+    TransactionEntryHandler::Output output;
+    output.tx = std::move(txn);
```
godexsoft, line 61: I think these "moves" may not "move" because of the `const` in `auto const [txn, meta]` above.

unittests/rpc/handlers/TransactionEntryTest.cpp
```diff
@@ -0,0 +1,276 @@
...
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+constexpr static auto CURRENCY = "0158415500000000C1F76FF6ECB0BAC600000000";
+
+class RPCTransactionEntryHandlerTest : public HandlerBaseTest
+{
+};
+
+TEST_F(RPCTransactionEntryHandlerTest, TxHashNotProvide)
+{
+    runSpawn([this](boost::asio::yield_context yield) {
```
godexsoft: minor: This can be `auto& yield` for less typing and less copying (probably very cheap eitherway).

godexsoft (changes requested):
Left a couple of comments. All looks good overall! 💪

src/rpc/ngHandlers/TransactionEntry.cpp
cindyyan317: ha. nice to know this

godexsoft (approved):
LGTM 🚀

PR #554, opened 2023-03-24
noripple_check implementation of new RPC system 

cindyyan317:
noripple_check implementation of new RPC system

### DIFF
--- a/src/rpc/ngHandlers/NoRippleCheck.cpp
+++ b/src/rpc/ngHandlers/NoRippleCheck.cpp
@@ -74,7 +74,7 @@ NoRippleCheckHandler::process(
         return tx;
     };
 
-    if (bDefaultRipple & !input.roleGateway)
+    if (bDefaultRipple && !input.roleGateway)
     {
         output.problems.push_back(
             "You appear to have set your default ripple flag even though "
@@ -82,7 +82,7 @@ NoRippleCheckHandler::process(
             "are not a gateway. This is not recommended unless you are "
             "experimenting");
     }
-    else if (input.roleGateway & !bDefaultRipple)
+    else if (input.roleGateway && !bDefaultRipple)
     {
         output.problems.push_back(
             "You should immediately set your default ripple flag");
@@ -114,12 +114,12 @@ NoRippleCheckHandler::process(
 
                 std::string problem;
                 bool needFix = false;
-                if (bNoRipple & input.roleGateway)
+                if (bNoRipple && input.roleGateway)
                 {
                     problem = "You should clear the no ripple flag on your ";
                     needFix = true;
                 }
-                else if (!bNoRipple & !input.roleGateway)
+                else if (!bNoRipple && !input.roleGateway)
                 {
                     problem =
                         "You should probably set the no ripple flag on "
@@ -171,30 +171,30 @@ tag_invoke(
 {
     auto const& jsonObject = jv.as_object();
     NoRippleCheckHandler::Input input;
-    input.account = jv.at(JS(account)).as_string().c_str();
-    input.roleGateway = jv.at(JS(role)).as_string() == "gateway";
+    input.account = jsonObject.at(JS(account)).as_string().c_str();
+    input.roleGateway = jsonObject.at(JS(role)).as_string() == "gateway";
     if (jsonObject.contains(JS(limit)))
     {
-        input.limit = jv.at(JS(limit)).as_int64();
+        input.limit = jsonObject.at(JS(limit)).as_int64();
     }
     if (jsonObject.contains(JS(transactions)))
     {
-        input.transactions = jv.at(JS(transactions)).as_bool();
+        input.transactions = jsonObject.at(JS(transactions)).as_bool();
     }
     if (jsonObject.contains(JS(ledger_hash)))
     {
-        input.ledgerHash = jv.at(JS(ledger_hash)).as_string().c_str();
+        input.ledgerHash = jsonObject.at(JS(ledger_hash)).as_string().c_str();
     }
     if (jsonObject.contains(JS(ledger_index)))
     {
         if (!jsonObject.at(JS(ledger_index)).is_string())
         {
-            input.ledgerIndex = jv.at(JS(ledger_index)).as_int64();
+            input.ledgerIndex = jsonObject.at(JS(ledger_index)).as_int64();
         }
         else if (jsonObject.at(JS(ledger_index)).as_string() != "validated")
         {
             input.ledgerIndex =
-                std::stoi(jv.at(JS(ledger_index)).as_string().c_str());
+                std::stoi(jsonObject.at(JS(ledger_index)).as_string().c_str());
         }
     }
 
--- a/src/rpc/ngHandlers/NoRippleCheck.cpp
+++ b/src/rpc/ngHandlers/NoRippleCheck.cpp
@@ -1,0 +1,220 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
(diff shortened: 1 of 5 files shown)


src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,221 @@
...
+        input.ledgerHash,
+        input.ledgerIndex,
+        range->maxSequence);
+
+    if (auto status = std::get_if<RPC::Status>(&lgrInfoOrStatus))
+        return Error{*status};
+
+    auto const lgrInfo = std::get<ripple::LedgerInfo>(lgrInfoOrStatus);
+    auto const accountID = RPC::accountFromStringStrict(input.account);
+    auto keylet = ripple::keylet::account(*accountID).key;
```
godexsoft: Could this also be `const`?

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,221 @@
...
+        sharedPtrBackend_->fetchLedgerObject(keylet, lgrInfo.seq, yield);
+    if (!accountObj)
+        return Error{RPC::Status{
+            RPC::RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
+
+    ripple::SerialIter it{accountObj->data(), accountObj->size()};
+    ripple::SLE sle{it, keylet};
+    std::uint32_t accountSeq = sle.getFieldU32(ripple::sfSequence);
+
+    bool bDefaultRipple =
```
godexsoft: Looks like can be `const` also.

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,221 @@
...
+NoRippleCheckHandler::Input
+tag_invoke(
+    boost::json::value_to_tag<NoRippleCheckHandler::Input>,
+    boost::json::value const& jv)
+{
+    auto const& jsonObject = jv.as_object();
+    NoRippleCheckHandler::Input input;
+    input.account = jv.at("account").as_string().c_str();
+    input.roleGateway = jv.at("role").as_string() == "gateway";
+    if (jsonObject.contains("limit"))
```
godexsoft: minor: Consider using JStrings for all these also

src/rpc/ngHandlers/NoRippleCheck.h
```diff
@@ -0,0 +1,98 @@
...
+    NoRippleCheckHandler(
+        std::shared_ptr<BackendInterface> const& sharedPtrBackend)
+        : sharedPtrBackend_(sharedPtrBackend)
+    {
+    }
+
+    RpcSpecConstRef
+    spec() const
+    {
+        static const RpcSpec rpcSpec = {
```
godexsoft: minor: This was my original mistake but i tried to fix in new PRs so consider also fixing: `static auto const rpcSpec = RpcSpec{...}`

src/rpc/ngHandlers/NoRippleCheck.h
```diff
@@ -0,0 +1,98 @@
...
+            {"account", validation::Required{}, validation::AccountValidator},
```
godexsoft: minor: Can probably utilise JStrings here too

godexsoft (changes requested):
Leaving a few suggestions. Overall very good 👍

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+    auto const getBaseTx = [&](ripple::AccountID const& accountID,
+                               std::uint32_t accountSeq) {
+        boost::json::object tx;
+        tx[JS(Sequence)] = accountSeq;
+        tx[JS(Account)] = ripple::toBase58(accountID);
+        tx[JS(Fee)] = RPC::toBoostJson(fees->units.jsonClipped());
+        return tx;
+    };
+
+    if (bDefaultRipple & !input.roleGateway)
```
godexsoft: Although this yields the same result in this case, I suppose the intention was to use `&&`, not `&` here.

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+    {
+        output.problems.push_back(
+            "You appear to have set your default ripple flag even though "
+            "you "
+            "are not a gateway. This is not recommended unless you are "
+            "experimenting");
+    }
+    else if (input.roleGateway & !bDefaultRipple)
```
godexsoft: Same here.

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+                    (bLow ? ripple::lsfLowNoRipple : ripple::lsfHighNoRipple);
+
+                std::string problem;
+                bool needFix = false;
+                if (bNoRipple & input.roleGateway)
+                {
+                    problem = "You should clear the no ripple flag on your ";
+                    needFix = true;
+                }
+                else if (!bNoRipple & !input.roleGateway)
```
godexsoft: And here.

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+            {
+                bool const bLow = accountID ==
+                    ownedItem.getFieldAmount(ripple::sfLowLimit).getIssuer();
+
+                bool const bNoRipple = ownedItem.getFieldU32(ripple::sfFlags) &
+                    (bLow ? ripple::lsfLowNoRipple : ripple::lsfHighNoRipple);
+
+                std::string problem;
+                bool needFix = false;
+                if (bNoRipple & input.roleGateway)
```
godexsoft: And here.

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+NoRippleCheckHandler::Input
+tag_invoke(
+    boost::json::value_to_tag<NoRippleCheckHandler::Input>,
+    boost::json::value const& jv)
+{
+    auto const& jsonObject = jv.as_object();
+    NoRippleCheckHandler::Input input;
+    input.account = jv.at(JS(account)).as_string().c_str();
```
godexsoft: I wonder, why not use jsonObject here too?

src/rpc/ngHandlers/NoRippleCheck.cpp
```diff
@@ -0,0 +1,220 @@
...
+    input.roleGateway = jv.at(JS(role)).as_string() == "gateway";
+    if (jsonObject.contains(JS(limit)))
+    {
+        input.limit = jv.at(JS(limit)).as_int64();
```
godexsoft: And also here and in similar situations below.

src/rpc/ngHandlers/NoRippleCheck.h
```diff
@@ -0,0 +1,99 @@
...
+{
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Output
+    {
+        std::string ledgerHash;
+        uint32_t ledgerIndex;
+        std::vector<std::string> problems;
+        std::optional<boost::json::array> transactions;
```
godexsoft, line 43: Please add `TODO` to get rid of json in here

unittests/rpc/handlers/NoRippleCheckTest.cpp
```diff
@@ -0,0 +1,737 @@
...
+    NormalPathRoleUserDefaultRippleUnsetTrustLineNoRippleUnSet)
+{
+    static auto constexpr seq = 30;
+    static auto constexpr expectedOutput =
+        R"({
+            "ledger_hash":"4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652",
+            "ledger_index":30,
+            "problems":[
+                "You should probably set the no ripple flag on your USD line to rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun",
+                "You should probably set the no ripple flag on your USD line to rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun"
```
godexsoft, line 325: Is it really supposed to output the same problem twice or is this a bug in the original code?

godexsoft (changes requested):
Found some more little things to improve and/or check 👍

unittests/rpc/handlers/NoRippleCheckTest.cpp
cindyyan317: Mock the same trustline twice , so it should be two problems here

src/rpc/ngHandlers/NoRippleCheck.cpp
cindyyan317: json object and value both have at() function. jsonObject here is mainly for contains() function.

godexsoft (dismissed):
LGTM! 🚀

godexsoft (approved):
Go Go 🔥

godexsoft (approved):
Go go

PR #558, opened 2023-03-24
Implement nextgen nft_info handler

### LINKED ISSUE #557: Implement nextgen nft_info handler
Need to implement `nft_info` handler for new RPC system

godexsoft:
Fixes #557


src/rpc/ngHandlers/NFTInfo.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace RPCng {
+
+NFTInfoHandler::Result
+NFTInfoHandler::process(
+    NFTInfoHandler::Input input,
+    boost::asio::yield_context& yield) const
+{
+    ripple::uint256 tokenID;
+    [[maybe_unused]] auto const _ = tokenID.parseHex(input.nftID);
```
cindyyan317: ripple::uint256 tokenID{input.nftID.c_str()}; does the same thing.
can be move to line 48, after the ledger check.

godexsoft: it was crashing.. let me try with `.c_str()`.

unittests/rpc/handlers/NFTInfoTest.cpp
```diff
@@ -0,0 +1,449 @@
...
+    mockBackendPtr->updateRange(10);  // min
+    mockBackendPtr->updateRange(30);  // max
+    auto ledgerinfo = CreateLedgerInfo(LEDGERHASH, 30);
+    ON_CALL(*rawBackendPtr, fetchLedgerByHash(ripple::uint256{LEDGERHASH}, _))
+        .WillByDefault(Return(ledgerinfo));
+    EXPECT_CALL(*rawBackendPtr, fetchLedgerByHash).Times(1);
+    // fetch nft return emtpy
+    ON_CALL(*rawBackendPtr, fetchNFT)
+        .WillByDefault(Return(std::optional<NFT>{}));
+    EXPECT_CALL(*rawBackendPtr, fetchNFT).Times(1);
```
cindyyan317: I suggest we can add matcher here to verify the handler calls fetchNFT correctly. Without it , we can not catch the error like token is parsed incorrectly (I made this mistake once by using ripple::uint256{string})or not use the specific ledger seq

cindyyan317 (approved):
💯

PR #562, opened 2023-03-27
account_tx of new RPC framework

cindyyan317:


src/backend/Types.h
```diff
@@ -70,6 +70,12 @@ struct TransactionsCursor
 {
     std::uint32_t ledgerSequence;
     std::uint32_t transactionIndex;
+    bool
+    operator==(const TransactionsCursor& other) const
```
godexsoft: `TransactionsCursor const&` please :D

src/rpc/ngHandlers/AccountTx.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+    else
+    {
+        if (input.forward)
+            cursor = {minIndex, 0};
+        else
+            cursor = {maxIndex, INT32_MAX};
+    }
+    auto constexpr limitDefault = 50;
+    auto const limit = input.limit ? *input.limit : limitDefault;
```
godexsoft: Consider `input.limit.value_or(limitDefault);` - it is more readable.

src/rpc/ngHandlers/AccountTx.cpp
```diff
@@ -0,0 +1,223 @@
...
+        {
+            gLog.debug() << "Skipping over transactions from incomplete ledger";
+            continue;
+        }
+
+        boost::json::object obj;
+        if (!input.binary)
+        {
+            auto [txn, meta] = RPC::toExpandedJson(txnPlusMeta);
+            obj[JS(meta)] = meta;
```
godexsoft: minor: These two probably should be moved

src/rpc/ngHandlers/AccountTx.cpp
```diff
@@ -0,0 +1,223 @@
...
+            obj[JS(tx_blob)] = ripple::strHex(txnPlusMeta.transaction);
+            obj[JS(ledger_index)] = txnPlusMeta.ledgerSequence;
+            // only clio has this field
+            obj[JS(date)] = txnPlusMeta.date;
+        }
+        obj[JS(validated)] = true;
+        response.transactions.push_back(obj);
+    }
+
+    if (input.limit)
```
godexsoft, line 141: I wonder if it's better to copy unconditionally here. Result is the same but one less branch.

src/rpc/ngHandlers/AccountTx.cpp
```diff
@@ -0,0 +1,223 @@
...
+{
+    jv = {
+        {JS(account), output.account},
+        {JS(ledger_index_min), output.ledgerIndexMin},
+        {JS(ledger_index_max), output.ledgerIndexMax},
+        {JS(transactions), output.transactions},
+        {JS(validated), output.validated}};
+    if (output.marker)
+        jv.as_object()[JS(marker)] = {
+            {JS(ledger), output.marker->ledger}, {JS(seq), output.marker->seq}};
```
godexsoft: minor: Why not overload `tag_invoke` for `Marker`?

src/rpc/ngHandlers/AccountTx.cpp
```diff
@@ -0,0 +1,223 @@
...
+    boost::json::value_to_tag<AccountTxHandler::Input>,
+    boost::json::value const& jv)
+{
+    auto const& jsonObject = jv.as_object();
+    AccountTxHandler::Input input;
+    input.account = jv.at(JS(account)).as_string().c_str();
+    if (jsonObject.contains(JS(ledger_index_min)) &&
+        jv.at(JS(ledger_index_min)).as_int64() != -1)
+    {
+        input.ledgerIndexMin = jv.at(JS(ledger_index_min)).as_int64();
```
godexsoft: minor: I'm not a huge fan of mixing `jsonObject` and `jv` - i'd prefer to just use `jsonObject` if we have it already. Up to you.

src/rpc/ngHandlers/AccountTx.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include <boost/asio/spawn.hpp>
+
+namespace RPCng {
+class AccountTxHandler
+{
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Marker
```
godexsoft, line 35: This could be useful for other APIs too or not really?

unittests/rpc/handlers/AccountTxTest.cpp
```diff
@@ -0,0 +1,778 @@
...
+                }
+            })",
+            "invalidParams",
+            "Invalid parameters."},
+        AccountTxParamTestCaseBundle{
+            "MarkerSeqNotInt",
+            R"({
+                "account":"rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
+                 "marker": 
+                    {
```
godexsoft: Formatting is off here compared to above

godexsoft (changes requested):
A few comments and suggestions 🖖

src/rpc/ngHandlers/AccountTx.cpp
cindyyan317: The original logic is : when there is no limit in input , we don't have limit in output.

src/rpc/ngHandlers/AccountTx.h
cindyyan317: So far, we only have this API use this format of marker. will move it to common place when other API use it.

src/rpc/ngHandlers/AccountTx.cpp
godexsoft: since `output.limit` is `std::optional` this logic would remain the same if you just copy the value over (e.g. if `input.limit` is nullopt => `output.limit` will be too). I don't mind if you leave it as is either - it's very minor.

godexsoft (approved):
LGTM :)

PR #568, opened 2023-03-28
Implement nft_buy_offers and tests

### LINKED ISSUE #564: Implement nft_buy_offers in nextgen RPC system
Implement nextgen handler for `nft_buy_offers`.

godexsoft:
Fixes #564

### DIFF
--- a/src/rpc/handlers/NFTOffers.cpp
+++ b/src/rpc/handlers/NFTOffers.cpp
@@ -33,7 +33,7 @@ namespace json = boost::json;
 
 namespace ripple {
 
-void
+inline void
 tag_invoke(json::value_from_tag, json::value& jv, SLE const& offer)
 {
     auto amount = ::RPC::toBoostJson(
--- a/src/rpc/ngHandlers/NFTBuyOffers.cpp
+++ b/src/rpc/ngHandlers/NFTBuyOffers.cpp
@@ -1,0 +1,226 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <rpc/RPCHelpers.h>
+#include <rpc/ngHandlers/NFTBuyOffers.h>
+
+#include <ripple/app/tx/impl/details/NFTokenUtils.h>
+#include <ripple/protocol/Indexes.h>
+
+using namespace ripple;
+using namespace ::RPC;
+
+namespace ripple {
+
+// TODO: move to some common serialization impl place
+inline void
+tag_invoke(
+    boost::json::value_from_tag,
+    boost::json::value& jv,
+    SLE const& offer)
+{
+    auto amount = ::RPC::toBoostJson(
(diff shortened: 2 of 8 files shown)


src/rpc/handlers/NFTOffers.cpp
```diff
@@ -33,7 +33,7 @@ namespace json = boost::json;
 
 namespace ripple {
 
-void
+inline void
```
cindyyan317, line 36: does "inline" really work if it is used in cpp file?

cindyyan317:
1 Can we put the common logic to base class ,then nft_sell_offers can reuse that part?
2 The marker logic seems not right for both old implementation and new implementation. If I understand correctly, the marker is the last element returning to the requester, the following response also contains this offer. I don't understand why offers size changes according to the marker status. WDYT?
 **ha i saw the logic, it pops the last one out before return**

src/rpc/ngHandlers/NFTBuyOffers.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    if (input.marker)
+    {
+        cursor = uint256(input.marker->c_str());
+
+        // We have a start point. Use limit - 1 from the result and use the very
+        // last one for the resume.
+        auto const sle =
+            [this, &cursor, &lgrInfo, &yield]() -> std::shared_ptr<SLE const> {
+            auto key = keylet::nftoffer(cursor).key;
```
cindyyan317: const?

src/rpc/ngHandlers/NFTBuyOffers.cpp
```diff
@@ -0,0 +1,226 @@
...
+        *sharedPtrBackend_,
+        directory,
+        cursor,
+        startHint,
+        lgrInfo.seq,
+        reserve,
+        {},
+        yield,
+        [&offers](ripple::SLE&& offer) {
+            if (offer.getType() == ripple::ltNFTOKEN_OFFER)
```
cindyyan317, line 139: the directory is only for the nft buy offers, do we still need to check the type here?
If other type is possible, we can not use size check to include marker field (line 151).

src/rpc/ngHandlers/NFTBuyOffers.cpp
```diff
@@ -0,0 +1,226 @@
...
+                return true;
+            }
+
+            return false;
+        });
+
+    if (auto status = std::get_if<Status>(&result))
+        return Error{*status};
+
+    if (offers.size() == reserve)
```
cindyyan317, line 151: If there are other types in this directory, the size might be less than reserve, but we still need to include the marker.

cindyyan317 (commented):
Send back just to confirm if we need a baseclass for sell offer and buy offer

src/rpc/ngHandlers/NFTBuyOffers.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    boost::json::object obj = {
+        {JS(nft_offer_index), to_string(offer.key())},
+        {JS(flags), offer[sfFlags]},
+        {JS(owner), toBase58(offer.getAccountID(sfOwner))},
+        {JS(amount), std::move(amount)},
+    };
+
+    if (offer.isFieldPresent(sfDestination))
+        obj.insert_or_assign(
```
cindyyan317, line 49: other places use [], wonder why we use insert_or_assign here?

src/rpc/ngHandlers/NFTBuyOffers.cpp
godexsoft: I have no idea - i just copy the previous implementation and make it testable :)

src/rpc/ngHandlers/NFTBuyOffers.cpp
godexsoft: Good point. I copied the code tho so not sure if this is a bug or not. We can deal with it after all rpc is migrated and tested i think.

src/rpc/ngHandlers/NFTBuyOffers.cpp
godexsoft: I think it's more efficient than [] either way and we probably should use it more. But as with other things - this code is just copied as is.

src/rpc/ngHandlers/NFTBuyOffers.cpp
godexsoft: yep, could be 👍

godexsoft:
> Send back just to confirm if we need a baseclass for sell offer and buy offer

Yes or some other way for common logic. But i think it's better refactored in the next issue that covers `nft_sell_offers` rpc so i left it in one place for now.

godexsoft:
> 1 Can we put the common logic to base class ,then nft_sell_offers can reuse that part? 2 The marker logic seems not right for both old implementation and new implementation. If I understand correctly, the marker is the last element returning to the requester, the following response also contains this offer. I don't understand why offers size changes according to the marker status. WDYT? **ha i saw the logic, it pops the last one out before return**

1 - yes, in the next issue :)
2 - idk. i'm not sure whether `nft*` marker/paging is tested by integration or not but there are some tests that refer to `nft_buy_offers` so i'd imagine this part is tested to some degree. eitherway i don't think we need to focus on this now while our goal is to make rpc testable and cover with tests as much as we can. let's leave fixing paging/marker till after migration is done completely, wdyt?

cindyyan317 (approved):
LGTM. Just checked the integration test, we have tested on marker and limit field.

PR #572, opened 2023-03-30
Add experimental code integrating nextgen

### LINKED ISSUE #592: Integrate nextgen RPC into main code
Currently nextgen RPC handlers are tested by unit tests but not actually used by `clio`. 
This issue is for 
* Integrate all nextgen RPC handlers into `clio`
* Rename RPCng to RPC (namespace)
* Remove old code entirely

godexsoft:
Not the cleanest integration but it will do for now. Later when we rewrite some other parts we can add unit-tests for those - workqueue, counters, rpcEngine, etc.

Fixes #592


godexsoft:
On hold until all remaining RPC handlers are migrated and we can drop old stuff.

src/rpc/Counters.cpp
```diff
@@ -69,27 +65,26 @@ Counters::rpcComplete(std::string const& method, std::chrono::microseconds const
 void
 Counters::rpcForwarded(std::string const& method)
 {
-    if (!validHandler(method))
```
cindyyan317, line 72: The method check is after the forward checking. If user set "ledger_index" : current, it will forward to rippled regardless the method is valid or not.

godexsoft: I guess that's ok, wdyt? it may result in creating extra methods in the counters however. maybe i can add this check on the call site instead.

src/rpc/HandlerTable.h
```diff
@@ -17,33 +17,45 @@
...
+#pragma once
 
+#include <backend/BackendInterface.h>
+#include <rpc/common/AnyHandler.h>
+#include <rpc/common/Types.h>
+
+#include <memory>
 #include <optional>
+#include <string>
+#include <unordered_map>
```
cindyyan317: we are not using this header in this file

src/rpc/HandlerTable.h
```diff
@@ -17,33 +17,45 @@
 */
 //==============================================================================
 
-#include <rpc/RPCHelpers.h>
-#include <rpc/ngHandlers/LedgerRange.h>
+#pragma once
 
+#include <backend/BackendInterface.h>
```
cindyyan317: not need this one?

src/rpc/RPCHelpers.h
```diff
@@ -94,7 +89,7 @@ generatePubLedgerMessage(
     std::uint32_t txnCount);
 
 std::variant<Status, ripple::LedgerInfo>
-ledgerInfoFromRequest(Context const& ctx);
+ledgerInfoFromRequest(std::shared_ptr<Backend::BackendInterface const> const& backend, Web::Context const& ctx);
```
cindyyan317, line 93: no one is using traverseOwnedNodes(L108) function, we can remove it now.

cindyyan317 (approved):

PR #573, opened 2023-03-31
account_info implementation in new RPC framework

cindyyan317:


src/rpc/ngHandlers/AccountInfo.cpp
```diff
@@ -0,0 +1,145 @@
...
+    jv = boost::json::object{
+        {JS(account_data), RPC::toJson(output.accountData)},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(ledger_index), output.ledgerIndex}};
+    if (output.signerLists)
+    {
+        jv.as_object()[JS(signer_lists)] = boost::json::array();
+        for (auto const& signerList : output.signerLists.value())
+            jv.as_object()[JS(signer_lists)].as_array().push_back(
+                RPC::toJson(signerList));
```
godexsoft, line 102: Let's add a todo for later here to 
- replace `toJson` with `tag_invoke`
- not look up `signer_lists` + `as_array` each iteration
This really should be just `jv.as_object().insert_or_assign(JS(signer_lists)), output.signerLists);` if all `tag_invoke`s are correctly added.

src/rpc/ngHandlers/AccountInfo.h
```diff
@@ -0,0 +1,112 @@
...
+            ripple::STLedgerEntry sle)
+            : ledgerIndex(ledgerId)
+            , ledgerHash(std::move(ledgerHash))
+            , accountData(std::move(sle))
+        {
+        }
+    };
+
+    // "queue" is not available in Reporting mode
+    // TODO: ident is not in document
```
godexsoft: Consider adding an issue for the doc update and posting a link to it here in this todo 🙇

src/rpc/ngHandlers/AccountInfo.h
```diff
@@ -0,0 +1,112 @@
...
+    AccountInfoHandler(
+        std::shared_ptr<BackendInterface> const& sharedPtrBackend)
+        : sharedPtrBackend_(sharedPtrBackend)
+    {
+    }
+
+    RpcSpecConstRef
+    spec() const
+    {
+        static const RpcSpec rpcSpec = {
```
godexsoft: `static auto const rpcSpec = RpcSpec{` is nicer and and also puts `const` in the usual spot 👍

unittests/rpc/handlers/AccountInfoTest.cpp
```diff
@@ -0,0 +1,446 @@
...
+
+constexpr static auto ACCOUNT = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn";
+constexpr static auto ACCOUNT1 = "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW";
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+// constexpr static auto ISSUER = "rK9DrarGKnVEo2nYp5MfVRXRYf5yRX3mwD";
+constexpr static auto LEDGERHASH =
+    "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+constexpr static auto INDEX1 =
+    "1B8590C01B0006EDFA9ED60296DD052DC5E90F99659B25014D08E1BC983515BC";
+// constexpr static auto INDEX2 =
```
godexsoft: Can remove these commented out lines :)

godexsoft (changes requested):
Adding a few small suggestions. Overall looks great as always ⭐

src/rpc/ngHandlers/AccountInfo.cpp
cindyyan317: I think we have to postpone this change. 
Currently we have SLE tag_invoke here -> https://github.com/XRPLF/clio/blob/232acaeff293a43f97d332eb196263930ddedeac/src/rpc/ngHandlers/NFTOffersCommon.cpp#L33

If we put this function in a common place, convert SLE to json value according to the SLE type, it will work, but not ideal. 
because we need more information to distinguish the ledger detail,
for example, current https://github.com/XRPLF/clio/blob/232acaeff293a43f97d332eb196263930ddedeac/src/rpc/ngHandlers/NFTOffersCommon.cpp#L33 is only for nft offer not general offer.

godexsoft: I'm ok with that. Just add a TODO comment about this desired change then and let's move on :)

godexsoft (approved):
LGTM 🚀

PR #577, opened 2023-04-04
Fix compile error on clang14.0.3

cindyyan317:
-std::result_of has been removed from this version. 
Add DBOOST_ASIO_HAS_STD_INVOKE_RESULT to force boost not use it

-move() throw "unqualified call to 'std::move' " error

- new version clang finds more static issue.


godexsoft (approved):
LGTM 👍

PR #579, opened 2023-04-04
Create ngContext

cindyyan317:


src/rpc/common/Concepts.h
```diff
@@ -50,17 +49,17 @@ concept Requirement = requires(T a) {
  */
 // clang-format off
 template <typename T>
-concept CoroutineProcess = requires(T a, typename T::Input in, typename T::Output out, boost::asio::yield_context& y) {
+concept ContextProcess = requires(T a, typename T::Input in, typename T::Output out, Context& y) {
```
godexsoft: probably could be const ref here too

src/rpc/common/Types.h
```diff
@@ -57,6 +59,12 @@ struct VoidOutput
 {
 };
 
+struct Context
+{
+    boost::asio::yield_context* pYield = nullptr;
+    std::shared_ptr<WsBase> pWs;
```
godexsoft: may want to include <memory> above or perhaps forward declare Context and have the actual Context in its own file. this would however require us to include it in multiple places

godexsoft: im also not too sure about the nami g here: consider session or connection

src/rpc/common/impl/Processors.h
```diff
@@ -34,7 +34,7 @@ struct DefaultProcessor final
     operator()(
         HandlerType const& handler,
         boost::json::value const& value,
-        boost::asio::yield_context* ptrYield = nullptr) const
+        Context const& ctx = Context()) const
```
godexsoft: i thought you would have a pointer to context here and default it to nullptr but inside of context store an actual reference wrapper and shared ptr. this also works tho but now we expose the raw pointer business to clients instead of keeping it in detail

src/rpc/ngHandlers/AccountChannels.cpp
```diff
@@ -57,8 +57,9 @@ AccountChannelsHandler::addChannel(
 AccountChannelsHandler::Result
 AccountChannelsHandler::process(
     AccountChannelsHandler::Input input,
-    boost::asio::yield_context& yield) const
+    Context const& ctx) const
 {
+    auto& yield = *(ctx.pYield);
```
godexsoft: this should be auto const& even tho it resolves to that already - better specify const for visibility

src/rpc/ngHandlers/AccountCurrencies.cpp
```diff
@@ -23,8 +23,9 @@ namespace RPCng {
 AccountCurrenciesHandler::Result
 AccountCurrenciesHandler::process(
     AccountCurrenciesHandler::Input input,
-    boost::asio::yield_context& yield) const
+    Context const& ctx) const
 {
+    auto& yield = *(ctx.pYield);
```
godexsoft: same here

src/rpc/ngHandlers/AccountInfo.cpp
```diff
@@ -21,13 +21,13 @@
...
-AccountInfoHandler::process(
-    AccountInfoHandler::Input input,
-    boost::asio::yield_context& yield) const
+AccountInfoHandler::process(AccountInfoHandler::Input input, Context const& ctx)
+    const
 {
     if (!input.account && !input.ident)
         return Error{RPC::Status{RPC::RippledError::rpcACT_MALFORMED}};
 
+    auto& yield = *(ctx.pYield);
```
godexsoft: and here, and probably many more places 😀

unittests/rpc/handlers/AccountChannelsTest.cpp
```diff
@@ -55,7 +55,7 @@ TEST_F(RPCAccountHandlerTest, NonHexLedgerHash)
                 "ledger_hash": "xxx"
             }})",
             ACCOUNT));
-        auto const output = handler.process(input, yield);
+        auto const output = handler.process(input, Context{&yield});
```
godexsoft: i'd prefer std::cref(yield) here if you can make it work with the idea i described above

unittests/rpc/handlers/impl/FakesAndMocks.h
```diff
@@ -122,7 +122,7 @@ class CoroutineHandlerFake
     }
 
     Result
-    process(Input input, boost::asio::yield_context& yield) const
+    process(Input input, RPCng::Context ctx) const
```
godexsoft: probably should match the signature of the original one - consider adding const 😃

godexsoft (changes requested):
if it is possible to get it to work without exposing raw pointers to the entire codebase thru Context it is best we get it sorted sooner than later.. let me know if it does not work and we can look at it together again🙇‍♂️

godexsoft (approved):
Looking good 🔥

PR #584, opened 2023-04-06
Rough draft of nfts_by_issuer

ledhed2222:
Implementation of this spec:

https://docs.google.com/document/d/e/2PACX-1vS396CnsS2K2hvyniOHTwQbrQ-1joiVHHFSfnnNB-EJh2nmRXm6UxYFtqiEPvKOajZs44RqYuNXr-aM/pub


src/backend/CassandraBackend.cpp
```diff
@@ -1496,6 +1580,16 @@ CassandraBackend::open(bool readOnly)
         if (!selectNFT_.prepareStatement(query, session_.get()))
             continue;
 
+        query.str("");
+        query << "SELECT token_id,sequence,owner,is_burned"
+              << " FROM " << tablePrefix << "nf_tokens WHERE"
+              << " token_id IN ? AND"
+              << " sequence <= ?"
+              << " ORDER BY sequence DESC"
+              << " PER PARTITION LIMIT 1";
```
ledhed2222, line 1588: this query could actually replace the existing nf_token select query used for `nft_info`. not sure we really need to save on memory but it's possible.

src/backend/CassandraBackend.cpp
```diff
@@ -1519,6 +1613,15 @@ CassandraBackend::open(bool readOnly)
         if (!selectNFTURI_.prepareStatement(query, session_.get()))
             continue;
 
+        query.str("");
+        query << "SELECT token_id,uri FROM " << tablePrefix << "nf_token_uris"
+              << " WHERE token_id IN ? AND"
+              << " sequence <= ?"
+              << " ORDER BY sequence DESC"
+              << " PER PARTITION LIMIT 1";
```
ledhed2222, line 1620: this query could actually replace the existing nf_token_uri select query used for `nft_info`. not sure we really need to save on memory but it's possible.

src/backend/CassandraBackend.h
```diff
@@ -271,6 +302,40 @@ class CassandraStatement
         curBindingIndex_++;
     }
 
+    // TODO: tuple binding desperately needs to be implemented using templates
+    void
```
ledhed2222, line 306: i don't think we should even address this TODO since we have the new framework coming up

src/backend/cassandra/impl/Statement.h
```diff
@@ -91,6 +92,8 @@ class Statement : public ManagedObject<CassStatement>
         using decayed_t = std::decay_t<Type>;
         using uchar_vec_t = std::vector<unsigned char>;
         using uint_tuple_t = std::tuple<uint32_t, uint32_t>;
+        using uint_bytes_tuple_t = std::tuple<uint32_t, ripple::uint256>;
+        using bytes_vector_t = std::vector<ripple::uint256>;
```
ledhed2222, line 96: @godexsoft this is where i could use your help with better application of generics. i think it's possible to get tuples abstracted better

unittests/rpc/handlers/NFTsByIssuerTest.cpp
```diff
@@ -0,0 +1,359 @@
+//------------------------------------------------------------------------------
+/*
```
ledhed2222, line 2: @godexsoft generally these tests need help. could use your help on this too

src/backend/CassandraBackend.cpp
```diff
@@ -661,6 +661,90 @@ CassandraBackend::fetchNFTTransactions(
...
+{
+    NFTsAndCursor ret;
+
+    CassandraStatement const idQueryStatement = [&taxon, &issuer, &cursorIn, &limit, this]() {
+        if (taxon.has_value())
+        {
+            auto statement = CassandraStatement{selectNFTIDsByIssuerTaxon_};
+            statement.bindNextBytes(issuer);
+            // TODO no idea why binding UInt fails
+            statement.bindNextInt(*taxon);
```
ledhed2222, line 680: @godexsoft

src/backend/CassandraBackend.cpp
```diff
@@ -661,6 +661,90 @@ CassandraBackend::fetchNFTTransactions(
...
+    std::vector<ripple::uint256> nftIDs;
+    do
+        nftIDs.push_back(idQueryResult.getUInt256());
+    while (idQueryResult.nextRow());
+
+    if (nftIDs.size() == limit)
+        ret.cursor = nftIDs.back();
+
+    // TODO these two queries should happen in parallel
+    auto nftQueryStatement = CassandraStatement{selectNFTBulk_};
```
ledhed2222, line 708: @godexsoft - possibly not a blocker for this PR, but these two can happen in parallel. not sure if we have an API for that ATM. I'd be OK with leaving this as a todo until there is an API for parallel reads

src/backend/CassandraBackendNew.h
```diff
@@ -488,6 +488,81 @@ class BasicCassandraBackend : public BackendInterface
...
+            nftIDs.push_back(nftID);
+
+        if (nftIDs.size() == 0)
+            return ret;
+
+        if (nftIDs.size() == limit)
+            ret.cursor = nftIDs.back();
+
+        /// TODO these two queries should happen in parallel
+        auto nftQueryStatement = schema_->selectNFTBulk.bind(nftIDs);
```
ledhed2222, line 534: ditto same as above

src/backend/cassandra/Schema.h
```diff
@@ -560,6 +560,19 @@ class Schema
...
+            return handle_.get().prepare(fmt::format(
+                R"(
+                SELECT token_id, sequence, owner, is_burned
+                  FROM {}    
+                 WHERE token_id IN ?
+                   AND sequence <= ?
+              ORDER BY sequence DESC
+   PER PARTITION LIMIT 1
+                )",
+                qualifiedTableName(settingsProvider_.get(), "nf_tokens")));
```
ledhed2222, line 573: ditto same as above WRT to replacing old query

src/backend/cassandra/Schema.h
```diff
@@ -573,6 +586,19 @@ class Schema
...
+                SELECT token_id, uri
+                  FROM {}    
+                 WHERE token_id IN ?
+                   AND sequence <= ?
+              ORDER BY sequence DESC
+   PER PARTITION LIMIT 1
+                )",
+                qualifiedTableName(settingsProvider_.get(), "nf_token_uris")));
+        }();
+
```
ledhed2222, line 601: ditto same as above WRT to replacing old query

src/backend/cassandra/impl/Collection.h
```diff
@@ -0,0 +1,88 @@
...
+        auto const rc = cass_collection_append_bool(*this, value ? cass_true : cass_false);
+        throwErrorIfNeeded(rc, "Bind bool");
+    }
+
+    void
+    append(int64_t const value) const
+    {
+        auto const rc = cass_collection_append_int64(*this, value);
+        throwErrorIfNeeded(rc, "Bind int64");
+    }
```
ledhed2222, line 78: @godexsoft we don't strictly need collections for anything other than `std::vector<ripple::uint256>` for this PR, but it was easy to add these. AFAIK, collections with heterogeneous types are not a thing.

src/rpc/ngHandlers/NFTInfo.cpp
```diff
@@ -31,7 +30,7 @@ namespace RPCng {
 NFTInfoHandler::Result
 NFTInfoHandler::process(NFTInfoHandler::Input input, Context const& ctx) const
 {
-    auto const tokenID = ripple::uint256{input.nftID.c_str()};
+    auto const tokenID = uint256{input.nftID.c_str()};
```
ledhed2222: @godexsoft these changes are to fix the bug i mentioned to you in your new implementation of `nft_info`. while i was here :)

src/backend/CassandraBackendNew.h
```diff
@@ -488,6 +488,81 @@ class BasicCassandraBackend : public BackendInterface
...
+        auto nftURIQueryStatement = schema_->selectNFTURIBulk.bind(nftIDs);
+        nftURIQueryStatement.bindAt(1, ledgerSequence);
+
+        auto const& nftURIQueryResults = executor_.read(yield, nftURIQueryStatement).value();
+        ///
+        //
+
+        std::unordered_map<std::string, Blob> nftURIMap;
+        for (auto const [nftID, uri] : extract<ripple::uint256, Blob>(nftURIQueryResults))
+            nftURIMap.insert({ripple::strHex(nftID), uri});
```
ledhed2222, line 547: note - this is not ideal to have the keys be `std::string` instead of `ripple::uint256` but it seems to me that the latter has no implemented hash function so it cannot be used as the key to a map.

src/backend/CassandraBackend.cpp
godexsoft: Because `taxon` is bigint (int64) but `bindUInt` binds using `cass_statement_bind_int32` so you get a type mismatch on DB side.

src/backend/CassandraBackend.cpp
godexsoft: Defo not a blocker for this PR - we will handle this in the future with the new backend.

src/backend/CassandraBackend.cpp
godexsoft: Not as part of this PR - feel free to add a comment in the relevant place in nft_info handlers tho.

src/backend/CassandraBackend.cpp
godexsoft: Same as above.

src/backend/CassandraBackend.h
godexsoft: Yea, definitely not for the old backend.. in the new one we can implement it properly.

src/backend/CassandraBackendNew.h
godexsoft: You could use a `std::map` instead. I assume we don't expect millions of URIs.

src/backend/cassandra/impl/Collection.h
godexsoft: No harm. Just remove the `const` in params here and for bool.

src/backend/cassandra/impl/Statement.h
godexsoft: Yes. Let's discuss over a call later.

src/rpc/ngHandlers/NFTInfo.cpp
```diff
@@ -66,7 +63,12 @@ void
 tag_invoke(boost::json::value_from_tag, boost::json::value& jv, NFTInfoHandler::Output const& output)
 {
     // TODO: use JStrings when they become available
-    auto object = boost::json::object{
+    // TODO - this formatting is exactly the same and SHOULD REMAIN THE SAME
+    // for each element of the `nfts_by_issuer` API. We should factor this out
+    // so that the formats don't diverge. In the mean time, do not make any
+    // changes to this formatting without making the same changes to that
+    // formatting.
+    auto const object = boost::json::object{
```
godexsoft: Now that there is no "injection" there is no need to jump thru hoops with `std::move` and temporary object. Safe to rewrite to something like
```cpp
jv = { ... };
```

Btw., I appreciate that you try to add `const` everywhere but in this particular case making `object` const actually prevents `std::move` below from doing its job and makes a copy instead :)

src/rpc/ngHandlers/NFTInfo.cpp
```diff
@@ -55,9 +54,7 @@ NFTInfoHandler::process(NFTInfoHandler::Input input, Context const& ctx) const
     output.issuer = toBase58(nft::getIssuer(nft.tokenID));
     output.taxon = nft::toUInt32(nft::getTaxon(nft.tokenID));
     output.serial = nft::getSerial(nft.tokenID);
-
-    if (not nft.isBurned)
-        output.uri = strHex(nft.uri);
+    output.uri = strHex(nft.uri);
```
godexsoft: I assume you checked it is safe to call `strHex` on an empty string?

unittests/rpc/handlers/NFTInfoTest.cpp
```diff
@@ -288,7 +288,7 @@ TEST_F(RPCNFTInfoHandlerTest, NonExistNFT)
 // normal case when only provide nft_id
 TEST_F(RPCNFTInfoHandlerTest, DefaultParameters)
 {
-    constexpr static auto correntOutput = R"({
+    constexpr static auto currentOutput = R"({
```
godexsoft: I'm not sure about this but i think it meant to say `correctOutput` 🤣 
Either way works 👍

unittests/rpc/handlers/NFTsByIssuerTest.cpp
godexsoft: Yep, can go thru them together on the call.

godexsoft (commented):
Adding first batch of replies/comments.

cjcobb23:
> Implementation of this spec:
> 
> https://docs.google.com/document/d/1MIiz0K6Um4eKZF4XygcKjxm4EfAmHrJuY4b1d00VQRU/edit

This is a private google doc FYI, would be nice if the spec was public since this is an open source project

ledhed2222:
> This is a private google doc FYI, would be nice if the spec was public since this is an open source project

Updated

cjcobb23:
> > This is a private google doc FYI, would be nice if the spec was public since this is an open source project
> 
> Updated

still no access for me

ledhed2222:
> > > This is a private google doc FYI, would be nice if the spec was public since this is an open source project
> > 
> > 
> > Updated
> 
> still no access for me

via the updated link? it should be a public link now. regardless i can paste it in the PR if necessary

tequdev:
It doesn't appear to be public yet.

godexsoft:
Closing this as it was implemented in #948

PR #586, opened 2023-04-06
Implement better error handling and datastax unauthorized workaround

### LINKED ISSUE #585: Datastax backend connection error handling
Error handling for secure bundle connection to datastax cassandra cloud needs fixing. Seeing ManagedObject throwing exceptions (meaning that we get nullptr results sometimes) as well as creation of keyspace can fail with UNATHORIZED error since on datastax you have to configure the keyspaces thru their admin interface, not thru code.

godexsoft:
Fixes #585


cindyyan317 (approved):
LGTM

PR #588, opened 2023-04-07
add default ordering to issuer_nf_tokens_v2

### LINKED ISSUE #589: No ordering specified for `issuer_nf_tokens_v2`
The new `issuer_nf_tokens_v2` table doesn't explicitly mark ordering on the table. This is fine because we actually want to use the default ordering of `(taxon ASC, token_id ASC)`, but it should be specified for maximal clarity, IMO

ledhed2222:
title says it all. adding this ordering shouldn't actually be strictly necessary since I've just specified what is the default ordering anyway (which is what we want). however, I think adding it adds to the clarity of what we're doing here.

Fix #589

### DIFF
--- a/src/backend/cassandra/Schema.h
+++ b/src/backend/cassandra/Schema.h
@@ -214,15 +214,15 @@ public:
 
         statements.emplace_back(fmt::format(
             R"(
-            CREATE TABLE IF NOT EXISTS {}
-                ( 
-                    issuer blob,
-                    taxon bigint,
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                      issuer blob,
+                       taxon bigint,
                     token_id blob,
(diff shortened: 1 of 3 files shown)


src/backend/cassandra/Schema.h
```diff
@@ -214,15 +214,18 @@ class Schema
 
         statements.emplace_back(fmt::format(
             R"(
-           CREATE TABLE IF NOT EXISTS {}
-                  ( 
-                      issuer blob,
-                       taxon bigint,
+            CREATE TABLE IF NOT EXISTS {}
```
godexsoft: Fix the formatting of the query and we good 👍

godexsoft (changes requested):
The alignment is all messed up now, please fix 🥇

godexsoft (approved):
LGTM 🚀

PR #590, opened 2023-04-11
Implement nextgen server_info handler

### LINKED ISSUE #587: Implement server_info nextgen rpc handler
Port server_info to nextgen RPC system.

godexsoft:
Fixes #587

### DIFF
--- a/src/rpc/common/impl/Processors.h
+++ b/src/rpc/common/impl/Processors.h
@@ -64,11 +64,23 @@ struct DefaultProcessor final
         }
         else if constexpr (HandlerWithoutInput<HandlerType>)
         {
+            using OutType = HandlerReturnType<typename HandlerType::Output>;
+
             // no input to pass, ignore the value
-            if (auto const ret = handler.process(); not ret)
-                return Error{ret.error()};  // forward Status
+            if constexpr (ContextProcessWithoutInput<HandlerType>)
+            {
+                if (auto const ret = handler.process(*ctx); not ret)
+                    return Error{ret.error()};  // forward Status
+                else
+                    return value_from(ret.value());
+            }
             else
-                return value_from(ret.value());
+            {
+                if (auto const ret = handler.process(); not ret)
+                    return Error{ret.error()};  // forward Status
+                else
+                    return value_from(ret.value());
+            }
         }
         else
         {
--- a/src/rpc/ngHandlers/ServerInfo.h
+++ b/src/rpc/ngHandlers/ServerInfo.h
@@ -1,0 +1,243 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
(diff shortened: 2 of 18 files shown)


src/rpc/common/impl/Processors.h
```diff
@@ -64,11 +64,29 @@ struct DefaultProcessor final
...
         else if constexpr (HandlerWithoutInput<HandlerType>)
         {
+            using OutType = HandlerReturnType<typename HandlerType::Output>;
+
             // no input to pass, ignore the value
-            if (auto const ret = handler.process(); not ret)
-                return Error{ret.error()};  // forward Status
+            // clang-format off
+            if constexpr (
+                requires(HandlerType h, Context const& c) {
```
cindyyan317: we can make it a concept like ContextProcess

src/rpc/ngHandlers/ServerInfo.h
```diff
@@ -0,0 +1,237 @@
...
+    {
+    }
+
+    Result
+    process(Context const& ctx) const
+    {
+        using namespace RPC;
+        using namespace std::chrono;
+
+        auto const range = backend_->fetchLedgerRange();
```
cindyyan317, line 121: I understand it is from old code. We can remove this check , should be handled in RPC framework

src/rpc/ngHandlers/ServerInfo.h
```diff
@@ -0,0 +1,237 @@
...
+        if (!range)
+            return Error{Status{RippledError::rpcNOT_READY, "emptyDatabase", "The server has no data in the database"}};
+
+        auto const lgrInfo = backend_->fetchLedgerBySequence(range->maxSequence, ctx.yield);
+        auto const fees = backend_->fetchFees(lgrInfo->seq, ctx.yield);
+        if (!lgrInfo || !fees)
```
cindyyan317: !lgrInfo check should be earlier

src/rpc/ngHandlers/ServerInfo.h
```diff
@@ -0,0 +1,237 @@
...
+        if (info.rippledInfo)
+        {
+            try
+            {
+                auto const& rippledInfo = info.rippledInfo.value();
+                jv.as_object()[JS(load_factor)] = rippledInfo.at(JS(load_factor));
+                jv.as_object()[JS(validation_quorum)] = rippledInfo.at(JS(validation_quorum));
+                jv.as_object()["rippled_version"] = rippledInfo.at(JS(build_version));
+            }
+            catch (std::exception const&)
```
cindyyan317, line 202: is for keyNotExist?

src/rpc/common/impl/Processors.h
godexsoft: I wonder what the requirements for such concept would be, considering Context has nothing but member variables.. not sure i can come up with one.

godexsoft: Could do but i did not see the value in that tbh. for this case

src/rpc/ngHandlers/ServerInfo.h
godexsoft: Aha, good catch!

src/rpc/ngHandlers/ServerInfo.h
godexsoft: Ha! i did not notice there is a bug here when copying this code - good catch.

src/rpc/ngHandlers/ServerInfo.h
godexsoft: For any sort of issue, could be different type, could be keyNotExist, etc.
I agree we could catch specific exceptions here but it's ok to leave as is i think

src/rpc/ngHandlers/ServerInfo.h
godexsoft: Turns out there is no such place yet as we did not integrate the new framework into clio yet. So this will have to be done later. Added a TODO comment for that.

godexsoft: And tbh. i'm not even sure it should be done in framework level - not all RPC handlers will need range/db access. This is not parsing of the range from input, it's the actual range fetching code.. so we will see later how we handle it.

cindyyan317 (dismissed):
good to go

PR #591, opened 2023-04-12
Subscribe handler

cindyyan317:


src/rpc/common/Validators.cpp
```diff
@@ -134,7 +134,7 @@ CustomValidator AccountValidator =
         // remove all old handler, this function can be moved to here
         if (!RPC::accountFromStringStrict(value.as_string().c_str()))
         {
-            return Error{RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "Malformed"}};
+            return Error{RPC::Status{RPC::RippledError::rpcACT_MALFORMED, std::string(key) + "Malformed"}};
```
cindyyan317, line 137: rpcACT_MALFORMED make more sense.

src/rpc/common/Validators.cpp
```diff
@@ -199,4 +199,73 @@ CustomValidator IssuerValidator =
         return MaybeError{};
     }};
 
+CustomValidator SubscibeStreamValidator =
```
cindyyan317: Put here for now. I will work on unsubscribe, I might move it to common place of subscribe and unsubscribe , like a base class

src/rpc/common/Validators.cpp
godexsoft: I agree it does. I guess rippled had `rpcINVALID_PARAMS` for this tho?

src/rpc/common/Validators.cpp
```diff
@@ -199,4 +199,71 @@ CustomValidator IssuerValidator =
...
+    CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_array())
+            return Error{RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "NotArray"}};
+        if (value.as_array().size() == 0)
+            return Error{RPC::Status{RPC::RippledError::rpcACT_MALFORMED, std::string(key) + " malformed."}};
+
+        for (auto const& v : value.as_array())
+        {
+            auto obj = boost::json::object();
+            auto const keyItem = std::string(key) + "'sItem";
```
godexsoft, line 230: Why the `"'sItem"` instead of just the `key`?

src/rpc/ngHandlers/Subscribe.cpp
```diff
@@ -0,0 +1,194 @@
...
+    std::shared_ptr<WsBase> const& session,
+    std::shared_ptr<SubscriptionManager> const& manager) const
+{
+    boost::json::object response;
+    for (auto const& stream : streams)
+    {
+        if (stream == "ledger")
+            response = manager->subLedger(yield, session);
+        else if (stream == "transactions")
+            manager->subTransactions(session);
```
godexsoft: Just checking: Only `ledger` subscription returns some response, others are empty, is this correct?

src/rpc/ngHandlers/Subscribe.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    for (auto const& internalBook : books)
+    {
+        if (internalBook.snapshot)
+        {
+            if (!rng)
+                rng = sharedPtrBackend_->fetchLedgerRange();
+            auto const getOrderBook = [&](auto const& book) {
+                auto const bookBase = getBookBase(book);
+                auto const [offers, _] = sharedPtrBackend_->fetchBookOffers(bookBase, rng->maxSequence, 200, yield);
```
godexsoft: This `200` would be great to have as a meaningful constant somewhere in the class/function

src/rpc/ngHandlers/Subscribe.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+                // the taker is not really uesed, same issue with https://github.com/XRPLF/xrpl-dev-portal/issues/1818
```
godexsoft: 1) uesed -> used
2) we can do better on formatting (it would fit in two lines now i think)

src/rpc/ngHandlers/Subscribe.h
```diff
@@ -0,0 +1,116 @@
...
+
+namespace RPCng {
+class SubscribeHandler
+{
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Output
+    {
+        // response of stream "ledger"
```
godexsoft, line 37: This comment and the ones bellow can use better formatting - looks like too many lines atm, wdyt?

src/subscriptions/SubscriptionManager.cpp
```diff
@@ -329,7 +329,7 @@ SubscriptionManager::unsubProposedTransactions(std::shared_ptr<WsBase> session)
     txProposedSubscribers_.unsubscribe(session);
 }
 void
```
godexsoft, line 333: Could use a spacer line in between the functions to follow the same style with above

src/subscriptions/SubscriptionManager.h
```diff
@@ -276,10 +276,10 @@ class SubscriptionManager
     pubTransaction(Backend::TransactionAndMetadata const& blobs, ripple::LedgerInfo const& lgrInfo);
 
     void
-    subAccount(ripple::AccountID const& account, session_ptr& session);
+    subAccount(ripple::AccountID const& account, session_ptr const& session);
```
godexsoft: While you are here, maybe rename `session_ptr` to `SessionPtrType` or something like that?

unittests/rpc/BaseTests.cpp
```diff
@@ -440,3 +440,83 @@ TEST_F(RPCBaseTest, IssuerValidator)
     err = spec.validate(failingInput);
     ASSERT_FALSE(err);
 }
+
+TEST_F(RPCBaseTest, SubscibeStreamValidator)
+{
+    auto const spec = RpcSpec{{"streams", SubscibeStreamValidator}};
```
godexsoft: I notice it's all `Subscibe` here instead of `Subscribe` - probably need to change everywhere (search + replace in the project).

unittests/rpc/BaseTests.cpp
```diff
@@ -440,3 +440,83 @@ TEST_F(RPCBaseTest, IssuerValidator)
...
+    auto passingInput = json::parse(
+        R"({ "streams": ["ledger", "transactions_proposed","validations","transactions","manifests","transactions","book_changes"]})");
```
godexsoft: Formatting of JSON please :) You can leverage the multiline nature of `R"()"`

unittests/rpc/handlers/SubscribeTest.cpp
```diff
@@ -0,0 +1,884 @@
...
+using namespace RPCng;
+namespace json = boost::json;
+using namespace testing;
+
+constexpr static auto MINSEQ = 10;
+constexpr static auto MAXSEQ = 30;
+constexpr static auto ACCOUNT = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn";
+constexpr static auto ACCOUNT2 = "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun";
+constexpr static auto LEDGERHASH = "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+// 20 USD : 10 XRP
```
godexsoft: I think this is in the name of the below two constants - consider removing

unittests/rpc/handlers/SubscribeTest.cpp
```diff
@@ -0,0 +1,884 @@
...
+        EXPECT_EQ(report.at("accounts_proposed").as_uint64(), 2);
+    });
+}
+
+TEST_F(RPCSubscribeHandlerTest, JustBooks)
+{
+    auto const input = json::parse(fmt::format(
+        R"({{
+            "books": [
+                    {{
```
godexsoft, line 660: Looks like spacing is messed up a little?

godexsoft: And in many similar JSON strings below too..

src/rpc/common/Types.h
```diff
@@ -65,6 +66,7 @@ struct Context
     // update backend interfaces to use const& yield
     const std::reference_wrapper<boost::asio::yield_context> yield;
     const std::shared_ptr<WsBase> session;
+    const std::shared_ptr<SubscriptionManager> subManager;
```
godexsoft: SubscriptionManager just like Backend should be injected thru constructor of the Handler rather than thru Context.

godexsoft (changes requested):
Leaving a few comments and suggestions :)

src/rpc/common/Validators.cpp
cindyyan317: No. It is just a bug

src/rpc/common/Validators.cpp
cindyyan317: Because this is a hint for the item of this array. like "stream's item ", so the error message makes more sense. eg, stream's item not a string

unittests/rpc/BaseTests.cpp
cindyyan317: Good eyesight

src/rpc/ngHandlers/Subscribe.cpp
cindyyan317: yes . ledger will return ledger info

godexsoft (dismissed):
LGTM! 🚀

godexsoft (approved):

PR #595, opened 2023-04-18
Unsubscribe

cindyyan317:


src/rpc/ngHandlers/Subscribe.h
```diff
@@ -71,11 +71,42 @@ class BaseSubscribeHandler
...
+            validation::CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+                if (!value.is_array())
+                    return Error{RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "NotArray"}};
+                for (auto const& book : value.as_array())
+                {
+                    if (!book.is_object())
+                        return Error{
+                            RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "ItemNotObject"}};
+                    if (book.as_object().contains("both") && !book.as_object().at("both").is_bool())
+                    {
```
godexsoft: May as well remove the extra scope here same as on the line above as there is no complexity to the statement. same for below if statements :)

src/rpc/ngHandlers/Unsubscribe.h
```diff
@@ -0,0 +1,212 @@
...
+            validation::CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+                if (!value.is_array())
+                    return Error{RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "NotArray"}};
+                for (auto const& book : value.as_array())
+                {
+                    if (!book.is_object())
+                        return Error{
+                            RPC::Status{RPC::RippledError::rpcINVALID_PARAMS, std::string(key) + "ItemNotObject"}};
+                    if (book.as_object().contains("both") && !book.as_object().at("both").is_bool())
+                    {
```
godexsoft: same as above - probably no need for the extra braces here

src/rpc/ngHandlers/Unsubscribe.h
```diff
@@ -0,0 +1,212 @@
...
+            {JS(books), booksValidator}};
+
+        return rpcSpec;
+    }
+
+    Result
+    process(Input input, Context const& ctx) const
+    {
+        if (input.streams)
+        {
```
godexsoft: I'd omit braces here too, especially now that we have more characters per column and it won't look bad if you omit the braces

src/rpc/ngHandlers/Unsubscribe.h
```diff
@@ -0,0 +1,212 @@
...
+        if (input.books)
+        {
+            unsubscribeToBooks(*(input.books), ctx.session);
+        }
+        return Output{};
+    }
+
+private:
+    void
+    unsubscribeToStreams(std::vector<std::string> const& streams, std::shared_ptr<WsBase> const& session) const
```
godexsoft: it's probably old code that you copied but consider renaming to something like 'unsubscribeFromStreams' :)

src/rpc/ngHandlers/Unsubscribe.h
```diff
@@ -0,0 +1,212 @@
...
+    {
+        for (auto const& s : streams)
```
godexsoft: perhaps 'stream' is a better name here

src/rpc/ngHandlers/Unsubscribe.h
```diff
@@ -0,0 +1,212 @@
...
+    {
+        for (auto const& account : accounts)
+        {
+            auto accountID = RPC::accountFromStringStrict(account);
+            subscriptions_->unsubAccount(*accountID, session);
+        }
+    }
+
+    void
+    unsubscribeToProposedAccounts(std::vector<std::string> accountsProposed, std::shared_ptr<WsBase> const& session)
```
godexsoft: Probably should be 'From' here and below too

unittests/rpc/handlers/UnsubscribeTest.cpp
```diff
@@ -0,0 +1,640 @@
...
+        auto const output = handler.process(input, Context{std::ref(yield), session_});
+        ASSERT_TRUE(output);
+        EXPECT_TRUE(output->as_object().empty());
+    });
+}
+
+TEST_F(RPCUnsubscribeTest, Books)
+{
+    auto const input = json::parse(fmt::format(
+        R"({{
```
godexsoft, line 573: Formatting of JSON again 😄 here and below 🙏

godexsoft (changes requested):
Leaving small suggestions - nothing major 🙇‍♂️

src/rpc/ngHandlers/Unsubscribe.h
cindyyan317: yea, 'from' makes more sense

godexsoft (approved):
LGTM 🔥

PR #596, opened 2023-04-19
Ledgerdata

cindyyan317:
"ledger_data" in new RPC framework.
Worth to mention: "out_of_order" field is only in Clio, I just copy the logic from old handler.

### DIFF
--- a/src/rpc/ngHandlers/LedgerData.cpp
+++ b/src/rpc/ngHandlers/LedgerData.cpp
@@ -87,7 +87,7 @@ LedgerDataHandler::process(Input input, Context const& ctx) const
     if (input.diffMarker)
     {
         // keep the same logic as previous implementation
-        auto const diff = sharedPtrBackend_->fetchLedgerDiff(*(input.diffMarker), ctx.yield);
+        auto diff = sharedPtrBackend_->fetchLedgerDiff(*(input.diffMarker), ctx.yield);
         std::vector<ripple::uint256> keys;
         for (auto& [key, object] : diff)
         {
@@ -119,9 +119,9 @@ LedgerDataHandler::process(Input input, Context const& ctx) const
     }
     auto const end = std::chrono::system_clock::now();
 
-    auto time = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
+    log_.debug() << "Number of results = " << results.size() << " fetched in "
+                 << std::chrono::duration_cast<std::chrono::microseconds>(end - start).count() << " microseconds";
 
-    log_.debug() << "Number of results = " << results.size() << " fetched in " << time << " microseconds";
     output.states.reserve(results.size());
     for (auto const& [key, object] : results)
     {
@@ -140,10 +140,10 @@ LedgerDataHandler::process(Input input, Context const& ctx) const
     }
     if (input.outOfOrder)
         output.cacheFull = sharedPtrBackend_->cache().isFull();
-    auto const end2 = std::chrono::system_clock::now();
 
-    time = std::chrono::duration_cast<std::chrono::microseconds>(end2 - end).count();
-    log_.debug() << "Number of results = " << results.size() << " serialized in " << time << " microseconds";
+    auto const end2 = std::chrono::system_clock::now();
+    log_.debug() << "Number of results = " << results.size() << " serialized in "
+                 << std::chrono::duration_cast<std::chrono::microseconds>(end2 - end).count() << " microseconds";
 
     return output;
 }
--- a/src/rpc/ngHandlers/LedgerData.cpp
+++ b/src/rpc/ngHandlers/LedgerData.cpp
@@ -1,0 +1,221 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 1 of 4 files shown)


src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,233 @@
...
+//==============================================================================
+
+#include <rpc/ngHandlers/LedgerData.h>
+
+#include <ripple/app/ledger/LedgerToJson.h>
+
+#include <algorithm>
+// local to compilation unit loggers
+namespace {
+clio::Logger gLog{"RPC"};
```
godexsoft: this can be a private member now :)

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,233 @@
...
+}  // namespace
+
+namespace RPCng {
+
+LedgerDataHandler::Result
+LedgerDataHandler::process(Input input, Context const& ctx) const
+{
+    // marker must be int if outOfOrder is true
+    if (input.outOfOrder && input.marker)
+    {
```
godexsoft: can remove extra braces here and below

src/rpc/ngHandlers/LedgerData.h
```diff
@@ -0,0 +1,93 @@
...
+             validation::IfType<std::string>{validation::Uint256HexStringValidator}},
+        };
+        return rpcSpec;
+    }
+
+    Result
+    process(Input input, Context const& ctx) const;
+
+private:
+    static uint32_t constexpr limitIfBinary = 2048;
```
godexsoft: I think we use all caps for constants in other places? let's keep it uniform :)

godexsoft (changes requested):
I know this is still a draft but i thought i'd leave a few comments while i was already here 🫡 Just small stuff, nothing major.

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,221 @@
...
+    output.ledgerIndex = lgrInfo.seq;
+
+    auto const start = std::chrono::system_clock::now();
+    std::vector<Backend::LedgerObject> results;
+    if (input.diffMarker)
+    {
+        // keep the same logic as previous implementation
+        auto const diff = sharedPtrBackend_->fetchLedgerDiff(*(input.diffMarker), ctx.yield);
+        std::vector<ripple::uint256> keys;
+        for (auto& [key, object] : diff)
```
cindyyan317, line 92: Old handler uses " for (auto&& [key, object] : diff)" . change to auto&, auto && is confusing here

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,221 @@
...
+        {
+            if (!object.size())
+                keys.push_back(std::move(key));
+        }
+        auto objs = sharedPtrBackend_->fetchLedgerObjects(keys, lgrInfo.seq, ctx.yield);
+        for (size_t i = 0; i < objs.size(); ++i)
+        {
+            auto& obj = objs[i];
```
cindyyan317, line 100: change auto&& to auto&

src/rpc/ngHandlers/LedgerData.cpp
godexsoft: can you check if 'auto const diff' above breaks move below?

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,221 @@
...
+        auto page = sharedPtrBackend_->fetchLedgerPage(input.marker, lgrInfo.seq, limit, input.outOfOrder, ctx.yield);
+        results = std::move(page.objects);
+        if (page.cursor)
+            output.marker = ripple::strHex(*(page.cursor));
+        else if (input.outOfOrder)
+            output.diffMarker = sharedPtrBackend_->fetchLedgerRange()->maxSequence;
+    }
+    auto const end = std::chrono::system_clock::now();
+
+    auto time = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
```
godexsoft: can be const?

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,221 @@
...
+        else
+        {
+            output.states.push_back(RPC::toJson(sle));
+        }
+    }
+    if (input.outOfOrder)
+        output.cacheFull = sharedPtrBackend_->cache().isFull();
+    auto const end2 = std::chrono::system_clock::now();
+
+    time = std::chrono::duration_cast<std::chrono::microseconds>(end2 - end).count();
```
godexsoft: no need to try reuse, just create a new const here :) i understand this is old code but let's improve it

godexsoft: you can even put it straight into the log line.. no real need for temporaries here

src/rpc/ngHandlers/LedgerData.cpp
godexsoft: or inline into log line even better

src/rpc/ngHandlers/LedgerData.h
```diff
@@ -0,0 +1,94 @@
...
+        return rpcSpec;
+    }
+
+    Result
+    process(Input input, Context const& ctx) const;
+
+private:
+    static uint32_t constexpr LIMITBINARY = 2048;
+    static uint32_t constexpr LIMITJSON = 256;
+    friend void
```
godexsoft, line 89: needs newline before the function declaration

godexsoft (changes requested):
Adding more small suggestions. Almost there :)

src/rpc/ngHandlers/LedgerData.cpp
cindyyan317: indeed . it should not be const in this case. will fix

src/rpc/ngHandlers/LedgerData.cpp
```diff
@@ -0,0 +1,221 @@
...
+    auto const start = std::chrono::system_clock::now();
+    std::vector<Backend::LedgerObject> results;
+    if (input.diffMarker)
+    {
+        // keep the same logic as previous implementation
+        auto diff = sharedPtrBackend_->fetchLedgerDiff(*(input.diffMarker), ctx.yield);
+        std::vector<ripple::uint256> keys;
+        for (auto& [key, object] : diff)
+        {
+            if (!object.size())
```
cindyyan317, line 94: Not really sure why the check is !object.size(), maybe just wanna the removed item? there is no document whatsoever regarding the "out_of_order". any idea?@godexsoft

godexsoft: i suppose it is checking if the object is empty, whatever that means for this object type. i'd leave the logic as is if you copied it from the old impl 🤷‍♂️

godexsoft (approved):
LGTM 🚀

Issue #603, opened 2023-04-25
Throw error when "full"/"accounts"/"owner_funds"/"queue" exist AND set to true


cindyyan317:
XRPL.py have them in request even the value of “full“ , “accounts“ are default. Clio returns “NotSupported“ if the fields be present in the request, which causes the response is different between clio and rippled.

To make Clio more robust and friendly to the existing library, Clio will return “NotSupported“ only for full/accounts exists AND the value is true.

Same fix will apply to the other fields “owner_funds“ and “queue“ which are ignored in the current Clio. We shall give a proper warning to users if users intend to set them as true.

PR #604, opened 2023-04-25
Ledger

cindyyan317:
1 Implement Ledger API in the new RPC framework
2 Fix the issue #603

### DIFF
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -23,6 +23,8 @@
 #include <rpc/common/Specs.h>
 #include <rpc/common/Types.h>
 
+#include <fmt/core.h>
+
 namespace RPCng::validation {
 
 /**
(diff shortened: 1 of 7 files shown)


src/rpc/common/Validators.h
```diff
@@ -107,6 +109,66 @@ struct Required final
...
+                return Error{RPC::Status{
+                    RPC::RippledError::rpcNOT_SUPPORTED,
+                    fmt::format("Not supported field '{}'s value '{}'", std::string{key}, res)}};
+        }
+        return {};
+    }
+
+    NotSupported(T t)
+    {
+        val = t;
```
godexsoft: probably should be in initialization list instead? `: value_{value}` or similar.

src/rpc/common/Validators.h
```diff
@@ -107,6 +109,66 @@ struct Required final
...
+            auto const res = value_to<T>(value.as_object().at(key.data()));
+            if (val == res)
+                return Error{RPC::Status{
+                    RPC::RippledError::rpcNOT_SUPPORTED,
+                    fmt::format("Not supported field '{}'s value '{}'", std::string{key}, res)}};
+        }
+        return {};
+    }
+
+    NotSupported(T t)
```
godexsoft: Let's use better name for `val` and `t` :)

src/rpc/common/Validators.h
```diff
@@ -107,6 +109,66 @@ struct Required final
...
+template <typename... T>
+class NotSupported;
+
+/**
+ * @brief A specialized NotSupported validator that forbids a field to be present when the value equals the given value
+ */
+template <typename T>
+class NotSupported<T> final
+{
+    T val;
```
godexsoft: Convention is `value_` (with underscore) for private members 👍

src/rpc/ngHandlers/Ledger.cpp
```diff
@@ -0,0 +1,170 @@
...
+                        auto [txn, meta] = RPC::toExpandedJson(obj);
+                        entry = std::move(txn);
+                        entry[JS(metaData)] = std::move(meta);
+                    }
+                    else
+                    {
+                        entry[JS(tx_blob)] = ripple::strHex(obj.transaction);
+                        entry[JS(meta)] = ripple::strHex(obj.metadata);
+                    }
+                    // entry[JS(ledger_index)] = obj.ledgerSequence;
```
godexsoft: Can remove if we don't need this

src/rpc/ngHandlers/Ledger.cpp
```diff
@@ -0,0 +1,170 @@
...
+        boost::json::array& jsonDiff = output.header.at("diff").as_array();
+        auto diff = sharedPtrBackend_->fetchLedgerDiff(lgrInfo.seq, ctx.yield);
+        for (auto const& obj : diff)
+        {
+            boost::json::object entry;
+            entry["object_id"] = ripple::strHex(obj.key);
+            if (input.binary)
+                entry["object"] = ripple::strHex(obj.blob);
+            else if (obj.blob.size())
+            {
```
godexsoft, line 114: I'd rather not mix curly braces and no braces in one if-else statement block. If one of the clauses have enough complexity - others should probably have the same style (i.e. with braces in this case).

src/rpc/ngHandlers/Ledger.cpp
```diff
@@ -0,0 +1,170 @@
...
+                ripple::STLedgerEntry const sle{ripple::SerialIter{obj.blob.data(), obj.blob.size()}, obj.key};
+                entry["object"] = RPC::toJson(sle);
+            }
+            else
+                entry["object"] = "";
+            jsonDiff.push_back(std::move(entry));
```
godexsoft, line 123: Always good to add spacer line after end of blocks (like if-else blocks).

godexsoft (changes requested):
Adding a few things i found. Logically everything seems good 👍

godexsoft (approved):
LGTM! 🚀

PR #614, opened 2023-04-28
book_changes in the new RPC framework

cindyyan317:


src/subscriptions/SubscriptionManager.cpp
```diff
@@ -247,6 +247,7 @@ SubscriptionManager::pubBookChanges(
     ripple::LedgerInfo const& lgrInfo,
     std::vector<Backend::TransactionAndMetadata> const& transactions)
 {
+    // TODO: change to ngRPC after old RPC is removed
```
godexsoft, line 250: Actually, we will rename `ngRPC` to `RPC` once all code is migrated and old one is deleted 👍

godexsoft (approved):
LGTM! 🚀

PR #619, opened 2023-05-02
self host mac

cindyyan317:
Use self host mac node to run build-dev.
One timer related testcase failed at this node.


godexsoft (approved):
Let's try with this 👍

PR #622, opened 2023-05-03
Fix gmock warning of uninterested function 

cindyyan317:
Forgot to list all the expect call.
Add this to suppress the warning "Uninteresting mock function call"


godexsoft (approved):
Thanks!

PR #625, opened 2023-05-05
Fix bugs

cindyyan317:
1 tx and transaction_entry missed the "validated" field in response
2 account_channels and account_lines failed to process invalid marker
3 account_lines missed the "account" field in response


godexsoft (approved):
Code looks fine to me 👍

PR #627, opened 2023-05-09
Add handlers comments

cindyyan317:
Add handlers' brief and link

### DIFF
--- a/src/rpc/handlers/AccountCurrencies.h
+++ b/src/rpc/handlers/AccountCurrencies.h
@@ -27,10 +27,12 @@
 #include <set>
 
 namespace RPC {
+
 /**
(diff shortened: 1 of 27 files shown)


src/rpc/handlers/AccountCurrencies.h
```diff
@@ -27,6 +27,11 @@
 #include <set>
 
 namespace RPC {
+/**
```
godexsoft, line 31: Let's add a space line to each of these comments if it's missing - to match the format everywhere.
Also, I think the last link should be `For more details see https://...html` and be separated by a line of space from the brief.

godexsoft (changes requested):
The contents look alright. Let's fix the formatting and it's good to go. 👍

godexsoft (approved):
LGTM! 🚀

PR #629, opened 2023-05-09
Improve performance of Counters 

### LINKED ISSUE #478: RPC Counters - refactor with a simpler lock-based implementation
Currently RPC counters are implemented using both a mutex and atomics. This was presumably done this way because it was thought it is more performant. In reality however, atomics are rarely the answer (unless you really know what you are doing) and as [this benchmark](https://quick-bench.com/q/kDzGDBPKMhkG5LjLZJWiKHuSwdM) shows, a simple locked implementation is unsurprisingly ~2x faster. 

Let's use the much more readable and performant lock-based implementation.
Please also implement unit-tests for it as part of this task.

godexsoft:
Fixes #478 

Also introduces a single test for counters - to be enhanced with mocked WorkQueue later on.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,7 @@ if(BUILD_TESTS)
     unittests/rpc/ErrorTests.cpp
     unittests/rpc/BaseTests.cpp
     unittests/rpc/RPCHelpersTest.cpp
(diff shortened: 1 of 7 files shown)


cindyyan317 (approved):
LGTM

PR #631, opened 2023-05-10
Various small fixes

godexsoft:
A few small imrpovements
- Fix a race condition
- Refactor RPCEngine to use a strategy for verifying user is admin or not
- Fix small bug in UT due to uninitialized field

### DIFF
--- a/src/backend/SimpleCache.cpp
+++ b/src/backend/SimpleCache.cpp
@@ -96,9 +96,9 @@ SimpleCache::getPredecessor(ripple::uint256 const& key, uint32_t seq) const
 std::optional<Blob>
 SimpleCache::get(ripple::uint256 const& key, uint32_t seq) const
 {
+    std::shared_lock lck{mtx_};
     if (seq > latestSeq_)
         return {};
-    std::shared_lock lck{mtx_};
     objectReqCounter_++;
     auto e = map_.find(key);
     if (e == map_.end())
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -30,6 +30,7 @@
 #include <etl/ReportingETL.h>
 #include <log/Logger.h>
 #include <rpc/Counters.h>
+#include <rpc/RPCEngine.h>
 #include <rpc/common/impl/HandlerProvider.h>
 #include <webserver/Listener.h>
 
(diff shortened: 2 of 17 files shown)


src/backend/SimpleCache.cpp
```diff
@@ -96,9 +96,9 @@ SimpleCache::getPredecessor(ripple::uint256 const& key, uint32_t seq) const
 std::optional<Blob>
 SimpleCache::get(ripple::uint256 const& key, uint32_t seq) const
 {
+    std::shared_lock lck{mtx_};
```
cindyyan317, line 99: Good catch~

src/main/main.cpp
```diff
@@ -30,6 +30,7 @@
 #include <etl/ReportingETL.h>
 #include <log/Logger.h>
 #include <rpc/Counters.h>
+#include <rpc/RPCEngine.h>
```
cindyyan317, line 33: Why we add this header here?

src/webserver/HttpBase.h
```diff
@@ -41,6 +41,7 @@
 #include <main/Build.h>
 #include <rpc/Counters.h>
 #include <rpc/RPC.h>
+#include <rpc/RPCEngine.h>
```
cindyyan317, line 44: it is not used in this header?

src/webserver/WsBase.h
```diff
@@ -25,6 +25,7 @@
 #include <log/Logger.h>
 #include <rpc/Counters.h>
 #include <rpc/RPC.h>
+#include <rpc/RPCEngine.h>
```
cindyyan317, line 28: it is not used in this header?

src/main/main.cpp
godexsoft: For `RPCEngine::make_RPCEngine()` factory.

src/webserver/HttpBase.h
godexsoft: `RPC::RPCEngine` is used. Could maybe forward declare but meh.

src/webserver/WsBase.h
godexsoft: Same as for HttpBase.h :)

cindyyan317 (approved):
Nice to catch that race condition!
All looks good. Please check some headers seem no need to be added.

cindyyan317 (approved):
Nice to catch that race condition!
All looks good. Please check some headers seem no need to be added.

PR #635, opened 2023-05-12
Fix nft_history error to match rippled

### LINKED ISSUE #633: nft_history with ledger_index_min value greater than ledger_index_max do not return expected error.
Expected response: `lgrIdxsInvalid`

Sample request:
`{
    "method": "nft_history",
    "params": [
      {
          "nft_id": "0000000087B884851861E59B1445EFEE926FD28C62BC560F0000099B00000000",
          "ledger_index_min": 28495274,
          "ledger_index_max": 21377274
      }
    ]
}`

Clio response:
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "invalidIndex",
        "status": "error",
        "type": "response",
        "request": {
            "method": "nft_history",
            "params": [
                {
                    "nft_id": "0000000087B884851861E59B1445EFEE926FD28C62BC560F0000099B00000000",
                    "ledger_index_min": 28495274,
                    "ledger_index_max": 21377274
                }
            ]
        }
    },
    "warnings": [
        {
            "id": 2001,
        }
    ]
}`

godexsoft:
Fixes #633

### DIFF
--- a/src/rpc/handlers/NFTHistory.cpp
+++ b/src/rpc/handlers/NFTHistory.cpp
@@ -47,7 +47,7 @@ NFTHistoryHandler::process(NFTHistoryHandler::Input input, Context const& ctx) c
     }
 
     if (minIndex > maxIndex)
-        return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidIndex"}};
+        return Error{Status{RippledError::rpcLGR_IDXS_INVALID, "invalidIndex"}};
 
     if (input.ledgerHash || input.ledgerIndex)
     {
--- a/src/rpc/handlers/NFTHistory.cpp
+++ b/src/rpc/handlers/NFTHistory.cpp
@@ -47,7 +47,7 @@ NFTHistoryHandler::process(NFTHistoryHandler::Input input, Context const& ctx) c
     }
 
     if (minIndex > maxIndex)
-        return Error{Status{RippledError::rpcLGR_IDXS_INVALID, "invalidIndex"}};
+        return Error{Status{RippledError::rpcLGR_IDXS_INVALID}};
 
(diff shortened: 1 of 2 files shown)


godexsoft:
@cindyyan317 i did this just in case, don't remember if we decided that our version of the error was more correct or not. let me know if you think we should not merge this but instead adjust the tests.

Seems like there is also `account_tx` which has the same issue: 
```
if (minIndex > maxIndex)
        return Error{Status{RippledError::rpcLGR_IDXS_INVALID}};
```

I reckon we should match all 3 places of `if (minIndex > maxIndex)`. Currently two out of three use `rpcINVALID_PARAMS`.

cindyyan317:
You are correct. We have this issue in account_tx as well, which had been fixed already. 
account_tx is using `Error{Status{RippledError::rpcLGR_IDXS_INVALID}};` without specifying another error message. Now account_tx testcase passes.

src/rpc/handlers/NFTHistory.cpp
```diff
@@ -47,7 +47,7 @@ NFTHistoryHandler::process(NFTHistoryHandler::Input input, Context const& ctx) c
     }
 
     if (minIndex > maxIndex)
-        return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidIndex"}};
+        return Error{Status{RippledError::rpcLGR_IDXS_INVALID, "invalidIndex"}};
```
cindyyan317: Let us remove "invalidIndex" error message. It will use the ripped specified error message: "Ledger indexes invalid."

cindyyan317 (dismissed):

cindyyan317 (approved):
💯

PR #636, opened 2023-05-14
Rename functions to camelCase

godexsoft:
Some code was copy-pasted from examples which had snake_case functions. Renamed those to camelCase to match other code in clio.


src/webserver/Listener.h
```diff
@@ -141,7 +141,7 @@ class Detector : public std::enable_shared_from_this<Detector<PlainSession, SslS
 };
 
 void
-make_websocket_session(
+make_WebsocketSession(
```
cindyyan317, line 144: Why not makeWebsocketSession ?

godexsoft: Same like we have other `make_*` functions, don't want to change that as well in this pr - maybe another time :)

cindyyan317 (approved):

PR #646, opened 2023-05-24
Update the default limit

cindyyan317:
This PR is to fix some handlers' default limit not align with rippled's.


godexsoft (approved):
LGTM! 🔥

PR #649, opened 2023-05-25
Refactor ETL into smaller, testable components

godexsoft:
This pr splits old ETL code into smaller components that can be tested separately.
It also appears to fix the old TSAN warnings we were seeing.
As far as unit tests go, only one was added in this pr and the idea is to add other tests in other prs to prevent review fatigue and make prs smaller.


src/etl/ETLHelpers.h
```diff
@@ -89,10 +98,13 @@ class NetworkValidatedLedgers
     }
 };
 
-/// Generic thread-safe queue with an optional maximum size
-/// Note, we can't use a lockfree queue here, since we need the ability to wait
-/// for an element to be added or removed from the queue. These waits are
-/// blocking calls.
+// TODO: does the note make sense? lockfree queues provide the same blocking behaviour just without mutex, don't they?
```
cindyyan317, line 101: i don't quite understand this. lockfree queue can not wait?

src/etl/ETLHelpers.h
```diff
@@ -103,17 +115,26 @@ class ThreadSafeQueue
...
 public:
-    /// @param maxSize maximum size of the queue. Calls that would cause the
-    /// queue to exceed this size will block until free space is available
+    /**
+     * @brief Create an instance of the queue
+     *
+     * @param maxSize maximum size of the queue. Calls that would cause the queue to exceed this size will block until
+     * free space is available
+     */
     ThreadSafeQueue(uint32_t maxSize) : maxSize_(maxSize)
```
cindyyan317, line 124: It is not the issue of this PR, just found the maxSize_ can not be false, but member functions keep checking it.

src/etl/ETLHelpers.h
godexsoft: That's what the old comment was suggesting. It's worth checking but i think lockfree has nothing to do with waiting or not - it's just a way to implement the same async queue but using `compare_exchange_*` and without a `mutex`.

src/etl/ETLService.h
```diff
@@ -0,0 +1,240 @@
...
+    using LedgerFetcherType = clio::detail::LedgerFetcher<LoadBalancerType>;
+    using ExtractorType = clio::detail::Extractor<DataPipeType, NetworkValidatedLedgersType, LedgerFetcherType>;
+    using LedgerLoaderType = clio::detail::LedgerLoader<LoadBalancerType, LedgerFetcherType>;
+    using LedgerPublisherType = clio::detail::LedgerPublisher<SubscriptionManagerType>;
+    using TransformerType = clio::detail::Transformer<DataPipeType, LedgerLoaderType, LedgerPublisherType>;
+
+    clio::Logger log_{"ETL"};
+
+    std::shared_ptr<BackendInterface> backend_;
+    std::shared_ptr<SubscriptionManagerType> subscriptions_;
```
cindyyan317: not being used

src/etl/ETLHelpers.h
godexsoft: Well, it can be if the default constructor is used. However, in this codebase we do not use the default constructor so it could be simplified to just `uint32_t maxSize_` i think.

src/etl/ETLService.cpp
```diff
@@ -0,0 +1,260 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
cindyyan317: 2023

src/etl/impl/CacheLoader.h
```diff
@@ -0,0 +1,433 @@
...
+        {
+            loadCacheFromDb(seq);
+        }
+
+        // If loading synchronously, poll cache until full
+        while (cacheLoadStyle_ == LoadStyle::SYNC && not cache_.get().isFull())
+        {
+            log_.debug() << "Cache not full. Cache size = " << cache_.get().size() << ". Sleeping ...";
+            std::this_thread::sleep_for(std::chrono::seconds(10));
+            log_.info() << "Cache is full. Cache size = " << cache_.get().size();
```
cindyyan317: I think this code just copy from old clio, but the log content just not make sense, right?

src/etl/impl/ForwardCache.h
```diff
@@ -0,0 +1,83 @@
...
+class ForwardCache
+{
+    using ResponseType = std::optional<boost::json::object>;
+
+    clio::Logger log_{"ETL"};
+
+    mutable std::shared_mutex mtx_;
+    std::unordered_map<std::string, ResponseType> latestForwarded_;
+    boost::asio::io_context::strand strand_;
+    boost::asio::steady_timer timer_;
```
cindyyan317: I guess this part is just from the old code. i am curious how forward cache work, so took a look at the source. timer_ is not used ? i think we can remove it.

src/etl/impl/LedgerLoader.h
```diff
@@ -0,0 +1,253 @@
...
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <backend/BackendInterface.h>
+#include <backend/DBHelpers.h>  // for deserializeHeader: should be moved to util?
```
cindyyan317: agree, not anything about DB. You have a ledgerUtil, more suitable.

src/etl/impl/ForwardCache.h
godexsoft: Correct. I'm going to remove it - good catch.

src/etl/impl/LedgerPublisher.h
```diff
@@ -0,0 +1,253 @@
...
+    setLastClose(std::chrono::time_point<ripple::NetClock> lastCloseTime)
+    {
+        std::scoped_lock lck(closeTimeMtx_);
+        lastCloseTime_ = lastCloseTime;
+    }
+
+    void
+    setLastPublishTime()
+    {
+        std::scoped_lock lck(publishTimeMtx_);
```
cindyyan317, line 240: another old issue: since the publish run in one strand, does not need lock.

src/etl/impl/LedgerPublisher.h
```diff
@@ -0,0 +1,253 @@
...
+                    subscriptions_->pubTransaction(txAndMeta, lgrInfo);
+
+                subscriptions_->pubBookChanges(lgrInfo, transactions);
+
+                log_.info() << "Published ledger " << std::to_string(lgrInfo.seq);
+            }
+            else
+                log_.info() << "Skipping publishing ledger " << std::to_string(lgrInfo.seq);
+
+            setLastPublishTime();
```
cindyyan317: old issue : should be put in the age<600 check, otherwise the info not accurate.

src/etl/impl/LedgerPublisher.h
godexsoft: I added this lock because while writing it is only done from the strand, reading it is still done from another thread. You have to use a lock here. This actually fixed an ancient warning/error from TSAN :)

src/etl/impl/LedgerPublisher.h
godexsoft: Seems like this is only used for server_info and does not actually affect any logic so i'll go ahead and apply your suggestion!

src/etl/impl/LedgerPublisher.h
```diff
@@ -0,0 +1,253 @@
...
+     * @param lgrInfo the ledger to publish
+     */
+    void
+    publish(ripple::LedgerInfo const& lgrInfo)
+    {
+        boost::asio::post(publishStrand_, [this, lgrInfo = lgrInfo]() {
+            log_.info() << "Publishing ledger " << std::to_string(lgrInfo.seq);
+
+            if (!state_.get().isWriting)
+            {
```
cindyyan317, line 139: Can we move the cache update to ETLService, i can see two points:
1 make the functionality more separate.Publisher just publish to subscribers. We also can remove shared state isWriting.
2 I remember in future we want to  remove backend cache() function.

src/etl/impl/LedgerPublisher.h
```diff
@@ -0,0 +1,253 @@
...
+    setLastPublishTime()
+    {
+        std::scoped_lock lck(publishTimeMtx_);
+        lastPublish_ = std::chrono::system_clock::now();
+    }
+
+    void
+    setLastPublishedSequence(std::optional<uint32_t> lastPublishedSequence)
+    {
+        std::scoped_lock lck(lastPublishedSeqMtx_);
```
cindyyan317, line 247: do we need this variable lastPublishedSequence_? is it not the latest seq in backend?

src/etl/impl/LedgerPublisher.h
godexsoft: Actually, this lock was already there which was correct i think. I added the seq one.

src/etl/impl/LedgerPublisher.h
godexsoft: I added it because we still provide a getter for it and it is called from a different thread via publish().. this fixed a TSAN warning (actually it appears we don't have any TSAN messages after this fix).

src/etl/impl/LedgerPublisher.h
```diff
@@ -0,0 +1,253 @@
...
+
+    /**
+     * @brief Publish the passed in ledger
+     *
+     * All ledgers are published thru publishStrand_ which ensures that all publishes are performed in a serial fashion.
+     *
+     * @param lgrInfo the ledger to publish
+     */
+    void
+    publish(ripple::LedgerInfo const& lgrInfo)
```
godexsoft, line 133: It should be, yes. However, it is used from `Transformer` and therefore needs to stay public (for now at least). I don't want to change too much logic in the first pass of refactoring just so we don't end up with something that does not quite work the same way it did before the refactoring... Let's leave this one for later if you don't mind.

src/etl/impl/Transformer.h
```diff
@@ -0,0 +1,394 @@
...
+#include <chrono>
+#include <memory>
+#include <thread>
+
+namespace clio::detail {
+
+/*
+ * TODO:
+ *
+ * 1) loading of data into db should not really be part of transform right?
```
cindyyan317, line 41: loader_ here just to call the insertTransactions right? i feel these are some similar things happening in loader and transformer.

cindyyan317 (dismissed):
just small questions and old issues. we can discuss offline

src/etl/impl/LedgerPublisher.h
godexsoft: Would love to but that would mean it's no longer done from the `publishStrand_` and who knows what kind of side effect that will have. I'd rather do that in the future just so we don't change too much flow/logic with this pr.

src/etl/impl/Transformer.h
godexsoft: Yea, we currently write to cache/db from multiple places.. of course i'd like to change that, hence the todo. But at the moment i'd rather not change this fragile-looking logic and flow

src/etl/impl/LedgerPublisher.h
cindyyan317: Ha i see

cindyyan317 (approved):
Push push push

cindyyan317 (approved):
Push push push

cindyyan317 (approved):
Push push push

PR #671, opened 2023-06-06
Implement extractor tests

godexsoft:
Implements unit tests for ETL's Extractor component.
This is part of a bigger refactoring effort but can be reviewed and merged separately.

### DIFF
--- a/unittests/etl/ExtractorTest.cpp
+++ b/unittests/etl/ExtractorTest.cpp
@@ -82,7 +82,7 @@ TEST_F(ETLExtractorTest, StopsWhenCurrentSequenceExceedsFinishSequence)
     EXPECT_CALL(dataPipe_, push).Times(3);
     EXPECT_CALL(dataPipe_, finish(0)).Times(1);
 
-    // run the thread. expected to invoke for seq 0, 4, 8 and finally stop as seq will be greater than finishing seq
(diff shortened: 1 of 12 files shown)


unittests/etl/ExtractorTest.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    // expect to finish after just one response due to finishSequence set to 1
+    extractor_ = std::make_unique<ExtractorType>(dataPipe_, networkValidatedLedgers_, ledgerFetcher_, 0, 1, state_);
+    extractor_->waitTillFinished();  // this is what clio does too. waiting for the thread to join
+
+    EXPECT_TRUE(optionalResponse.has_value());
+    EXPECT_EQ(optionalResponse.value(), response);
+}
+
+TEST_F(ETLExtractorTest, CallsPipeFinishWithInitialSequenceAtExit)
```
cindyyan317, line 164: Is this one the same as StopsOnServerShutdown?

cindyyan317 (approved):
All good!

unittests/etl/ExtractorTest.cpp
godexsoft: Essentially yes :) but it checks for different thing so logically it's a different test.. although the code is basically the same. It's all microseconds so i thought it's better to have it than not 👍

PR #676, opened 2023-06-08
Move the json check to RPC executor

cindyyan317:


src/webserver/details/HttpBase.h
```diff
@@ -207,21 +207,10 @@ class HttpBase : public ConnectionBase
...
-            return sender_(httpResponse(
-                http::status::ok,
-                "application/json",
-                boost::json::serialize(RPC::makeError(RPC::RippledError::rpcBAD_SYNTAX))));
-        }
-        try
-        {
-            (*handler_)(std::move(request), derived().shared_from_this());
+            std::string reqStr = req_.body();
+            (*handler_)(std::move(reqStr), derived().shared_from_this());
```
godexsoft: Looks like we are attempting to move the request string here but the receiving end is now `const&`. I think we can actually do the move as `reqStr` is not used anymore in this scope. Consider updating either here or in the executor 👍

src/webserver/details/WsBase.h
```diff
@@ -221,12 +221,19 @@ class WsSession : public ConnectionBase, public std::enable_shared_from_this<WsS
...
-        auto sendError = [this](auto error, boost::json::value const& request) {
+        auto sendError = [this](auto error, std::string const& requestStr) {
             auto e = RPC::makeError(error);
-
-            if (request.is_object() && request.as_object().contains("id"))
-                e["id"] = request.as_object().at("id");
-            e["request"] = request;
+            try
+            {
+                auto request = boost::json::parse(requestStr);
```
godexsoft, line 228: I thought we are trying to move all json work out of here? Maybe we can introduce something on the handler object that will do this, possibly without actually sending but just computing the response instead, leaving the sending part for WsSession. Wdyt?

src/webserver/details/WsBase.h
```diff
@@ -221,12 +221,19 @@ class WsSession : public ConnectionBase, public std::enable_shared_from_this<WsS
...
+                if (request.is_object() && request.as_object().contains("id"))
+                    e["id"] = request.as_object().at("id");
+                e["request"] = std::move(request);
+            }
+            catch (std::exception&)
+            {
+                e["request"] = std::move(requestStr);
```
godexsoft, line 235: Trying to move a const ref here

src/webserver/interface/Concepts.h
```diff
@@ -33,9 +32,9 @@ namespace Server {
  */
 // clang-format off
 template <typename T>
-concept ServerHandler = requires(T handler, boost::json::object&& req, std::shared_ptr<ConnectionBase> const& ws, boost::beast::error_code ec) {
+concept ServerHandler = requires(T handler, std::string const& req, std::shared_ptr<ConnectionBase> const& ws, boost::beast::error_code ec) {
```
godexsoft, line 35: I think you don't need to specify `const&` or anything like that here, imagine that this is a variable of some type (e.g. `string` or `shared_ptr`), it will already bind just fine to the actual `const&` we have in the implementation.

unittests/webserver/RPCExecutorTest.cpp
```diff
@@ -673,3 +651,24 @@ TEST_F(WebRPCExecutorTest, HTTPTooBusy)
     (*rpcExecutor2)(std::move(request), session);
     EXPECT_EQ(boost::json::parse(session->message), boost::json::parse(response));
 }
+
+TEST_F(WebRPCExecutorTest, HTTPRequestNotJson)
+{
+    static auto constexpr request = "not json";
+    static auto constexpr response =
+        R"({"error":"badSyntax","error_code":1,"error_message":"Syntax error.","status":"error","type":"response"})";
```
godexsoft: Formatting please 🚀

unittests/webserver/RPCExecutorTest.cpp
```diff
@@ -673,3 +651,24 @@ TEST_F(WebRPCExecutorTest, HTTPTooBusy)
...
+    (*rpcExecutor)(std::move(request), session);
+    EXPECT_EQ(boost::json::parse(session->message), boost::json::parse(response));
+}
+
+TEST_F(WebRPCExecutorTest, WsRequestNotJson)
+{
+    session->upgraded = true;
+    static auto constexpr request = "not json";
+    static auto constexpr response =
+        R"({"error":"badSyntax","error_code":1,"error_message":"Syntax error.","status":"error","type":"response"})";
```
godexsoft: And here 🥇

unittests/webserver/ServerTest.cpp
```diff
@@ -207,73 +207,40 @@ TEST_F(WebServerTest, Ws)
...
+    auto e = std::make_shared<ExceptionExecutor>();
     auto const server = Server::make_HttpServer(cfg, ctx, std::nullopt, dosGuard, e);
-    wsClient.connect("localhost", "8888");
-    auto const res = wsClient.syncPost(R"({)");
-    wsClient.disconnect();
+    auto const res = HttpSyncClient::syncPost("localhost", "8888", R"({})");
     EXPECT_EQ(
         res,
-        R"({"error":"badSyntax","error_code":1,"error_message":"Syntax error.","status":"error","type":"response","request":["{"]})");
+        R"({"error":"internal","error_code":73,"error_message":"Internal error.","status":"error","type":"response"})");
```
godexsoft, line 217: As well as here and below please 🙏

godexsoft (changes requested):
Some move vs const ref business needs fixing ⭐

src/webserver/details/WsBase.h
cindyyan317: server still need to send the error when dosguard overload or error happens, for web socket, it needs to attach the id.

src/webserver/details/HttpBase.h
cindyyan317: Good catch . the last minute i changed the && to const&. missed this change, oops

godexsoft (approved):
Looks alright 👍

Issue #679, opened 2023-06-08
Unsubscribe to unsupported streams do not return reportingUnsupported error


mounikakun:
As per doc reporting Mode servers return the error reportingUnsupported for server, peer_status, consensus streams but Clio returns malformedStream

PR #682, opened 2023-06-09
Subscribe ReportingNotSupported error

cindyyan317:
Rippled's reporting mode returns "ReportingNotSupported" for "peer_status" and "consensus" streams while clio returns streamMalformed like other invalid stream name.

This PR is to add ReportingNotSupported error for "peer_status" and "consensus"
#678  #679


godexsoft (approved):
LGTM 👍

PR #683, opened 2023-06-09
Feature/count real errors

### LINKED ISSUE #665: Add counter for real RPC errors
Currently we have counters for started, finished, errored and forwarded invocations per RPC handler that we periodically monitor to produce graphs. However, `errored` is a counter that tracks wrong user input, missing data, etc.
However, we want to distinguish between such errors and actual server issues such as db timeouts, internal errors, etc.
Let's introduce a new counter for those edge cases.

godexsoft:
Fixes #665 

Adds a few new counters to distinguish user errors from system errors as well as allow for more fine grained stats.


cindyyan317 (approved):
Great work! I like the webserver's unawareness of counter.

src/rpc/RPCEngine.h
```diff
@@ -149,24 +156,23 @@ class RPCEngineBase
             if (v)
                 return v->as_object();
             else
+            {
+                notifyErrored(ctx.method);
```
cindyyan317, line 160: One thing, do we need to change the name to logicError ? suggest it is not a real error

godexsoft: We could change this `notifyErrored` and inside Counters but we can't change the json we generate in the report i think. So i left it as it was to match the key in the report itself.

PR #684, opened 2023-06-09
Fix mismatch : subscribe book offers when both is true

cindyyan317:
There is no document regarding this. 
According to rippled's code,
When snapshot is true, we should return the list of the current offers.
When both is false, these offers put into offers field
When both is true, the original book's offers go to the 'bids' field, the reversed book's offers go to the 'asks' field.
'offers' will be present if any book in subscribed book list has snapshot being true but both being false.
'bids' 'asks' will be present if any book in subscribed book list has snapshot being true and both being true.


godexsoft (approved):
Looks good! 🚀

PR #687, opened 2023-06-12
Fix marker issue

cindyyan317:
The first tx of account_tx's second page overlaps the last tx of the first page. This should be a old issue for clio. 

SQL query should exclude the marker.


godexsoft (approved):
Looks fine 👍

PR #689, opened 2023-06-12
Implement basic transformer tests

godexsoft:
This pr adds a bunch of mocks and fakes to support ETL tests as well as implements a few tests for the Transformer.
By no means is this complete but it may be just enough for a start. Next iteration it would be good to separate some of the transformer logic into another entity that can be tested separately from the threading facilities it provides. Transform is doing way too much atm. so it's not worth spending more time adding tests for what it is atm.

### DIFF
--- a/src/rpc/RPCEngine.h
+++ b/src/rpc/RPCEngine.h
@@ -186,10 +186,7 @@ public:
     bool
     post(Fn&& func, std::string const& ip)
     {
-        if (!workQueue_.get().postCoro(std::forward<Fn>(func), dosGuard_.get().isWhiteListed(ip)))
-            return false;
-
-        return true;
+        return workQueue_.get().postCoro(std::forward<Fn>(func), dosGuard_.get().isWhiteListed(ip));
     }
 
     /**
--- a/src/rpc/RPCEngine.h
+++ b/src/rpc/RPCEngine.h
@@ -187,10 +187,7 @@ public:
     post(Fn&& func, std::string const& ip)
     {
         if (!workQueue_.get().postCoro(std::forward<Fn>(func), dosGuard_.get().isWhiteListed(ip)))
-        {
-            notifyTooBusy();
             return false;
-        }
 
         return true;
     }
--- a/unittests/backend/cassandra/BackendTests.cpp
+++ b/unittests/backend/cassandra/BackendTests.cpp
@@ -18,13 +18,13 @@
 //==============================================================================
 
 #include <util/Fixtures.h>
+#include <util/StringUtils.h>
 
 #include <backend/CassandraBackend.h>
 #include <config/Config.h>
 #include <etl/NFTHelpers.h>
 #include <rpc/RPCHelpers.h>
 
-#include <ripple/basics/base_uint.h>
 #include <boost/json/parse.hpp>
 #include <fmt/compile.h>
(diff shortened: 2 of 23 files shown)


src/rpc/RPCEngine.h
```diff
@@ -187,10 +187,7 @@ class RPCEngineBase
     post(Fn&& func, std::string const& ip)
     {
         if (!workQueue_.get().postCoro(std::forward<Fn>(func), dosGuard_.get().isWhiteListed(ip)))
-        {
-            notifyTooBusy();
```
godexsoft, line 191: Unrelated change but decided to do it asap because we were basically counting this twice. Here and in the function that calls this.

godexsoft: This looks like now it can also be simplified to just `return workQueue_...`

unittests/backend/cassandra/BackendTests.cpp
```diff
@@ -89,29 +89,6 @@ TEST_F(BackendCassandraTest, Basic)
             "3E2232B33EF57CECAC2816E3122816E31A0A00F8377CD95DFA484CFAE282656A58"
             "CE5AA29652EFFD80AC59CD91416E4E13DBBE";
 
-        auto hexStringToBinaryString = [](auto const& hex) {
```
godexsoft, line 92: Removing duplicates and introducing some StringUtils that can be used from multiple places.

unittests/etl/TransformerTest.cpp
```diff
@@ -0,0 +1,151 @@
...
+    EXPECT_CALL(*rawBackendPtr, writeAccountTransactions).Times(AtLeast(1));
+    EXPECT_CALL(*rawBackendPtr, writeNFTs).Times(AtLeast(1));
+    EXPECT_CALL(*rawBackendPtr, writeNFTTransactions).Times(AtLeast(1));
+    EXPECT_CALL(*rawBackendPtr, doFinishWrites).Times(AtLeast(1));
+
+    // should not call publish
+    EXPECT_CALL(ledgerPublisher_, publish(_)).Times(0);
+
+    transformer_ =
+        std::make_unique<TransformerType>(dataPipe_, mockBackendPtr, ledgerLoader_, ledgerPublisher_, 0, state_);
```
cindyyan317, line 150: There is no sleep and explicit waiting . I am puzzled why EXPECT_CALL still pass.

unittests/util/StringUtils.cpp
```diff
@@ -0,0 +1,52 @@
...
+binaryStringToUint256(std::string const& bin)
+{
+    ripple::uint256 uint;
+    return uint.fromVoid((void const*)bin.data());
+}
+
+std::string
+ledgerInfoToBinaryString(ripple::LedgerInfo const& info)
+{
+    auto blob = RPC::ledgerInfoToBlob(info, true);
```
cindyyan317: const

cindyyan317 (dismissed):
LGTM ~ So many mocks 👍

unittests/etl/TransformerTest.cpp
godexsoft: Both transformer and extractor have `join` in the destructor :)

cindyyan317 (approved):

PR #691, opened 2023-06-13
Read only for backend

cindyyan317:
Not to create keyspace and tables if clio is running read-only mode.

If there is no keyspace and tables created, read-only clio will exit with the hint:

**backend/cassandra/Schema.h:649 [2023-06-13 10:31:29.369997] Backend:NFO Preparing cassandra statements
src/backend/CassandraBackend.h:94 [2023-06-13 10:31:29.371811] Backend:ERR Failed to prepare the statements for Clio-readonly:true
src/main/main.cpp:224 [2023-06-13 10:31:29.373172] General:FTL Exit on exception: Invalid query: keyspace clio does not exist**

### DIFF
--- a/CMake/deps/gtest.cmake
+++ b/CMake/deps/gtest.cmake
@@ -17,4 +17,6 @@ enable_testing()
 
 include(GoogleTest)
 
-gtest_discover_tests(clio_tests)
+#increase timeout for tests discovery to 10 seconds, by default it is 5s. As more unittests added, we start to hit this issue
+#https://github.com/google/googletest/issues/3475
+gtest_discover_tests(clio_tests DISCOVERY_TIMEOUT 10)
--- a/src/backend/CassandraBackend.h
+++ b/src/backend/CassandraBackend.h
@@ -72,7 +72,7 @@ public:
         if (auto const res = handle_.connect(); not res)
             throw std::runtime_error("Could not connect to Cassandra: " + res.error());
 
-        if (!readOnly)
+        if (not readOnly)
         {
             if (auto const res = handle_.execute(schema_.createKeyspace); not res)
             {
@@ -85,15 +85,17 @@ public:
             if (auto const res = handle_.executeEach(schema_.createSchema); not res)
                 throw std::runtime_error("Could not create schema: " + res.error());
         }
+
         try
         {
             schema_.prepareStatements(handle_);
         }
-        catch (std::exception const& ex)
+        catch (std::runtime_error const& ex)
         {
-            log_.error() << "Failed to prepare the statements for Clio-Readonly:" << readOnly;
+            log_.error() << "Failed to prepare the statements: " << ex.what() << " readOnly: " << readOnly;
             throw;
         }
+
         log_.info() << "Created (revamped) CassandraBackend";
     }
 
--- a/src/backend/CassandraBackend.h
+++ b/src/backend/CassandraBackend.h
@@ -63,7 +63,7 @@ public:
      *
      * @param settingsProvider
      */
-    BasicCassandraBackend(SettingsProviderType settingsProvider)
+    BasicCassandraBackend(SettingsProviderType settingsProvider, bool readOnly)
         : settingsProvider_{std::move(settingsProvider)}
         , schema_{settingsProvider_}
         , handle_{settingsProvider_.getSettings()}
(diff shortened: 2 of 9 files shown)


src/backend/CassandraBackend.h
```diff
@@ -72,18 +72,28 @@ class BasicCassandraBackend : public BackendInterface
...
+        }
+        try
+        {
+            schema_.prepareStatements(handle_);
+        }
+        catch (std::exception const& ex)
+        {
+            log_.error() << "Failed to prepare the statements for Clio-Readonly:" << readOnly;
+            throw;
+        }
```
godexsoft, line 97: Whitespace needed

src/backend/CassandraBackend.h
```diff
@@ -72,18 +72,28 @@ class BasicCassandraBackend : public BackendInterface
...
+                // on datastax, creation of keyspaces can be configured to only be done thru the admin interface.
+                // this does not mean that the keyspace does not already exist tho.
+                if (res.error().code() != CASS_ERROR_SERVER_UNAUTHORIZED)
+                    throw std::runtime_error("Could not create keyspace: " + res.error());
+            }
 
-        schema_.prepareStatements(handle_);
+            if (auto const res = handle_.executeEach(schema_.createSchema); not res)
+                throw std::runtime_error("Could not create schema: " + res.error());
+        }
```
godexsoft, line 87: Whitespace needed

src/backend/CassandraBackend.h
```diff
@@ -72,18 +72,28 @@ class BasicCassandraBackend : public BackendInterface
         if (auto const res = handle_.connect(); not res)
             throw std::runtime_error("Could not connect to Cassandra: " + res.error());
 
-        if (auto const res = handle_.execute(schema_.createKeyspace); not res)
+        if (!readOnly)
```
godexsoft: Perfect place to use `if (not readOnly)` syntax :octocat:

src/backend/CassandraBackend.h
```diff
@@ -72,18 +72,28 @@ class BasicCassandraBackend : public BackendInterface
...
 
-        schema_.prepareStatements(handle_);
+            if (auto const res = handle_.executeEach(schema_.createSchema); not res)
+                throw std::runtime_error("Could not create schema: " + res.error());
+        }
+        try
+        {
+            schema_.prepareStatements(handle_);
+        }
+        catch (std::exception const& ex)
```
godexsoft: Is there maybe a finer exception type we can catch here? What if some unrelated exception happens and readOnly was not actually set to true?

src/backend/CassandraBackend.h
```diff
@@ -72,18 +72,28 @@ class BasicCassandraBackend : public BackendInterface
...
+        {
+            log_.error() << "Failed to prepare the statements for Clio-Readonly:" << readOnly;
```
godexsoft: Maybe something like `"Failed to prepare statements: " << ex.what() << "; readonly: " << readOnly` would be slightly nicer?

godexsoft (changes requested):
Few small things can be made slightly nicer. Other than that it looks like it works 👍

src/backend/CassandraBackend.h
cindyyan317: I change the type to runtime_error which is raised here -> https://github.com/XRPLF/clio/blob/a76194d299f2b853ba865447a5f45f17a9ebac56/src/backend/cassandra/Handle.cpp#L152. 
I think we may have other exception happens but readonly is not true. So I print out the readonly value.

src/backend/CassandraBackend.h
cindyyan317: Fixed 👍 : The error msg is :
backend/cassandra/Schema.h:649 [2023-06-13 16:29:33.971595] Backend:NFO Preparing cassandra statements
src/backend/CassandraBackend.h:95 [2023-06-13 16:29:33.973286] Backend:ERR Failed to prepare the statements: Invalid query: keyspace clio does not exist readOnly: true
src/main/main.cpp:224 [2023-06-13 16:29:33.974685] General:FTL Exit on exception: Invalid query: keyspace clio does not exist

godexsoft: This is better but still unreadable `Failed to prepare the statements: Invalid query: keyspace clio does not exist readOnly: true`.
Please separate the `readOnly: true` part from the exception message (my recommendation is to use `;`) otherwise it looks like part of the same message 👍

src/backend/CassandraBackend.h
godexsoft: Looks like whitespace (empty line) still missing?

src/backend/CassandraBackend.h
godexsoft: It's good you print it - i'm only concerned with the formatting of the error message.. users will have a hard time if it's all collapsed and have no spaces.

src/backend/CassandraBackend.h
godexsoft: Looks like still needed.

godexsoft (changes requested):
Still a few things to iron out ⭐

CMake/deps/gtest.cmake
```diff
@@ -17,4 +17,6 @@ enable_testing()
 
 include(GoogleTest)
 
-gtest_discover_tests(clio_tests)
+#increase timeout for tests discovery to 10 seconds, by default it is 5s. As more unittests added, we start to hit this issue
+#https://github.com/google/googletest/issues/3475
+gtest_discover_tests(clio_tests DISCOVERY_TIMEOUT 10)
```
godexsoft, line 22: Wow. Great find 🥇

godexsoft (dismissed):
Looks good! thanks for taking the time to fix the small details and writing the new tests 🌟

godexsoft (approved):
👍

Issue #693, opened 2023-06-14
account_tx bug: return all the transactions when ledger_index is "validated"


cindyyan317:
Should only return the transactions of the latest validated ledger.

PR #694, opened 2023-06-14
account_tx bug fix

cindyyan317:
Two issues:
1 When ledger_index is specified as "validated", clio returns all the transactions in the db. The correct behavior is to return the transactions only in the latest validated ledger. #693 
2 When forward is specified as "true", clio does not return the transactions of the minimum seq with transaction idx 0.

### DIFF
--- a/src/backend/cassandra/Schema.h
+++ b/src/backend/cassandra/Schema.h
@@ -528,19 +528,21 @@ public:
                 SELECT hash, seq_idx 
                   FROM {}               
                  WHERE account = ?
-                   AND seq_idx < ?
+                   AND seq_idx <= ?
                  LIMIT ?
                 )",
                 qualifiedTableName(settingsProvider_.get(), "account_tx")));
         }();
 
+        // the smallest transaction idx is 0, we use uint to store the transaction index, so we shall use ">=" to
+        // include it(the transaction with idx 0) in the result
         PreparedStatement selectAccountTxForward = [this]() {
             return handle_.get().prepare(fmt::format(
                 R"(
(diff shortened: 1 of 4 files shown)


src/backend/cassandra/Schema.h
```diff
@@ -528,19 +528,21 @@ class Schema
                 SELECT hash, seq_idx 
                   FROM {}               
                  WHERE account = ?
-                   AND seq_idx < ?
+                   AND seq_idx <= ?
```
cindyyan317, line 531: Keep the same as selectAccountTxForward

src/backend/cassandra/Schema.h
```diff
@@ -528,19 +528,21 @@ class Schema
...
                  LIMIT ?
                 )",
                 qualifiedTableName(settingsProvider_.get(), "account_tx")));
         }();
 
+        // the smallest transaction idx is 0, we use uint to store the transaction index, so we shall use ">=" to
+        // include it(the transaction with idx 0) in the result
```
godexsoft, line 538: This seems to be changing the output.. is this the same behaviour rippled has as far as output goes?

godexsoft (approved):
Looks good to me 👍 Please just make sure that the output produced after the change in included sequences is matching rippled 🥇

src/backend/cassandra/Schema.h
cindyyan317: This is to fix an old issue.
Imagine if we query all the transaction  > ( seq1 , 0) which old clio did, we actually missed all the transactions of seq1 with transaction idx 0 .

PR #695, opened 2023-06-14
Support `api_version`

### LINKED ISSUE #64: Support API versioning
rippled actually has a version 2 of the API that is not yet enabled. We need to implement those.

godexsoft:
Fixes #64 

Adds a parser utility for parsing `api_version` out of request and validate it against configured values.
Supports reading configuration from config file with fallback to hardcoded values on misconfiguration.

### DIFF
--- a/src/rpc/Errors.cpp
+++ b/src/rpc/Errors.cpp
@@ -18,6 +18,7 @@
 //==============================================================================
 
 #include <rpc/Errors.h>
+#include <rpc/JS.h>
 
 #include <algorithm>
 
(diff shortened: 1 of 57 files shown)


src/rpc/Errors.cpp
```diff
@@ -77,7 +78,8 @@ getErrorInfo(ClioError code)
         {ClioError::rpcMALFORMED_REQUEST, "malformedRequest", "Malformed request."},
         {ClioError::rpcMALFORMED_OWNER, "malformedOwner", "Malformed owner."},
         {ClioError::rpcMALFORMED_ADDRESS, "malformedAddress", "Malformed address."},
-        {ClioError::rpcINVALID_HOT_WALLET, "invalidHotWallet", "Invalid hot wallet."}};
+        {ClioError::rpcINVALID_HOT_WALLET, "invalidHotWallet", "Invalid hot wallet."},
+        {ClioError::rpcINVALID_API_VERSION, JS(invalid_API_version), "Invalid API version."}};
```
godexsoft, line 82: I'd like to use `invalidAPIVersion` or `invalidVersion` instead but this is what `rippled` returns so it may have to stay this way.

src/rpc/common/APIVersion.h
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @brief Default API version to use if no version is specified by clients
+ */
+static constexpr uint32_t API_VERSION_DEFAULT = 1u;
+
+/**
+ * @brief Minimum API version supported by this build
+ */
+static constexpr uint32_t API_VERSION_MIN = 1u;
```
godexsoft, line 39: Hm. we should probably sanity check that whatever is specified by config is within the hardcoded values here.

cindyyan317 (approved):
LGTM

PR #699, opened 2023-06-16
Fix account_tx marker issue

cindyyan317:
There 2 two issues in account_tx.
1 The last tx in the page overlaps the first tx in the following page when marker is used. To fix it, we change the SQL to exclusive  < >
2 The tx of transaction index 0 can not be filtered out. To fix it , we change the SQL back to <= ; >=

This fix is to change the SQL to < and >. But decrease the seq number (seq-1, int32_max) to include the (seq,0) tx. According to the doc, the earliest ledger index is 1. https://xrpl.org/basic-data-types.html#ledger-index

### DIFF
--- a/src/backend/cassandra/Schema.h
+++ b/src/backend/cassandra/Schema.h
@@ -528,21 +528,19 @@ public:
                 SELECT hash, seq_idx 
                   FROM {}               
                  WHERE account = ?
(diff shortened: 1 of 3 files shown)


godexsoft (approved):
👍

PR #701, opened 2023-06-16
Remove sensitive info from log

cindyyan317:


src/util/JsonUtils.h
```diff
@@ -0,0 +1,51 @@
...
+    {
+        for (auto const& secretField : secretFields)
+        {
+            newObject.at("params").as_array()[0].as_object().erase(secretField);
+        }
+    }
+    // for websocket requests
+    for (auto const& secretField : secretFields)
+    {
+        newObject.erase(secretField);
```
godexsoft: I think instead of erasing it would be better to "mask" it. So change the actual secret with something like "<removed>". Just so when something goes wrong we can still see in the log that they really did send the secret field and did not forget to send it.

src/util/JsonUtils.h
```diff
@@ -0,0 +1,51 @@
...
+namespace util {
+
+inline boost::json::object
+removeSecret(boost::json::object const& object)
+{
+    auto newObject = object;
+    auto const secretFields = {"secret", "seed", "seed_hex", "passphrase"};
+
+    if (newObject.contains("params") and newObject.at("params").is_array() and
+        newObject.at("params").as_array().size() > 0 and newObject.at("params").as_array()[0].is_object())
```
godexsoft: `.size() > 0` could probably be `not .empty()` instead?

godexsoft (changes requested):
Pretty cool 👍 leaving a few small suggestions.

godexsoft (approved):
LGTM! 🌟

PR #707, opened 2023-06-21
Handle runtime error when deserialise the unknown fields

cindyyan317:
This PR protects the crash when deserialising the ledger and tx.
For RPC handlers part, almost every handle will deserialise the blob. When crash happen at RPC handle, it will be caught as internal error. We need more time to fix RPC side.


cindyyan317:
This approach will corrupt the db.

PR #708, opened 2023-06-21
Stop etl when crash

cindyyan317:


src/etl/ETLService.cpp
```diff
@@ -83,31 +83,39 @@ ETLService::monitor()
...
+                    log_.info() << "The wait for the next validated "
+                                << "ledger has been aborted. "
+                                << "Exiting monitor loop";
+                    return;
+                }
             }
         }
+        catch (std::runtime_error const& e)
+        {
+            log_.error() << "Failed to load initial ledger, Exiting monitor loop : " << e.what();
```
godexsoft: Maybe this should be a fatal error message for extra visibility. And we should probably say something like "Unknown fields detected in ledger data. Perhaps a new amendment was enabled? ETL is not possible until Clio is updated with a compatible version. Stopping ETL". or something along these lines.

src/etl/NFTHelpers.cpp
```diff
@@ -326,15 +326,21 @@ std::vector<NFTsData>
...
+        ripple::STLedgerEntry const sle =
+            ripple::STLedgerEntry(ripple::SerialIter{blob.data(), blob.size()}, ripple::uint256::fromVoid(key.data()));
+        if (sle.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltNFTOKEN_PAGE)
+            return nfts;
+
+        auto const owner = ripple::AccountID::fromVoid(key.data());
+        for (ripple::STObject const& node : sle.getFieldArray(ripple::sfNFTokens))
+            nfts.emplace_back(node.getFieldH256(ripple::sfNFTokenID), seq, owner, node.getFieldVL(ripple::sfURI));
+    }
+    catch (std::runtime_error const& e)
```
godexsoft: `e` is unused here and `return nfts;` can be omitted (empty body to ignore exception is fine).

godexsoft (changes requested):
Few small suggestions 👍

godexsoft (approved):
LGTM 👍

PR #711, opened 2023-06-22
Use InvalidLgrRange for type error

cindyyan317:
Usually Clio returns "invalidParam" error for parameter type error, for example, int is given for a field expecting string.
But for tx, the doc specifies 
"invalidLgrRange - The specified min_ledger is larger than the max_ledger, or one of those parameters is not a valid ledger index."

### DIFF
--- a/unittests/rpc/handlers/TxTest.cpp
+++ b/unittests/rpc/handlers/TxTest.cpp
@@ -47,7 +47,7 @@ TEST_F(RPCTxTest, ExcessiveLgrRange)
                 "transaction": "{}",
                 "min_ledger": 1,
                 "max_ledger":1002
-                }})",
+            }})",
             TXNID));
         auto const output = handler.process(req, Context{std::ref(yield)});
         ASSERT_FALSE(output);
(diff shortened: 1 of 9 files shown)


unittests/rpc/handlers/TxTest.cpp
```diff
@@ -79,6 +79,48 @@ TEST_F(RPCTxTest, InvalidLgrRange)
...
+{
+    runSpawn([this](auto& yield) {
+        auto const handler = AnyHandler{TxHandler{mockBackendPtr}};
+        auto const req = json::parse(fmt::format(
+            R"({{ 
+                "command": "tx",
+                "transaction": "{}",
+                "max_ledger": "xx",
+                "min_ledger": 10
+                }})",
```
godexsoft: Let's fix formatting here and below 👍

godexsoft (commented):
As usual, let's try to keep formatting uniform. 
Otherwise looks good 🚀

godexsoft (approved):
Nice 👍

PR #716, opened 2023-06-23
Forward api_version 1 requests to rippled

godexsoft:
Introduces forwarding for anything that is marked as api_version 1.


src/rpc/common/impl/ForwardingProxy.h
```diff
@@ -0,0 +1,136 @@
...
+
+        if (specifiesCurrentOrClosedLedger(request))
+            return true;
+
+        if (ctx.method == "account_info" && request.contains("queue") && request.at("queue").is_bool() &&
+            request.at("queue").as_bool())
+            return true;
+
+        // TODO: if needed, make configurable with json config option
+        if (ctx.apiVersion == 1)
```
cindyyan317: This condition check is simple. Maybe we can put it earlier, do a shortcut.

src/rpc/common/impl/ForwardingProxy.h
```diff
@@ -0,0 +1,136 @@
...
+    ForwardingProxy(
+        std::shared_ptr<LoadBalancerType> const& balancer,
+        CountersType& counters,
+        std::shared_ptr<HandlerProviderType const> const& handlerProvider)
+        : balancer_{balancer}, counters_{std::ref(counters)}, handlerProvider_{handlerProvider}
+    {
+    }
+
+    bool
+    shouldForward(Web::Context const& ctx) const
```
cindyyan317, line 54: It 's not this PR's issue. I am thinking we should not forward the request "subscribe" "unsubscribe" no matter what, right?

cindyyan317 (dismissed):

cindyyan317 (approved):
More unittest added ~ 👍

PR #720, opened 2023-06-27
Implement custom http errors

### LINKED ISSUE #697: Look into JSON-RPC 400 errors
For rippled: `curl -v -d '{"params": [{"api_version": 1}]}' 127.0.0.1:5005`
```
*   Trying 127.0.0.1:5005...
* Connected to 127.0.0.1 (127.0.0.1) port 5005 (#0)
> POST / HTTP/1.1
> Host: 127.0.0.1:5005
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Length: 32
> Content-Type: application/x-www-form-urlencoded
>
< HTTP/1.1 400 Bad Request
< Date: Wed, 14 Jun 2023 19:18:44 +0000
< Connection: Keep-Alive
< Content-Length: 13
< Content-Type: application/json; charset=UTF-8
< Server: ripple-json-rpc/rippled-1.11.0-rc3
<
Null method
```

But Clio: `curl -v -d '{"params": [{"api_version": 1}]}' 127.0.0.1:51233`
```
*   Trying 127.0.0.1:51233...
* Connected to 127.0.0.1 (127.0.0.1) port 51233 (#0)
> POST / HTTP/1.1
> Host: 127.0.0.1:51233
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Length: 32
> Content-Type: application/x-www-form-urlencoded
>
< HTTP/1.1 200 OK
< Server: clio-server-20230614163346-feature/api-version-support-a76194d+DEBUG
< Content-Type: application/json
< Content-Length: 185
<
{"result":{"error":"badSyntax","error_code":1,"error_message":"Method is not specified or is not a string.","status":"error","type":"response","request":{"params":[{"api_version":1}]}}}
```

According to [the documentation](https://xrpl.org/error-formatting.html#json-rpc-format) rippled is doing the right thing in this case. We should probably do the same.

godexsoft:
Fixes #697 

Added as many as i could find. Needs to be tested by integration to see if something broke or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,7 @@ if(BUILD_TESTS)
     unittests/backend/cassandra/ExecutionStrategyTests.cpp
     unittests/backend/cassandra/AsyncExecutorTests.cpp
     unittests/webserver/ServerTest.cpp
-    unittests/webserver/RPCExecutorTest.cpp)
+    unittests/webserver/RPCServerHandlerTest.cpp)
   include(CMake/deps/gtest.cmake)
 
   # fix for dwarf5 bug on ci 
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -31,7 +31,7 @@
 #include <rpc/Counters.h>
 #include <rpc/RPCEngine.h>
 #include <rpc/common/impl/HandlerProvider.h>
-#include <webserver/RPCExecutor.h>
+#include <webserver/RPCServerHandler.h>
 #include <webserver/Server.h>
 
 #include <boost/date_time/posix_time/posix_time_types.hpp>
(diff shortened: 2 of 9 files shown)


godexsoft:
@cindyyan317 when you have time, take a look please. This is definitely not the design i had in mind but it would take way too much effort to do what i thought so I'd rather leave a mental note for future refactorings and do it the simpler way for now.

cindyyan317:
**1** curl -v -d '{' 127.0.0.1:xxx. return code is different
**2** clio returns 400, rippled return 200: 
{
    "method":"server_info",
    "params":[]
}
**3** busy, clio retun 200, rippled returns 503 https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/ServerHandlerImp.cpp#L305
**4**
{
    123
}
clio returns 200, rippled returns 400
**5**
123
clio has internal eror, rippled returns 400
**6**
subscribe on http:
clio returns 400, rippled returns 200. (in this case, clio maybe right)

godexsoft:
@cindyyan317  

1, 3, 4 and 5) fixed in the latest commit.
2) is  expected - we are a little more strict than rippled and do what the doc says in this case.
6) clio is correct as you suggested so leaving it as is.

cindyyan317 (approved):
Looks good to me

PR #723, opened 2023-07-03
Fix max limit for account tx

cindyyan317:
Refer to rippled's implementation, 
Screenshot 2023-07-03 at 09 25 12
Adjust the max and default value .
@ckniffen

### DIFF
--- a/src/rpc/handlers/AccountTx.h
+++ b/src/rpc/handlers/AccountTx.h
@@ -91,7 +91,7 @@ public:
             {JS(ledger_index_max), validation::Type<int32_t>{}},
             {JS(binary), validation::Type<bool>{}},
             {JS(forward), validation::Type<bool>{}},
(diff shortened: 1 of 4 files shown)


src/rpc/handlers/AccountTx.h
```diff
@@ -91,7 +91,7 @@ class AccountTxHandler
             {JS(ledger_index_max), validation::Type<int32_t>{}},
             {JS(binary), validation::Type<bool>{}},
             {JS(forward), validation::Type<bool>{}},
-            {JS(limit), validation::Type<uint32_t>{}, validation::Between{1, 100}},
+            {JS(limit), validation::Type<uint32_t>{}, validation::Between{1, INT32_MAX}},
```
godexsoft: Let's use `std::numeric_limits::max<int32_t>` instead 👍

godexsoft (changes requested):
Just one small thing and it's good to go 🥇

godexsoft (approved):

PR #731, opened 2023-07-04
Add strict field support

### LINKED ISSUE #729: Raise "unsupported" error when 'strict' is false
Implement this logic in
- account_currencies
- account_info
- account_offers
- account_tx
- gateway_balances
- tx
- nft_history

godexsoft:
Fixes #729

### DIFF
--- a/src/rpc/handlers/AccountTx.h
+++ b/src/rpc/handlers/AccountTx.h
@@ -57,7 +57,6 @@ public:
         bool validated = true;
     };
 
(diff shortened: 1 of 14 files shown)


src/rpc/handlers/AccountTx.h
```diff
@@ -57,7 +57,7 @@ class AccountTxHandler
         bool validated = true;
     };
 
-    // TODO:we did not implement the "strict" field
```
cindyyan317, line 60: I must copied the comment from other place, account_tx's document does not mention "strict".

cindyyan317 (approved):
Perfect!

PR #732, opened 2023-07-04
Remove date from "account_tx" when "binary" is true

cindyyan317:
Remove the "date" field which was particularly added in old "clio"


godexsoft (approved):
🚀

PR #734, opened 2023-07-04
Use "invalidParam" when "book_offers" taker format is wrong

cindyyan317:
Clio uses the "AccountValidator" for all account content field. This field needs to be taken care of. According to document and rippled, we should "invalidParam" error.


godexsoft (approved):
LGTM. I wonder why coverage went down tho..

PR #736, opened 2023-07-04
account_object supports nft page

cindyyan317:


src/rpc/RPCHelpers.cpp
```diff
@@ -405,61 +407,111 @@ getStartHint(ripple::SLE const& sle, ripple::AccountID const& accountID)
     return sle.getFieldU64(ripple::sfOwnerNode);
 }
 
+// traverse account's nfts
+// return Status if error occurs
+// return [nextpage, count of nft already found] if success
 std::variant<Status, AccountCursor>
-traverseOwnedNodes(
```
cindyyan317, line 409: We don't call this function except in unittests.

src/rpc/RPCHelpers.cpp
```diff
@@ -470,7 +522,6 @@ traverseOwnedNodes(
     std::uint32_t const startHint,
     std::uint32_t sequence,
     std::uint32_t limit,
-    std::optional<std::string> jsonCursor,
```
cindyyan317, line 473: not used

src/rpc/RPCHelpers.cpp
godexsoft: If it's meant to only be used in unit tests and if we still need it.. I suggest we move it to some helper file in unittests folder instead.

src/rpc/RPCHelpers.cpp
```diff
@@ -405,61 +407,111 @@ getStartHint(ripple::SLE const& sle, ripple::AccountID const& accountID)
...
+        else  // marker is in the right range, but still invalid
+            return Status{RippledError::rpcINVALID_PARAMS, "invalid Marker"};
+    }
+
+    // the object exists and the key is in right range, must be nft page
+    ripple::SLE pageSLE{ripple::SLE{ripple::SerialIter{page->data(), page->size()}, currentPage}};
+
+    auto count = 0;
+    // traverse the nft page linked list until the start of the list or reach the limit
+    while (true)
```
godexsoft, line 449: while(true) is a bit scary :) 
Let's rewrite this as a simple for loop:
```
for(auto count = 0u; count < limit; ++count) { ... break early if previous page is zero ... }
```

src/rpc/RPCHelpers.cpp
```diff
@@ -405,61 +407,111 @@ getStartHint(ripple::SLE const& sle, ripple::AccountID const& accountID)
...
     std::optional<std::string> jsonCursor,
     boost::asio::yield_context& yield,
-    std::function<void(ripple::SLE&&)> atOwnedNode)
+    std::function<void(ripple::SLE&&)> atOwnedNode,
+    bool nftIncluded)
 {
     auto const maybeCursor = parseAccountCursor(jsonCursor);
+
+    if (!maybeCursor)
+        return Status{RippledError::rpcINVALID_PARAMS, "Malformed cursor"};
```
godexsoft: I think other errors usually add a dot at the end of the error message. If that's the case we should probably try to keep it uniform. Unless of course rippled does it without the dot for this particular case.

src/rpc/RPCHelpers.cpp
```diff
@@ -405,61 +407,111 @@ getStartHint(ripple::SLE const& sle, ripple::AccountID const& accountID)
...
     // the format is checked in RPC framework level
-    auto const [hexCursor, startHint] = *maybeCursor;
+    auto [hexCursor, startHint] = *maybeCursor;
+
+    // if we need to traverse nft objects and this is the first request -> traverse nft objects
+    // if we need to traverse nft objects and the marker is still in nft page -> traverse nft objects
+    // if we need to traverse nft objects and the marker is still in nft page but next page is zero -> owned nodes
+    // if we need to traverse nft objects and the marker is not in nft page -> traverse owned nodes
+    if (nftIncluded and
+        (!jsonCursor or (startHint == std::numeric_limits<uint32_t>::max() and hexCursor != beast::zero)))
```
godexsoft: For readability sake it may be better to create a variable or a few of them for the condition(s).

godexsoft (changes requested):
Very cool! leaving just a few small suggestions for readability improvements. Everything else looks good to me 🚀

src/rpc/RPCHelpers.cpp
cindyyan317: In unittest, we want to test it. since we don't need it, i port the test to new function

src/rpc/RPCHelpers.cpp
cindyyan317: You are right. Shall be consistent. I will add dot

src/rpc/RPCHelpers.cpp
cindyyan317:
```
    auto count = 1;
    ripple::uint256 nftPreviousPage = beast::zero;
    // traverse the nft page linked list until the start of the list or reach the limit
    for (; count <= limit; count++)
    {
        nftPreviousPage = pageSLE.getFieldH256(ripple::sfPreviousPageMin);
        atOwnedNode(std::move(pageSLE));

        if (count == limit or nftPreviousPage == beast::zero)
            break;

        page = backend.fetchLedgerObject(nftPreviousPage, sequence, yield);
        pageSLE = ripple::SLE{ripple::SerialIter{page->data(), page->size()}, nftPreviousPage};
    }

    return AccountCursor{nftPreviousPage, count};
```

Not much better. The count check is still inside the for loop. I will keep the while loop, but simplify the inner logic bit.

src/rpc/RPCHelpers.cpp
```diff
@@ -505,7 +558,7 @@ traverseOwnedNodes(
             std::find(std::begin(indexes), std::end(indexes), hexMarker) == std::end(indexes))
         {
             // the index specified by marker is not in the page specified by marker
-            return Status(ripple::rpcINVALID_PARAMS, "Invalid marker");
+            return Status(ripple::rpcINVALID_PARAMS, "Invalid marker.");
```
godexsoft, line 561: "Invalid marker." vs. "Invalid Marker." above. Choose one please :)

unittests/rpc/RPCHelpersTest.cpp
```diff
@@ -348,7 +297,7 @@ TEST_F(RPCHelpersTest, TraverseOwnedNodesWithUnexistingIndexMarker)
         auto status = std::get_if<Status>(&ret);
         EXPECT_TRUE(status != nullptr);
         EXPECT_EQ(*status, ripple::rpcINVALID_PARAMS);
-        EXPECT_EQ(status->message, "Invalid marker");
+        EXPECT_EQ(status->message, "Invalid marker.");
```
godexsoft, line 300: same

unittests/rpc/handlers/AccountOffersTest.cpp
```diff
@@ -589,6 +589,6 @@ TEST_F(RPCAccountOffersHandlerTest, MarkerNotExists)
         ASSERT_FALSE(output);
         auto const err = RPC::makeError(output.error());
         EXPECT_EQ(err.at("error").as_string(), "invalidParams");
-        EXPECT_EQ(err.at("error_message").as_string(), "Invalid marker");
+        EXPECT_EQ(err.at("error_message").as_string(), "Invalid marker.");
```
godexsoft, line 592: same

unittests/rpc/handlers/AccountObjectsTest.cpp
```diff
@@ -123,6 +126,13 @@ generateTestValuesForParametersTest()
             R"({"account":"rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun", "marker":"xxxx"})",
             "invalidParams",
             "Malformed cursor"},
+        AccountObjectsParamTestCaseBundle{
+            "NFTMarkerInvalid",
+            fmt::format(
+                R"({{"account":"rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun", "marker":"wronghex256,{}"}})",
+                std::numeric_limits<uint32_t>::max()),
+            "invalidParams",
+            "Malformed cursor"},
```
godexsoft: Looks like this should also have a dot at the end?

godexsoft (changes requested):
Found a small inconsistency. Other than that it looks good. Let's finish this 👍

godexsoft (approved):

PR #737, opened 2023-07-05
Add deletion_blockers_only support

### LINKED ISSUE #730: Support "account_objects" field "deletion_blockers_only"
https://xrpl.org/account_objects.html
Need to filter out the objects which will affect the account deletion.

godexsoft:
Fixes #730

### DIFF
--- a/src/rpc/handlers/AccountObjects.cpp
+++ b/src/rpc/handlers/AccountObjects.cpp
@@ -53,7 +53,8 @@ AccountObjectsHandler::process(AccountObjectsHandler::Input input, Context const
     if (!accountLedgerObject)
         return Error{Status{RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
 
-    std::vector<ripple::LedgerEntryType> typeFilter;
+    auto typeFilter = std::optional<std::vector<ripple::LedgerEntryType>>{};
+
     if (input.deletionBlockersOnly)
     {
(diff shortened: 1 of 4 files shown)


src/rpc/handlers/AccountObjects.cpp
```diff
@@ -53,10 +53,40 @@ AccountObjectsHandler::process(AccountObjectsHandler::Input input, Context const
...
+            ripple::ltCHECK,
+            ripple::ltESCROW,
+            ripple::ltNFTOKEN_PAGE,
+            ripple::ltPAYCHAN,
+            ripple::ltRIPPLE_STATE,
+        };
+
+        typeFilter.reserve(std::size(deletionBlockers));
+
+        for (auto type : deletionBlockers)
```
cindyyan317, line 71: If input.type is "offer" which does not block deletion and deletionBlockersOnly is true -> typeFilter will be empty, right? it ends up including all the objects, more than "offer" and against deletionBlockersOnly too. in such case, i think the result should be empty.

cindyyan317 (dismissed):

src/rpc/handlers/AccountObjects.cpp
godexsoft: You are right. I fix it now :)

cindyyan317 (approved):
LGTM

PR #740, opened 2023-07-06
Implement a Clamp modifier/validator

godexsoft:
This implements support for _modifiers_ such as `Clamp`.
See `ClampingValidator` test for usage examples.

If we decide to go this path we probably will want to rename `Validator` and `validate` to something that implies _possible_ modification to the original json value.

Most of the validators can continue receiving a `const&` of the json value. Only _modifiers_ and _containers_ for other validators/modifiers need to remove the const.

I'm certainly not a huge fan of this "design" but i don't see another way to achieve clamping if we need it.
Sometime in the future perhaps we could integrate the parsing steps into this framework instead of relying on user-provided `value_from` and `value_to` but for now this approach is least blood for modifiers support.

Let me know your thoughts @cindyyan317 and @ximinez.

### DIFF
--- a/src/rpc/common/MetaProcessors.cpp
+++ b/src/rpc/common/MetaProcessors.cpp
@@ -1,0 +1,73 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/basics/base_uint.h>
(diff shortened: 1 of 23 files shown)


cindyyan317:
Thanks for quickly fixing this!
The mutation in validation looks okay for me. You change the name to "Clamp" which is clear enough. For some validation like account , the validation and parsing is the same operation, currently we do it on accounts twice. In future, with proper parsing step, we can even convert to the different type.

godexsoft:
Handlers and other parts of clio can remain using const object as request now that the default processor makes a copy of the request and validation framework operates on the mutable copy instead.
Modifiers, validators and meta handlers were all refactored out to separate namespaces and files.

src/rpc/common/MetaProcessors.cpp
```diff
@@ -0,0 +1,73 @@
...
+*/
+//==============================================================================
+
+#include <ripple/basics/base_uint.h>
+#include <rpc/RPCHelpers.h>
+#include <rpc/common/MetaProcessors.h>
+
+#include <boost/json/value.hpp>
+
+#include <charconv>
```
cindyyan317: don't need this header

src/rpc/common/MetaProcessors.cpp
```diff
@@ -0,0 +1,73 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/basics/base_uint.h>
```
cindyyan317: don't need this header

src/rpc/common/impl/Factories.h
```diff
@@ -28,23 +28,33 @@
...
 namespace RPC::detail {
 
-template <Requirement... Requirements>
+template <typename>
+static constexpr bool unsupported_v = false;
+
+template <Processor... Processors>
 [[nodiscard]] auto
-makeFieldValidator(std::string const& key, Requirements&&... requirements)
+makeFieldValidator(std::string const& key, Processors&&... procs)
```
cindyyan317: this function can also be renamed. :)

cindyyan317 (dismissed):
LGTM 👍

cindyyan317 (approved):

PR #752, opened 2023-07-10
Use clamp modifier on limit field instead of between validator

godexsoft:
This implements rippled-like clamping behaviour for all `limit` fields other than the special one in `ledger_data`.


cindyyan317 (approved):
GO!

PR #755, opened 2023-07-11
Remove "strict"

cindyyan317:
Since the "strict" got removed from document, Clio also remove all "strict" fields.


godexsoft (approved):
Go go 🚀

PR #757, opened 2023-07-11
Forward not supported fields

cindyyan317:


src/rpc/common/impl/ForwardingProxy.h
```diff
@@ -71,8 +71,12 @@ class ForwardingProxy
         if (specifiesCurrentOrClosedLedger(request))
             return true;
```
godexsoft, line 73: This is quickly becoming too messy to read.. i suggest we split each individual block between `or`s to a simple variable. Or if you want to avoid performing extra checks you can wrap each of them into a lambda and invoke them in this `if` statement separated by `or`.

Could look something like 
```cpp
auto checkQueue = [&]{ return request.contains("queue") and request.at("queue").is_bool() and request.at("queue").as_bool(); };
// ... same for full and accounts

if (... or checkQueue() or ...) { ... }
```

godexsoft (changes requested):
Let's make the condition more readable. Otherwise all good 👍

godexsoft (approved):

PR #758, opened 2023-07-11
Missing "tx_hash" for transaction_entry

cindyyan317:
According to document, we shall return "fieldNotFoundTransaction" when tx_hash is missing
https://xrpl.org/transaction_entry.html#possible-errors


src/rpc/Errors.cpp
```diff
@@ -85,6 +85,7 @@ getErrorInfo(ClioError code)
         {ClioError::rpcCOMMAND_NOT_STRING, "commandNotString", "Method is not a string."},
         {ClioError::rpcCOMMAND_IS_EMPTY, "emptyCommand", "Method is an empty string."},
         {ClioError::rpcPARAMS_UNPARSEABLE, "paramsUnparseable", "Params must be an array holding exactly one object."},
+        {ClioError::rpcFIELD_NOT_FOUND_TRANSACTION, "fieldNotFoundTransaction", "Missing field."},
```
godexsoft: Let's put this after `rpcUNKNOWN_OPTION` to maintain original ordering..

godexsoft (changes requested):
Just one small thing to improve 👍

godexsoft (approved):

PR #761, opened 2023-07-12
Add "network_id" to server_info

cindyyan317:
[EASY]
Clio does not have this field. This PR forward it to server_info's response.
  
main    -> 0   testnet -> 1. devnet  -> 2


godexsoft (approved):
This is good for what needed to be done. Later we will have to cache this value and check every reconnect that the version is compatible and the same for all ETL sources 👍 We can make a separate issue for that tho.

PR #770, opened 2023-07-14
Rename WsSession to WsBase

cindyyan317:
Rename WsSession to WsBase to match the filename.
WsBase is more suitable , consider its counterpart is called HttpBase


godexsoft (approved):
Fine, let's rename 👍

PR #782, opened 2023-07-25
added new RPC Handler "version"

PeterChen13579:
Added a new RPC Handler named "version". This will show the version api formatted exactly how rippled does it.

### DIFF
--- a/src/rpc/common/impl/APIVersionParser.h
+++ b/src/rpc/common/impl/APIVersionParser.h
@@ -27,10 +27,6 @@
 #include <algorithm>
 #include <string>
 
-namespace RPC {
-class VersionHandler;
-};
-
 namespace RPC::detail {
 
 class ProductionAPIVersionParser : public APIVersionParser
@@ -79,7 +75,23 @@ public:
     util::Expected<uint32_t, std::string>
     parse(boost::json::object const& request) const override;
 
-    friend class RPC::VersionHandler;
+    inline int
+    getDefaultVersion() const
+    {
+        return defaultVersion_;
+    }
+
+    inline int
+    getMinVersion() const
+    {
+        return minVersion_;
+    }
+
+    inline int
+    getMaxVersion() const
+    {
+        return maxVersion_;
+    }
 };
 
 }  // namespace RPC::detail
--- a/src/rpc/common/impl/APIVersionParser.h
+++ b/src/rpc/common/impl/APIVersionParser.h
@@ -75,19 +75,19 @@ public:
     util::Expected<uint32_t, std::string>
     parse(boost::json::object const& request) const override;
 
-    inline int
+    inline uint32_t
     getDefaultVersion() const
     {
         return defaultVersion_;
     }
 
-    inline int
+    inline uint32_t
     getMinVersion() const
     {
         return minVersion_;
     }
 
-    inline int
+    inline uint32_t
     getMaxVersion() const
     {
         return maxVersion_;
--- a/src/rpc/common/impl/APIVersionParser.h
+++ b/src/rpc/common/impl/APIVersionParser.h
@@ -27,6 +27,10 @@
 #include <algorithm>
 #include <string>
 
+namespace RPC {
+class VersionHandler;
+};
+
 namespace RPC::detail {
 
 class ProductionAPIVersionParser : public APIVersionParser
@@ -74,6 +78,8 @@ public:
 
     util::Expected<uint32_t, std::string>
     parse(boost::json::object const& request) const override;
+
+    friend class RPC::VersionHandler;
 };
 
 }  // namespace RPC::detail
--- a/src/rpc/handlers/VersionHandler.h
+++ b/src/rpc/handlers/VersionHandler.h
@@ -75,7 +75,8 @@ private:
                  {"first", output.minVersion},
                  {"last", output.maxVersion},
                  {"good", output.currVersion},
-             }}};
+             }},
+        };
     }
 };
 
--- a/src/rpc/handlers/VersionHandler.h
+++ b/src/rpc/handlers/VersionHandler.h
@@ -1,0 +1,94 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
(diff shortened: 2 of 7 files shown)


src/rpc/handlers/VersionHandler.h
```diff
@@ -0,0 +1,94 @@
...
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <cassert>
```
godexsoft: The groups of headers are 
1) Local headers (from clio codebase)
2) 3rd party libraries (boost, xrpl, fmt, etc)
3) standard libraries (including `cassert`)

So it should be 
```cpp
#include <rpc/Errors.h>
#include <rpc/RPCHelpers.h>
#include <rpc/common/APIVersion.h>
#include <rpc/common/Types.h>
#include <rpc/common/impl/APIVersionParser.h>

#include <cassert>
```

src/rpc/handlers/VersionHandler.h
```diff
@@ -0,0 +1,94 @@
...
+private:
+    friend void
+    tag_invoke(boost::json::value_from_tag, boost::json::value& jv, Output const& output)
+    {
+        using boost::json::value_from;
+
+        jv = {{"version", value_from(output.info)}};
+    }
+
+    friend void
```
godexsoft: Do we really need a separate struct here? Consider moving the 3 fields to `Output` and adjust the `tag_invoke` above to output it as a sub object under the "version" key.

unittests/rpc/handlers/VersionHandlerTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+            "max": {},
+            "default": {}
+        }})",
+        MIN_API_VERSION,
+        MAX_API_VERSION,
+        DEFAULT_API_VERSION))};
+
+    runSpawn([&](auto& yield) {
+        auto const handler = AnyHandler{VersionHandler{cfg}};
+        RPC::ReturnType const output = handler.process(static_cast<json::value>(cfg), Context{std::ref(yield)});
```
godexsoft: should be able to use `auto const output` here, no need to type out `RPC::ReturnType` 👍

unittests/rpc/handlers/VersionHandlerTest.cpp
```diff
@@ -0,0 +1,64 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <rpc/common/AnyHandler.h>
```
godexsoft: Grouping in tests should be similar to main codebase except for test-local headers are even higher in the list 👍

src/rpc/handlers/VersionHandler.h
```diff
@@ -0,0 +1,94 @@
...
+
+    using Result = HandlerReturnType<Output>;
+
+    Result
+    process([[maybe_unused]] Context const& ctx) const
+    {
+        using namespace RPC;
+
+        auto output = Output{};
+        output.info.currVersion = apiVersionParser_.defaultVersion_;
```
godexsoft: I'd not `friend` an unrelated class but rather create public `const` getters for the fields you want to access.

src/rpc/common/impl/APIVersionParser.h
```diff
@@ -74,6 +78,8 @@ class ProductionAPIVersionParser : public APIVersionParser
 
     util::Expected<uint32_t, std::string>
     parse(boost::json::object const& request) const override;
+
+    friend class RPC::VersionHandler;
```
godexsoft: This is probably not the best approach here - `friend` is best used when two or more classes only make sense together as one system. In our case that's not the case so i'd recommend adding getters instead 🥇

src/rpc/common/impl/APIVersionParser.h
```diff
@@ -27,6 +27,10 @@
 #include <algorithm>
 #include <string>
 
+namespace RPC {
+class VersionHandler;
```
godexsoft: This should not be needed here if you drop the `friend` idea 👍

godexsoft (changes requested):
Looks good but needs a few minor adjustments 🥇 Leaving some suggestions.

src/rpc/common/impl/APIVersionParser.h
```diff
@@ -74,6 +74,24 @@ class ProductionAPIVersionParser : public APIVersionParser
 
     util::Expected<uint32_t, std::string>
     parse(boost::json::object const& request) const override;
+
+    inline int
```
godexsoft: The type of these is not `int`, please fix to `uint32_t`.

src/rpc/handlers/VersionHandler.h
```diff
@@ -0,0 +1,82 @@
...
+        output.minVersion = apiVersionParser_.getMinVersion();
+        output.maxVersion = apiVersionParser_.getMaxVersion();
+        return output;
+    }
+
+private:
+    friend void
+    tag_invoke(boost::json::value_from_tag, boost::json::value& jv, Output const& output)
+    {
+        jv = {
```
godexsoft, line 72: Formatting is strange here, maybe let's add a `,`  before the last closing curly, it will probably look more readable 👍 Alternatively you can extract the inner object into a local variable and then move it into the resulting json object.

godexsoft (changes requested):
A few new small issues. nearly there 👍

godexsoft (approved):
LGTM 👍

PR #788, opened 2023-07-26
Fix a few tsan issues

godexsoft:
This is part 1 of probably many fixes that we will have to iterate over in order to get rid of all the TSAN warnings (and some SEGV/BUS sometimes) that we are now seeing after upgrading to boost 1.82.


arihantkothari (approved):
lgtm 👍

PR #790, opened 2023-07-27
Fix connect_timeout request_timeout not work + tsan in RPCServerTestSuite

cindyyan317:
1 Fix :WebRPCServerHandlerTest tsan thread problem:
   The issue is the other thread writes the lastStatus while main thread reads it to compare.
2 Fix :Timeout parameter not read

### DIFF
--- a/src/backend/cassandra/SettingsProvider.cpp
+++ b/src/backend/cassandra/SettingsProvider.cpp
@@ -123,6 +123,7 @@ SettingsProvider::parseSettings() const
     settings.maxConcurrentRequestsThreshold = config_.valueOr<uint32_t>(
         "max_concurrent_requests_threshold",
         (settings.maxReadRequestsOutstanding + settings.maxWriteRequestsOutstanding) / settings.coreConnectionsPerHost);
+
     auto const timeoutSecond = config_.maybeValue<uint32_t>("connect_timeout");
     if (timeoutSecond)
         settings.connectionTimeout = std::chrono::milliseconds{*timeoutSecond * 1000};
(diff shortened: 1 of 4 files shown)


src/backend/cassandra/SettingsProvider.cpp
```diff
@@ -123,6 +123,9 @@ SettingsProvider::parseSettings() const
     settings.maxConcurrentRequestsThreshold = config_.valueOr<uint32_t>(
         "max_concurrent_requests_threshold",
         (settings.maxReadRequestsOutstanding + settings.maxWriteRequestsOutstanding) / settings.coreConnectionsPerHost);
+    auto const timeoutSecond = config_.maybeValue<uint32_t>("connect_timeout");
```
godexsoft: Let's add the same thing for `request_timeout` while we are at it 🙇 
Also i'd separate the 4 timeout parsing lines by space line on 126, similar to how we have a line on 129 👍

unittests/webserver/RPCServerHandlerTest.cpp
```diff
@@ -399,8 +399,9 @@ TEST_F(WebRPCServerHandlerTest, HTTPInvalidAPIVersion)
 
     (*handler)(std::move(request), session);
     std::this_thread::sleep_for(200ms);
+    rpcEngine->post(
```
godexsoft: This feels and looks like a hack 🤣
Let's make MockAsyncRPCEngine::post synchronous on the caller thread:
```cpp
template <typename Fn>
bool
post(Fn&& func, [[maybe_unused]] std::string const& ip = "")
{
    using namespace boost::asio;
    io_context ioc;

    spawn(ioc, [handler = std::forward<Fn>(func), _ = make_work_guard(ioc.get_executor())](auto yield) mutable {
        handler(yield);
    });

    ioc.run();
    return true;
}
```

and perform the EXPECT_EQ on main thread as before + remove the 200ms sleeping which would no longer be needed.

godexsoft (changes requested):
Leaving a few ideas on how to improve this 👍

godexsoft (approved):
LGTM 👍

PR #792, opened 2023-07-27
Fix backend factory test and remove cout from base tests

godexsoft:
Finally remove couts as well as fix the error in Factory test.
The issue was that by default we have 2 core connections and max is also set to 2. We were trying to create 3 however.


cindyyan317 (approved):
💯

PR #795, opened 2023-08-01
Fix linux/gcc compilation and many TSAN fixes

godexsoft:
This PR 
- Fixes a bunch of places that were triggering TSAN warnings before
- Implements WorkQueue tests
- Fixes Linux/gcc compilation (tested on gcc11)
- Uses yield_context as a value throughout the codebase
- Many other small improvements


CMakeLists.txt
```diff
@@ -214,11 +209,17 @@ if(tests)
...
+  # Tmp: uncomment for TSAN
+  # target_compile_options(${TEST_TARGET} PRIVATE -fsanitize=thread)
+  # target_link_options(${TEST_TARGET} PRIVATE -fsanitize=thread)
+
   target_compile_definitions(${TEST_TARGET} PUBLIC UNITTEST_BUILD)
   target_include_directories(${TEST_TARGET} PRIVATE unittests)
   target_link_libraries(${TEST_TARGET} PUBLIC clio gtest::gtest)
 
   # Generate `clio_test-ccov` if coverage is enabled
+  # Note: use `make clio_test-ccov` to generate report
```
cindyyan317, line 222: 👍

README.md
```diff
@@ -57,9 +62,11 @@ cmake --build . --parallel 8 # or without the number if you feel extra adventuro
 If all goes well, `conan install` will find required packages and `cmake` will do the rest. you should end up with `clio_server` and `clio_unittests` in the `build` directory (the current directory).
 Please note that a few unittests are currently failing. See below.
 
+> **Tip:** You can omit the `-o tests=True` in `conan install` command above if you don't want to build `clio_unittests`.
```
cindyyan317: wrong target name : clio_tests

src/backend/BackendInterface.h
```diff
@@ -94,24 +94,22 @@ auto
...
+    using R = typename boost::result_of<F(boost::asio::yield_context)>::type;
     if constexpr (!std::is_same<R, void>::value)
     {
         R res;
-        boost::asio::spawn(
-            strand, [&f, &res, _ = boost::asio::make_work_guard(strand)](boost::asio::yield_context yield) {
-                res = f(yield);
-                ;
-            });
+        boost::asio::spawn(ctx, [&f, &res, _ = boost::asio::make_work_guard(ctx)](boost::asio::yield_context yield) {
```
cindyyan317: not sure i understand the code . why we need the extra parameter , but we don't use it inside?
Please remove the extra ';' on the spot.

src/subscriptions/SubscriptionManager.h
```diff
@@ -41,7 +41,7 @@ class Subscription
     Subscription(Subscription&) = delete;
     Subscription(Subscription&&) = delete;
 
-    explicit Subscription(boost::asio::io_context& ioc) : strand_(ioc.get_executor())
+    explicit Subscription(boost::asio::io_context& ioc) : strand_(boost::asio::make_strand(ioc))
```
cindyyan317, line 44: we need to call extra copy constructor after change?

src/webserver/Server.h
```diff
@@ -130,11 +125,11 @@ class Server : public std::enable_shared_from_this<Server<PlainSession, SslSessi
     using std::enable_shared_from_this<Server<PlainSession, SslSession, Handler>>::shared_from_this;
 
     clio::Logger log_{"WebServer"};
-    std::reference_wrapper<boost::asio::io_context> const ioc_;
-    std::optional<std::reference_wrapper<boost::asio::ssl::context>> const ctx_;
-    util::TagDecoratorFactory const tagFactory_;
-    std::reference_wrapper<clio::DOSGuard> const dosGuard_;
-    std::shared_ptr<Handler> const handler_;
+    std::reference_wrapper<boost::asio::io_context> ioc_;
```
cindyyan317, line 128: why remove these const?

cindyyan317 (dismissed):
LOOKS GOOD~ Please check my questions regarding the strand.

src/backend/BackendInterface.h
godexsoft: i think in this particular case we may not actually need this work guard - it was added for a test and i must have forgotten about it. thanks for noticing 👍 

as for the extra `;` - it's there to make clang-format format the body onto its own line instead of squashing everything into one unreadable line.. if i remove the work guard then maybe we can remove `;` too.

src/subscriptions/SubscriptionManager.h
godexsoft: thanks to guaranteed copy-elision this is ok. it's just a nicer way to say the same thing without explicitly using get_executor.

src/webserver/Server.h
godexsoft: const and reference members are generally a code smell. in this case it may not matter for copy-assignability but i still prefer to follow core guidelines whenever we can. one day we may integrate clang-tidy and then this would come up as a failed check anyways: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

cindyyan317 (approved):

PR #796, opened 2023-08-01
Support whitelisting for IPV4/IPV6 with CIDR

PeterChen13579:


src/webserver/DOSGuard.h
```diff
@@ -19,15 +19,25 @@
...
+#include <memory>
+#include <optional>
+#include <random>
+#include <regex>
 #include <string>
 #include <unordered_map>
 #include <unordered_set>
 
-#include <config/Config.h>
+using namespace boost::asio;
```
godexsoft: `using namespace` lines should not be in header files. 
This is because if some developer of another source file includes `DOSGuard.h` they may not expect that everything from `boost::asio` is now at the same level as their own code - it's possible to get name clashing and what not. You can do this in `.cpp` files or at the top your functions in this header if typing the extra `boost::asio::` is annoying 👍

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
+        return false;
+    }
+
+private:
+    bool
+    isV4(std::string_view net) const
+    {
+        static const std::regex ipv4CidrRegex(R"(^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/\d{1,2}$)");
+        return std::regex_match(std::string(net), ipv4CidrRegex);
+    }
```
godexsoft: Let's add a line of space between functions, here and below 🥇

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
     clear() noexcept = 0;
 };
 
+class Whitelist
```
godexsoft: This could maybe live in a detail namespace somewhere in `src/webserver/impl`.

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
+    bool
+    isV6(std::string_view net) const
+    {
+        static const std::regex ipv6CidrRegex(R"(^([0-9A-Fa-f]{1,4}:){7}[0-9A-Fa-f]{1,4}/\d{1,3}$)");
+        return std::regex_match(std::string(net), ipv6CidrRegex);
+    }
+    bool
+    isMask(std::string_view net) const
+    {
+        return net.find('/') != std::string_view::npos;
```
godexsoft: Not using regex because mask can be either v4 or v6, right?

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
+    }
+};
+
+class WhitelistHandler
```
godexsoft: This can probably also be a `detail`.
It's also nice to write some doxygen-style comments like `BasicDOSGuard` has.

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
+
+    static bool
+    isInV6Subnet(ip::address const& addr, ip::network_v6 const& subnet)
+    {
+        auto const range = subnet.hosts();
+        return range.find(addr.to_v6()) != range.end();
+    }
+
+public:
+    // could throw, handle with care
```
godexsoft: This comment on throwing could be part of doxygen instead:

```
/**
 * @brief ...
 * 
 * @param net ...
 * @throws std::runtime_error ...
 */
```

godexsoft: Also in other places where the same comment was left 👍

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
     clear() noexcept = 0;
 };
 
+class Whitelist
+{
+    std::vector<ip::network_v4> subnetsV4_;
+    std::vector<ip::network_v6> subnetsV6_;
+    std::vector<ip::address> ips_;
+
+    static bool
```
godexsoft: I'd move the private functions into a `private:` block at the bottom instead 👍

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
+    isWhiteListed(std::string_view ip) const
+    {
+        return whitelist_.isWhiteListed(ip);
+    }
+};
+
 /**
  * @brief A simple denial of service guard used for rate limiting.
  *
  * @tparam SweepHandler Type of the sweep handler
```
godexsoft: You probably want to update the docs here with the new tparam and other things.

src/webserver/DOSGuard.h
```diff
@@ -40,12 +50,110 @@ class BaseDOSGuard
...
  */
-template <typename SweepHandler>
+template <typename WhitelistHandlerType, typename SweepHandler>
```
godexsoft: Feel free to also change `SweepHandler` to `SweepHandlerType` to match the other one.

unittests/DOSGuard.cpp
```diff
@@ -59,7 +64,7 @@ constexpr static auto IP = "127.0.0.2";
 class FakeSweepHandler
 {
 private:
-    using guard_type = BasicDOSGuard<FakeSweepHandler>;
+    using guard_type = BasicDOSGuard<WhitelistHandler, FakeSweepHandler>;
```
godexsoft: This test is for BasicDOSGuard not for the whitelist. Let's create a `Fake` or `Mock` for the `WhitelistHandler` and pass it in here. You can then just check that if your `isWhiteListed()` function of DOSguard returns the same thing as the mock/fake and that's enough here.

For the WhitelistHandler you would have a separate file with a separate test suite 🥇

unittests/DOSGuard.cpp
```diff
@@ -91,6 +96,17 @@ TEST_F(DOSGuardTest, Whitelisting)
     EXPECT_FALSE(guard.isWhiteListed(IP));
 }
 
+TEST_F(DOSGuardTest, CIDRWhitelistTest)
```
godexsoft: This is not a bad test but **ideally** it belongs in its own test suite in a separate file. Or at least at the bottom of the file like `IntervalSweepHandler` is done atm. 
If you are not afraid to get your hands dirty then feel free to create a separate file for the `IntervalSweepHandler` as well 🥇

src/webserver/DOSGuard.h
```diff
@@ -19,15 +19,25 @@
 
 #pragma once
 
+#include <config/Config.h>
+
 #include <boost/asio.hpp>
 #include <boost/iterator/transform_iterator.hpp>
+#include <boost/system/error_code.hpp>
 
+#include <algorithm>
```
godexsoft, line 29: I doubt we need so many includes here 🔢

godexsoft (changes requested):
Leaving some suggestions how you can improve this 🥇 It's pretty good already!

src/webserver/DOSGuard.h
```diff
@@ -92,9 +98,9 @@ class BasicDOSGuard : public BaseDOSGuard
      * @return false
      */
     [[nodiscard]] bool
-    isWhiteListed(std::string const& ip) const noexcept
+    isWhiteListed(std::string_view const& ip) const noexcept
```
godexsoft: `string_view` is best used by value (no need for `const&`).

src/webserver/DOSGuard.h
```diff
@@ -107,7 +113,7 @@ class BasicDOSGuard : public BaseDOSGuard
     [[nodiscard]] bool
     isOk(std::string const& ip) const noexcept
```
godexsoft, line 114: Just wondering, why change `isWhiteListed` to `string_view` but not others? consider others too.

src/webserver/impl/WhitelistHandler.h
```diff
@@ -0,0 +1,166 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
godexsoft: 2023 now 🔢

src/webserver/impl/WhitelistHandler.h
```diff
@@ -0,0 +1,166 @@
...
+
+#include <regex>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+
+namespace clio {
+
+/**
+ * @brief A whitelist to check for
```
godexsoft: Looks like this could slightly be improved by providing more info about the whitelist, for example what it does and why it's there.

src/webserver/impl/WhitelistHandler.h
```diff
@@ -0,0 +1,166 @@
...
+ */
+class Whitelist
+{
+    std::vector<boost::asio::ip::network_v4> subnetsV4_;
+    std::vector<boost::asio::ip::network_v6> subnetsV6_;
+    std::vector<boost::asio::ip::address> ips_;
+
+public:
+    /**
+     * @brief add network address to whitelist
```
godexsoft: When you write these comments, unlike regular comments, they become documentation via `Doxygen`. It's better to keep the capitalisation of words and sentences the same way any doc/tutorial/book would.

src/webserver/impl/WhitelistHandler.h
```diff
@@ -0,0 +1,166 @@
...
+            ips_.push_back(ip::make_address(net));
+            return;
+        }
+
+        if (isV4(net))
+            subnetsV4_.push_back(ip::make_network_v4(net));
+        else if (isV6(net))
+            subnetsV6_.push_back(ip::make_network_v6(net));
+        else
+            throw std::runtime_error(std::string{"malformed network: "} + net.data());
```
godexsoft: `string{...} + str` pattern can be replaced by `fmt::format("malformed network: {}", net.data())`.

unittests/webserver/WhitelistHandlerTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+            ]
+        }
+    }
+)JSON";
+
+class WhitelistHandlerTest : public NoLoggerFixture
+{
+};
+
+TEST_F(WhitelistHandlerTest, testWhiteListIPV4)
```
godexsoft: In Clio, names of tests usually start from capital letter 🔠

unittests/webserver/SweepHandlerTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+            "max_fetches": 100,
+            "sweep_interval": 0.1,
+            "max_connections": 2,
+            "whitelist": ["127.0.0.1"]
+        }
+    }
+)JSON";
+
+template <typename SweepHandler>
+struct BasicDOSGuardMock : public BaseDOSGuard
```
godexsoft: This mock could be moved to FakesAndMocks or a similar file. It may be useful in other places, even tho it's not used anywehere else atm.

godexsoft (changes requested):
A few more suggestions, mostly small ones 👍

unittests/DOSGuard.cpp
```diff
@@ -38,28 +39,26 @@ constexpr static auto JSONData = R"JSON(
...
+    MOCK_METHOD(bool, isWhiteListed, (std::string_view ip), (const));
+};
+
+using MockWhitelistHandlerType = NiceMock<MockWhitelistHandler>;
+
 class FakeSweepHandler
 {
 private:
-    using guard_type = BasicDOSGuard<FakeSweepHandler>;
+    using guard_type = BasicDOSGuard<MockWhitelistHandlerType, FakeSweepHandler>;
```
godexsoft, line 61: This is not too important and i know i'm guilty of leaving a few of those in the codebase but consider renaming to `GuardType` instead of the snake case notation here 🥇

godexsoft (approved):
LGTM! there is one thing that can be renamed but you can also do that in your next PR where you want to refactor a bunch of things anyways 👍

PR #804, opened 2023-08-02
Return error when limit<=0

cindyyan317:
Have confirmed with @PeterChen13579 
When limit is 0 or negative value , we raise "invalidParameter" error.

### DIFF
--- a/src/rpc/common/Validators.h
+++ b/src/rpc/common/Validators.h
@@ -221,6 +221,90 @@ public:
 };
 
 /**
+ * @brief Validate that value is equal or greater than the specified min
+ */
+template <typename Type>
+class Min final
+{
+    Type min_;
+
+public:
+    /**
+     * @brief Construct the validator storing min value
+     *
+     * @param min
+     */
+    explicit Min(Type min) : min_{min}
+    {
+    }
+
+    /**
+     * @brief Verify that the JSON value is not smaller than min
+     *
+     * @param value The JSON value representing the outer object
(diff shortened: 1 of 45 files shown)


src/rpc/common/Validators.h
```diff
@@ -214,6 +220,90 @@ class Between final
...
+     * @brief Construct the validator storing max value
+     *
+     * @param max
+     */
+    explicit Max(Type max) : max_{max}
+    {
+    }
+
+    /**
+     * @brief Verify that the JSON value is within a certain range
```
godexsoft: looks like a copy-paste leftover :)

src/rpc/common/Validators.h
```diff
@@ -214,6 +220,90 @@ class Between final
...
+    explicit Min(Type min) : min_{min}
+    {
+    }
+
+    /**
+     * @brief Verify that the JSON value is not smaller than min
+     *
+     * @param value The JSON value representing the outer object
+     * @param key The key used to retrieve the tested value from the outer
+     * object
```
godexsoft, line 246: this and similar copy-pasted doxygen can be improved by putting this back onto the same line as the `@param` - after we increased max line width in clang-format this should now be fitting on one line.

godexsoft (dismissed):
Wow Cindy, great job! If you want you can fix the comments but i'm also really happy with this as is 👍

godexsoft (approved):

PR #808, opened 2023-08-02
Expose advanced options from cassandra-cpp-driver thru the config

godexsoft:
This should be useful for fine-tuning the way Clio interacts with the DB.

### DIFF
--- a/example-config.json
+++ b/example-config.json
@@ -18,7 +18,7 @@
             // ---
             "max_connections_per_host": 1, // Defaults to 2
             "core_connections_per_host": 1, // Defaults to 2
-            "max_concurrent_requests_threshold": 55000, // Defaults to ((max_read + max_write) / core_connections_per_host)
+            "max_concurrent_requests_threshold": 55000 // Defaults to ((max_read + max_write) / core_connections_per_host)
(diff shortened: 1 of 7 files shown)


example-config.json
```diff
@@ -9,10 +12,27 @@
...
-            "core_connections_per_host": 1, // defaults to 2
-            "max_concurrent_requests_threshold": 55000, // defaults to max_read + max_write / core_connections_per_host
-            "threads": 8
+            "threads": 8,
+            //
+            // Advanced options. USE AT OWN RISK:
+            // ---
+            "max_connections_per_host": 1, // Defaults to 2
+            "core_connections_per_host": 1, // Defaults to 2
+            "max_concurrent_requests_threshold": 55000, // Defaults to ((max_read + max_write) / core_connections_per_host)
```
cindyyan317: This comma will cause json format error

godexsoft: That's true. I'll fix it even tho at the top of the file it says not to use this example as-is 👍

cindyyan317 (dismissed):
LGTM

cindyyan317 (approved):

PR #813, opened 2023-08-07
Build on linux

legleux:
Build and test Clio on linux


godexsoft (approved):
Looks like a good start! 🚀 
Eventually would love to have this back:
- Clang build for Linux, not only gcc
- Backend tests (need db in docker)
- Get coverage working again

PR #815, opened 2023-08-07
Update readme and compiler requirements

godexsoft:
Finalize readme for Conan migration.


godexsoft:
@thejohnfreeman @shawnxie999 I'm removing the sections that were using John's custom fork and local packages for now. I think it's fine to provide minimal build steps for now. The default use case is to just build Clio in release mode and at this moment the best way to do it is via Conan+Artifactory+Cmake combo.

Please feel free to review and help me come up with better wording if you think something is unclear in the build instructions.

README.md
```diff
@@ -30,28 +69,23 @@ build_type=Release
...
+conan remove -f cassandra-cpp-driver
 ```
-this will export a local package `xrpl/1.12.0-b2`.
 
-4. Navigate to clio's root directory and perform
+## Building Clio
+
+Navigate to Clio's root directory and perform
 ```sh
 conan export external/cassandra # export our "custom" cassandra driver package
```
cindyyan317: cassandra dependency seems available at remote artifactory . The local one may not be used actually

godexsoft: Good point, it's on our artifactory already so should be sufficient. I'm a bit afraid that there are other dependencies with the same name (https://conan.io/center/recipes/cassandra-cpp-driver?version=2.15.3) but it does look like it only would resolve the older version (we have 2.16.3) so it should be forced to use ours (especially if artifactory was setup correctly with the `-insert 0` option).

cindyyan317 (approved):
LGTM

PR #816, opened 2023-08-09
Move connection state logic to an earlier point

godexsoft:
Seems like boost 1.82 treats the suggested timeout option slightly different than before so the 20 sec idle timeout hack is no longer needed. 
Also the connected state flag was previously set after receiving the initial message from the subscription. While this is probably correct, it's possible that we will not receive the message soon enough and the connection state will remain set to `false` for longer than needed leading to forward requests potentially failing until clio realizes it is connected again.


PeterChen13579 (approved):
LGTM!

PR #817, opened 2023-08-09
populate/change namespace + refactor

PeterChen13579:
- moved logger and config folder/files to src/util folder; all the files are under clio::util namespace
- got rid of everything with only "clio" namespace. Everything with clio should have something after it
- added "test" suffix to every .cpp files that includes tests
- changed "Server" namespace to "web"
- under src/webserver/impl the namespace is web::detail
- changed "Backend" namespace to "data". Same with folder name. "Backend::Cassandra" is now "data::cassandra"
- under src/etl/impl, the namespace is now clio::etl::detail


CMakeLists.txt
```diff
@@ -134,12 +134,12 @@ if(tests)
   add_executable(${TEST_TARGET}
     # Common
     unittests/Main.cpp
-    unittests/Playground.cpp
```
godexsoft, line 137: This is probably the only exception and i'd leave it as just Playground.cpp 👍

CMakeLists.txt
```diff
@@ -192,13 +192,13 @@ if(tests)
...
     unittests/rpc/handlers/VersionHandlerTest.cpp
     # Backend
-    unittests/backend/BackendFactoryTest.cpp
-    unittests/backend/cassandra/BaseTests.cpp
-    unittests/backend/cassandra/BackendTests.cpp
-    unittests/backend/cassandra/RetryPolicyTests.cpp
-    unittests/backend/cassandra/SettingsProviderTests.cpp
-    unittests/backend/cassandra/ExecutionStrategyTests.cpp
-    unittests/backend/cassandra/AsyncExecutorTests.cpp
+    unittests/data/BackendFactoryTest.cpp
```
godexsoft: Let's pick either Test or Tests filename ending and rename all of the mismatches while we are at it 🙏

src/data/README.md
```diff
@@ -121,7 +121,7 @@ CREATE TABLE clio.successor (
...
 
-![Successor Table Trace Diagram](https://raw.githubusercontent.com/Shoukozumi/clio/9b2ea3efb6b164b02e9a5f0ef6717065a70f078c/src/backend/README.png)
+![Successor Table Trace Diagram](https://raw.githubusercontent.com/Shoukozumi/clio/9b2ea3efb6b164b02e9a5f0ef6717065a70f078c/src/data/README.png)
```
godexsoft: I have a feeling this is not an intended change 🫡

src/main/Main.cpp
```diff
@@ -178,12 +179,12 @@ try
     io_context ioc{threads};
 
     // Rate limiter, to prevent abuse
-    auto sweepHandler = IntervalSweepHandler{config, ioc};
-    auto whitelistHandler = WhitelistHandler{config};
-    auto dosGuard = DOSGuard{config, whitelistHandler, sweepHandler};
+    auto sweepHandler = web::IntervalSweepHandler{config, ioc};
+    auto whitelistHandler = web::detail::WhitelistHandler{config};
```
godexsoft: I'd avoid detail namespace for WhitelistHandler if we are using it here in main.

godexsoft (changes requested):
I found a few little things that probably should be updated 🔬 
On top of the few comments I think I'm not sure if having `clio` namespace is useful at all and would not just become an extra burden to type every time. What do you think if we completely remove `clio` namespace and leave `web`, `data`, `etl`, `rpc` and `util` as the top level namespaces? I think with the new physical layout the namespace names will neatly match the folders under `src` and it will be easier to know what to type for both namespaces and includes if you are looking for something from a specific subsystem.

Other than that, we can do similar improvements to `rpc` and `etl` namespaces/subsystems but that can be a separate PR if that's easier.

src/webserver/impl/WhitelistHandler.h
```diff
@@ -27,7 +27,7 @@
 #include <unordered_map>
 #include <unordered_set>
 
-namespace web::detail {
+namespace web {
```
godexsoft, line 30: I think if you move it out of the detail namespace you probably want to move the file out of impl too?

godexsoft (approved):
Found one small issue but i think you can leave that for later. If you want to merge this - go ahead 🚀

PR #823, opened 2023-08-15
Add LOG macro to prevent unnecessary evaluations

godexsoft:
This is basically copying what JLOG in `rippled` is doing. It's a trick that saves us evaluating potentially expensive argument lists used for logging on levels that we rarely enable on servers (trace and the like).


PeterChen13579 (approved):
LGTM ! 😄 💯

PR #825, opened 2023-08-16
Fix for empty log lines

godexsoft:
#823 introduced a problem with an extra empty log line for each time we use LOG. This is because `logger.debug()` and the like return by value and the LOG macro did not take that into account.
There would be a better way to fix this with another proxy object but sadly boost.log does not give us a way to ask whether the filter would pass the given severity for the channel or not.. it only provides a way to try and open a log record which can return an `empty` record which means the log line is going to be discarded.


PeterChen13579 (approved):
👍 💯

PR #826, opened 2023-08-21
Filtering Account_tx by TransactionType

PeterChen13579:
Tackled new feature #685

### DIFF
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -155,8 +155,10 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
             obj[JS(tx)] = std::move(txn);
 
             auto objTransactionType = obj[JS(tx)].as_object()[JS(TransactionType)];
+            auto it = TYPESMAP.find(objTransactionType.as_string());
+
             // if transactionType does not match
-            if (input.transactionType.has_value() &&
+            if (input.transactionType.has_value() && it != TYPESMAP.end() &&
                 AccountTxHandler::TYPESMAP.at(objTransactionType.as_string()) != input.transactionType.value())
                 continue;
 
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -22,6 +22,34 @@
 
 namespace rpc {
 
+// found here : https://xrpl.org/transaction-types.html
+std::unordered_map<std::string_view, ripple::TxType> const AccountTxHandler::TYPESMAP{
+    {"AccountSet", ripple::ttACCOUNT_SET},
+    {"AccountDelete", ripple::ttACCOUNT_DELETE},
+    {"CheckCancel", ripple::ttCHECK_CANCEL},
+    {"CheckCash", ripple::ttCHECK_CASH},
+    {"CheckCreate", ripple::ttCHECK_CREATE},
+    {"DepositPreauth", ripple::ttDEPOSIT_PREAUTH},
+    {"EscrowCancel", ripple::ttESCROW_CANCEL},
+    {"EscrowCreate", ripple::ttESCROW_CREATE},
+    {"EscrowFinish", ripple::ttESCROW_FINISH},
+    {"NFTokenAcceptOffer", ripple::ttNFTOKEN_ACCEPT_OFFER},
+    {"NFTokenBurn", ripple::ttNFTOKEN_BURN},
+    {"NFTokenCancelOffer", ripple::ttNFTOKEN_CANCEL_OFFER},
+    {"NFTokenCreateOffer", ripple::ttNFTOKEN_CREATE_OFFER},
+    {"NFTokenMint", ripple::ttNFTOKEN_MINT},
+    {"OfferCancel", ripple::ttOFFER_CANCEL},
+    {"OfferCreate", ripple::ttOFFER_CREATE},
+    {"Payment", ripple::ttPAYMENT},
+    {"PaymentChannelClaim", ripple::ttPAYCHAN_CLAIM},
+    {"PaymentChannelCreate", ripple::ttCHECK_CREATE},
+    {"PaymentChannelFund", ripple::ttPAYCHAN_FUND},
+    {"SetRegularKey", ripple::ttREGULAR_KEY_SET},
+    {"SignerListSet", ripple::ttSIGNER_LIST_SET},
+    {"TicketCreate", ripple::ttTICKET_CREATE},
+    {"TrustSet", ripple::ttTRUST_SET},
+};
+
 // TODO: this is currently very similar to nft_history but its own copy for time
 // being. we should aim to reuse common logic in some way in the future.
 AccountTxHandler::Result
@@ -116,6 +144,13 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
             auto [txn, meta] = toExpandedJson(txnPlusMeta, NFTokenjson::ENABLE);
             obj[JS(meta)] = std::move(meta);
             obj[JS(tx)] = std::move(txn);
+
+            auto objTransactionType = obj[JS(tx)].as_object()[JS(TransactionType)];
+            // if transactionType does not match
+            if (input.transactionType.has_value() &&
+                AccountTxHandler::TYPESMAP.at(objTransactionType.as_string()) != input.transactionType.value())
+                continue;
+
             obj[JS(tx)].as_object()[JS(ledger_index)] = txnPlusMeta.ledgerSequence;
             obj[JS(tx)].as_object()[JS(date)] = txnPlusMeta.date;
         }
@@ -207,6 +242,12 @@ tag_invoke(boost::json::value_to_tag<AccountTxHandler::Input>, boost::json::valu
         input.marker = AccountTxHandler::Marker{
             jsonObject.at(JS(marker)).as_object().at(JS(ledger)).as_int64(),
             jsonObject.at(JS(marker)).as_object().at(JS(seq)).as_int64()};
+
+    if (jsonObject.contains(JS(TransactionType)))
+    {
+        auto objTransactionType = jsonObject.at(JS(TransactionType));
+        input.transactionType = AccountTxHandler::TYPESMAP.at(objTransactionType.as_string());
+    }
 
     return input;
 }
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -154,11 +154,11 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
             obj[JS(meta)] = std::move(meta);
             obj[JS(tx)] = std::move(txn);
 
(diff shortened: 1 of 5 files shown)


src/rpc/handlers/AccountTx.h
```diff
@@ -109,6 +112,34 @@ class AccountTxHandler
                  {JS(ledger), validation::Required{}, validation::Type<uint32_t>{}},
                  {JS(seq), validation::Required{}, validation::Type<uint32_t>{}},
              }},
+            {JS(TransactionType),
+             validation::Type<std::string>{},
+             validation::OneOf<std::string>{
```
godexsoft: Let's not duplicate the strings by hand and do the same way `LedgerDataHandler::TYPES_KEYS` is done 👍

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -725,6 +725,61 @@ TEST_F(RPCAccountTxHandlerTest, SpecificLedgerIndexValidated)
...
+    mockBackendPtr->updateRange(MAXSEQ);  // max
+    MockBackend* rawBackendPtr = static_cast<MockBackend*>(mockBackendPtr.get());
+    // adjust the order for forward->false
+    auto const transactions = genTransactions(MAXSEQ, MAXSEQ - 1);
+    auto const transCursor = TransactionsAndCursor{transactions, TransactionsCursor{12, 34}};
+    ON_CALL(*rawBackendPtr, fetchAccountTransactions).WillByDefault(Return(transCursor));
+    EXPECT_CALL(
+        *rawBackendPtr,
+        fetchAccountTransactions(
+            testing::_,
```
godexsoft: I believe the `testing` namespace is already available - you can simply use `_`.

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -725,6 +725,61 @@ TEST_F(RPCAccountTxHandlerTest, SpecificLedgerIndexValidated)
...
+        EXPECT_EQ(output->at("transactions").as_array().size(), 0);
+    });
+
+    runSpawn([&, this](auto yield) {
+        auto const handler = AnyHandler{AccountTxHandler{mockBackendPtr}};
+        auto const static input = boost::json::parse(fmt::format(
+            R"({{
+                "account":"{}",
+                "ledger_index":"validated",
+                "TransactionType": "Payment"
```
godexsoft: It would be great to keep the formatting  uniform - I personally prefer with the space between `:` and `"` 🥇

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -725,6 +725,61 @@ TEST_F(RPCAccountTxHandlerTest, SpecificLedgerIndexValidated)
...
+    EXPECT_CALL(*rawBackendPtr, fetchLedgerBySequence).Times(2);
+    ON_CALL(*rawBackendPtr, fetchLedgerBySequence(MAXSEQ, _)).WillByDefault(Return(ledgerinfo));
+
+    runSpawn([&, this](auto yield) {
+        auto const handler = AnyHandler{AccountTxHandler{mockBackendPtr}};
+        auto const static input = boost::json::parse(fmt::format(
+            R"({{
+                "account":"{}",
+                "ledger_index":"validated",
+                "TransactionType": "EscrowCancel"
```
godexsoft: So we only test two out of many types here, is that deliberate? 
If you would like to test that at least each of the supported parameters is passing but anything else don't, you can implement something similar to `LedgerEntryParameterTest` for your new `TransactionType` parameters.

godexsoft: Another thing is that i would have called it `transaction_type` instead to keep the API fields in the same style.
It does look like `TransactionType` is used already in some responses (along other CamelCase fields) but i don't see any requests using CamelCase, only snake_case. 

@intelliot @injaelee Thoughts?

godexsoft (changes requested):
Few small fixes and improvements are possible - left some comments.
Also I'm not happy with the field name - let's see if we can get it changed to snake_case.

src/rpc/handlers/AccountTx.cpp
```diff
@@ -116,6 +153,13 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
             auto [txn, meta] = toExpandedJson(txnPlusMeta, NFTokenjson::ENABLE);
             obj[JS(meta)] = std::move(meta);
             obj[JS(tx)] = std::move(txn);
+
+            auto objTransactionType = obj[JS(tx)].as_object()[JS(TransactionType)];
+            // if transactionType does not match
```
godexsoft, line 159: Could use an empty line before the comment - always better 👍

src/rpc/handlers/AccountTx.cpp
```diff
@@ -116,6 +153,13 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
...
+            if (input.transactionType.has_value() &&
+                AccountTxHandler::TYPESMAP.at(objTransactionType.as_string()) != input.transactionType.value())
```
godexsoft, line 162: If an unknown transaction type appears in the json this code will start crashing i think. Let's add a check that the TYPESMAP/TYPES_KEYS actually knows about this type before we use `.at()`.

src/rpc/handlers/AccountTx.h
```diff
@@ -39,6 +39,10 @@ class AccountTxHandler
     util::Logger log_{"RPC"};
     std::shared_ptr<BackendInterface> sharedPtrBackend_;
 
+    static std::unordered_map<std::string_view, ripple::TxType> const TYPESMAP;
+
```
godexsoft: Probably no need for extra empty line here. The two are related to each other and so it makes sense to group them.

src/rpc/handlers/AccountTx.h
```diff
@@ -109,6 +114,13 @@ class AccountTxHandler
                  {JS(ledger), validation::Required{}, validation::Type<uint32_t>{}},
                  {JS(seq), validation::Required{}, validation::Type<uint32_t>{}},
              }},
+            {JS(TransactionType),
+             meta::WithCustomError{
+                 validation::Type<std::string>{},
+                 Status{ripple::rpcINVALID_PARAMS, "Invalid field 'type', not string."}},
```
godexsoft: If the `meta::WithCustomError` is used just for the message to be aligned with rippled - no need to do it. 
As long as the error code matches we consider it the same. 
In Clio we actually prefer to have universal error messages generated by the framework instead of customizing each one.

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -1028,6 +1028,314 @@ TEST_F(RPCAccountTxHandlerTest, NFTTxs)
...
+            })",
+            "[]"},
+        AccountTxTransactionBundle{
+            "Payment",
+            R"({
+                "account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
+                "ledger_index": "validated",
+                "TransactionType": "Payment"
+            })",
+            R"([
```
godexsoft, line 1197: Looks like this json is misaligned in multiple places compared to others. Please fix 🥇

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -1028,6 +1028,314 @@ TEST_F(RPCAccountTxHandlerTest, NFTTxs)
...
+    mockBackendPtr->updateRange(MAXSEQ);  // max
+    MockBackend* rawBackendPtr = static_cast<MockBackend*>(mockBackendPtr.get());
+
+    auto const transactions = genTransactions(MAXSEQ, MAXSEQ - 1);
+    auto const transCursor = TransactionsAndCursor{transactions, TransactionsCursor{12, 34}};
+    ON_CALL(*rawBackendPtr, fetchAccountTransactions).WillByDefault(Return(transCursor));
+    EXPECT_CALL(
+        *rawBackendPtr,
+        fetchAccountTransactions(
+            testing::_,
```
godexsoft: no need for `testing::` here i believe.

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -1028,6 +1028,314 @@ TEST_F(RPCAccountTxHandlerTest, NFTTxs)
...
+    EXPECT_CALL(*rawBackendPtr, fetchLedgerBySequence).Times(1);
+
+    ON_CALL(*rawBackendPtr, fetchLedgerBySequence(MAXSEQ, _)).WillByDefault(Return(ledgerinfo));
+
+    auto const testBundle = GetParam();
+    runSpawn([&, this](auto yield) {
+        auto const handler = AnyHandler{AccountTxHandler{mockBackendPtr}};
+        auto const req = json::parse(testBundle.testJson);
+        auto const output = handler.process(req, Context{yield});
+        ASSERT_TRUE(output);
```
godexsoft: I'd use EXPECT_TRUE here so that if this ever breaks it does not halt the entire test run.

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -1028,6 +1028,314 @@ TEST_F(RPCAccountTxHandlerTest, NFTTxs)
...
+
+        auto transactions = output->at("transactions").as_array();
+        // parse to json object
```
godexsoft: This comment can be removed as the code already says the same thing 💯 
I'd also use `auto const` instead of `json::value` below as well as `auto const transactions` above.

godexsoft (changes requested):
Adding one concern that i have not spotted before as well as a few small changes that can improve this further 🥇

godexsoft (approved):
LGTM! 🚀

godexsoft:
The question about CamelCase vs snake_case in the request still remains tho. 
@injaelee @intelliot Please add your opinion 👍

src/rpc/handlers/AccountTx.cpp
```diff
@@ -22,6 +22,43 @@
 
 namespace rpc {
 
+// found here : https://xrpl.org/transaction-types.html
+std::unordered_map<std::string_view, ripple::TxType> const AccountTxHandler::TYPESMAP{
```
intelliot, line 26: note: looks like this will need updating when additional transaction types are added in the future

unittests/rpc/handlers/AccountTxTests.cpp
intelliot: snake_case is fine for the request field name

intelliot (commented):
snake_case is fine for the request field name

godexsoft:
This continues here: https://github.com/XRPLF/clio/pull/851

src/rpc/handlers/AccountTx.cpp
godexsoft: Sure, we should keep track of those somehow.

godexsoft: Unless there is a way to get all supported types from libxrpl. In which case we should probably do that instead of hardcoding our own list.

PR #828, opened 2023-08-23
Add working dir to git command executions

godexsoft:
This may be unnecessary but better safe than sorry. Setting the working directory can help in case cmake ran from outside the git repo.


cindyyan317 (approved):
👍

PR #829, opened 2023-08-24
Implement amendment block handler and test

### LINKED ISSUE #364: clio should repeatedly report when it is amendment blocked
If an amendment is activated on the network from which clio is extracting data, but clio is not linked against a version of rippled that understands the amendment, clio will crash. we should detect this situation, and log an error continuously, instead of crashing.

godexsoft:
Fixes #364

Implements a handler for when Clio detects it's amendment blocked.
The handler sets a global flag (as before) but also initiates a repeating timer that logs the error as `FATAL`.
ETL is blocked from processing but RPC can continue handling requests.

### DIFF
--- a/src/etl/ETLService.cpp
+++ b/src/etl/ETLService.cpp
@@ -111,8 +111,8 @@ ETLService::monitor()
         }
         catch (std::runtime_error const& e)
         {
-            amendmentBlockHandler_.onAmendmentBlock();
-            return;
+            LOG(log_.fatal()) << "Failed to load initial ledger: " << e.what();
+            return amendmentBlockHandler_.onAmendmentBlock();
         }
 
         if (ledger)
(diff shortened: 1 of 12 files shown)


src/etl/ETLService.cpp
```diff
@@ -110,11 +111,7 @@ ETLService::monitor()
         }
         catch (std::runtime_error const& e)
```
godexsoft, line 112: Catching `runtime_error` here also would catch the ones we throw ourselves. This may or may not be desired. 
Certainly the `e.what()` part should still be logged - will fix.

src/etl/impl/Transformer.h
```diff
@@ -238,7 +243,7 @@ class Transformer
                 LOG(log_.debug()) << "object neighbors not included. using cache";
 
                 if (!backend_->cache().isFull() || backend_->cache().latestLedgerSequence() != lgrInfo.seq - 1)
-                    throw std::runtime_error("Cache is not full, but object neighbors were not included");
+                    throw std::logic_error("Cache is not full, but object neighbors were not included");
```
godexsoft, line 248: Previously this would have been caught by the try/catch that attempts to catch amendment block situation. I don't think this error here can happen if Clio is setup correctly so I'd consider this a logic error and we can either have a separate catch clause for it or let it bubble up to main (i believe that was the original behaviour).

cindyyan317 (dismissed):
LGTM ~ 
I agree we shall log the e.what()

cindyyan317 (approved):
GOGOGO

PR #832, opened 2023-08-29
Fix synchronous paths

godexsoft:
We have the `data::synhronous` function that is used to call coroutine-based code synchronously. 
While it is a nice idea that allows for less code duplication it's both not most efficient and more importantly it seems to be broken since we moved to boost 1.82 (used to work fine with boost 1.75).

A bunch of testing revealed that 
- There appears to be no issue with the `FutureWithCallback` and how we use it with `async_compose` - for each request there is always exactly one callback invocation whether the result is an error or some data from the DB; unless
- When used thru `synchronous`, sometimes `async_compose` will not return and the thread that is using `synchronous` gets stuck entirely

This PR currently adds `sync` versions of some of the coroutine-based functions that were used thru `synchronous`. The change appears to help the situation (have not seen the freezing behaviour yet but this could change when perf testing).

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -24,7 +24,7 @@ class Clio(ConanFile):
         'fmt/10.0.0',
         'grpc/1.50.1',
         'openssl/1.1.1u',
-        'xrpl/1.12.0-b2',
+        'xrpl/1.12.0-rc1',
     ]
 
     default_options = {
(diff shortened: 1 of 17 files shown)


godexsoft:
Notes:
- There is still one place using `synchronous` - should probably be removed but requires a bunch of changes
- A lot of code duplication that needs to be addressed somehow

godexsoft:
Discarding this. The problem is not really in the way we synchronize otherwise async code but in other code around async_compose.

PR #836, opened 2023-09-07
Better clangd support

### LINKED ISSUE #839: Better clangd support
It would be nice to export `compile_commands.json` from cmake by default. Also clangd creates `.cache` directory, so it should be ignored by git.

kuznetsss:
Clangd requires file `build/compile_commands.json` to be able to index project and provide features like "Find usages". I added the option to urn on generating the file by `cmake`.
Also clangd creates local `.cache` directory so I made git to ignore it.

Fixed #839

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *clio*.log
 build*/
 .build
+.cache
 .vscode
 .python-version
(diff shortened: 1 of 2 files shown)


godexsoft (approved):
LGTM! 👍

PR #837, opened 2023-09-07
Fix missing lock

kuznetsss:


godexsoft (approved):
Can't believe we missed this before 🥇

PR #838, opened 2023-09-11
Add log to better debug DB workload

cindyyan317:
We may issue lots of db requests through this call. The last week , we observed that even the requests traffic not change much, the db access increased vastly. 
We will change this part for better db throttling in future (but not in 2.0). So this PR adds the number of objects in the log for bettering understanding clio behavior from log.
The log will be like 
"
src/rpc/RPCHelpers.cpp:657 [2023-09-08 15:11:26.352054] RPC:DBG Time loading owned directories: 28361 milliseconds, entries size: 200051
"


godexsoft (approved):
LGTM! ⭐

PR #841, opened 2023-09-11
Add workaround for async_compose

### LINKED ISSUE #840: Locking of async_compose code
We are seeing `async_compose` code locking up randomly in `DefaultExecutionStrategy` read code.
The issue is [described in some detail here](http://stackoverflow.com/questions/77004137/boost-asio-async-compose-gets-stuck-under-load).
This needs to be fixed before 2.0 can be released.

godexsoft:
Fixes #840

This is a combination of two workarounds found by @cindyyan317. It appears that just the "lifetime extension" workaround is not enough and still fails under load while seemingly working fine in unit-tests. After a lot of debugging and experimentation we found out how to write the work_guards correctly so the code works well on both linux and mac. However, it does feel like a workaround still. Especially because valgrind and asan report errors on a minimal example.

There is a [discussion about this issue on here](http://stackoverflow.com/questions/77004137/boost-asio-async-compose-gets-stuck-under-load).

The problem appears to be that `async_compose` does something to the stack (maybe coroutine implementation detail) that makes our `FutureWithCallback` release prematurely, even before leaving the scope where it was defined. This workaround makes sure that the original callback survives until the very end so that the inner `post` in the read functions can be scheduled and executed.


conanfile.py
```diff
@@ -24,7 +24,7 @@ class Clio(ConanFile):
         'fmt/10.0.0',
         'grpc/1.50.1',
         'openssl/1.1.1u',
-        'xrpl/1.12.0-b2',
+        'xrpl/1.12.0',
```
kuznetsss, line 27: Is it related to the issue?

godexsoft:
This survived performance testing - seems like a working solution 🎉

kuznetsss (approved):
👍

cindyyan317 (approved):
Let's go

PR #847, opened 2023-09-12
Add admin password check

### LINKED ISSUE #846: Verify admin password
Currently clio will verify anyone coming from 127.0.0.1 as admin. We should add another way to handle this thru a password that is added to configuration file and check against the auth header in the request.

kuznetsss:
Fixes #846


src/web/impl/AdminVerificationStrategy.h
```diff
@@ -0,0 +1,82 @@
...
+    std::string password_;
+
+public:
+    PasswordAdminVerificationStrategy(std::string password);
+
+    /**
+     * @brief Checks whether request is from a host that is considered authorized as admin using
+     * the password (if any) from the request.
+     *
+     * @param passwordFromRequest The password from request
```
godexsoft: Doc does not match signature anymore ☝️

unittests/web/ServerTests.cpp
```diff
@@ -359,3 +359,76 @@ TEST_F(WebServerTest, WsPayloadOverload)
         res,
         R"({"payload":"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","warning":"load","warnings":[{"id":2003,"message":"You are about to be rate limited"}]})");
 }
+
+constexpr static auto JSONServerConfigWithAdminPassword = R"JSON(
```
godexsoft: Let's try to keep `const` and `constexpr` on east side? 🥇 
Ideally this would be `static auto constexpr` but we have all styles mixed atm. 
We can change them all in one separate PR perhaps.

unittests/web/ServerTests.cpp
```diff
@@ -359,3 +359,76 @@ TEST_F(WebServerTest, WsPayloadOverload)
...
+    {
+        "server":{
```
godexsoft, line 366: Minor: formatting of JSON could be slightly better 🔢

godexsoft (changes requested):
Overall looks good! Spotted only a few minor things ⭐ 
Oh, and we probably should describe this new feature in README as well.

unittests/web/ServerTests.cpp
kuznetsss: We should move to a newer clang-format so this could be forced
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment

#848

cindyyan317 (dismissed):

godexsoft (approved):

PR #851, opened 2023-09-13
AccountTx filtering by transaction type

godexsoft:
Implements #685 

Original PR by @PeterChen13579: https://github.com/XRPLF/clio/pull/826

This PR adds on top of that
- Change request field name used to `tx_type`
- Introduces ToLower util and RPC field modifier
- Matches `tx_type` without case-sensitivity

### DIFF
--- a/src/rpc/common/Modifiers.h
+++ b/src/rpc/common/Modifiers.h
@@ -22,6 +22,9 @@
 #include <rpc/common/Concepts.h>
 #include <rpc/common/Specs.h>
 #include <rpc/common/Types.h>
+#include <util/JsonUtils.h>
+
+#include <string_view>
 
 namespace rpc::modifiers {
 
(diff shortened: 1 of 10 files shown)


src/rpc/common/Modifiers.h
```diff
@@ -68,4 +71,32 @@ class Clamp final
...
+{
+    /**
+     * @brief Update the input string to lower case.
+     *
+     * @param value The JSON value representing the outer object
+     * @param key The key used to retrieve the modified value from the outer object
+     * @return Possibly an error
+     */
+    [[nodiscard]] MaybeError
+    modify(boost::json::value& value, std::string_view key) const
```
kuznetsss, line 89: Why not moving the value and returning it back?
Current version makes usage shorter, but I think it is easier to miss that `v` is being modified here
```cpp
boost::json::value v;
if (!ToLower().modify(v)) {
 . . .
}
```
Comparing to this
```cpp
boost::json::value v;
auto expected = ToLower().modify(std::move(v));
if (!expected.has_value()) {
 . . .
 }
 v = std::move(expected.value());
 ```

**Review by kuznetsss — COMMENTED** [review]
All good, just one question

File: src/rpc/common/Modifiers.h (hunk 1 above)
Comment (godexsoft) [reply src/rpc/common/Modifiers.h:89]: Modifiers were added late. At first we had only Validators and they had this signature already. I'm not against changing it later but i'd not touch this for now 👍

File: src/rpc/handlers/AccountTx.cpp [hunk 2]
```diff
@@ -18,10 +18,48 @@
 //==============================================================================
 
 #include <rpc/handlers/AccountTx.h>
+#include <util/JsonUtils.h>
 #include <util/Profiler.h>
 
 namespace rpc {
 
+// found here : https://xrpl.org/transaction-types.html
```
cindyyan317, line 26: Since we have supported "Clawback", should we also add "Clawback"?

cindyyan317 (dismissed):

cindyyan317 (approved):
👍  LOOKS GOOD

PR #852, opened 2023-09-13
Remove deprecated cassandra options

godexsoft:
Unfortunately all the options we added for testing turned out to be deprecated. Removing them so that this does not make it to 2.0 release.


cindyyan317:
Code LGTM
They don't support these setting anymore? Is there any option we are using?

kuznetsss (commented):
LGTM

PR #854, opened 2023-09-15
Add libxrpl version to server_info output

### LINKED ISSUE #853: Add libxrpl version to server_info
It would be useful to know what version of libxrpl clio is built against. For example to know what amendments are supported etc. Let's expose it thru `server_info`.
See https://github.com/ximinez/rippled/blob/develop/src/ripple/protocol/impl/BuildInfo.cpp#L36

godexsoft:
Fixes #853


kuznetsss (approved):
👍

cindyyan317:
LGTM

PR #857, opened 2023-09-18
Amm info handler new

arihantkothari:
- Adds `amm_info` handler to clio


src/rpc/common/Validators.cpp
```diff
@@ -196,4 +196,32 @@ CustomValidator SubscribeAccountsValidator =
         return MaybeError{};
     }};
 
+CustomValidator AssetValidator =
```
godexsoft: Is this potentially something that can be unified with `ammAssetValidator` in LedgerEntry.h?

arihantkothari: Seems like `ammAssetValidator` in ledgerEntry.h is a better implementation because it uses rippled function `ripple::issueFromJson`. If rippled code changes, there'll be less chances of clio breaking. I'm going to put this method in the `validators.cpp` file and use it instead.

arihantkothari: I've pushed the changes

src/rpc/AMMHelpers.cpp
```diff
@@ -0,0 +1,81 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <data/BackendInterface.h>
+#include <rpc/AMMHelpers.h>
+#include <util/log/Logger.h>
+
+// #include <ripple/app/misc/AMMUtils.h>
```
godexsoft: Let's remove this commented include

src/rpc/AMMHelpers.h
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma once
+
+#include <data/BackendInterface.h>
+#include <rpc/RPCHelpers.h>
+
+namespace rpc {
+
+std::pair<ripple::STAmount, ripple::STAmount>
+getAmmPoolHolds(
```
godexsoft, line 31: Eeach function needs a doxygen comment 🔤

src/rpc/RPCHelpers.cpp
```diff
@@ -944,11 +948,10 @@ accountHolds(
 )
 {
     ripple::STAmount amount;
-    if (ripple::isXRP(currency)) {
+    if (ripple::isXRP(currency))
         return {xrpLiquid(backend, sequence, account, yield)};
-    }
-    auto key = ripple::keylet::line(account, issuer, currency).key;
 
+    auto key = ripple::keylet::line(account, issuer, currency).key;
```
godexsoft: Probably could be made const?

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,208 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <data/DBHelpers.h>
+#include <rpc/AMMHelpers.h>
+#include <rpc/handlers/AMMInfo.h>
+
+// #include <ripple/app/misc/AMMUtils.h>
```
godexsoft: Let's remove the commented include

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <ripple/protocol/AMMCore.h>
+
+namespace {
+
+std::string
+to_iso8601(ripple::NetClock::time_point tp)
```
godexsoft: All Clio code is in camelCase. This should probably be called `toISO8601` or `asISO8601`. I don't like how it looks in camelCase but it's still better than mixing both styles in one file 👍

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,95 @@
...
+
+        uint32_t ledgerIndex = 0;
+        bool validated = true;
+    };
+
+    struct Input {
+        std::optional<ripple::AccountID> accountID = std::nullopt;
+        ripple::Issue issue1 = ripple::noIssue();
+        ripple::Issue issue2 = ripple::noIssue();
+        std::optional<std::string> ledgerHash;
```
godexsoft, line 60: `accountID` is set to `nullopt` but `ledgerHash` is not. I wonder why not both or neither?

godexsoft (changes requested):

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,95 @@
...
+
+    static RpcSpecConstRef
+    spec([[maybe_unused]] uint32_t apiVersion)
+    {
+        static auto const rpcSpec = RpcSpec{
+            {JS(ledger_hash), validation::Uint256HexStringValidator},
+            {JS(ledger_index), validation::LedgerIndexValidator},
+            {JS(asset), validation::Required{}, validation::ammAssetValidator},
+            {JS(asset2), validation::Required{}, validation::ammAssetValidator},
+            {JS(account), validation::AccountValidator},
```
cindyyan317, line 78: The field name is wrong. It should be "amm_account"?

cindyyan317:
The amm_info can be fetch by "amm_account" field. But I did not see the implementation of it. Is it in the PR's scope?
https://xrpl.org/amm_info.html#amm_info
Since it is already in the document, I think we need to implement it .

src/rpc/RPCHelpers.cpp
```diff
@@ -904,14 +904,18 @@ xrpLiquid(
     ripple::SLE const sle{it, key};
 
     std::uint32_t const ownerCount = sle.getFieldU32(ripple::sfOwnerCount);
-
     auto const reserve = backend.fetchFees(sequence, yield)->accountReserve(ownerCount);
```
cindyyan317: Since reserve is used conditionally, we can save one backend call if it is looking for amm

src/rpc/RPCHelpers.cpp
cindyyan317: since u have been here, add 'const'

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    auto ammKeylet = keylet::amm(input.issue1, input.issue2);
+    auto const ammBlob = sharedPtrBackend_->fetchLedgerObject(ammKeylet.key, lgrInfo.seq, ctx.yield);
+
+    if (not ammBlob)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND}};
+
+    auto const amm = SLE{SerialIter{ammBlob->data(), ammBlob->size()}, ammKeylet.key};
+    auto const accID = amm.getAccountID(sfAccount);
```
cindyyan317: 1 there are too many account here. I suggest to use more describable name like 
'lpAccount' means the account to check lp
'ammAccount' means the amm's account
2 I think we don't need to check the existence of amm account. It should exist if amm object exists? That being said, i am not against to check it just in case, but the error message should be more specific , like "amm object not found"? because the input might contain the other account, the error message may misleading.

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,207 @@
...
+    auto const accBlob = sharedPtrBackend_->fetchLedgerObject(keylet::account(accID).key, lgrInfo.seq, ctx.yield);
+    if (not accBlob)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND}};
+
+    auto const [asset1Balance, asset2Balance] =
+        getAmmPoolHolds(*sharedPtrBackend_, lgrInfo.seq, accID, input.issue1, input.issue2, ctx.yield);
+    auto const lptAMMBalance = input.accountID
```
cindyyan317, line 78: this variable is not used , please check if it is buggy

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+void
+tag_invoke(boost::json::value_from_tag, boost::json::value& jv, AMMInfoHandler::Output const& output)
+{
+    boost::json::object amm = {
+        {JS(lp_token), output.lpToken},
+        {JS(amount), output.amount1},
+        {JS(amount2), output.amount2},
+        {JS(account), output.ammAccount},
+        {JS(trading_fee), output.tradingFee},
```
cindyyan317, line 163: please make sure the trading_fee type, should be number not string

src/rpc/AMMHelpers.cpp
```diff
@@ -0,0 +1,80 @@
...
+    BackendInterface const& backend,
+    std::uint32_t sequence,
+    ripple::AccountID const& ammAccountID,
+    ripple::Issue const& issue1,
+    ripple::Issue const& issue2,
+    boost::asio::yield_context yield
+)
+{
+    auto const assetInBalance =
+        accountHolds(backend, sequence, ammAccountID, issue1.currency, issue1.account, true, yield);
```
cindyyan317: please double confirm the flag is true not false. I checked rippled's implementation , we seem should pass false. Please do confirm this one.

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,95 @@
...
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+
+public:
+    struct Output {
+        // todo: use better type than json types
+        boost::json::value amount1;
+        boost::json::value amount2;
+        boost::json::value lpToken;
+        boost::json::array voteSlots;
+        boost::json::value auctionSlot = nullptr;
```
cindyyan317: no need nullptr, you can use auctionSlot.is_null() ,default constructed represents a null JSON element

src/rpc/RPCHelpers.cpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2022, the clio developers.
+    Copyright (c) 2023, the clio developers.
```
kuznetsss: The year here is the year the file appeared, so you don't need this change.

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,95 @@
...
+#include <rpc/common/Validators.h>
+
+namespace rpc {
+
+/**
+ * @brief AMMInfoHandler returns information about AMM pools.
+ *
+ * This API is not documented in the rippled API documentation.
+ */
+class AMMInfoHandler {
```
kuznetsss, line 35: There should be tests for new handler

kuznetsss (changes requested):

src/rpc/handlers/AMMInfo.h
arihantkothari: I'm adding these soon.

arihantkothari:
Thanks for the reviews, I'll work on each comment one by one. So apologies for replying late to them. @cindyyan317 @kuznetsss

src/rpc/RPCHelpers.cpp
arihantkothari: added, thanks! There's actually bunch of other functions `transferRate`, `isGlobalFrozen`, and `isFrozen` that has the same issue. But I will leave it for now since this is outside the scope of this PR and to avoid any potential side effects.

src/rpc/RPCHelpers.cpp
arihantkothari: done

src/rpc/RPCHelpers.cpp
arihantkothari: Hey @cindyyan317 , that is a good point.
There's 2 options (or maybe more...): 
1. putting the reserve inside the lambda like this
```
ripple::STAmount const amount = [&]() {
        // AMM doesn't require the reserves
        if ((sle.getFlags() & ripple::lsfAMMNode) != 0u)
            return balance;
        auto const reserve = backend.fetchFees(sequence, yield)->accountReserve(ownerCount);
        ripple::STAmount amount = balance - reserve;
        if (balance < reserve)
            amount.clear();
        return amount;
    }();
```
Pros: No extra check required and saves one call if it is looking for amm
Cons: It is inside the lambda, in this case lambda is just called once so it's fine but if it is called multiple times - we will be doing multiple calls.

2. Just adding a simple if statement check to do a backend fetch for reserve outside the lambda
```
if ((sle.getFlags() & ripple::lsfAMMNode) == 0u)
  reserve = backend.fetchFees(sequence, yield)->accountReserve(ownerCount);
```
Pros: Outside lambda and saves one backend call in case of amms
Cons: Now there's two checks instead one. But `(sle.getFlags() & ripple::lsfAMMNode)` can be just computed once

What do you think ? I'm leaning towards the second option

cindyyan317: I think both are okay.  maybe simple if is more readable.

arihantkothari: Sounds good, thank you. Marking this as resolved!

src/rpc/handlers/AMMInfo.h
arihantkothari: Fixed, thanks!

src/rpc/handlers/AMMInfo.h
arihantkothari: Fixed, sorry about that.

src/rpc/AMMHelpers.cpp
arihantkothari: I think you are right, thanks for finding this bug. I've fixed it and pushed it. But I wonder if you know what's a good test to test this up ?

cindyyan317: Please refer to the unittest "PaysUSDGetsXRPFrozen" how to mock a global frozen currency

arihantkothari:
> The amm_info can be fetch by "amm_account" field. But I did not see the implementation of it. Is it in the PR's scope? https://xrpl.org/amm_info.html#amm_info Since it is already in the document, I think we need to implement it .

I implemented this in my last commit, thanks for looking into it carefully.

src/rpc/handlers/AMMInfo.cpp
arihantkothari: fixed

src/rpc/AMMHelpers.h
arihantkothari: done!

src/rpc/handlers/AMMInfo.cpp
arihantkothari: How does my new changes look like now ?

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+}  // namespace
+
+namespace rpc {
+AMMInfoHandler::Result
+AMMInfoHandler::process(AMMInfoHandler::Input input, Context const& ctx) const
+{
+    using namespace ripple;
+
+    if (!input.ammAccount && (input.issue1 == ripple::noIssue() || input.issue2 == ripple::noIssue()))
```
cindyyan317, line 48: we need to check :
1 ammAccount and issue1/2 can not be set at the same time
2 ammAccount or issue1/2 must be set

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,225 @@
...
+    if (input.ammAccount) {
+        auto accountKeylet = ripple::keylet::account(*input.ammAccount);
+        auto const accountLedgerObject =
+            sharedPtrBackend_->fetchLedgerObject(accountKeylet.key, lgrInfo.seq, ctx.yield);
+        ripple::STLedgerEntry const sle{
+            ripple::SerialIter{accountLedgerObject->data(), accountLedgerObject->size()}, accountKeylet.key};
+        ammID = sle.getFieldH256(ripple::sfAMMID);
+    }
+
+    if (input.accountID) {
```
cindyyan317, line 70: I think we can move this check above the ammAccount

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,225 @@
...
+        *sharedPtrBackend_, ctx.yield, input.ledgerHash, input.ledgerIndex, range->maxSequence
+    );
+
+    if (auto const status = std::get_if<Status>(&lgrInfoOrStatus))
+        return Error{*status};
+
+    auto const lgrInfo = std::get<LedgerInfo>(lgrInfoOrStatus);
+    ripple::uint256 ammID;
+    if (input.ammAccount) {
+        auto accountKeylet = ripple::keylet::account(*input.ammAccount);
```
cindyyan317, line 62: we need to check the validity of the account , make sure 
1 it is a valid account
2 it is a valid amm account
otherwise the Clio may crash

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+    auto ammKeylet = ammID != 0 ? keylet::amm(ammID) : keylet::amm(input.issue1, input.issue2);
+    auto const ammBlob = sharedPtrBackend_->fetchLedgerObject(ammKeylet.key, lgrInfo.seq, ctx.yield);
+
+    if (not ammBlob)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND, "Amm account not found"}};
+
+    auto const amm = SLE{SerialIter{ammBlob->data(), ammBlob->size()}, ammKeylet.key};
+    auto const ammAccountID = amm.getAccountID(sfAccount);
+    auto const accBlob =
```
cindyyan317, line 84: do we need to check the existence of ammAccountID which is from the amm ledger object 🤔

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,96 @@
...
+    {
+    }
+
+    static RpcSpecConstRef
+    spec([[maybe_unused]] uint32_t apiVersion)
+    {
+        static auto const rpcSpec = RpcSpec{
+            {JS(ledger_hash), validation::Uint256HexStringValidator},
+            {JS(ledger_index), validation::LedgerIndexValidator},
+            {JS(asset), validation::ammAssetValidator},
```
cindyyan317, line 76: I understand the ammAssetValidator is from ledger_entry . You may need to bespoke the error code here.

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <data/DBHelpers.h>
+#include <rpc/AMMHelpers.h>
+#include <rpc/handlers/AMMInfo.h>
+
+#include <ripple/protocol/AMMCore.h>
+
+namespace {
+
+std::string
+toIso8601(ripple::NetClock::time_point tp)
```
cindyyan317, line 29: Can be replaced by **to_string_iso** ? u can find its usage in our codebase.

src/rpc/handlers/AMMInfo.h
arihantkothari: I removed the ammAssetValidator from ledger_entry and added it to validators. What do you mean by "bespoke the error code" ?

src/rpc/handlers/AMMInfo.cpp
arihantkothari: I wonder why they made a new method 

https://github.com/XRPLF/rippled/blob/8ce85a9750568b8f41f24e65e6103cc256a4f5b0/src/ripple/rpc/handlers/AMMInfo.cpp#L62-L71

From what I understand to_string_iso is more general. But I need another pair of eyes to look into this... I am not 100% sure yet

arihantkothari: Actually it looks like to_string_iso is new code https://github.com/XRPLF/rippled/commit/32ced493de22601b70234ba215d11e3666cc14cb

arihantkothari: Feels like it should be safe to replace...

src/rpc/handlers/AMMInfo.h
cindyyan317: Try to input invalid asset , you will see rippled raises different error for the ledger_entry and amm_info.

src/rpc/handlers/AMMInfo.cpp
cindyyan317: Looks like the format is different. Let us keep this local implementation for amm_info for now to match rippled.

arihantkothari: Okay, thanks for looking into it cindy.. I'm gonna resolve this convo for now

arihantkothari:
Some example payloads that might be useful:

```
curl -d '{
    "method": "amm_info",
    "params": [{
    "amm_account":"rLcS7XL6nxRAi7JcbJcn1Na179oF3vdfbh"
    }]
}' http://localhost:51233

curl -d '{
    "method": "amm_info",
    "params": [{
      "asset": {
        "currency": "CNY", "issuer":"rBdLS7RVLqkPwnWQCT2bC6HJd6xGoBizq8"
      },
      "asset2": {
        "currency": "USD",
        "issuer": "rBdLS7RVLqkPwnWQCT2bC6HJd6xGoBizq8"
      }
    }]
}' http://localhost:51233

curl -d '{
    "method": "amm_info",
    "params": [{
    "amm_account":"rspfN8ESWkkDLxBD9MT6fvxbytnCijnYh9"
    }]
}' http://localhost:51233
```

arihantkothari:
I have pushed the amm_info unit tests but they are really incomplete. This is the only thing that is incomplete in the PR.

godexsoft:
Closing this one in favour of #1060

PR #859, opened 2023-09-19
Add unittests for ledger publisher and bug fixes 

cindyyan317:
1 Add unittests to ledger publisher (100%)
2 Inject cache module, remove backend->cache().
3 Bug fix: Use SubscriptionManagerType; maxAttempts calculation

### DIFF
--- a/src/etl/impl/LedgerPublisher.h
+++ b/src/etl/impl/LedgerPublisher.h
@@ -22,7 +22,6 @@
 #include <data/BackendInterface.h>
 #include <etl/SystemState.h>
 #include <util/LedgerUtils.h>
-#include <util/Profiler.h>
 #include <util/log/Logger.h>
 
 #include <ripple/protocol/LedgerHeader.h>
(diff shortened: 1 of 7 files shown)


src/etl/impl/LedgerPublisher.h
```diff
@@ -97,6 +102,7 @@ class LedgerPublisher
 
             if (!range || range->maxSequence < ledgerSequence)
             {
+                ++numAttempts;
```
cindyyan317, line 105: Fix bug, when numAttemps is x, we actually accessed db x+1 times.

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+constexpr static auto LEDGERHASH = "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
+constexpr static auto SEQ = 30;
+constexpr static auto AGE = 800;
+
+class ETLLedgerPublisherTest : public MockBackendTest, public SyncAsioContextTest, public MockSubscriptionManagerTest
+{
+    void
+    SetUp() override
+    {
+        MockBackendTest::SetUp();
```
kuznetsss, line 46: Do you actually need to do this manually? Or do you need specific sequence?

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+    EXPECT_TRUE(publisher.lastPublishAgeSeconds() <= 1);
+}
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerInfoCloseTimeGreaterThanNow)
+{
+    SystemState dummyState;
+    dummyState.isWriting = true;
+    // age is -10
+    ripple::LedgerInfo dummyLedgerInfo = CreateLedgerInfo(LEDGERHASH, SEQ, 0);
+    auto const now = duration_cast<seconds>(system_clock::now().time_since_epoch());
```
kuznetsss, line 159: It would be better to fix the time point here. Tests shouldn't depend on time.

kuznetsss (changes requested):

unittests/etl/LedgerPublisherTests.cpp
cindyyan317: This is deliberately create a ledger with the future timestamp to trigger the error handling in lastCloseAgeSeconds function.

unittests/etl/LedgerPublisherTests.cpp
cindyyan317: 1 We do need to call these manually. It inherits multiple classes. 
2 Keeping the setup and teardown order does not matter for this usercase. Just my preference to keep such construction/ destruction actions symmetrical.

PR #860, opened 2023-09-19
Add unittests for ledger publisher and bug fixes

cindyyan317:
1 Add unittests to ledger publisher (100%)
2 Inject cache module, remove backend->cache().
3 Bug fix: Use SubscriptionManagerType; maxAttempts calculation

### DIFF
--- a/src/etl/impl/LedgerPublisher.h
+++ b/src/etl/impl/LedgerPublisher.h
@@ -23,7 +23,6 @@
 #include <etl/SystemState.h>
 #include <feed/SubscriptionManager.h>
 #include <util/LedgerUtils.h>
-#include <util/Profiler.h>
 #include <util/log/Logger.h>
 
 #include <ripple/protocol/LedgerHeader.h>
@@ -44,14 +43,18 @@ namespace etl::detail {
  * includes reading all of the transactions from the database) is done from the application wide asio io_service, and a
  * strand is used to ensure ledgers are published in order.
  */
-template <typename SubscriptionManagerType>
+template <typename SubscriptionManagerType, typename CacheType>
 class LedgerPublisher
 {
+    // we are still catching up if not within PUBLISH_THRESHOLD_SEC from the last close,
+    constexpr static auto PUBLISH_THRESHOLD_SEC = 600;
+
     util::Logger log_{"ETL"};
 
     boost::asio::strand<boost::asio::io_context::executor_type> publishStrand_;
 
     std::shared_ptr<BackendInterface> backend_;
+    std::reference_wrapper<CacheType> cache_;
     std::shared_ptr<SubscriptionManagerType> subscriptions_;
     std::reference_wrapper<SystemState const> state_;  // shared state for ETL
 
@@ -71,12 +74,15 @@ public:
     LedgerPublisher(
         boost::asio::io_context& ioc,
         std::shared_ptr<BackendInterface> backend,
-        std::shared_ptr<feed::SubscriptionManager> subscriptions,
+        CacheType& cache,
+        std::shared_ptr<SubscriptionManagerType> subscriptions,
         SystemState const& state)
         : publishStrand_{boost::asio::make_strand(ioc)}
         , backend_{std::move(backend)}
         , subscriptions_{std::move(subscriptions)}
-        , state_{std::cref(state)}
+        , state_{std::cref(state)},
+                 cache_{cache}
+
     {
     }
 
@@ -99,6 +105,7 @@ public:
 
             if (!range || range->maxSequence < ledgerSequence)
             {
+                ++numAttempts;
                 LOG(log_.debug()) << "Trying to publish. Could not find "
                                      "ledger with sequence = "
                                   << ledgerSequence;
@@ -110,7 +117,6 @@ public:
                     return false;
                 }
                 std::this_thread::sleep_for(std::chrono::seconds(1));
-                ++numAttempts;
                 continue;
             }
 
@@ -126,7 +132,7 @@ public:
     }
 
     /**
-     * @brief Publish the passed in ledger
+     * @brief Publish the passed ledger asynchronously.
      *
      * All ledgers are published thru publishStrand_ which ensures that all publishes are performed in a serial fashion.
      *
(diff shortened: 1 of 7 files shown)


unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerInfoCloseTimeGreaterThanNow)
+{
+    SystemState dummyState;
+    dummyState.isWriting = true;
+    // age is -10
+    ripple::LedgerInfo dummyLedgerInfo = CreateLedgerInfo(LEDGERHASH, SEQ, 0);
+    auto const now = duration_cast<seconds>(system_clock::now().time_since_epoch());
+    auto const closeTime = (now - seconds(-10)).count() - rippleEpochStart;
```
kuznetsss: Why not `now + seconds(10)`?

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+    EXPECT_TRUE(publisher.lastPublishAgeSeconds() <= 1);
+}
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerInfoCloseTimeGreaterThanNow)
+{
+    SystemState dummyState;
+    dummyState.isWriting = true;
+    // age is -10
+    ripple::LedgerInfo dummyLedgerInfo = CreateLedgerInfo(LEDGERHASH, SEQ, 0);
+    auto const now = duration_cast<seconds>(system_clock::now().time_since_epoch());
```
kuznetsss: > @[kuznetsss](https://github.com/kuznetsss) kuznetsss [3 hours ago](https://github.com/XRPLF/clio/pull/859#discussion_r1329867870)
It would be better to fix the time point here. Tests shouldn't depend on time.

> @[cindyyan317](https://github.com/cindyyan317) cindyyan317 [1 hour ago](https://github.com/XRPLF/clio/pull/859#discussion_r1330046015)
This is deliberately create a ledger with the future timestamp to trigger the error handling in lastCloseAgeSeconds function.

It should be fine here, but in general it is better to avoid changing input data in tests. I'm ok with this code, but if you want you use a far future time point to avoid `now()`:
```cpp
std::chrono::system_clock::time_point tp{std::chrono::years(200)}; // Dec 31 12:00:00 2169
```

kuznetsss (dismissed):
👍

src/etl/impl/LedgerPublisher.h
```diff
@@ -69,10 +72,12 @@ class LedgerPublisher
     LedgerPublisher(
         boost::asio::io_context& ioc,
         std::shared_ptr<BackendInterface> backend,
-        std::shared_ptr<feed::SubscriptionManager> subscriptions,
+        std::reference_wrapper<CacheType> cache,
```
godexsoft: minor: We usually pass argument by reference and then use `std::ref` or `std::cref` (see SystemState below). Consider changing for consistency 👍

src/etl/impl/LedgerPublisher.h
```diff
@@ -214,14 +219,21 @@ class LedgerPublisher
...
-        auto now = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch())
-                       .count();
-        auto closeTime = lastCloseTime_.time_since_epoch().count();
-        if (now < (rippleEpochStart + closeTime))
+        auto const now =
+            std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch())
+                .count();
+        auto const closeTime = lastCloseTime_.time_since_epoch().count();
+
+        if (auto const age = now - (rippleEpochStart + closeTime); age < 0)
```
godexsoft: Looks like you could use `std::max` here for a one liner. Up to you:
```cpp
return std::max(0, now - (rippleEpochStart + closeTime));
```

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+#include <gtest/gtest.h>
+
+#include <chrono>
+
+using namespace testing;
+using namespace etl;
+namespace json = boost::json;
+using namespace std::chrono;
+
+constexpr static auto ACCOUNT = "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn";
```
godexsoft: This probably will be fixed by clang-tidy or even newer clang-format. As far as i remember we decided to use east-const/constexpr everywhere though 🔢

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+    util::Config cfg{json::parse("{}")};
+    MockCache mockCache;
+};
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerInfoIsWritingFalse)
+{
+    SystemState dummyState;
+    dummyState.isWriting = false;
+    ripple::LedgerInfo dummyLedgerInfo = CreateLedgerInfo(LEDGERHASH, SEQ, AGE);
+    detail::LedgerPublisher publisher(ctx, mockBackendPtr, std::ref(mockCache), mockSubscriptionManagerPtr, dummyState);
```
godexsoft: See, now you have to write `std::ref` here. Let's change the constructor so it takes a regular reference and hides this `std::ref` stuff 👍

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+    ctx.run();
+    EXPECT_FALSE(rawBackendPtr->fetchLedgerRange());
+}
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerInfoInRange)
+{
+    SystemState dummyState;
+    dummyState.isWriting = true;
+    // age is 0
```
godexsoft: When you have tiny one line comments like this, consider putting them on the same line with the code - it looks strange when there is a comment line between code without any spacing unless it's on the same line as the code it comments on 🙇

unittests/util/MockCache.h
```diff
@@ -25,7 +25,14 @@
 
 struct MockCache
 {
-    MOCK_METHOD(void, update, (std::vector<data::LedgerObject> const& a, uint32_t b, bool c), ());
+    MOCK_METHOD(void, updateImp, (std::vector<data::LedgerObject> const& a, uint32_t b, bool c), ());
+
+    // mock update has a parameter with default value
```
godexsoft: This comment does not seem very useful as we can clearly see it from the code. Consider removing 👍

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,247 @@
...
+    EXPECT_FALSE(publisher.publish(SEQ, {}));
+}
+
+TEST_F(ETLLedgerPublisherTest, PublishLedgerSeqMaxAttampt)
+{
+    SystemState dummyState;
+    dummyState.isStopping = false;
+    detail::LedgerPublisher publisher(ctx, mockBackendPtr, std::ref(mockCache), mockSubscriptionManagerPtr, dummyState);
+
+    static constexpr auto MAX_ATTEMPT = 2;
```
godexsoft: Our constexpr is all over the place 😃 Now constexpr got one place closer to the right spot but still not quite there 🤣 
```cpp
static auto constexpr ...
```

Probably will be fixed by clang-tidy/clang-format 👍

godexsoft (changes requested):
Adding a few small comments, all minor stuff ⭐

src/etl/impl/LedgerPublisher.h
cindyyan317: I tried this way . but if std::max < uint > the result will be wrong, if std::max < int > , we will update the return type.

godexsoft (dismissed):

kuznetsss (dismissed):

godexsoft (dismissed):

kuznetsss (approved):

Issue #848, opened 2023-09-13
Move to a newer clang format


kuznetsss:
We are using quite outdated clang-format-11. E.g. clang-format 14 allows to force east const https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifieralignment

godexsoft:
Hopefully also concepts code can be formatted properly now? and some lambdas in our codebase could benefit from better formatting too if we can now tweak it.

kuznetsss:
Try to adjust config to have formatting for
```cpp
if
     <one line code>
ellse
   <one line code>
``` 
statements without `{ }`

kuznetsss:
Force east const style

kuznetsss:
https://github.com/XRPLF/clio/pull/864#discussion_r1331573273

kuznetsss:
https://github.com/XRPLF/clio/pull/864#discussion_r1331587916
https://github.com/XRPLF/clio/pull/864#discussion_r1331595691

PR #864, opened 2023-09-20
Add clang tidy

### LINKED ISSUE #863: Add clang-tidy checks
Add clang-tidy checks for the code. Fix everything. Add clang-tidy to CI.

kuznetsss:
Fixes #863


src/data/BackendInterface.cpp
```diff
@@ -67,16 +67,18 @@ BackendInterface::fetchLedgerObject(
         LOG(gLog.trace()) << "Cache hit - " << ripple::strHex(key);
         return *obj;
     }
+
+    LOG(gLog.trace()) << "Cache miss - " << ripple::strHex(key);
+    auto dbObj = doFetchLedgerObject(key, sequence, yield);
+    if (!dbObj)
+    {
```
godexsoft, line 74: I'm not a huge fan of always forcing curlies. Many of the times, if the statement under the if/else is a legitimate one-liner it's nice to be able to omit curlies. Is there no setting to force only for when the statement is so long it has to be split into multiple lines?

src/data/BackendInterface.cpp
```diff
@@ -181,11 +191,12 @@ BackendInterface::fetchBookOffers(
         while (keys.size() < limit)
         {
             ++numPages;
-            ripple::STLedgerEntry sle{ripple::SerialIter{offerDir->blob.data(), offerDir->blob.size()}, offerDir->key};
+            ripple::STLedgerEntry const sle{
```
godexsoft, line 194: Is there maybe a setting that would force and automatically update to something like 
```cpp
auto const sle = ripple::STLedgerEntry{...};
```

src/data/BackendInterface.h
```diff
@@ -48,6 +48,7 @@ class DatabaseTimeout : public std::exception
     }
 };
 
+static constexpr std::size_t DEFAULT_WAIT_BETWEEN_RETRY = 500;
```
godexsoft, line 51: Can we have the same eastern rules for `const` and `constexpr`?
```cpp
static std::size_t constexpr 
```

src/data/BackendInterface.cpp
godexsoft: I also noticed down the line that only `if`+`else` are forced but one-liner `if` are still allowed without curlies - maybe it's fine then. I would still prefer to force it based on length of the longest statement in the if/else. i.e. if any of the statements exceeds maximum line-length.

src/data/cassandra/SettingsProvider.cpp
```diff
@@ -135,11 +138,11 @@ SettingsProvider::parseSettings() const
 
     auto const connectTimeoutSecond = config_.maybeValue<uint32_t>("connect_timeout");
     if (connectTimeoutSecond)
-        settings.connectionTimeout = std::chrono::milliseconds{*connectTimeoutSecond * 1000};
+        settings.connectionTimeout = std::chrono::milliseconds{*connectTimeoutSecond * util::MILISECONDS_PER_SECOND};
```
godexsoft: Should be `MILLISECONDS_PER_SECOND` 🔢

src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -305,15 +302,14 @@ class DefaultExecutionStrategy
                 init, token, boost::asio::get_associated_executor(token));
             --numReadRequestsOutstanding_;
 
-            if (auto res = future->get(); res)
+            auto res = future->get();
+            if (res)
             {
```
godexsoft: The curlies should not be here for this one, right? I thought this is automatically forced/converted..

src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -417,18 +413,18 @@ class DefaultExecutionStrategy
...
         {
             // mutex lock required to prevent race condition around spurious
             // wakeup
-            std::lock_guard lck(throttleMutex_);
+            std::lock_guard const lck(throttleMutex_);
             throttleCv_.notify_one();
         }
         if (cur == 0)
         {
             // mutex lock required to prevent race condition around spurious
```
godexsoft, line 407: Is there a way to automatically fix up multi-line comments that can now fit on one line? There is multiple in the codebase since we moved from 80 character to 120 character limit.

src/etl/ETLService.cpp
```diff
@@ -20,6 +20,9 @@
 #include <etl/ETLService.h>
 
 #include <ripple/protocol/LedgerHeader.h>
+#include <util/Constants.h>
```
godexsoft: This should be in the top block - local to Clio code

src/rpc/common/MetaProcessors.h
```diff
@@ -100,27 +102,28 @@ class IfType final
...
-            std::optional<Status> firstFailure = std::nullopt;
+        : processor_(
+              [... r = std::forward<Requirements>(
+                   requirements)](boost::json::value& j, std::string_view key) -> MaybeError {
+                  std::optional<Status> firstFailure = std::nullopt;
 
-            // the check logic is the same as fieldspec
-            // clang-format off
+                  // the check logic is the same as fieldspec
+                  // clang-format off
```
godexsoft, line 111: If this new version of clang-format produces nice output for fold expressions then let's remove `// clang-format off/on` here and allow clang-format to do its thing 👍

CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ option (tests     "Build tests"                                       FALSE)
 option (docs      "Generate doxygen docs"                             FALSE)
 option (coverage  "Build test coverage report"                        FALSE)
 option (packaging "Create distribution packages"                      FALSE)
+option (clang_tidy "Run clang-tidy checks during compilation"         FALSE)
```
godexsoft: Perhaps `lint` would be a better name?

src/util/Taggable.h
```diff
@@ -226,15 +226,18 @@ class TagDecoratorFactory final
...
+
+        if (boost::iequals(style, "uuid"))
             return TagDecoratorFactory::Type::UUID;
-        else
-            throw std::runtime_error(
-                "Could not parse `log_tag_style`: expected `uint`, `uuid` or "
-                "`null`");
+
+        throw std::runtime_error(
+            "Could not parse `log_tag_style`: expected `uint`, `uuid` or "
```
godexsoft, line 238: Can fit on one line.. is there some way to automate this?

src/rpc/common/MetaProcessors.h
godexsoft: There is also a bunch of `// clang-format off` in `Concept.h` files - i'd love to see if clang-format now can make them look good as well 🥇

godexsoft (changes requested):
Overall this is a MASSIVE improvement - great job!
I am leaving a few suggestions/questions. Let's see if we can improve even more. If not possible - i'm happy with most of the proposed changes 👍

CMakeLists.txt
godexsoft: Also the string is misaligned compared to others 😢

conanfile.py
```diff
@@ -16,6 +16,7 @@ class Clio(ConanFile):
         'docs': [True, False],      # doxygen API docs; create custom target 'docs'
         'packaging': [True, False], # create distribution packages
         'coverage': [True, False],  # build for test coverage report; create custom target `clio_tests-ccov`
+        'clang_tidy': [True, False], # run clang-tidy checks during compilation
```
godexsoft: Comment misaligned a little bit

CMake/ClangTidy.cmake
```diff
@@ -0,0 +1,31 @@
...
+    if (NOT _CLANG_TIDY_PATH)
+      message (FATAL_ERROR
+        "clang-tidy binary not found. Please set the CLIO_CLANG_TIDY_BIN environment variable or install clang-tidy.")
+    endif ()
+  endfunction ()
+
+  # Support for https://github.com/matus-chochlik/ctcache
+  find_program (CLANG_TIDY_CACHE_PATH NAMES "clang-tidy-cache")
+  if (CLANG_TIDY_CACHE_PATH)
+    find_clang_tidy (_CLANG_TIDY_PATH)
```
cindyyan317: find_clang_tidy can run outside the if statement? Do we need this variable :CLANG_TIDY_PATH ?

CMake/ClangTidy.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+  function (find_clang_tidy _CLANG_TIDY_PATH)
+    if (DEFINED ENV{CLIO_CLANG_TIDY_BIN})
+      set (_CLANG_TIDY_PATH $ENV{CLIO_CLANG_TIDY_BIN})
+      if (NOT EXISTS ${_CLANG_TIDY_PATH})
+        message (FATAL_ERROR "$ENV{CLIO_CLANG_TIDY_BIN} no such file. Check CLIO_CLANG_TIDY_BIN env variable")
+      endif ()
+      message (STATUS "Using clang-tidy from CLIO_CLANG_TIDY_BIN")
+    else ()
+      find_program (_CLANG_TIDY_PATH NAMES "clang-tidy-16" "clang-tidy-15" "clang-tidy-14" "clang-tidy" REQUIRED)
```
cindyyan317: find_program is supposed to create the var to carry the result. not sure if existing variable can reflect the result correctly? 
I wonder if you change this line to "      find_program (_CLANG_TIDY_PATH NAMES "bb" "cc" "dd" "ee" REQUIRED)
",  if (NOT _CLANG_TIDY_PATH) still is false

cindyyan317: The below snippet is to try to carry the update from the argument, but it fails for me 
```  function (mmm TMPMM)
  set(TMPMM, "123")
  endfunction ()

  mmm(TMPMM)  
  message(print "${TMPMM}")
```
I might misunderstand this part.

src/data/CassandraBackend.h
```diff
@@ -288,7 +288,8 @@ class BasicCassandraBackend : public BackendInterface
     std::optional<LedgerRange>
     hardFetchLedgerRange(boost::asio::yield_context yield) const override
     {
-        if (auto const res = executor_.read(yield, schema_->selectLedgerRange); res)
+        auto const res = executor_.read(yield, schema_->selectLedgerRange);
```
cindyyan317, line 291: Curious why clang-tidy does not like this line :?

src/data/CassandraBackend.h
```diff
@@ -715,7 +713,7 @@ class BasicCassandraBackend : public BackendInterface
                 std::back_inserter(statements),
                 [this, &record](auto&& account) {
                     return schema_->insertAccountTx.bind(
-                        std::move(account),
+                        std::forward<decltype(account)>(account),
```
cindyyan317, line 716: Good one 👍

src/data/cassandra/impl/Cluster.cpp
```diff
@@ -207,7 +215,7 @@ Cluster::setupCertificate(Settings const& settings)
         return;
 
     LOG(log_.debug()) << "Configure SSL context";
-    SslContext context = SslContext(*settings.certificate);
+    SslContext const context = SslContext(*settings.certificate);
```
cindyyan317, line 141: why not auto ?

src/data/BackendInterface.cpp
kuznetsss: We could try to adjust it in #848

src/data/BackendInterface.cpp
kuznetsss: We could try to use auto in the next clang-tidy iteration. There is an option for that
https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/use-auto.html

src/data/BackendInterface.h
kuznetsss: It will be done in https://github.com/XRPLF/clio/issues/848

src/data/cassandra/impl/ExecutionStrategy.h
kuznetsss: Yeah. I assume clang-tidy doesn't warn about extra bracket but checks that there are brackets for multiline blocks.

src/data/cassandra/impl/ExecutionStrategy.h
kuznetsss: Will try in #848

src/rpc/common/MetaProcessors.h
kuznetsss: Will try in #848

src/util/Taggable.h
kuznetsss: Will try in #848

CMake/ClangTidy.cmake
kuznetsss: > find_clang_tidy can run outside the if statement?

It turns out that functions in cmake has always global scope, so I will rename it to` _find_clang_tidy()`.
But it wouldn't be not defined if the option `lint` is false.

> Do we need this variable :CLANG_TIDY_PATH ?

`_CLANG_TIDY_PATH` is needed to merge it with arguments for clang-tidy on the line 29.
But there is also `CLANG_TIDY_PATH` which is a bug.

I have refactored this part a lot, so it should be better now.

CMake/ClangTidy.cmake
kuznetsss: Now it should be better

src/data/CassandraBackend.h
kuznetsss: There is `else` after `return` below

src/data/cassandra/impl/Cluster.cpp
kuznetsss: It automatically puts `const` where it is possible. For better auto there is an option:
https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/use-auto.html

We will turn it on at the next clang-tidy iteration to not bring too many changes.

src/data/cassandra/impl/Result.h
```diff
@@ -86,23 +86,23 @@ extractColumn(CassRow const* row, std::size_t idx)
...
         output = std::string{value, len};
     }
     else if constexpr (std::is_same_v<DecayedType, bool>)
     {
-        cass_bool_t flag;
+        cass_bool_t flag = cass_bool_t::cass_false;
         auto const rc = cass_value_get_bool(cass_row_get_column(row, idx), &flag);
         throwErrorIfNeeded(rc, "Extract bool");
-        output = flag ? true : false;
+        output = flag != 0;
```
cindyyan317: Since flag is cass_bool_t, maybe compare with cass_bool_t::cass_false is more readable.

src/feed/SubscriptionManager.cpp
```diff
@@ -348,18 +355,18 @@ SubscriptionManager::subscribeHelper(
...
 void
 SubscriptionManager::cleanup(SessionPtrType session)
 {
-    std::scoped_lock lk(cleanupMtx_);
+    std::scoped_lock const lk(cleanupMtx_);
     if (!cleanupFuncs_.contains(session))
         return;
 
-    for (auto f : cleanupFuncs_[session])
+    for (const auto& f : cleanupFuncs_[session])
```
cindyyan317: auto const&

src/util/Taggable.h
```diff
@@ -226,15 +226,18 @@ class TagDecoratorFactory final
         auto const& style = value.as_string();
 
         if (boost::iequals(style, "int") || boost::iequals(style, "uint"))
+        {
```
cindyyan317: curious why this if needs brackets?

src/util/log/Logger.cpp
```diff
@@ -52,27 +52,28 @@ tag_invoke(boost::json::value_to_tag<Severity>, boost::json::value const& value)
     auto const& logLevel = value.as_string();
 
     if (boost::iequals(logLevel, "trace"))
+    {
```
cindyyan317: is the first if any special? 👁️‍🗨️

src/util/Taggable.h
kuznetsss: It actually doesn't need. It was automatically fixing brackets first when there was `else` but when I was removing `else` after `return`, I missed  deleting some brackets. I will fix all the places you found.

CMake/Settings.cmake
```diff
@@ -25,15 +25,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
   )
 endif ()
 
-if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
kuznetsss, line 28: Have to remove this because it causes error for clang-tidy

godexsoft: Hm, that's annoying 😢 Maybe we can keep them if `lint` is not enabled?

godexsoft (dismissed):

godexsoft:
Approved 👍  Feel free to ignore my comment if it's not worth.

CMake/Settings.cmake
cindyyan317: Curious what the error is ?

kuznetsss:
```
error: unknown warning option '-Wduplicated-branches' [clang-diagnostic-unknown-warning-option]
error: unknown warning option '-Wduplicated-cond' [clang-diagnostic-unknown-warning-option]
error: unknown warning option '-Wlogical-op'; did you mean '-Wlong-long'? [clang-diagnostic-unknown-warning-option]
error: unknown warning option '-Wuseless-cast' [clang-diagnostic-unknown-warning-option]
```
https://github.com/XRPLF/clio/actions/runs/6383690629/job/17324827601#step:11:430

cindyyan317 (approved):

PR #867, opened 2023-09-21
CI improvements

### LINKED ISSUE #865: Improve CI
There are a few problems with CI now:
- clang 16 is using on mac while it should be clang 14
- ccache could be used at least on mac to speed up recompilation
- redundant failing steps on linux
- it would be nice to use ccache and conan cache on linux

kuznetsss:
Fixes #865, #869

Improvements:
- move duplicated code into actions
- use ccache to speed up compilation
- isolate build on mac (don't use home directory)
- generate conan profile on each run
- cache `.conan/data` 
- cache `.ccache`

Doubts:
- there is still duplicated code on the top level. I tried to create an action from it, but log doesn't look nice after that
- ~~key for conan data cache is hash from conan file. It may bring troubles because it is using only the dependencies from the first level~~
- this PR overrides docker image for linux. But conan's docker image is build on ubuntu 20.04 while packages binaries from artifactory are built using ubuntu 22.04.


.github/workflows/build.yml
```diff
@@ -18,41 +18,54 @@ jobs:
...
+            conan profile update settings.compiler=clang $CONAN_PROFILE
+            conan profile update settings.compiler.version=14 $CONAN_PROFILE
+            conan profile update "conf.tools.build:compiler_executables={\"c\": \"$clang_path\", \"cpp\": \"$clang_cxx_path\"}" $CONAN_PROFILE
+            conan profile update env.CC="$clang_path" $CONAN_PROFILE
+            conan profile update env.CXX="$clang_cxx_path" $CONAN_PROFILE
+
+      - name: Restore conan cache
+        uses: actions/cache/restore@v3
+        with:
+          path: path/to/dependencies
```
godexsoft: Looks like some sort of placeholder path?

kuznetsss: Fixed, thanks

.github/actions/save_cache/action.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Cleanup conan directory from extra data
+      shell: bash
+      run: |
+        conan remove "*" -s -b -f
+
+    - name: Save conan cache
+      uses: actions/cache/save@v3
+      with:
+        path: ${{ inputs.conan_dir }}/data
+        key: clio-conan_data-${{ runner.os }}-develop-${{ hashFiles('./conanfile.py') }}
```
cindyyan317: i am thinking if there is any chance , the runner.os does not change and conanfile.py does not change, but the cache is outdated? for example, the compiler gets updated? 
Do we need to also consider to add build mode (debug/release)?

kuznetsss: We can always manually delete cache here: https://github.com/XRPLF/clio/actions/caches
We have only release builds now, we could change key later if needed later. Cache miss costs only about 1 hour of CI work

legleux: The Conan packages' hashes encapsulate the compiler version so that should invalidate the cache right?

kuznetsss: If we update compiler, conan will not use packages in local cache. But with the current implementation because `conanfile.py` didn't change, CI will not upload a new cache to the github storage.

I found a better solution:
```bash
conan info . -j info.json
cat info.json | jq '.[] | "\(.id), \(.package_revision)"' | shasum -a 256 -
```
This will take id and revision from every conan package used and calculate sha256 from it. So if we change compiler packages revisions will be different and the sha256 hash will be different.

godexsoft (approved):
LGTM 👍

Issue #845, opened 2023-09-12
Code coverage not accurate due to LOG


cindyyan317:
We have macro LOG() to avoid the extra evaluation if log is off or log level does not match , which gain performance win👍 . However, small side effect of it is code coverage not accurate due to the log line might not be executed.

![Screenshot 2023-09-12 at 10 01 08](https://github.com/XRPLF/clio/assets/120398799/84a2add5-8fd3-4e91-88c7-6315392fac16)

Need to update the fixture .

godexsoft:
Actually, this is a good side effect. We should not count logging towards code coverage, wdyt?

PR #868, opened 2023-09-21
Fix coverage

cindyyan317:
This PR is for fixing https://github.com/XRPLF/clio/issues/845 
 LOG macro causes the coverage result being inaccurate. Even the code is tested fully, but it will still skip the log in the unittest.
I think for unittest, we'd better also cover the evaluation part. 
In this PR, I define different implementation for LOG according to COVERAGE_ENABLED. That will require to disable one unittest.
Let us discuss here, welcome new idea 🤙

### DIFF
--- a/unittests/LoggerTests.cpp
+++ b/unittests/LoggerTests.cpp
@@ -57,6 +57,7 @@ TEST_F(LoggerTest, Filtering)
     checkEqual("Trace:TRC Trace line logged for 'Trace' component");
 }
 
+#ifndef COVERAGE_ENABLED
 TEST_F(LoggerTest, LOGMacro)
 {
     Logger log{"General"};
@@ -73,6 +74,7 @@ TEST_F(LoggerTest, LOGMacro)
     log.trace() << compute();
     EXPECT_TRUE(computeCalled);
 }
+#endif
 
 TEST_F(NoLoggerTest, Basic)
 {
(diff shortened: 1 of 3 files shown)


unittests/LoggerTests.cpp
```diff
@@ -57,6 +57,7 @@ TEST_F(LoggerTest, Filtering)
     checkEqual("Trace:TRC Trace line logged for 'Trace' component");
 }
 
+#ifndef COVERAGE_ENABLED
```
kuznetsss, line 60: You could make it a bit more compact:
```cpp
#ifndef COVERAGE_ENABLED
TEST_F(LoggerTest, DISABLED_LOGMacro)
#else
TEST_F(LoggerTest, LOGMacro)
#endif
{
...
```
but current version is also fine

kuznetsss (commented):
I don't know how coverage works but the solution looks simple and nice 👍 
Just one idea how disable test in a different way

unittests/LoggerTests.cpp
godexsoft: Both solutions are fine. Keep in mind that `DISABLED_*` will still make the test compile but not run. Given how small the test is it does not really matter too much which way you decide to do it in the end.

godexsoft (approved):

PR #872, opened 2023-09-25
Add tests for characters in currency validator

### LINKED ISSUE #835: Accept all valid currency codes in API
https://github.com/XRPLF/rippled/pull/4566

kuznetsss:
Fixes #835 

We are using xrpl's function to check currency code is valid so there is no need to change our code.
I added more test cases to be sure that clio supports characters added in xrpl.


cindyyan317 (approved):
👏  Thanks .
LGTM

godexsoft (approved):
Looks fine. I assume it's checking [this](https://xrpl.org/currency-formats.html#currency-codes)

PR #880, opened 2023-10-03
Fixes broken counters for broken pipe connections

godexsoft:
When the client breaks the connection, Clio does not always cleanup correctly. This extra bit of code makes sure we call the handler's error function that cleans up the subscriptions associated with this connection.
Ideally this would be done from the destructor but `shared_from_this` is too late to call from the destructor (and illegal). If there is a better place for this - let me know.


src/web/impl/WsBase.h
```diff
@@ -121,6 +121,10 @@ class WsBase : public ConnectionBase, public std::enable_shared_from_this<WsBase
     void
     maybeSendNext()
     {
+        // cleanup if needed. can't do this in destructor so it's here
+        if (derived().dead())
```
kuznetsss: Why `derived()`? Isn't `dead()` in `ConnectionBase`?

kuznetsss (changes requested):

src/web/impl/WsBase.h
godexsoft: Agreed

kuznetsss (approved):
👍

cindyyan317 (approved):

PR #882, opened 2023-10-03
Fix trans order of subscription transactions stream

### LINKED ISSUE #833: Transaction events delivered out of order
If you subscribe to transaction events with https://xrpl.org/subscribe.html#transaction-streams, they are often delivered out of order. The `ledger_index` is always incrementing. But within each `ledger_index`, the value `Meta.TransactionIndex` is not necessarily incrementing. The events seems to be delivered in somewhat random order. The rippled server always delivers the transaction events in the order they are applied to the ledger.

cindyyan317:
Fix #833 
The transactions should be in order pushing to subscribers.
Sort the transactions in order with transaction index.

### DIFF
--- a/src/etl/impl/LedgerPublisher.h
+++ b/src/etl/impl/LedgerPublisher.h
@@ -174,7 +174,7 @@ public:
 
                 subscriptions_->pubLedger(lgrInfo, *fees, range, transactions.size());
 
-                // order via transaction index
+                // order with transaction index
                 std::sort(transactions.begin(), transactions.end(), [](auto const& t1, auto const& t2) {
                     ripple::SerialIter iter1{t1.metadata.data(), t1.metadata.size()};
                     ripple::STObject const object1(iter1, ripple::sfMetadata);
--- a/src/etl/impl/LedgerPublisher.h
+++ b/src/etl/impl/LedgerPublisher.h
@@ -163,7 +163,7 @@ public:
                     [&](auto yield) { return backend_->fetchFees(lgrInfo.seq, yield); });
                 assert(fees);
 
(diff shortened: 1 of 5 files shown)


src/etl/impl/LedgerPublisher.h
```diff
@@ -174,7 +174,17 @@ class LedgerPublisher
 
                 subscriptions_->pubLedger(lgrInfo, *fees, range, transactions.size());
 
-                for (auto const& txAndMeta : transactions)
+                // order with transaction index
+                std::sort(transactions.begin(), transactions.end(), [](auto const& t1, auto const& t2) {
```
godexsoft, line 178: Prefer `std::begin`, `std::end`, `std::size` and `std::data` instead of the member functions 🥇

unittests/etl/LedgerPublisherTests.cpp
```diff
@@ -247,3 +247,58 @@ TEST_F(ETLLedgerPublisherTest, PublishLedgerSeqStopIsFalse)
     EXPECT_TRUE(publisher.publish(SEQ, {}));
     ctx.run();
 }
+
+TEST_F(ETLLedgerPublisherTest, PublishMultipleTxInOrder)
+{
+    SystemState dummyState;
+    dummyState.isWriting = true;
```
godexsoft, line 254: Maybe this is a good place to use designated initializers: `auto dummyState = SystemState{.isWriting = true};`

kuznetsss (approved):
LGTM

godexsoft (approved):
Few minor suggestions. Otherwise all good 👍

unittests/etl/LedgerPublisherTests.cpp
cindyyan317: I will keep these in mind and fix it later

PR #892, opened 2023-10-05
Ctid draft

cindyyan317:


src/etl/ETLService.h
```diff
@@ -205,6 +206,34 @@ class ETLService
...
+    /**
+     * @brief Get the network id of ETL lazily
+     */
+    std::optional<uint32_t>
+    getNetworkID() const
+    {
+        static std::optional<uint32_t> networkID;
+
+        if (networkID)
+            return *networkID;
```
kuznetsss: Just `networkID`, because you are returning optional

src/rpc/RPCHelpers.cpp
```diff
@@ -1387,4 +1395,18 @@ isAmendmentEnabled(
     return std::find(listAmendments.begin(), listAmendments.end(), amendmentId) != listAmendments.end();
 }
 
+std::optional<std::string>
+encodeCTID(uint32_t ledgerSeq, uint16_t txnIndex, uint16_t networkId) noexcept
+{
+    if (ledgerSeq > 0x0FFF'FFFF)
+        return {};
+
+    uint64_t const ctidValue = ((0xC000'0000ULL + static_cast<uint64_t>(ledgerSeq)) << 32) +
```
kuznetsss, line 1404: It would be good to name the constants `0x...` here and leave a comment or link why we are doing exactly this operation

src/rpc/RPCHelpers.h
```diff
@@ -225,6 +229,49 @@ isAmendmentEnabled(
...
+std::optional<std::string>
+encodeCTID(uint32_t ledgerSeq, uint16_t txnIndex, uint16_t networkId) noexcept;
+
+template <typename T>
+inline std::optional<std::tuple<uint32_t, uint16_t, uint16_t>>
+decodeCTID(T const ctid) noexcept
+{
+    auto const getCTID64 = [](T const ctid) noexcept -> std::optional<uint64_t> {
+        if constexpr (
+            std::is_same_v<T, std::string> || std::is_same_v<T, char*> || std::is_same_v<T, const char*> ||
```
kuznetsss: Why didn't it work as `std::is_convertible_v<T, std::string>`?

src/rpc/RPCHelpers.h
```diff
@@ -225,6 +229,49 @@ isAmendmentEnabled(
...
+    };
+
+    auto const ctidValue = getCTID64(ctid).value_or(0);
+
+    if ((ctidValue & 0xF000'0000'0000'0000ULL) != 0xC000'0000'0000'0000ULL)
+        return {};
+
+    uint32_t const ledgerSeq = (ctidValue >> 32) & 0xFFFF'FFFUL;
+    uint16_t const txnIndex = (ctidValue >> 16) & 0xFFFFU;
+    uint16_t const networkId = ctidValue & 0xFFFFU;
```
kuznetsss, line 269: Also would be nice to have some comment or link about this if it is available

src/rpc/handlers/AccountTx.h
```diff
@@ -130,17 +132,285 @@ class AccountTxHandler
         return apiVersion == 1 ? rpcSpecForV1 : rpcSpec;
     }
 
+    // TODO: this is currently very similar to nft_history but its own copy for time
+    // being. we should aim to reuse common logic in some way in the future.
     Result
-    process(Input input, Context const& ctx) const;
+    process(BaseAccountTxHandler::Input input, Context const& ctx) const
+    {
+        auto const range = sharedPtrBackend_->fetchLedgerRange();
```
kuznetsss, line 140: `assert(range.has_value())`? Would be useful for #893

src/rpc/handlers/AccountTx.h
```diff
@@ -130,17 +132,285 @@ class AccountTxHandler
         return apiVersion == 1 ? rpcSpecForV1 : rpcSpec;
     }
 
+    // TODO: this is currently very similar to nft_history but its own copy for time
```
kuznetsss, line 135: Also looks very similar to the logic in `AccountTX`. Maybe we should create an issue for refactoring and add the link to the `TODO`?

src/rpc/handlers/TransactionEntry.h
```diff
@@ -77,14 +77,85 @@ class TransactionEntryHandler
     }
 
     Result
-    process(Input input, Context const& ctx) const;
+    process(Input input, Context const& ctx) const
+    {
+        auto const range = sharedPtrBackend_->fetchLedgerRange();
```
kuznetsss, line 82: Also maybe add `assert(range.has_value())`? Or maybe we should create an issue to change it everywhere in the code later?

src/rpc/handlers/Tx.h
```diff
@@ -24,23 +24,24 @@
 #include <rpc/common/Types.h>
 #include <rpc/common/Validators.h>
 
+namespace etl {
+class ETLService;
```
kuznetsss, line 28: It seems it is being used only at the line 245. Maybe moved closer to the usage? But also maybe better to include `ETLService.h`? Because I a user of this handler would want to include only the handler itself, but know they have to include both `ETLService.h` and `Tx.h`.

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -307,3 +307,35 @@ TEST_F(RPCHelpersTest, TraverseOwnedNodesWithUnexistingIndexMarker)
     });
     ctx.run();
 }
+
+TEST_F(RPCHelpersTest, EncodeCTID)
+{
+    auto const ctid = encodeCTID(0x1234, 0x67, 0x89);
+    EXPECT_TRUE(ctid);
```
kuznetsss, line 314: Better `ASSERT_TRUE()` to stop the test and don't have UB in the next line

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -355,15 +357,13 @@ TEST_P(AccountTxParameterTest, CheckParams)
     auto const testBundle = GetParam();
     auto* rawBackendPtr = dynamic_cast<MockBackend*>(mockBackendPtr.get());
     ASSERT_NE(rawBackendPtr, nullptr);
-    std::cout << "Before parse" << std::endl;
```
kuznetsss, line 358: Sorry for that 🙈

unittests/rpc/handlers/AccountTxTests.cpp
```diff
@@ -464,8 +464,12 @@ TEST_F(RPCAccountTxHandlerTest, IndexSpecificForwardTrue)
             testing::_))
         .Times(1);
 
+    auto const rawETLPtr = dynamic_cast<MockETLService*>(mockETLServicePtr.get());
```
kuznetsss, line 467: `ASSERT_NE(rawETLPtr)`, but clang-tidy should catch it 🙂

kuznetsss (changes requested):
All good, just a few ideas of some improvements

src/rpc/RPCHelpers.cpp
cindyyan317: This snippet is from rippled

src/rpc/handlers/AccountTx.h
cindyyan317: This part of code just move from cpp to h. Yes, we can create an issue for this TODO.

unittests/rpc/handlers/AccountTxTests.cpp
cindyyan317: weird . it does not catch for me

src/rpc/handlers/TransactionEntry.h
cindyyan317: This is from the existing cpp. All handlers should be like this. I suggest to create a separate issue for this.

PR #904, opened 2023-10-09
Fix account_tx response both both ledger range and ledger index/hash are specified

cindyyan317:
1 For the request which specifies both ledger range and ledger index/hash, itself is an invalid request.
V2 will return error for it.
We mimic rippled's behavior to take the range instead of the ledger index/hash.

2 Remove account_tx inLedger field when binary is true

### DIFF
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -109,8 +109,9 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
             return Error{Status{RippledError::rpcINVALID_PARAMS, "containsLedgerSpecifierAndRange"}};
         }
         else if (!input.ledgerIndexMax && !input.ledgerIndexMin)
-        {  // mimic rippled, when both range and index specified, respect the range.
-           // take ledger from ledgerHash or ledgerIndex only when range is not specified
+        {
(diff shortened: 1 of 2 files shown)


kuznetsss (dismissed):
LGTM 👍

src/rpc/handlers/AccountTx.cpp
```diff
@@ -105,15 +105,20 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
     if (input.ledgerHash || input.ledgerIndex || input.usingValidatedLedger)
     {
         if (ctx.apiVersion > 1u && (input.ledgerIndexMax || input.ledgerIndexMin))
+        {
             return Error{Status{RippledError::rpcINVALID_PARAMS, "containsLedgerSpecifierAndRange"}};
+        }
+        else if (!input.ledgerIndexMax && !input.ledgerIndexMin)
+        {  // mimic rippled, when both range and index specified, respect the range.
```
godexsoft: Hm, i'd hope `clang-format` would fix this automatically - comments should be on a separate line from `{` and aligned with the start of `auto const lgrInfoOrStatus` line.

godexsoft (changes requested):
Looks good other than the misaligned comment 👍

godexsoft (approved):

kuznetsss (approved):

PR #905, opened 2023-10-09
Fix error "readability-else-after-return"

cindyyan317:
Fix CI failure:

/__w/clio/clio/build/../src/rpc/handlers/AccountTx.cpp:111:9: error: do not use 'else' after 'return' [readability-else-after-return,-warnings-as-errors]

### DIFF
--- a/src/rpc/handlers/AccountTx.cpp
+++ b/src/rpc/handlers/AccountTx.cpp
@@ -108,7 +108,8 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
         {
             return Error{Status{RippledError::rpcINVALID_PARAMS, "containsLedgerSpecifierAndRange"}};
         }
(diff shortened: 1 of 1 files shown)


src/rpc/handlers/AccountTx.cpp
```diff
@@ -108,7 +108,8 @@ AccountTxHandler::process(AccountTxHandler::Input input, Context const& ctx) con
         {
             return Error{Status{RippledError::rpcINVALID_PARAMS, "containsLedgerSpecifierAndRange"}};
```
godexsoft, line 109: I'd remove the curlies now that it's a separate one-liner `if` 👍

godexsoft (approved):

src/rpc/handlers/AccountTx.cpp
cindyyan317: Let me do it in my next PR

PR #907, opened 2023-10-09
ctid for tx

cindyyan317:
1 Add the logic to fetch the network id when clio is starting

2 Add error check for:
   - no etl node cfg
   - etl node not available
   - network id of etl nodes not the same
   - 
3 search tx by ctid; return ctid for tx


src/etl/Source.h
```diff
@@ -339,13 +337,17 @@ class SourceImpl : public Source
...
             // https://github.com/ripple/rippled/blob/develop/cfg/rippled-example.cfg
-            ws->set_option(websocket::stream_base::decorator([&clientIp](websocket::request_type& req) {
-                req.set(http::field::user_agent, std::string(BOOST_BEAST_VERSION_STRING) + " websocket-client-coro");
-                req.set(http::field::forwarded, "for=" + clientIp);
-            }));
+            if (clientIp)
+            {
+                ws->set_option(websocket::stream_base::decorator([&clientIp](websocket::request_type& req) {
+                    req.set(
+                        http::field::user_agent, std::string(BOOST_BEAST_VERSION_STRING) + " websocket-client-coro");
```
kuznetsss: Why don't we need to set user agent for the request if client ip is unknown?

src/rpc/handlers/Tx.h
```diff
@@ -45,45 +43,218 @@ class TxHandler
...
+
+        if (!input.ctid && !input.transaction)  // at least one identifier must be supplied
+            return Error{Status{RippledError::rpcINVALID_PARAMS}};
+
+        static auto constexpr maxLedgerRange = 1000u;
+        auto const rangeSupplied = input.minLedger && input.maxLedger;
+
+        if (rangeSupplied)
+        {
+            if (*input.minLedger > *input.maxLedger)
```
kuznetsss, line 97: Maybe `input->minLedger`?
But it is a minor thing, feel free to ignore

unittests/etl/ETLStateTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+class ETLStateTest : public NoLoggerFixture
+{
+};
+
+TEST_F(ETLStateTest, Error)
+{
+    MockSource const source;
+    ON_CALL(source, forwardToRippled).WillByDefault(Return(std::nullopt));
+    EXPECT_CALL(source, forwardToRippled).Times(1);
```
kuznetsss: ` EXPECT_CALL(source, forwardToRippled).WillOnce(Return(std::nullopt));`

src/etl/ETLState.cpp
```diff
@@ -0,0 +1,45 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+    Permission to use, copy, modify, and distribute this software for any
```
godexsoft, line 6: Looks like this header lost a few spaces?

src/etl/ETLState.h
```diff
@@ -0,0 +1,43 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft, line 4: Same issue with spaces

src/etl/ETLState.cpp
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    ETLState state;
+    if (serverInfoRippled && !serverInfoRippled->contains(JS(error)))
+    {
+        if (serverInfoRippled->contains(JS(result)) && serverInfoRippled->at(JS(result)).as_object().contains(JS(info)))
+        {
+            auto const rippledInfo = serverInfoRippled->at(JS(result)).as_object().at(JS(info)).as_object();
+            if (rippledInfo.contains(JS(network_id)))
+                state.networkID.emplace(boost::json::value_to<int64_t>(rippledInfo.at(JS(network_id))));
```
godexsoft, line 52: We usually use `tag_invoke` to convert between json and our structs but this is fine for such a small parser.

src/etl/ETLState.h
```diff
@@ -0,0 +1,43 @@
...
+#include <cstdint>
+#include <optional>
+
+namespace etl {
+
+class Source;
+
+/**
+ * @brief This class is responsible for fetching and storing the state of the ETL information, such as the network id
+ *
```
godexsoft: Extra space line not needed

src/etl/ETLState.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+struct ETLState
+{
+    std::optional<uint32_t> networkID;
+
+    /**
+     * @brief Fetch the ETL state from the rippled server
+     *
```
godexsoft: Extra space line not needed

src/etl/LoadBalancer.cpp
```diff
@@ -87,9 +87,33 @@ LoadBalancer::LoadBalancer(
...
         std::unique_ptr<Source> source = make_Source(entry, ioc, backend, subscriptions, validatedLedgers, *this);
 
+        // checking etl node validity
+        auto const state = ETLState::fetchETLStateFromSource(*source);
+
+        if (!state.networkID)
+        {
+            LOG(log_.error()) << "Failed to fetch ETL state from source = " << source->toString()
+                              << " Please check the configuration and network";
+            throw std::runtime_error("ETL node not available");
```
godexsoft: Where is this caught if it happens?

godexsoft: Do all these `runtime_error`s bubble up to main?

src/etl/LoadBalancer.cpp
```diff
@@ -245,4 +269,12 @@ LoadBalancer::execute(Func f, uint32_t ledgerSequence)
     }
     return true;
 }
+
+ETLState
+LoadBalancer::getETLState() const
```
godexsoft: The other one was `noexcept`, why not this one? 
Keep in mind - in the future, we want to replace `assert` with our custom asserts that will throw at runtime.

src/etl/LoadBalancer.h
```diff
@@ -164,13 +166,21 @@ class LoadBalancer
...
+     * @param clientIp The IP address of the peer, if known.
      * @param yield The coroutine context
      * @return Response received from rippled node as JSON object on success; nullopt on failure
      */
     std::optional<boost::json::object>
-    forwardToRippled(boost::json::object const& request, std::string const& clientIp, boost::asio::yield_context yield)
-        const;
+    forwardToRippled(
+        boost::json::object const& request,
+        std::optional<std::string> clientIp,
```
godexsoft: Do we need to copy the string inside the optional here?

src/etl/ProbingSource.cpp
```diff
@@ -123,18 +123,24 @@ ProbingSource::fetchLedger(uint32_t sequence, bool getObjects, bool getObjectNei
...
-    std::string const& clientIp,
+    std::optional<std::string> clientIp,
     boost::asio::yield_context yield) const
 {
-    if (!currentSrc_)
-        return {};
+    if (!currentSrc_)  // Source may connect to rippled before the connection built to check the validity
+    {
+        if (auto res = plainSrc_->forwardToRippled(request, clientIp, yield))
+        {
```
godexsoft: I think no need for the curlies here?

unittests/etl/ETLStateTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+    auto const state = etl::ETLState::fetchETLStateFromSource(source);
+    EXPECT_FALSE(state.networkID.has_value());
+}
+
+TEST_F(ETLStateTest, NetworkIdValid)
+{
+    MockSource const source;
```
kuznetsss: This line is duplicated in each test. I would suggest to make it a field of `ETLStateTest`. For each `TEST_F` gtest creates a separate `ETLStateTest` object so tests will still be independent.

src/etl/Source.h
```diff
@@ -303,18 +306,13 @@ class SourceImpl : public Source
     std::optional<boost::json::object>
     requestFromRippled(
         boost::json::object const& request,
-        std::string const& clientIp,
+        std::optional<std::string> clientIp,
         boost::asio::yield_context yield) const override
     {
         LOG(log_.trace()) << "Attempting to forward request to tx. "
                           << "request = " << boost::json::serialize(request);
```
godexsoft: Let's collapse these small strings into one bigger one. I think this entire thing might fit on one line.

src/etl/Source.h
godexsoft: Not necessarily for this PR but we could change "websocket-client-coro" to something like "clio-[version]".

src/etl/Source.h
```diff
@@ -339,13 +337,17 @@ class SourceImpl : public Source
...
-            ws->set_option(websocket::stream_base::decorator([&clientIp](websocket::request_type& req) {
-                req.set(http::field::user_agent, std::string(BOOST_BEAST_VERSION_STRING) + " websocket-client-coro");
-                req.set(http::field::forwarded, "for=" + clientIp);
-            }));
+            if (clientIp)
+            {
+                ws->set_option(websocket::stream_base::decorator([&clientIp](websocket::request_type& req) {
+                    req.set(
+                        http::field::user_agent, std::string(BOOST_BEAST_VERSION_STRING) + " websocket-client-coro");
+                    req.set(http::field::forwarded, "for=" + *clientIp);
```
godexsoft, line 347: Also, maybe we should not set `clientIp` if unknown, but still set the `user_agent`?

src/rpc/RPCHelpers.cpp
```diff
@@ -218,9 +218,34 @@ toExpandedJson(data::TransactionAndMetadata const& blobs, NFTokenjson nftEnabled
...
+        }
+    }
+
     return {txnJson, metaJson};
 }
 
+std::optional<std::string>
+encodeCTID(uint32_t ledgerSeq, uint16_t txnIndex, uint16_t networkId) noexcept
+{
+    static constexpr uint32_t MAX_LEDGER_SEQ = 0x0FFF'FFFF;
```
godexsoft: We were trying to put `constexpr` after the type, like `const`. It's minor tho

unittests/rpc/handlers/TxTests.cpp
```diff
@@ -247,8 +303,12 @@ TEST_F(RPCTxTest, DefaultParameter_API_v2)
     ON_CALL(*rawBackendPtr, fetchTransaction(ripple::uint256{TXNID}, _)).WillByDefault(Return(tx));
     EXPECT_CALL(*rawBackendPtr, fetchTransaction).Times(1);
 
+    auto const rawETLPtr = dynamic_cast<MockETLService*>(mockETLServicePtr.get());
```
kuznetsss, line 300: `ASSERT_NE(rawETLPtr, nullptr);`

src/rpc/RPCHelpers.cpp
```diff
@@ -218,9 +218,34 @@ toExpandedJson(data::TransactionAndMetadata const& blobs, NFTokenjson nftEnabled
...
+    if (ledgerSeq > MAX_LEDGER_SEQ)
+        return {};
+
+    static constexpr uint64_t CTID_PREFIX = 0xC000'0000;
+    uint64_t const ctidValue =
+        ((CTID_PREFIX + static_cast<uint64_t>(ledgerSeq)) << 32) + (static_cast<uint64_t>(txnIndex) << 16) + networkId;
+
+    std::stringstream buffer;
+    buffer << std::hex << std::uppercase << std::setfill('0') << std::setw(16) << ctidValue;
```
godexsoft: This can probably be done more efficiently and nicely with `fmt::format`. If not easy to find then the current approach is fine too

src/rpc/RPCHelpers.h
```diff
@@ -62,8 +63,17 @@ deserializeTxPlusMeta(data::TransactionAndMetadata const& blobs);
 std::pair<std::shared_ptr<ripple::STTx const>, std::shared_ptr<ripple::TxMeta const>>
 deserializeTxPlusMeta(data::TransactionAndMetadata const& blobs, std::uint32_t seq);
 
+/**
+ *@brief Convert a TransactionAndMetadata to two JSON objects.
```
godexsoft: *[SPACE]@

src/rpc/RPCHelpers.h
```diff
@@ -225,6 +235,49 @@ isAmendmentEnabled(
...
+                return {};
+
+            if (!boost::regex_match(ctidString, boost::regex("^[0-9A-F]+$")))
+                return {};
+
+            return std::stoull(ctidString, nullptr, 16);
+        }
+
+        if constexpr (std::is_same_v<T, uint64_t>)
+        {
```
godexsoft: No curlies needed

unittests/util/MockSource.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include <etl/Source.h>
+
+#include <gmock/gmock.h>
+
+class MockSource : public etl::Source
+{
+public:
+    MOCK_METHOD(bool, isConnected, (), (const, override));
+    MOCK_METHOD(boost::json::object, toJson, (), (const override));
```
kuznetsss, line 29: There should be a comma between `const` and `override` for consistency

kuznetsss (changes requested):
All good, but some comments how to make tests a bit nicer

godexsoft (changes requested):
Leaving a few comments 👍

src/etl/Source.h
cindyyan317: This parameter is to notify the request's real owner. Rippled has the protection to avoid overwhelm the server. When client ip is not known, it is to say the request is from clio server itself. Usually clio server is in the secure_gateway cfg.

src/rpc/handlers/Tx.h
cindyyan317: minLedger and maxLedger are optional, input is not optional.

src/etl/ETLState.cpp
cindyyan317: laser eye

src/etl/LoadBalancer.cpp
cindyyan317: yes. and the clio will end. I think there is no need to continue if ETL cfg is wrong? WDYT?

src/etl/LoadBalancer.h
```diff
@@ -164,13 +166,21 @@ class LoadBalancer
      * @brief Forward a JSON RPC request to a randomly selected rippled node.
      *
      * @param request JSON-RPC request to forward
-     * @param clientIp The IP address of the peer
+     * @param clientIp The IP address of the peer, if known.
```
kuznetsss: Minor, feel free to ignore. There is no `.` in the other lines.

src/rpc/RPCHelpers.cpp
cindyyan317: ha . i saw other places are like that. I thought it is our new style

src/etl/ETLState.h
```diff
@@ -0,0 +1,48 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
kuznetsss: As I understand the year here should be the year when the file was created, so `2023`

src/rpc/handlers/Tx.h
kuznetsss: Yeah, sorry

unittests/rpc/handlers/TxTests.cpp
```diff
@@ -125,8 +128,14 @@ TEST_F(RPCTxTest, TxnNotFound)
     ON_CALL(*rawBackendPtr, fetchTransaction(ripple::uint256{TXNID}, _))
         .WillByDefault(Return(std::optional<TransactionAndMetadata>{}));
     EXPECT_CALL(*rawBackendPtr, fetchTransaction).Times(1);
+
+    auto const rawETLPtr = dynamic_cast<MockETLService*>(mockETLServicePtr.get());
+    ASSERT_NE(rawETLPtr, nullptr);
+    ON_CALL(*rawETLPtr, getETLState).WillByDefault(Return(etl::ETLState{}));
+    EXPECT_CALL(*rawETLPtr, getETLState).Times(1);
```
kuznetsss: Could you please also change two lines to ` EXPECT_CALL(...).WillOnce(Return(...))` in this file (or at least in the added lines)?

unittests/etl/ETLStateTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include <gtest/gtest.h>
+#include <util/Fixtures.h>
+
+namespace json = boost::json;
+using namespace util;
+using namespace testing;
+
+struct ETLStateTest : public NoLoggerFixture
+{
+    MockSource source;
```
kuznetsss: You could still make it `const`

unittests/etl/ETLStateTests.cpp
```diff
@@ -0,0 +1,74 @@
...
+        })JSON");
+    EXPECT_CALL(source, forwardToRippled).WillOnce(Return(json.as_object()));
+    auto const state = etl::ETLState::fetchETLStateFromSource(source);
+    ASSERT_TRUE(state.networkID.has_value());
+    EXPECT_EQ(state.networkID.value(), 12);
+}
+
+TEST_F(ETLStateTest, NetworkIdInvalid)
+{
+    MockSource const source;
```
kuznetsss: This should be deleted now

kuznetsss (changes requested):

src/etl/LoadBalancer.cpp
godexsoft: Yes, that's good. I'm only thinking that we should probably use `logic_error` instead since this is a `misconfiguration`.

src/rpc/RPCHelpers.cpp
```diff
@@ -738,7 +761,7 @@ keypairFromRequst(boost::json::object const& request)
 
     // All of the secret types we allow, but only one at a time.
     // The array should be constexpr, but that makes Visual Studio unhappy.
-    static std::string const secretTypes[]{"passphrase", "secret", "seed", "seed_hex"};
+    static const std::string secretTypes[]{"passphrase", "secret", "seed", "seed_hex"};
```
godexsoft: Something went wrong here 🤣

godexsoft (approved):
🚀

kuznetsss (approved):

PR #908, opened 2023-10-09
Feature/update clang format 16

godexsoft:
Test clang-format-16 on CI


src/etl/impl/AsyncData.h
```diff
@@ -94,8 +95,8 @@ class AsyncCallData
         }
         if (!next_->is_unlimited())
         {
-            LOG(log_.warn()) << "AsyncCallData is_unlimited is false. Make sure "
-                                "secure_gateway is set correctly at the ETL source";
+            LOG(log_.warn()
```
godexsoft: Hm, what happened here?

src/data/cassandra/impl/Cluster.cpp
```diff
@@ -50,14 +50,15 @@ Cluster::Cluster(Settings const& settings) : ManagedObject{cass_cluster_new(), c
     cass_cluster_set_token_aware_routing(*this, cass_true);
     if (auto const rc = cass_cluster_set_protocol_version(*this, CASS_PROTOCOL_VERSION_V4); rc != CASS_OK)
```
kuznetsss: Don't we want to to put `{` at the same line for loops and ifs. It is controlled by the option:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforebraces
I like `BS_linux` and `BS_WebKit`. WDYT?

src/data/cassandra/impl/Cluster.cpp
```diff
@@ -50,14 +50,15 @@ Cluster::Cluster(Settings const& settings) : ManagedObject{cass_cluster_new(), c
...
     {
-        throw std::runtime_error(
-            fmt::format("Error setting cassandra protocol version to v4: {}", cass_error_desc(rc)));
+        throw std::runtime_error(fmt::format("Error setting cassandra protocol version to v4: {}", cass_error_desc(rc))
```
kuznetsss, line 51: That's a bit weird formatting. I don't know whether it could be fixed.

kuznetsss (commented):
Code looks really good, just a few questions

src/data/cassandra/impl/Cluster.cpp
godexsoft: We could set a higher line length limit but i'm not sure it's worth it.

cindyyan317 (approved):

src/data/cassandra/impl/Cluster.cpp
kuznetsss: It is a bug of clang-format
llvm/llvm-project#55473

kuznetsss (approved):

PR #912, opened 2023-10-09
Add db counts

### LINKED ISSUE #911: Add statistics of database usage
It would be useful to have counts for db usage(pending read/write, completed read/write, etc.).

kuznetsss:
Fixes #911


src/data/BackendCounters.cpp
```diff
@@ -0,0 +1,150 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
godexsoft: 2023 :)

src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -118,7 +126,10 @@ class DefaultExecutionStrategy
     bool
     isTooBusy() const
     {
-        return numReadRequestsOutstanding_ >= maxReadRequestsOutstanding_;
+        const bool result = numReadRequestsOutstanding_ >= maxReadRequestsOutstanding_;
```
cindyyan317: bool const?

cindyyan317: I am also thinking if numReadRequestsOutstanding_ can be replaced by counter now? it should be equal to the pendings?

cindyyan317 (commented):
I wonder if we can remove the previous counter in strategy , it seems like can be replaced by our new counter

src/data/BackendCounters.h
```diff
@@ -0,0 +1,136 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2022, the clio developers.
```
godexsoft: 2023

src/data/BackendCounters.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+    assert(pendingCounter_ >= count);
+    pendingCounter_ -= count;
+    errorCounter_ += count;
+}
+
+boost::json::object
+BackendCounters::AsyncOperationCounters::report() const
+{
+    boost::json::object result;
```
godexsoft: Minor: When there is no conditionally added fields, we prefer to use the `initializer_list` constructor instead

src/data/BackendCounters.h
```diff
@@ -0,0 +1,136 @@
...
+// clang-format on
+
+/**
+ * @brief Holds statistics about the backend.
+ * @note This class is thread-safe.
+ */
+class BackendCounters
+{
+public:
+    using Ptr = std::shared_ptr<BackendCounters>;
```
godexsoft: When it's a Type, prefer to append `Type`. e.g. `PtrType`

src/data/cassandra/impl/AsyncExecutor.h
```diff
@@ -68,24 +70,36 @@ class AsyncExecutor : public std::enable_shared_from_this<AsyncExecutor<Statemen
...
     }
 
 private:
-    AsyncExecutor(boost::asio::io_context& ioc, StatementType&& data, CallbackType&& onComplete)
-        : data_{std::move(data)}, retryPolicy_{ioc}, onComplete_{std::move(onComplete)}
+    AsyncExecutor(
+        boost::asio::io_context& ioc,
+        StatementType&& data,
+        CallbackType&& onComplete,
+        RetryCallbackType onRetry)
```
godexsoft: Should we also copy-move `onComplete`? or make both rvalue ref

src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -230,6 +260,7 @@ class DefaultExecutionStrategy
         auto const numStatements = statements.size();
         std::optional<FutureWithCallbackType> future;
 
+        counters_->registerReadStarted(numStatements);
```
godexsoft, line 262: I'd prefer to put this on line 262 and add a spacer between the comment - like you do in the next read function :)

src/data/cassandra/impl/ExecutionStrategy.h
kuznetsss: They are the same counter but for different purposes. I feel like maybe we shouldn't mix the functionality, but I'm not sure. @godexsoft, WDYT?

src/rpc/handlers/ServerInfo.h
```diff
@@ -49,16 +49,24 @@ namespace rpc {
...
 public:
+    struct Input
+    {
+        bool backend_counters = false;
+    };
+
     struct AdminSection
     {
         boost::json::object counters = {};
+        boost::json::object backend_counters = {};
```
godexsoft: camelCase?

src/rpc/handlers/ServerInfo.h
```diff
@@ -223,6 +244,10 @@ class BaseServerInfoHandler
             jv.as_object()["etl"] = info.adminSection->etl;
             jv.as_object()[JS(counters)] = info.adminSection->counters;
             jv.as_object()[JS(counters)].as_object()["subscriptions"] = info.adminSection->subscriptions;
+            if (!info.adminSection->backend_counters.empty())
```
godexsoft: Don't we want to return zeroes at the start? chances are low but what if a script will expect the values there but at the very start there would be none?

unittests/data/BackendCountersTests.cpp
```diff
@@ -0,0 +1,186 @@
...
+            "write_async_pending": 0,
+            "write_async_completed": 0,
+            "write_async_retry": 0,
+            "write_async_error": 0,
+            "read_async_pending": 0,
+            "read_async_completed": 0,
+            "read_async_retry": 0,
+            "read_async_error": 0
+        }
+    )")
```
godexsoft: Something odd happened to the formatting here

kuznetsss:
> I wonder if we can remove the previous counter in strategy , it seems like can be replaced by our new counter

I think we can't, because the old counter contains a map method-to-counters and methods to register rpc related events

src/data/cassandra/impl/ExecutionStrategy.h
godexsoft: I'd not mix it for now. Later we can consider it again but for now let's use the counters for analytics - what they were designed for 👍

godexsoft (changes requested):
Leaving a few small minor things. Overall looks great! 👍

src/rpc/handlers/ServerInfo.h
kuznetsss: What do you mean? Here we are serialising `InfoSection` into json and checking whether `InfoSection` has backend counters because it will have it only if there was `backend_counters: true` parameter in request. I could make backendCounters `std::optional<boost::json::object>` to better express the logic here.

unittests/data/BackendCountersTests.cpp
kuznetsss: Because `.as_object()` is on the next line? For some reason clang-tidy does it 🤷‍♂️

godexsoft: I think the closing paren is misaligned, maybe that's why `.as_object()` is also moved.. idk

src/data/BackendCounters.h
godexsoft: Still 2022 here? :)

godexsoft (approved):
🚀

PR #913, opened 2023-10-09
Fix http params handling discrepancy 

### LINKED ISSUE #909: Not throw 400 error when params is empty
Some APIs don't need any parameter. We should allow params is empty in this case.

Request:
```
{"method":"server_info","params":[]}
{"method":"server_info","params":[null]}
{"method":"server_info","params": ""}
{"method":"server_info","params": null}
{"method":"server_info","params": {}}
```
Clio returns 400 error
Rippled returns response

godexsoft:
Fixes #909

### DIFF
--- a/src/web/RPCServerHandler.h
+++ b/src/web/RPCServerHandler.h
@@ -89,8 +89,8 @@ public:
             auto req = boost::json::parse(request).as_object();
             LOG(perfLog_.debug()) << connection->tag() << "Adding to work queue";
 
-            if (not connection->upgraded and not req.contains("params"))
-                req["params"] = boost::json::array({boost::json::object{}});
+            if (not connection->upgraded && shouldReplaceParams(req))
+                req[JS(params)] = boost::json::array({boost::json::object{}});
 
             if (!rpcEngine_->post(
                     [this, request = std::move(req), connection](boost::asio::yield_context yield) mutable {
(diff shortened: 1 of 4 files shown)


unittests/web/RPCServerHandlerTests.cpp
```diff
@@ -549,7 +549,7 @@ TEST_F(WebRPCServerHandlerTest, HTTPParamsUnparseableNotArray)
     EXPECT_EQ(session->lastStatus, boost::beast::http::status::bad_request);
 }
 
-TEST_F(WebRPCServerHandlerTest, HTTPParamsUnparseableEmptyArray)
+TEST_F(WebRPCServerHandlerTest, HTTPParamsUnparseableArrayWithDigit)
```
godexsoft, line 552: These should probably be made into a parameterised test and we check `[]`, `null`, `[{}]`, `[""]`, `""`, `null` and `{}` as success cases, `1` and `[1]` as failures.

src/web/RPCServerHandler.h
```diff
@@ -89,8 +89,8 @@ class RPCServerHandler
             auto req = boost::json::parse(request).as_object();
             LOG(perfLog_.debug()) << connection->tag() << "Adding to work queue";
 
-            if (not connection->upgraded and not req.contains("params"))
-                req["params"] = boost::json::array({boost::json::object{}});
+            if (not connection->upgraded && shouldReplaceParams(req))
```
cindyyan317: small thing : "and" is better here, because "not" had been used here.

cindyyan317 (dismissed):
LGTM 👍

src/web/RPCServerHandler.h
godexsoft: Agree!

PR #921, opened 2023-10-11
Fix messages pile up

cindyyan317:
wsFail is the function we called in async callback, it will release the session and cleanup the session from SubscriptionManager when the error is not **operation_aborted**.

However, when the error code is operation_aborted, we do nothing. Because individual read/write can be canceled, but the connection is still alive. 
When operation_aborted error happens in onWrite, we did not remove the previous payload, more importantly , we did not reset the **sending_** flag, which causing the messages_ can be pushed but it will never be sent again. So even the connection is dead, we don't know , because we will never call async_write.

How to cause the operation_aborted locally:  the peer calls close() after writing a chunk of data.
eg
```
        boost::beast::flat_buffer buffer;
        ws_.write(net::buffer(std::string(body)));
        ws_.close(boost::beast::websocket::close_code::normal);
```

### DIFF
--- a/src/web/impl/WsBase.h
+++ b/src/web/impl/WsBase.h
@@ -106,16 +106,12 @@ public:
     void
     onWrite(boost::system::error_code ec, std::size_t)
     {
+        messages_.pop();
+        sending_ = false;
         if (ec)
-        {
             wsFail(ec, "Failed to write");
-        }
         else
-        {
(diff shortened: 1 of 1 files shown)


kuznetsss (approved):
👍

godexsoft (approved):

PR #933, opened 2023-10-17
Add repair functionality

godexsoft:


src/main/main.cpp
```diff
@@ -173,23 +288,58 @@ doMigrationStepOne(
             txHashes.push_back(ripple::uint256::fromVoid(buf));
         }
 
-        auto const txs =
-            doTryFetchTransactions(timer, backend, txHashes, yield);
+        auto txs = doTryFetchTransactions(timer, backend, txHashes, yield);
+        assert(txs.size() == txHashes.size());
```
kuznetsss: This assert will not do anything in release. Maybe it is better to crash here instead of writing wrong data?

src/main/main.cpp
```diff
@@ -173,23 +288,58 @@ doMigrationStepOne(
...
+                if (!maybeTx.has_value())
+                {
+                    BOOST_LOG_TRIVIAL(fatal)
+                        << "Could not fetch written transaction for hash "
+                        << hash << "; Repair failed.";
+                    exit(-1);
+                }
+
+                txs[idx] = maybeTx.value();
+                --idx;  // repeat the try section for the repaired tx
```
kuznetsss, line 371: If idx = 0 here the loop will be finished

src/main/main.cpp
```diff
@@ -14,6 +18,115 @@ static std::uint32_t const MAX_RETRIES = 5;
...
+
+static void
+maybeWriteTransaction(
+    Backend::CassandraBackend& backend,
+    std::optional<boost::json::object> const& tx)
+{
+    if (!tx.has_value())
+        throw std::runtime_error("Could not repair transaction");
+
+    auto data = tx.value().at("result").as_object();
```
cindyyan317: better check the field existence  in case the tx is not fetched successfully

kuznetsss (approved):
👍

src/main/main.cpp
```diff
@@ -15,14 +19,151 @@ static std::chrono::seconds const WAIT_TIME = std::chrono::seconds(60);
...
+static void
+maybeWriteTransaction(
+    Backend::CassandraBackend& backend,
+    std::optional<boost::json::object> const& tx)
+{
+    if (!tx.has_value())
+        throw std::runtime_error("Could not repair transaction");
+
+    auto package = tx.value();
+    if (!package.contains("result") || !package.at("result").is_object())
```
cindyyan317, line 127: check the "error" field not null
```
{
  "result": {
    "error": "txnNotFound",
    "error_code": 29,
    "error_message": "Transaction not found.",
    "request": {
      "binary": true,
      "command": "tx",
      "transaction": "CCBBD2F5CEE90FA7A2ED13A7A343D015F57950B562185734E125B7BE740F2427"
    },
    "status": "error"
  }
}
```

Issue #939, opened 2023-10-18
Memory pile up caused by subscribing repeat topic


cindyyan317:
We are not checking the existing subscription when add subscription, the cleanup function list can keep increase if one session keep subscribes same topic, which we do observe such behaviour from log.
The session 1626711 subscribed 12824 accounts every minute.
![Screenshot 2023-10-18 at 11 31 31](https://github.com/XRPLF/clio/assets/120398799/a149e5f5-bbde-47d0-a8e7-8113a11ddd27)
1626711 left after:
```
 [2023-10-17 20:26:41.317036 (web/impl/WsBase.h:66) [0x00007f243cff9700] Performance:NFO [1626711] : read: End of file]
```
The increasing stoped.

https://github.com/XRPLF/clio/blob/develop/src/feed/SubscriptionManager.cpp#L291

PR #940, opened 2023-10-18
Subscribe cleanup fix

cindyyan317:
This PR adds fix for #939 
Check if the session has been already added to the subscribers list , if yes, don't push the cleanup function.


kuznetsss (approved):
👍

PR #941, opened 2023-10-18
Resume functionality

godexsoft:
- Refactor migrator to support saving and restoring the migration state.
- Refactor to use specified host:port for repair server instead of first element of the ETL sources in the Clio config.
- Use Boost.ProgramOptions to parse cli
    - resume
    - repair [host:port]
    - config (positional and -c/--conf)
    - help message

### DIFF
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -124,7 +124,8 @@ maybeWriteTransaction(
         throw std::runtime_error("Could not repair transaction");
 
     auto package = tx.value();
-    if (!package.contains("result") || !package.at("result").is_object())
+    if (!package.contains("result") || !package.at("result").is_object() ||
+        package.at("result").as_object().contains("error"))
         throw std::runtime_error("Received non-success response from rippled");
 
     auto data = package.at("result").as_object();
--- a/src/main/main.cpp
+++ b/src/main/main.cpp
@@ -9,6 +9,7 @@
 
 #include <boost/asio.hpp>
 #include <boost/json.hpp>
+#include <boost/program_options.hpp>
 #include <cassandra.h>
 
 #include <iostream>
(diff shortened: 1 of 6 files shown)


src/main/main.cpp
```diff
@@ -3,529 +3,386 @@
...
-        clio::LogService::info() << "There is no data to migrate";
-        return;
+        /*
+         * Step 3 - Drop the old `issuer_nf_tokens` table, which is replaced by
+         * `issuer_nf_tokens_v2`. Normally, we should probably not drop old
+         * tables in migrations, but here it is safe since the old table wasn't
+         * yet being used to serve any data anyway.
+         */
+        clio::LogService::info() << "Running " << tag_;
+
```
cindyyan317, line 320: don't forget to clear the resume state here

src/main/main.cpp
```diff
@@ -3,529 +3,386 @@
...
+         * found in any NFTokenPage object. Prior to this migration, we were not
+         * pulling out NFTs from the initial ledger, so all these NFTs would be
+         * missed. This will also record the URI of any NFTs minted prior to the
+         * start sequence.
+         */
+        std::vector<NFTsData> toWrite;
+        std::optional<ripple::uint256> cursor;
+
+        clio::LogService::info() << "Running " << tag_;
+        resumeProvider_.get().write({tag_, {}});  // at the start of step2
```
cindyyan317, line 249: Like it 👍

src/main/main.cpp
```diff
@@ -3,529 +3,386 @@
...
-{
-    auto const size = nfts.size();
-    if (size == 0)
-        return nfts;
-    backend.writeNFTs(std::move(nfts));
-    backend.sync();
-    clio::LogService::info() << tag << ": Wrote " << size << " records";
-    return {};
-}
+            toWrite = maybeDoNFTWrite(toWrite, backend_, tag_);
```
cindyyan317, line 180: maybeDoNFTWrite not really write to db. If it does not write through db, toWrite in memory can not be resumed.

src/main/main.cpp
```diff
@@ -3,529 +3,386 @@
...
 
-                if (not repairEnabled)
-                {
-                    clio::LogService::fatal()
-                        << "Not attempting to repair. Rerun with -repair to "
-                           "repair corrupted transactions.";
-                    exit(-1);
-                }
+            toWrite = maybeDoNFTWrite(toWrite, backend_, tag_);
+            cursor = page.cursor;
```
cindyyan317, line 279: same here, maybeDoNFTWrite may not write but cursor moves

src/main/main.cpp
godexsoft: Good catch. Let's check for `toWrite.empty()` and only save the cursor in case we actually stored the data in DB.

src/main/main.cpp
godexsoft: I did not do it deliberately - don't think it's important as step3 is tiny. If the last saved resume state was at the end of step2 then the resume operation would not take long to write the last page of step2 again and jump over to step3. But i can add it here for completeness 👍

cindyyan317 (approved):

PR #946, opened 2023-10-20
Add config to run without valid etl 

cindyyan317:
We previously added the logic to check the ETL nodes validity at startup. It caused some inconvenience of dev. 
This PR adds a config "allow_no_etl", if it is true, Clio is allowed to run without ETL. 
If the "allow_no_etl" is false, which by default, the error will be like:
Screenshot 2023-10-20 at 09 10 28

### DIFF
--- a/src/etl/ETLState.h
+++ b/src/etl/ETLState.h
@@ -19,14 +19,14 @@
 
 #pragma once
 
+#include <data/BackendInterface.h>
+
 #include <boost/json.hpp>
 
 #include <cstdint>
 #include <optional>
 
 namespace etl {
-
-class Source;
 
 /**
  * @brief This class is responsible for fetching and storing the state of the ETL information, such as the network id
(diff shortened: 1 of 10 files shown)


src/etl/ETLState.h
```diff
@@ -36,9 +36,22 @@ struct ETLState {
 
     /**
      * @brief Fetch the ETL state from the rippled server
+     * @param source The source to fetch the state from
+     * @return The ETL state, nullopt if source not available
      */
-    static ETLState
-    fetchETLStateFromSource(Source const& source) noexcept;
+    template <typename Forward>
```
kuznetsss, line 42: Why did this method become template?

kuznetsss (approved):
👍

src/etl/ETLState.h
cindyyan317: This function will be called by the single source at the beginning and by LoadBalancer at retry.

godexsoft (approved):

Issue #385, opened 2022-11-04
Add issuer_nfts API


shawnxie999:
`issuer_nfts` allows us to query for a list of NFTs that an account has issued. User can either query for a list of NFTokenIDs, or a list of expanded information showing details of each NFTokenIDs

shawnxie999:
Im working on this one

godexsoft:
@shawnxie999 any progress update on this?

shawnxie999:
> @shawnxie999 any progress update on this?

This is blocked by https://github.com/XRPLF/clio/pull/313

PR #948, opened 2023-10-24
`nfts_by_issuer`

shawnxie999:
Original PR:
https://github.com/XRPLF/clio/pull/584

Fixes issue https://github.com/XRPLF/clio/issues/385


src/data/cassandra/impl/Collection.h
```diff
@@ -0,0 +1,95 @@
...
+#include <string>
+#include <string_view>
+
+namespace data::cassandra::detail {
+
+class Collection : public ManagedObject<CassCollection> {
+    static constexpr auto deleter = [](CassCollection* ptr) { cass_collection_free(ptr); };
+
+    template <typename>
+    static constexpr bool unsupported_v = false;
```
cindyyan317: seems not used. can be removed.

src/data/cassandra/impl/Collection.h
```diff
@@ -0,0 +1,88 @@
...
+
+    void
+    throwErrorIfNeeded(CassError const rc, std::string_view const& label) const
```
cindyyan317: I think we can just pass std::string_view by value.

src/data/cassandra/impl/Collection.h
```diff
@@ -0,0 +1,95 @@
...
+
+    template <typename Type>
+    explicit Collection(std::vector<Type> const& value)
+        : ManagedObject{cass_collection_new(CASS_COLLECTION_TYPE_LIST, value.size()), deleter}
+    {
+        bind(value);
+    }
+
+    template <typename Type>
+    explicit Collection(std::vector<Type>&& value)
```
cindyyan317: QQ: since we don't really treat rvalue , lvalue differently in bind, maybe we can delete this constructor?

src/data/BackendInterface.h
```diff
@@ -290,6 +290,28 @@ class BackendInterface {
...
+     * last left off.
+     * @param yield Currently executing coroutine.
+     * @return std::vector<NFT> of NFTs issued by this account, or
+     * this issuer/taxon combination if taxon is passed and an optional marker
+     */
+    virtual NFTsAndCursor
+    fetchNFTsByIssuer(
+        ripple::AccountID const& issuer,
+        std::optional<std::uint32_t> const& taxon,
+        std::uint32_t const ledgerSequence,
```
cindyyan317: no need const.

src/data/CassandraBackend.h
```diff
@@ -441,6 +441,96 @@ class BasicCassandraBackend : public BackendInterface {
...
+
+        auto nftQueryStatement = schema_->selectNFTBulk.bind(nftIDs);
+        nftQueryStatement.bindAt(1, ledgerSequence);
+
+        // Fetch all the NFT data, meanwhile filtering out the NFTs that are not within the ledger range
+        auto const nftRes = executor_.read(yield, nftQueryStatement);
+        auto const& nftQueryResults = nftRes.value();
+
+        if (not nftQueryResults.hasRows()) {
+            LOG(log_.debug()) << "No rows returned";
```
cindyyan317, line 504: I am thinking maybe we should log error and the nft id here. Is it an error if "nf_tokens" does not contain the corresponding nftIDs nft?

src/data/CassandraBackend.h
```diff
@@ -441,6 +441,96 @@ class BasicCassandraBackend : public BackendInterface {
...
+        std::optional<std::uint32_t> const& taxon,
+        std::uint32_t const ledgerSequence,
+        std::uint32_t const limit,
+        std::optional<ripple::uint256> const& cursorIn,
+        boost::asio::yield_context yield
+    ) const override
+    {
+        NFTsAndCursor ret;
+
+        Statement idQueryStatement = [&taxon, &issuer, &cursorIn, &limit, this]() {
```
cindyyan317: Can idQuereyStatement and nftQueryStatement be const?

src/rpc/handlers/NFTsByIssuer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/protocol/nft.h>
+#include <rpc/RPCHelpers.h>
+#include <rpc/handlers/NFTsByIssuer.h>
+
+using namespace ripple;
+using namespace ::rpc;
```
cindyyan317: maybe we don't need this? 🤔

src/rpc/handlers/NFTsByIssuer.cpp
```diff
@@ -0,0 +1,130 @@
...
+    );
+    if (auto const status = std::get_if<Status>(&lgrInfoOrStatus))
+        return Error{*status};
+
+    auto const lgrInfo = std::get<LedgerInfo>(lgrInfoOrStatus);
+
+    auto const limit = input.limit.value_or(NFTsByIssuerHandler::LIMIT_DEFAULT);
+
+    auto const issuer = accountFromStringStrict(input.issuer);
+
```
cindyyan317, line 48: I can think two reasons to check the account existence here:
1 Error message is more descriptive than empty array.
2 May save a journey to db.

src/data/CassandraBackend.h
shawnxie999: > Is it an error if "nf_tokens" does not contain the corresponding nftIDs nft?

Yes, it would be an error if `nf_tokens` table does not contains the corresponding nftIDs, but that doesn't apply in this use case.

No row returned from `nftQueryResults`  isn't necessarily an error. The query `selectNFTBulk` does a _filter_ by ledger sequence on the  `nftIDs` vector that is passed in, so it's completely possible(and valid) that every entry in the vector is filtered out, and as a result, the query result is empty.

cindyyan317 (dismissed):
LGTM 🚀

src/data/CassandraBackend.h
```diff
@@ -441,6 +441,96 @@ class BasicCassandraBackend : public BackendInterface {
...
+        auto const res = executor_.read(yield, idQueryStatement);
+
+        auto const& idQueryResults = res.value();
+        if (not idQueryResults.hasRows()) {
+            LOG(log_.debug()) << "No rows returned";
+            return {};
+        }
+
+        std::vector<ripple::uint256> nftIDs;
+        for (auto const [nftID] : extract<ripple::uint256>(idQueryResults))
```
kuznetsss, line 487: Why do you need `[ ]` here?

src/data/CassandraBackend.h
```diff
@@ -441,6 +441,96 @@ class BasicCassandraBackend : public BackendInterface {
...
+        // Query for all the NFTs issued by the account, potentially filtered by the taxon
+        auto const res = executor_.read(yield, idQueryStatement);
+
+        auto const& idQueryResults = res.value();
+        if (not idQueryResults.hasRows()) {
+            LOG(log_.debug()) << "No rows returned";
+            return {};
+        }
+
+        std::vector<ripple::uint256> nftIDs;
```
kuznetsss, line 486: Maybe it is possible to avoid reallocations here?
```cpp
auto extracted = extract<ripple::uint256>(idQueryResults);
nftIDs.reserve(extracted.size());
```

unittests/rpc/handlers/NFTsByIssuerTest.cpp
```diff
@@ -0,0 +1,667 @@
...
+
+// error case ledger non exist via index
+TEST_F(RPCNFTsByIssuerHandlerTest, NonExistLedgerViaLedgerStringIndex)
+{
+    MockBackend* rawBackendPtr = dynamic_cast<MockBackend*>(mockBackendPtr.get());
+    mockBackendPtr->updateRange(10);  // min
+    mockBackendPtr->updateRange(30);  // max
+    // mock fetchLedgerBySequence return empty
+    ON_CALL(*rawBackendPtr, fetchLedgerBySequence).WillByDefault(Return(std::optional<ripple::LedgerInfo>{}));
+    EXPECT_CALL(*rawBackendPtr, fetchLedgerBySequence).Times(1);
```
kuznetsss: Could be shorter (here and below)
```cpp
EXPECT_CALL(*rawBackendPtr, fetchLedgerBySequence).WillOnce(Return(std::optional<ripple::LedgerInfo>{}));
```

unittests/rpc/handlers/NFTsByIssuerTest.cpp
```diff
@@ -0,0 +1,667 @@
...
+        ASSERT_FALSE(output);
+        auto const err = rpc::makeError(output.error());
+        EXPECT_EQ(err.at("error").as_string(), "lgrNotFound");
+        EXPECT_EQ(err.at("error_message").as_string(), "ledgerNotFound");
+    });
+}
+
+TEST_F(RPCNFTsByIssuerHandlerTest, NonExistLedgerViaLedgerIntIndex)
+{
+    MockBackend* rawBackendPtr = dynamic_cast<MockBackend*>(mockBackendPtr.get());
```
kuznetsss, line 241: It would be better (here and below) to check that it is not `nullptr`: 
```cpp
ASSERT_NE(rawBackendPtr, nullptr);
```

kuznetsss (dismissed):
All good 👍  Just a few ideas what could be improved

src/data/CassandraBackend.h
shawnxie999: I dont think so. `extract` returns a `ResultExtractor`, which does not support `size()`

src/data/CassandraBackend.h
shawnxie999: a tuple is returned from `extract`, need to use `[ ]` here

kuznetsss (approved):

shawnxie999:
@cindyyan317  @kuznetsss whats the routine for getting a PR merged? I don't have perm in this repo so i can't merge it myself. But pls feel free to merge it if you feel good about this PR.

PR #950, opened 2023-10-25
Add prometheus support

### LINKED ISSUE #888: Prometheus support
Export metrics to prometheus using https://github.com/jupp0r/prometheus-cpp

kuznetsss:
Fixes #888


godexsoft (commented):
examples misspelled 🔤

src/data/BackendCounters.h
```diff
@@ -116,16 +119,16 @@ class BackendCounters {
 
     private:
         std::string name_;
-        std::atomic_uint64_t pendingCounter_ = 0u;
-        std::atomic_uint64_t completedCounter_ = 0u;
-        std::atomic_uint64_t retryCounter_ = 0u;
-        std::atomic_uint64_t errorCounter_ = 0u;
+        util::prometheus::GaugeInt& pendingCounter_;
```
cindyyan317: we may use std::reference_wrapper here, like other places in Clio.

src/rpc/WorkQueue.h
```diff
@@ -127,13 +128,19 @@ class WorkQueue {
...
+        obj["queued"] = queued_.value();
+        obj["queued_duration_us"] = durationUs_.value();
+        obj["current_queue_size"] = curSize_.value();
         obj["max_queue_size"] = maxSize_;
 
         return obj;
     }
+
+    /**
+     * @brief Wait untill all the jobs in the queue are finished.
```
cindyyan317: 'until'

src/util/prometheus/Http.cpp
```diff
@@ -0,0 +1,61 @@
...
+bool
+isPrometheusRequest(http::request<http::string_body> const& req)
+{
+    return req.method() == http::verb::get && req.target() == "/metrics";
+}
+
+}  // namespace
+
+std::optional<http::response<http::string_body>>
+handlePrometheusRequest(http::request<http::string_body> const& req, bool const isAdmin)
```
cindyyan317, line 37: no need const?

src/util/prometheus/Label.cpp
```diff
@@ -0,0 +1,89 @@
...
+    std::string result;
+
+    if (!labels_.empty())
+        result.push_back('{');
+
+    for (auto& label : labels_) {
+        result += label.serialize();
+        result.push_back(',');
+    }
+    result.back() = '}';
```
cindyyan317: do we need to check emptiness for '}' , may cause error?

src/util/prometheus/Prometheus.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+std::string
+PrometheusImpl::collectMetrics()
+{
+    std::string result;
+
+    if (!isEnabled())
+        return result;
+
+    for (auto& [name, family] : metrics_) {
```
cindyyan317: + const

src/util/prometheus/Prometheus.cpp
```diff
@@ -0,0 +1,104 @@
...
+        getMetric(std::move(name), std::move(labels), std::move(description), MetricType::GAUGE_DOUBLE);
+    return convertBaseTo<GaugeDouble>(metricBase);
+}
+
+std::string
+PrometheusImpl::collectMetrics()
+{
+    std::string result;
+
+    if (!isEnabled())
```
cindyyan317, line 76: i am thinking maybe Prometheus does not need to know the config. The request will be rejected in webserver level anyway.

src/util/prometheus/Prometheus.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    return result;
+}
+
+MetricBase&
+PrometheusImpl::getMetric(
+    std::string name,
+    Labels labels,
+    std::optional<std::string> description,
+    MetricType const type
```
cindyyan317, line 90: const can be removed. It is not in declaration

src/util/prometheus/Prometheus.h
```diff
@@ -0,0 +1,161 @@
...
+    static PrometheusInterface&
+    instance()
+    {
+        assert(instance_);
+        return *instance_;
+    }
+
+    // Be careful with this method because there could be hanging references to counters
+    static void
+    replaceInstance(std::unique_ptr<PrometheusInterface> instance)
```
cindyyan317, line 149: This function seems only used for testing purpose.

src/util/prometheus/Metrics.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+MetricBase&
+MetricsFamily::getMetric(Labels labels)
+{
+    auto labelsString = labels.serialize();
+    auto it = metrics_.find(labelsString);
+    if (it == metrics_.end()) {
+        auto metric = metricBuilder_(name(), labelsString, type());
+        auto [it2, success] = metrics_.emplace(std::move(labelsString), std::move(metric));
```
cindyyan317, line 110: weird. why clang-tidy does not warn us about unused 'success'

unittests/util/MockPrometheus.h
```diff
@@ -0,0 +1,181 @@
...
+ */
+struct WithPrometheus : virtual ::testing::Test {
+    WithPrometheus()
+    {
+        PROMETHEUS_INIT();
+    }
+
+    ~WithPrometheus() override
+    {
+        PROMETHEUS_INIT();
```
cindyyan317: Curious why PROMETHEUS_INIT is called in destructor 🤔

unittests/util/MockPrometheus.h
```diff
@@ -0,0 +1,181 @@
...
+    {
+        PrometheusSingleton::replaceInstance(std::make_unique<MockPrometheusImpl>());
+    }
+
+    ~WithMockPrometheus() override
+    {
+        if (HasFailure()) {
+            std::cerr << "Registered metrics:\n";
+            for (auto const& [key, metric] : mockPrometheus().metrics) {
+                std::cout << key << "\n";
```
cindyyan317: std::cerr

cindyyan317:
Great work 🎉  🙌 
Left some small comments.
One thing worth to mention is:
I think we don't really need the 'enabled' config inside promethus. It should be able to be checked only in server layer. In that case, we can remove the config from promethus logic and move the 'promethus_enabled' into server , with 'admin_pwd' together.

src/util/prometheus/Prometheus.cpp
kuznetsss: `const` in declaration for a value parameter doesn't do anything. While `const` in definition is useful like any other const variable declaration. We even have clang-tidy rule to avoid const in declarations https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

unittests/util/MockPrometheus.h
kuznetsss: Because all expectations are being checked in mock destructors. Because counters are owned by static prometheus object they will live until the end of tests or until the next test reinitialising prometheus.  So without reinitialising prometheus in destructor expectations from current test could be checked in some other test.

src/util/prometheus/Prometheus.cpp
kuznetsss: I wanted to put together all the logic related to prometheus to keep server's code mostly unaffected. I think there will be at least one more option related to prometheus whether to compress reply or not. So maybe config is ok?

Ideally when we will refactor config I would want to see it as a number of structs representing objects in json. So in the future I hope prometheus will just take a struct `PrometheusConfig` which will contain only prometheus related options.

src/util/prometheus/Http.cpp
kuznetsss: Why? `const` shows here that `isAdmin` can't be changed in the body of the function

src/util/prometheus/Label.cpp
kuznetsss: Thank you! Maybe that's why tests are crashing in CI

src/util/prometheus/Prometheus.cpp
cindyyan317: ha , surprised to know declaration and definition can be different

src/util/prometheus/Metrics.cpp
kuznetsss: For some reason it counts by clang as one variable not as two separate

cindyyan317 (dismissed):

src/main/Main.cpp
```diff
@@ -156,6 +157,8 @@ try {
     LogService::init(config);
     LOG(LogService::info()) << "Clio version: " << Build::getClioFullVersionString();
 
+    PROMETHEUS_INIT(config);
```
godexsoft: Would much prefer this to be more similar to `LogService::init(config)` to keep things uniform and try avoid using macros 👍

src/rpc/Counters.h
```diff
@@ -34,28 +35,34 @@ namespace rpc {
...
-        std::uint64_t started = 0u;
-        std::uint64_t finished = 0u;
-        std::uint64_t failed = 0u;
-        std::uint64_t errored = 0u;
-        std::uint64_t forwarded = 0u;
-        std::uint64_t failedForward = 0u;
-        std::uint64_t duration = 0u;
+        MethodInfo(std::string const& method);
+
+        CounterType& started;
```
godexsoft: Reference members are discouraged. Although in this particular case perhaps there is no problem. I'd still prefer to use `std::reference_wrapper`, at least for consistency.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type

src/util/Constants.h
```diff
@@ -1,3 +1,21 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
 #pragma once
```
godexsoft, line 20: Missing space line between header and `#pragma once`

src/util/prometheus/Counter.h
```diff
@@ -0,0 +1,104 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+#pragma once
```
godexsoft, line 20: Space missing

src/util/prometheus/Gauge.h
```diff
@@ -0,0 +1,127 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+#pragma once
```
godexsoft, line 20: Space again. Probably in more places too 🔢

src/util/prometheus/Counter.h
```diff
@@ -0,0 +1,104 @@
...
+    {
+    }
+
+    /**
+     * @brief Increase the counter by one
+     */
+    AnyCounter&
+    operator++()
+    {
+        this->pimpl_->add(ValueType{1});
```
godexsoft, line 55: Do we need to use `this->` everywhere?

src/util/prometheus/Http.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (!prometheusRequest)
+        return std::nullopt;
+
+    if (!isAdmin) {
+        return http::response<http::string_body>(
+            http::status::unauthorized, req.version(), "Only admin is allowed to collect metrics"
+        );
+    }
+
+    if (!PROMETHEUS().isEnabled()) {
```
godexsoft: `not PrometheusService::isEnabled()` would read even better 👍

src/util/prometheus/Http.cpp
```diff
@@ -0,0 +1,61 @@
...
+        return http::response<http::string_body>(
+            http::status::forbidden, req.version(), "Prometheus is disabled in clio config"
+        );
+    }
+
+    auto response = http::response<http::string_body>(http::status::ok, req.version());
+    response.set(http::field::content_type, "text/plain; version=0.0.4");
```
godexsoft, line 57: Is this version not expected to ever change and is ok to be hardcoded? Just to make sure

src/util/prometheus/Label.cpp
```diff
@@ -0,0 +1,92 @@
...
+            }
+            default:
+                escapedValue.push_back(c);
+                break;
+        }
+    }
+    return fmt::format("{}=\"{}\"", name_, std::move(escapedValue));
+}
+
+Labels::Labels(std::vector<Label> labels) : labels_(std::move(labels))
```
godexsoft, line 68: Maybe a variadic of `Label...` would be a good choice here? It would make the call-site a little cleaner (one less level of curlies). e.g. `Labels{{args}, {secondArgs}}` or similar?

src/util/prometheus/Prometheus.h
```diff
@@ -0,0 +1,161 @@
...
+
+#include <util/config/Config.h>
+#include <util/prometheus/Counter.h>
+#include <util/prometheus/Gauge.h>
+
+#include <cassert>
+
+namespace util::prometheus {
+
+class PrometheusInterface {
```
godexsoft, line 30: Some doxygen would be nice here

src/util/prometheus/Prometheus.h
```diff
@@ -0,0 +1,161 @@
...
+
+private:
+    MetricBase&
+    getMetric(std::string name, Labels labels, std::optional<std::string> description, MetricType type);
+
+    std::unordered_map<std::string, MetricsFamily> metrics_;
+    friend class PrometheusSingleton;
+};
+
+class PrometheusSingleton {
```
godexsoft: Perhaps this could be called `PrometheusService` similarly to `LogService`

godexsoft: Also, doxygen missing

src/util/prometheus/Prometheus.h
```diff
@@ -0,0 +1,161 @@
...
+        instance_ = std::move(instance);
+    }
+
+private:
+    static std::unique_ptr<PrometheusInterface> instance_;
+};
+
+}  // namespace util::prometheus
+
+#define PROMETHEUS util::prometheus::PrometheusSingleton::instance
```
godexsoft: I'd prefer to avoid introducing macros. Let's try to make the API closer to `LogService`, wdyt?

src/util/prometheus/impl/AnyCounterBase.h
```diff
@@ -0,0 +1,81 @@
...
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+#pragma once
+
+#include <util/prometheus/impl/CounterImpl.h>
+
+#include <memory>
+
+namespace util::prometheus::impl {
```
godexsoft: In Clio we usually use `detail` as the namespace while `impl` is the folder with the detail code. I don't mind either of the names but would prefer to stay consistent with the rest of the codebase

src/util/prometheus/impl/CounterImpl.h
```diff
@@ -0,0 +1,97 @@
...
+#include <atomic>
+#include <cassert>
+#include <concepts>
+
+namespace util::prometheus::impl {
+
+template <typename T>
+concept SomeNumberType = std::is_arithmetic_v<T> && !std::is_same_v<T, bool> && !std::is_const_v<T>;
+
+// clang-format off
```
godexsoft: Maybe let clang-format deal with formatting this part too? We recently changed that with clang-format 16

src/util/prometheus/impl/CounterImpl.h
```diff
@@ -0,0 +1,97 @@
...
+
+    CounterImpl&
+    operator=(CounterImpl const&) = delete;
+    CounterImpl&
+    operator=(CounterImpl&&) = delete;
+
+    void
+    add(ValueType const value)
+    {
+        if constexpr (std::is_integral_v<ValueType>) {
```
godexsoft, line 70: Is this something we should have as a separate util and use everywhere we use atomics?

unittests/util/prometheus/CounterTests.cpp
```diff
@@ -0,0 +1,156 @@
...
+struct AnyCounterTests : ::testing::Test {
+    struct MockCounterImpl {
+        using ValueType = std::uint64_t;
+        MOCK_METHOD(void, add, (ValueType));
+        MOCK_METHOD(void, set, (ValueType));
+        MOCK_METHOD(ValueType, value, (), (const));
+    };
+
+    ::testing::StrictMock<MockCounterImpl> mockCounterImpl;
+    const std::string name = "test_counter";
```
godexsoft: East vs West

unittests/util/prometheus/HttpTests.cpp
```diff
@@ -0,0 +1,212 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+#include <util/prometheus/Http.h>
```
godexsoft, line 20: Space

godexsoft (changes requested):
Leaving some suggestions and questions 👍 Overall very cool 🔥

src/util/prometheus/Counter.h
kuznetsss: For some reason compiler doesn't recognise it without `this->`

src/util/prometheus/Http.cpp
kuznetsss: I saw both `!` and `not` (same for `&&`, `||` and `and`, `or`) in the codebase. Do we have a preference?

src/util/prometheus/Http.cpp
kuznetsss: It is fine to be hardcoded. It is the version of plain text response protocol of our reply.

src/util/prometheus/Http.cpp
godexsoft: There is no rule for this but my personal preference for using `not`, `or` and `and` is when it makes the statement read more like plain English.

src/util/prometheus/Label.cpp
kuznetsss: Decided to leave it as it is

src/util/prometheus/impl/CounterImpl.h
kuznetsss: Created issue for refactoring #966

godexsoft (dismissed):
Looks great! 🚀

godexsoft (approved):

PR #956, opened 2023-10-26
Docker compose and docker image upload

legleux:
Try it...
```
$ docker compose up -d
[+] Running 4/4
 ✔ clio 3 layers [⣿⣿⣿]      0B/0B      Pulled                                                                                                                           3.6s 
   ✔ e67fdae35593 Already exists                                                                                                                                        0.0s 
   ✔ 1240990d516c Pull complete                                                                                                                                         2.2s 
   ✔ 74c20035ab9d Pull complete                                                                                                                                         2.3s 
[+] Running 3/3
 ✔ Container docker-rippled-1  Healthy                                                                                                                                300.9s 
 ✔ Container cassandra         Healthy                                                                                                                                 61.9s 
 ✔ Container docker-clio-1     Started                                                                                                                                300.9s 

$ dps
NAMES              STATUS                   PORTS     IMAGE
docker-clio-1      Up 2 minutes                       rippleci/clio:develop
cassandra          Up 7 minutes (healthy)             cassandra:4.0.4
docker-rippled-1   Up 7 minutes (healthy)             rippleci/rippled:2.0.0-b3

$ docker exec cassandra bash -c "cqlsh -e 'use clio; select * from ledger_range;' "

 is_latest | sequence
-----------+----------
     False |  1072183
      True |  1072237
```


cindyyan317 (dismissed):
LGTM 🎉

docker/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM centos:7 as builder
```
kuznetsss, line 1: Why not ubuntu based or something more modern? Centos has very old package versions so if someone would need additional tools in the container they will not be so happy.

docker/cfg/config.json
```diff
@@ -0,0 +1,42 @@
...
+            "grpc_port":"50061"
+        }
+    ],
+    "dos_guard":
+    {
+        "whitelist":["127.0.0.1"]
+    },
+    "server":{
+        "ip":"127.0.0.1",
+        "port":51233
```
kuznetsss: Maybe add admin password here

docker/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: "3"
+
+services:
+  clio:
+    image: ${CLIO_IMAGE}
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+    network_mode: host
```
kuznetsss, line 11: Why network mode host for all containers? I think it is better to map ports. Also is there any point to expose cassandra to the host?

docker/env.example
```diff
@@ -0,0 +1,5 @@
+CLIO_IMAGE=rippleci/clio:develop
```
kuznetsss, line 1: I think better to put the latest release here. `develop` is not stable

docker/env.example
```diff
@@ -0,0 +1,5 @@
...
+RIPPLED_IMAGE=rippleci/rippled:2.0.0-b3
+CASSANDRA_IMAGE=cassandra:4.0.4
```
kuznetsss, line 3: Why not scylla?

docker/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+version: "3"
+
+services:
+  clio:
+    image: ${CLIO_IMAGE}
```
kuznetsss: Why is it better to have `docker-compose.yml` + `.env` instead of single `docker-compose.yml`?

docker/build_clio.sh
```diff
@@ -0,0 +1,27 @@
...
+
+pip install "conan<2"
+conan remote add --insert 0 conan-non-prod http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod || true
+# TODO: adjust --build parameters when remote is updated with binaries built from
+conan install . \
+    --build b2  \
+    --build missing \
+    --output-folder ${BUILD_DIR} \
+    --options tests=${TESTS} \
+    --options "xrpl/*:rocksdb=False" \
```
kuznetsss: Do you need it here if you added it to the conanfile?

docker/build_clio.sh
```diff
@@ -0,0 +1,27 @@
...
+source /opt/rh/rh-python38/enable
+
+pip install "conan<2"
+conan remote add --insert 0 conan-non-prod http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod || true
+# TODO: adjust --build parameters when remote is updated with binaries built from
+conan install . \
+    --build b2  \
+    --build missing \
+    --output-folder ${BUILD_DIR} \
+    --options tests=${TESTS} \
```
kuznetsss, line 27: Probably good to add `--options lint=False \`

kuznetsss (changes requested):
That's really cool to finally have docker images 👍 
I have a few questions.

kuznetsss:
I think it would be also useful to mention docker image and example of docker compose in the main README file

docker/docker-compose.yml
legleux: I don't think it adds anything really.
Definitely was very close to removing that so I will.

docker/build_clio.sh
legleux: Nope, it doesn't need to be here.

docker/docker-compose.yml
legleux: What I really would prefer would be the user bridge network so one can just use DNS names where applicable. I'm not too keen on using the host network usually but I think there's a field in rippled's config that doesn't allow for a DNS name (one in Clio too I think.)

I think quickly using host network to be simpler but I'm not going to defend that position too vehemently if you think port-mapping is more appropriate.

godexsoft:
@legleux is this going anywhere or should we close it?
cc @kuznetsss

kuznetsss:
I think it would be easier to close this PR and open a new one. Because we now have code to build and publish docker images which could be reused and we now export binaries as artefacts in CI. But whatever is more comfortable for @legleux.

legleux:
This could probably be refactored simpler now with the recent changes introduced.

PR #958, opened 2023-10-31
Admin password

### LINKED ISSUE #922: Use hash instead of plain text password
Admin password is stored in memory as plain text and client have to send password as plain text now. It is not secure so we should change it to hash.
Also add `local_admin` to config and check that either `admin_password` or `local_admin` is set.

local_admin -> only 127.0.0.1 is admin
admin_password-> only the request with valid password hash is admin

cindyyan317:
Compare hash of admin_password instead of the plain password
Fix #922

### DIFF
--- a/example-config.json
+++ b/example-config.json
@@ -66,11 +66,11 @@
         // Max number of requests to queue up before rejecting further requests.
         // Defaults to 0, which disables the limit.
         "max_queue_size": 500,
-        //if request contains header with authorization, clio will check if it matches this value's hash
-        //if matches, the request will be considered as admin request
+        // If request contains header with authorization, Clio will check if it matches this value's hash
+        // If matches, the request will be considered as admin request
         "admin_password": "xrp",
-        //if local_admin is true, clio will consider requests come from 127.0.0.1 as admin requests
-        //it's true by default unless admin_password is set,'local_admin':true and 'admin_password' can not be set at the same time
+        // If local_admin is true, Clio will consider requests come from 127.0.0.1 as admin requests
+        // It's true by default unless admin_password is set,'local_admin' : true and 'admin_password' can not be set at the same time
         "local_amdin": false
     },
     // Overrides log level on a per logging channel.
--- a/example-config.json
+++ b/example-config.json
@@ -66,7 +66,12 @@
         // Max number of requests to queue up before rejecting further requests.
         // Defaults to 0, which disables the limit.
         "max_queue_size": 500,
-        "admin_password": "xrp"
+        //if request contains header with authorization, clio will check if it matches this value's hash
+        //if matches, the request will be considered as admin request
+        "admin_password": "xrp",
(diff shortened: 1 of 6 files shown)


example-config.json
```diff
@@ -66,7 +66,12 @@
         // Max number of requests to queue up before rejecting further requests.
         // Defaults to 0, which disables the limit.
         "max_queue_size": 500,
-        "admin_password": "xrp"
+        //if request contains header with authorization, clio will check if it matches this value's hash
+        //if matches, the request will be considered as admin request
+        "admin_password": "xrp",
+        //if local_admin is true, clio will consider requests come from 127.0.0.1 as admin requests
+        //it's true by default unless admin_password is set,'local_admin':true and 'admin_password' can not be set at the same time
+        "local_amdin": false
```
cindyyan317, line 74: Maybe there is no need to add one more config "local_admin". Because local_admin must not equal with admin_password existence

example-config.json
```diff
@@ -66,7 +66,12 @@
         // Max number of requests to queue up before rejecting further requests.
         // Defaults to 0, which disables the limit.
         "max_queue_size": 500,
-        "admin_password": "xrp"
+        //if request contains header with authorization, clio will check if it matches this value's hash
```
godexsoft: Let's keep the style of comments the same - Capitalize and add spaces where applicable 🙇

src/web/Server.h
```diff
@@ -298,6 +298,20 @@ make_HttpServer(
     auto const address = boost::asio::ip::make_address(serverConfig.value<std::string>("ip"));
     auto const port = serverConfig.value<unsigned short>("port");
     auto adminPassword = serverConfig.maybeValue<std::string>("admin_password");
+    auto const localAdmin = serverConfig.maybeValue<bool>("local_admin");
+
+    // Throw config error when localAdmin is true and admin_password is also set
+    if (localAdmin && localAdmin.value() && adminPassword) {
+        LOG(log.error()) << "local_admin is true but admin_password is also set, please only specify only one method "
```
godexsoft: "please only specify only one" 🔢

src/web/impl/AdminVerificationStrategy.cpp
```diff
@@ -40,8 +47,9 @@ PasswordAdminVerificationStrategy::isAdmin(RequestType const& request, std::stri
         // No Authorization header
         return false;
     }
-
-    return it->value() == password_;
+    std::string userAuth(it->value());
+    std::transform(userAuth.begin(), userAuth.end(), userAuth.begin(), ::toupper);
```
godexsoft, line 53: Should we also have the same transform when we generate `passwordSha256_` just so we don't rely on the hasher to always produce upper-case strings?

godexsoft (changes requested):
Leaving a few small suggestions

godexsoft (approved):
LGTM 👍

PR #964, opened 2023-11-01
Fix exit of ETL on exception

godexsoft:
Fixes #708 

@cindyyan317 while this is fixing the original issue you found (i assume, when having both "admin_password" and "local_admin" set), this is not actually a full fix for any unexpected exception path. For proper clean exit we would have to refactor more code, especially ETL.


cindyyan317 (approved):

cindyyan317:
Thanks for the quick fix for ETLService part!

PR #965, opened 2023-11-01
Move AdminVerificationStrategy to Server

cindyyan317:
Following this PR #958 
This PR removes the password member from Server class. Also AdminVerificationStrategy can be shared between sessions. This PR move the construction from session to web server.

### DIFF
--- a/src/web/SslHttpSession.h
+++ b/src/web/SslHttpSession.h
@@ -56,7 +56,7 @@ public:
     explicit SslHttpSession(
         tcp::socket&& socket,
         std::string const& ip,
-        std::optional<std::string> adminPassword,
+        std::shared_ptr<detail::AdminVerificationStrategy> const& adminVerification,
         boost::asio::ssl::context& ctx,
         std::reference_wrapper<util::TagDecoratorFactory const> tagFactory,
         std::reference_wrapper<web::DOSGuard> dosGuard,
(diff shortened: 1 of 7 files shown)


src/web/impl/HttpBase.h
```diff
@@ -129,14 +129,14 @@ class HttpBase : public ConnectionBase {
     HttpBase(
         std::string const& ip,
         std::reference_wrapper<util::TagDecoratorFactory const> tagFactory,
-        std::optional<std::string> adminPassword,
+        std::shared_ptr<AdminVerificationStrategy> const& adminVerification,
```
kuznetsss: Better to take by copy and move it (will remove one reference creation)

kuznetsss (changes requested):

src/web/SslHttpSession.h
```diff
@@ -56,7 +56,7 @@ class SslHttpSession : public detail::HttpBase<SslHttpSession, HandlerType>,
     explicit SslHttpSession(
         tcp::socket&& socket,
         std::string const& ip,
-        std::optional<std::string> adminPassword,
+        std::shared_ptr<detail::AdminVerificationStrategy> const& adminVerification,
```
kuznetsss, line 59: Here also better to copy and move

cindyyan317: I hesitate to change this one, because there is no copy construction happening inside this function. I think it is okay to pass const& here, instead of move it while call other function.

kuznetsss (approved):

Issue #893, opened 2023-10-05
Improve asserts in the code


kuznetsss:
- Create a macro `ASSERT()` to log enough information and close clio
- For each `assert()` in our code:
  - replace it with the `ASSERT()` macro if it checks invariants violation
  - replace it with a proper exception throwing to recover at the right level of the application (probably at request processing level)
  - if this will be done after #888 we should report to prometheus recoverable errors

PR #969, opened 2023-11-02
Fix code inside assert

kuznetsss:


arihantkothari (approved):
Thanks for the fix.

arihantkothari:
I think you will also get this error (or warning i must say) after the changes, that's what I am getting locally...
```
/Users/ak/Dev/ak-forks/new/clio/src/data/BackendCounters.cpp:164:42: error: comparison of integers of different signs: 'ValueType' (aka 'long long') and 'const std::uint64_t' (aka 'const unsigned long long') [clang-diagnostic-sign-compare]
    assert(pendingCounter_.get().value() >= count);
```

godexsoft (approved):
Wow. Looks like it's really time to implement custom asserts that will work outside of debug builds 💯

arihantkothari:
Just for the sake of learning, I am just curious why did the compiler not pick this up when compiling on Release instead of Debug ? 
1. I get that ` assert(...); becomes void; in release ` but should asserts not be there in Release as well ?
2. Even if we have something like this, should we not make a github action for building in debug mode as well ?

arihantkothari:
Answering my own question,
1. I guess asserts are for debugging hence  it's only in Debug mode. But I guess there should be something to raise exceptions in Release mode instead of simply killing it ?

kuznetsss:
> 1. I guess asserts are for debugging hence it's only in Debug mode. But I guess there should be something to raise exceptions in Release mode instead of simply killing it ?

We have a task #893 about that.

arihantkothari:
> > 1. I guess asserts are for debugging hence it's only in Debug mode. But I guess there should be something to raise exceptions in Release mode instead of simply killing it ?
> 
> We have a task #893 about that.

Got it, thanks!

PR #971, opened 2023-11-02
[Trivial] .empty() instead of .size() for vectors

arihantkothari:
I've been getting this warning while compiling:
```
/Users/ak/Dev/ak-forks/new/clio/src/data/BackendInterface.cpp:203:16: error: the 'empty' method should be used to check for emptiness instead of 'size' [readability-container-size-empty,-warnings-as-errors]
        assert(objs[i].size());
               ^
/opt/homebrew/Cellar/llvm@16/16.0.6/bin/../include/c++/v1/vector:551:10: note: method 'vector'::empty() defined here
    bool empty() const _NOEXCEPT
```
This PR fixes this issue.


arihantkothari:
@godexsoft can you review this tiny PR when you get a chance ?

godexsoft (dismissed):
LGTM 👍

arihantkothari:
sorry, there were few more of these. Now it should be done ✅

godexsoft (approved):

PR #976, opened 2023-11-03
Fail if there is no `local_admin` or `admin_password`

kuznetsss:
I think we should force users to either explicitly add `local_admin` option if they are fine with not secure ip based admin verification or add `admin_password` option to use the new approach.


cindyyan317:
I don't see any issue of the code per se. 
 But this restriction may surprise the current users, they need to explicitly specify one config. 
 Because the currently Clio can not fully exit on exception, user might ignore the error log. 
WDYT

kuznetsss:
Yeah, probably we shouldn't merge this until clio can exit on exception.

kuznetsss:
Decided not to add breaking changes in the config.

PR #977, opened 2023-11-06
Fix lowercase ctid 

cindyyan317:
I am thinking to make lowercase convert to string util function. Now the stringutils are for unittests only. Do you think we need to move  it out from unittests?

### DIFF
--- a/src/web/impl/AdminVerificationStrategy.cpp
+++ b/src/web/impl/AdminVerificationStrategy.cpp
@@ -18,6 +18,7 @@
 //==============================================================================
 
 #include <fmt/format.h>
+#include <util/JsonUtils.h>
 #include <web/impl/AdminVerificationStrategy.h>
 
 #include <ripple/protocol/digest.h>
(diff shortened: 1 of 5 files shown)


godexsoft:
I'd rather add `toUpper` in JsonUtils which already features `toLower`. Arguably, JsonUtils is no longer the best name for the file though :) I don't think the StringUtils in unittests are very useful outside of unittests. Wdyt?

cindyyan317:
> I'd rather add `toUpper` in JsonUtils which already features `toLower`. Arguably, JsonUtils is no longer the best name for the file though :) I don't think the StringUtils in unittests are very useful outside of unittests. Wdyt?

LOL I will squeeze it to JsonUtils then. Maybe we can change the filename to a more generic name 'Utilities.cpp' later?

src/web/impl/AdminVerificationStrategy.cpp
```diff
@@ -39,7 +40,7 @@ PasswordAdminVerificationStrategy::PasswordAdminVerificationStrategy(std::string
     std::memcpy(sha256.data(), d.data(), d.size());
     passwordSha256_ = ripple::to_string(sha256);
     // make sure it's uppercase
-    std::transform(passwordSha256_.begin(), passwordSha256_.end(), passwordSha256_.begin(), ::toupper);
+    passwordSha256_ = util::toUpper(passwordSha256_);
```
kuznetsss: You could move `passwordSha256_` here

kuznetsss (approved):
👍

PR #979, opened 2023-11-09
DeliverMax alias of Payment tx

cindyyan317:
Fix #973
'Amount' maybe misleading for partial payment transaction.
-This PR add an alias for 'Amount' -> 'DeliverMax'. 
-For version 2, remove 'Amount' field
This change is only for 'Payment' transaction.

The affected API:
- tx 
- transaction_entry
- account_tx
- ledger

The remaining work:
- subscribe
 Currently subscription does not support api_version. We will solve it in separate [PR](https://github.com/XRPLF/clio/issues/978) .

Issue #978, opened 2023-11-08
Support api_version for subscribe 


cindyyan317:
Currently all the subscribers of one topic receive the same content.
The tx json remove "Amount" for version 2, we shall check api_version for subscriber too. Now it is just hardcoded to 1. see #980

cindyyan317:
Fixed at #1071

PR #980, opened 2023-11-09
"DeliverMax" alias for Payment tx in subscription streams

cindyyan317:
Following #979 
There are 3 places having payment tx json:
transaction stream
accounts stream
orderbook stream

This PR is adding "DeliverMax" there. "api_version" has not been supported for subscription, thus hardcode the version for now.
I was thinking to support api_version for subscription in this PR, but it seems like a big change. We will implement it in the upcoming redesign task.

### DIFF
--- a/src/feed/SubscriptionManager.cpp
+++ b/src/feed/SubscriptionManager.cpp
@@ -174,6 +174,8 @@ SubscriptionManager::pubTransaction(data::TransactionAndMetadata const& blobs, r
     pubObj["transaction"] = rpc::toJson(*tx);
     pubObj["meta"] = rpc::toJson(*meta);
     rpc::insertDeliveredAmount(pubObj["meta"].as_object(), tx, meta, blobs.date);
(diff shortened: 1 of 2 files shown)


src/feed/SubscriptionManager.cpp
```diff
@@ -174,6 +174,8 @@ SubscriptionManager::pubTransaction(data::TransactionAndMetadata const& blobs, r
     pubObj["transaction"] = rpc::toJson(*tx);
     pubObj["meta"] = rpc::toJson(*meta);
     rpc::insertDeliveredAmount(pubObj["meta"].as_object(), tx, meta, blobs.date);
+    // hardcode api_version to 1 for now, until https://github.com/XRPLF/clio/issues/978 fixed
```
kuznetsss, line 177: Maybe it would be useful to add a note about it in #978

kuznetsss (approved):
👍

PR #987, opened 2023-11-15
Add compression and histogram metric type for Prometheus

### LINKED ISSUE #932: Better prometheus support
- Add support for histogram and summary metric types
- Add gzip compression for replies with metrics

### LINKED ISSUE #966: Move CounterImpl into a separate util type
We created CounterImpl in `util/prometheus/impl` wrap around atomic number which has a special case for floating point atomics. It could be useful in some other places (currently we don't have any) so it is better to move it to a separate util class.

kuznetsss:
Fixes #932
Also fixes #966

Decided not to add Summary type because it has the same functionality as Histogram but makes more calculations on client side (Clio side). See https://prometheus.io/docs/practices/histograms for detailed comparison.


src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -140,10 +140,11 @@ class DefaultExecutionStrategy {
     ResultOrErrorType
     writeSync(StatementType const& statement)
     {
-        counters_->registerWriteSync();
+        auto const startTime = std::chrono::steady_clock::now();
```
cindyyan317, line 144: If we record start time here, we also include the retry duration ,right?

unittests/util/prometheus/HistogramTests.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+TEST_F(AnyHistogramTests, serializeValue)
+{
+    OStream stream{false};
+    EXPECT_CALL(mockHistogramImpl, serializeValue(name, labelsString, ::testing::_));
+    histogram.serializeValue(stream);
+}
+
+struct HistogramTests : ::testing::Test {
+    std::vector<std::int64_t> const buckets{1, 2, 3};
```
cindyyan317: not used?

src/data/cassandra/impl/ExecutionStrategy.h
kuznetsss: Yeah, duration for these counters is a total duration from start to end including retries

src/util/prometheus/Histogram.h
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @param name The name of the metric
+     * @param labelsString The labels of the metric in serialized format, e.g. {name="value",name2="value2"}
+     * @param buckets The buckets of the histogram
+     * @param impl The implementation of the histogram (has default value and need to be specified only for testing)
+     */
+    template <detail::SomeHistogramImpl ImplType = detail::HistogramImpl<ValueType>>
+        requires std::same_as<ValueType, typename std::remove_cvref_t<ImplType>::ValueType>
+    AnyHistogram(std::string name, std::string labelsString, Buckets const& buckets, ImplType&& impl = ImplType{})
+        : MetricBase(std::move(name), std::move(labelsString))
```
cindyyan317, line 48: the format must be "{name="value",name2="value2"}",right? Is there anyway compose this restriction. I saw in serializeValue function, we kinda rely on certain format.

kuznetsss: Yeah, or it could be empty string

src/util/prometheus/impl/HistogramImpl.h
```diff
@@ -0,0 +1,130 @@
...
+        sum_ += value;
+    }
+
+    void
+    serializeValue(std::string const& name, std::string labelsString, OStream& stream) const
+    {
+        if (labelsString.empty()) {
+            labelsString = "{";
+        } else {
+            assert(labelsString.size() > 1);
```
cindyyan317: no need

unittests/util/prometheus/OStreamTests.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+TEST(OStreamTests, compression)
+{
+    OStream stream{true};
+    std::string const str = "helloooooooooooooooooooooooooooooooooo";
+    stream << str;
+    auto const compressed = std::move(stream).data();
+    EXPECT_LT(compressed.size(), str.size());
+
+    const std::string decompressed = [&compressed]() {
```
cindyyan317: std::string const

src/util/prometheus/impl/HistogramImpl.h
kuznetsss: I was thinking that it could be good to say that this function can take either empty string or at least `"{}"`. But maybe it is not really helpful

unittests/util/prometheus/OStreamTests.cpp
kuznetsss: Interestring why clang-format is ok with it 🤔

unittests/util/prometheus/MetricsFamilyTests.cpp
```diff
@@ -17,66 +17,63 @@
 */
 //==============================================================================
 
-#include <util/prometheus/Counter.h>
-#include <util/prometheus/Gauge.h>
-#include <util/prometheus/Metrics.h>
+#include <util/prometheus/MetricsFamily.h>
+
+#include <fmt/format.h>
```
cindyyan317: no newline , fmt and gmock all external libraries

src/util/prometheus/Prometheus.h
```diff
@@ -200,6 +280,40 @@ class PrometheusService {
         std::optional<std::string> description = std::nullopt
     );
 
+    /**
+     * @brief Get a integer based histogram metric. It will be created if it doesn't exist
```
cindyyan317: typo: an

src/util/prometheus/Prometheus.h
```diff
@@ -76,10 +79,45 @@ class PrometheusInterface {
      * @param name The name of the metric
      * @param labels The labels of the metric
      * @param description The description of the metric
+     * @return The reference to the gauge object
      */
     virtual GaugeDouble&
     gaugeDouble(std::string name, Labels labels, std::optional<std::string> description = std::nullopt) = 0;
 
+    /**
+     * @brief Get a integer based histogram metric. It will be created if it doesn't exist
```
cindyyan317: typo : an

src/util/prometheus/Prometheus.h
```diff
@@ -99,8 +137,20 @@ class PrometheusInterface {
         return isEnabled_;
     }
 
+    /**
+     * @brief Whether to compress the reply
+     *
+     * @return true if the reply should be compressed
+     */
+    bool
+    compressReply() const
```
cindyyan317: This function name sounds like an action , WDYT

src/util/prometheus/impl/HistogramImpl.h
```diff
@@ -0,0 +1,130 @@
...
+    HistogramImpl() = default;
+    HistogramImpl(HistogramImpl const&) = delete;
+    HistogramImpl(HistogramImpl&&) = default;
+    HistogramImpl&
+    operator=(HistogramImpl const&) = delete;
+    HistogramImpl&
+    operator=(HistogramImpl&&) = default;
+
+    void
+    setBuckets(std::vector<ValueType> const& bounds)
```
cindyyan317, line 59: Curious , if the buckets can only be set once, how about set it in constructor? like AnyHistogram

cindyyan317:
👍  Great~ happy to see the compression is supported and the improvements . Left a few of QQ and discussion.

src/util/prometheus/impl/HistogramImpl.h
kuznetsss: `HistogramImpl` have to be default constructable to be able to make `impl = detail::HistogramImpl{}` in AnyHistogram's constructor.

src/util/prometheus/Histogram.h
kuznetsss: I added assertion in `serializeValue()`. I will replace it to the better `ASSERT()` when #994 is merged.

cindyyan317 (dismissed):
LGTM

src/util/prometheus/MetricBuilder.cpp
```diff
@@ -32,11 +33,11 @@ MetricBuilder::operator()(
     std::vector<std::int64_t> const& buckets
 )
 {
-    ASSERT(type != MetricType::HISTOGRAM_DOUBLE);
+    ASSERT(type != MetricType::HISTOGRAM_DOUBLE, "Wrong metric type. Probably wrong bucket type was used.");
     if (type == MetricType::HISTOGRAM_INT) {
         return makeHistogram(std::move(name), std::move(labelsString), type, buckets);
     }
-    ASSERT(buckets.empty());
+    ASSERT(buckets.empty(), "Histogram must have at least one bucket.");
```
cindyyan317: info should be "non-histogram must have empty bucket"?

cindyyan317 (dismissed):

cindyyan317 (approved):

Issue #984, opened 2023-11-14
tx with invalid binary do not replicate rippled's response


mounikakun:
Rippled and Clio should not error for invalid binary value for default api_version:1 and error out for api_version:2 but Clio is not replicating rippled's response for default api_version.

sample request:
`{
    "method": "tx",
    "params": [
        {
            "ctid": "C015C2BA00010002",
            "binary": "fhgj"
        }
    ]
}`

Rippled response:
`{
    "result": {
        "ctid": "C015C2BA00010002",
        "date": 752711912,
        "hash": "3050F11FB53A56B506B73A20CCD04400F747514EFF592459492B81B773AB7209",
        "inLedger": 1426106,
        "ledger_index": 1426106,
        "meta": "201C00000001F8E5110061250015C2B3558AE0C20F427DB146AD16D184E680F1B91EF74E27CD85916005F297CF54DCAD0B569B6B6D225D5AAF7517E6532177978889AB7A1DDD8356F0ABE0DE1B9DAF8E5B5CE6240015C2B36240000002540BE400E1E72200000000240015C2B42D0000000062400000024E1602F681144280875B1B0435334FCA9CC68152860A5DF34BEBE1E1E5110061250015C2B855DC68D3A0A43F30811A8EA6D6DC9C8BB31DEECBFBF4005E96BE3427B7A169920756BC0DAE09C0BFBC4A49AA94B849266588BFD6E1F554B184B5788AC55D6E07EB95E66240029FE98A5F327BE1E72200000000240007263A2D000000006240029FE99055137B81142A6ADC782DAFDDB464E434B684F01416B8A33B20E1E1F1031000",
        "status": "success",
        "tx": "1200002280000000240015C2B3614000000005F5E10068400000000000000A7321ED712FD2065885A3581B27810E162BABF1D4353B1CF064CA34BDE11BD61C4AA23C7440F4A8804BD821D1E6E584D4F0FC01E983DDF28B4FE4354D189C5C4325CD764302A14DA7A0371810DE050BDEDA8A0B94E4DBBC4C80C1AE73BA994491DA222F6E0D81144280875B1B0435334FCA9CC68152860A5DF34BEB83142A6ADC782DAFDDB464E434B684F01416B8A33B20",
        "validated": true
    }
}`

Clio response:
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "Invalid parameters.",
        "status": "error",
        "type": "response",
        "request": {
[...]

PR #993, opened 2023-11-20
Fix invalid binary for tx

cindyyan317:
#984 
Rippled tx does not restrict the 'binary' must be a boolean.
So Clio matches the rippled for v1. For v2, Clio will still raise error for invalid 'binary', like other 'binary' fields.


kuznetsss (approved):
👍

PR #994, opened 2023-11-20
Add assertion that terminate clio

### LINKED ISSUE #893: Improve asserts in the code

### LINKED ISSUE #929: Use Boost.Stacktrace or std equivalent
Add stacktrace to crash log.

See https://www.boost.org/doc/libs/1_82_0/doc/html/stacktrace.html 
and https://en.cppreference.com/w/cpp/utility/basic_stacktrace

kuznetsss:
Fixes #893.

Also added termination handler to print backtrace on crash, so fixes #929.


CMakeLists.txt
```diff
@@ -56,11 +57,14 @@ target_link_libraries (clio
...
   PUBLIC Boost::log
   PUBLIC Boost::log_setup
+  PUBLIC Boost::stacktrace_backtrace
   PUBLIC cassandra-cpp-driver::cassandra-cpp-driver
   PUBLIC fmt::fmt
   PUBLIC OpenSSL::Crypto
   PUBLIC OpenSSL::SSL
   PUBLIC xrpl::libxrpl
+  PUBLIC dl
+  PUBLIC libbacktrace::libbacktrace
```
godexsoft, line 67: Do we still need these two even with Boost.Stacktrace?

conanfile.py
```diff
@@ -27,6 +27,7 @@ class Clio(ConanFile):
         'grpc/1.50.1',
         'openssl/1.1.1u',
         'xrpl/2.0.0-rc1',
+        'libbacktrace/cci.20210118'
```
godexsoft, line 30: Is this needed with Boost.Stacktrace still?

src/data/cassandra/impl/ExecutionStrategy.h
```diff
@@ -455,10 +466,7 @@ class DefaultExecutionStrategy {
     decrementOutstandingRequestCount()
     {
         // sanity check
-        if (numWriteRequestsOutstanding_ == 0) {
-            assert(false);
-            throw std::runtime_error("decrementing num outstanding below 0");
-        }
+        ASSERT(numWriteRequestsOutstanding_ > 0, "decrementing num outstanding below 0");
```
godexsoft: Let's keep the messages in the same format. i.e. capitalised 👍

src/etl/Source.h
```diff
@@ -794,7 +794,7 @@ class SourceImpl : public Source {
                 uint32_t const sequence = std::stoll(minAndMax[0]);
                 pairs.emplace_back(sequence, sequence);
             } else {
-                assert(minAndMax.size() == 2);
+                ASSERT(minAndMax.size() == 2, "Min and max should be of size 2. Got size ={}", minAndMax.size());
```
godexsoft: Please add a space between `=` and `{}` 👍

src/etl/impl/AsyncData.h
```diff
@@ -60,7 +61,12 @@ class AsyncCallData {
                           << " . prefix = " << ripple::strHex(std::string(1, prefix))
                           << " . nextPrefix_ = " << ripple::strHex(std::string(1, nextPrefix_));
 
-        assert(nextPrefix_ > prefix || nextPrefix_ == 0x00);
+        ASSERT(
+            nextPrefix_ > prefix || nextPrefix_ == 0x00,
+            "nextPrefix_ must be greater than prefix. Got: nextPrefix_ = {}, prefix = {} ",
```
godexsoft: I think it's better to say "Next prefix must be greater than current prefix." to keep the formatting. 
Also let's remove the last space 🔢

src/etl/impl/LedgerLoader.h
```diff
@@ -189,50 +189,49 @@ class LedgerLoader {
...
-                                    assert(succ);
-                                    if (succ->key == cur->key) {
-                                        LOG(log_.debug()) << "Writing book successor = " << ripple::strHex(base)
-                                                          << " - " << ripple::strHex(cur->key);
-
-                                        backend_->writeSuccessor(
-                                            uint256ToString(base), sequence, uint256ToString(cur->key)
-                                        );
-                                    }
+                auto seconds = ::util::timed<std::chrono::seconds>([this, edgeKeys = &edgeKeys, sequence, &numWrites](
```
godexsoft: If you remove the unnecessary `()` from the lambda it may format better 👍

src/main/Main.cpp
```diff
@@ -218,4 +220,8 @@ try {
     return EXIT_SUCCESS;
 } catch (std::exception const& e) {
     LOG(LogService::fatal()) << "Exit on exception: " << e.what();
+    return EXIT_FAILURE;
+} catch (...) {
```
godexsoft, line 224: Good idea 👍

src/util/Assert.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string_view>
+
+#include <boost/stacktrace.hpp>
+#include <fmt/format.h>
+
+namespace util {
+
+template <typename... Args>
+constexpr void
+assert_impl(
```
godexsoft: CamelCase please 🔤

src/util/Assert.h
```diff
@@ -0,0 +1,52 @@
...
+    char const* expression,
+    bool const condition,
+    fmt::format_string<Args...> format,
+    Args&&... args
+)
+{
+    if (!condition) {
+        fmt::println(stderr, "Assertion '{}' failed at {}:{}:", expression, location.file_name(), location.line());
+        fmt::println(stderr, format, std::forward<Args>(args)...);
+        std::cerr << "Stacktrace:\n" << boost::stacktrace::stacktrace() << std::endl;
```
godexsoft: Why not use the same `println` function for this? In `libfmt` you can use everything that's compatible with streams if you include `fmt/std.h` if i remember correctly.

unittests/Main.cpp
```diff
@@ -16,13 +16,15 @@
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */
 //==============================================================================
+#include <util/TerminationHandler.h>
```
godexsoft, line 20: There should always be a line of space after the `//==` line.

godexsoft (changes requested):
Very cool 🔥 Let's fix the nitpicks; otherwise looks very useful 🙇

src/util/Assert.h
```diff
@@ -0,0 +1,52 @@
...
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <util/SourceLocation.h>
+
+#include <iostream>
+#include <string_view>
```
cindyyan317: do we need this header?

CMakeLists.txt
kuznetsss: Yes, Boost.stacktrace requires these two. https://www.boost.org/doc/libs/1_83_0/doc/html/stacktrace/configuration_and_build.html

conanfile.py
kuznetsss: As I understand Boost.Stacktrace is a wrapper around this library (if Boost.stacktrace is used in  backtrace mode).

src/util/Assert.h
kuznetsss: Still one more line is needed to define `template<> fmt::formatter<boost::stacktrace::stacktrace>`.

godexsoft (dismissed):
LGTM 🚀

godexsoft (approved):

PR #996, opened 2023-11-21
Add server_definitions handler

### LINKED ISSUE #942: New RPC :server_definitions
https://github.com/XRPLF/rippled/pull/4703

kuznetsss:
Fixes #942


src/rpc/handlers/ServerDefinitions.h
```diff
@@ -0,0 +1,71 @@
...
+    {
+        static const RpcSpec rpcSpec = {};
+        return rpcSpec;
+    }
+
+    Result
+    process(Context const& ctx) const
+    {
+        auto serverDefinitionsOpt =
+            balancer_->forwardToRippled({{"command", "server_definitions"}}, ctx.clientIp, ctx.yield);
```
kuznetsss: It seems there is a parameter `hash` in rippled. Probably we should accept any parameters and forward them.

cindyyan317: no need to forward the parameter manually . See the list of commands we shall forward here -> https://github.com/XRPLF/clio/blob/8ebe2d6a802e5a6c2b078b339fedcdfa08984524/src/rpc/common/impl/ForwardingProxy.h#L103
Add the new one here , the framework will do it for you ~

cindyyan317:
it will be good to add test in RPCForwardingProxyTest

cindyyan317 (approved):

PR #1000, opened 2023-11-24
Add hostname resolving to dosguard

### LINKED ISSUE #983: Resolve hostnames from config
It would be useful to be able to put DNS names instead of IP addresses in config. Also after adding these feature we should fix docker-compose file with clio and rippled to use docker hostnames.

kuznetsss:
Fixes #983. 

Cassandra, ETL sorces and cache already support hostname resolving.


cindyyan317 (dismissed):
Nice ~

src/web/Resolver.cpp
```diff
@@ -0,0 +1,63 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <web/Resolver.h>
```
godexsoft: Didn't you want to start using quotes for local includes?

src/web/Resolver.h
```diff
@@ -0,0 +1,60 @@
...
+#include <boost/asio.hpp>
+
+#include <concepts>
+#include <string>
+#include <vector>
+
+namespace web {
+
+template <typename T>
+concept SomeResolver = requires(T t) {
```
godexsoft, line 31: We usually put these in Concepts.h files and the 'web' part of the codebase already seems to have one. I don't mind either way tho.

src/web/Resolver.h
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @brief Simple hostnames to IP addresses resolver.
+ */
+class Resolver {
+public:
+    /**
+     * @brief Resolve hostname to IP addresses.
+     *
+     * @note This method throws an exception when the hostname cannot be resolved.
```
godexsoft: Maybe more clear with a `@throw`?

godexsoft (dismissed):
Just a few suggestions. Great job 👍

src/web/Resolver.cpp
kuznetsss: I think I will do it separately for the whole project.

src/web/Resolver.h
kuznetsss: I don't think it is a good idea to put all the concepts together. Because it is a rare situation when you need more then one concept at once and often you need a concept in the place where you need the class it is describing.

src/web/WhitelistHandler.h
```diff
@@ -159,17 +163,27 @@ class WhitelistHandler {
...
         auto const transform = [](auto const& elem) { return elem.template value<std::string>(); };
 
-        return SetType{
+        std::unordered_set<std::string> const hostnames{
             boost::transform_iterator(std::begin(whitelist), transform),
             boost::transform_iterator(std::end(whitelist), transform)};
+
+        // resolve hostnames to ips
+        std::unordered_set<std::string> ips;
+        for (auto const& hostname : hostnames) {
```
godexsoft, line 180: Does this code work fine with entries that are in CIDR format (e.g. `182.168.0.1/22`)?

godexsoft (dismissed):
I think all good. Just one question (thought it's worth confirming just in case, even though I think we test for it in unit tests).

src/web/WhitelistHandler.h
kuznetsss: Yes, because resolver checks whether it is an IP or network address and actually resolves only hostnames. See function `isAddress()` in `Resolver`.
Also I put mock resolver only into one test while two others are using the real one.

godexsoft (approved):

PR #1004, opened 2023-11-28
Basic execution context framework

godexsoft:
Async framework to be used to eradicate plain `std::thread` use from the codebase.
Let's try and get this merged and start using it in actual code. We will likely discover things that can be done better by actually using it outside of tests.


src/util/async/context/BasicExecutionContext.hpp
```diff
@@ -0,0 +1,334 @@
...
+    using Timer = SteadyTimer<Executor>;
+    using Strand = AsioPoolStrandContext;
+
+    Strand
+    makeStrand()
+    {
+        return {boost::asio::make_strand(executor)};
+    }
+
+    boost::asio::thread_pool executor;
```
cindyyan317: Executor

src/util/async/AnyOperation.hpp
```diff
@@ -0,0 +1,108 @@
...
+                return util::Unexpected(std::move(data).error());
+
+            if constexpr (std::is_void_v<RetType>) {
+                return {};
+            } else {
+                return std::any_cast<RetType>(std::move(data).value());
+            }
+
+        } catch (std::bad_any_cast const& e) {
+            return util::Unexpected{ExecutionError(fmt::format("{}", std::this_thread::get_id()), "Bad any cast")};
```
cindyyan317, line 100: I was thinking if e.what() is more explanatory. But never mind, it is just a string "bad any_cast" 😿

src/util/async/Operation.hpp
```diff
@@ -0,0 +1,169 @@
...
+            return op_.value();
+        }
+    };
+
+    std::shared_ptr<State> state_ = std::make_shared<State>();
+    typename CtxType::Timer timer_;
+
+    BasicScheduledOperation(auto& executor, auto delay, auto&& fn)
+        : timer_(executor, delay, [state = state_, fn = std::forward<decltype(fn)>(fn)](auto ec) mutable {
+            state->emplace(fn(ec));
```
cindyyan317, line 93: Shall we check ec here , if canceled : boost::asio::error::operation_aborted, not to call fn at all?

src/util/async/context/BasicExecutionContext.hpp
godexsoft: "An executor is someone who is responsible for executing, or following through on, an assigned task or duty.".
It makes sense, although it may be better to call it something else, idk.
Any ideas?

src/util/async/AnyOperation.hpp
godexsoft: Unfortunately there is not much more detail we can get from the exception itself.
The good news is that i don't think this should ever fire as we can only get here thru the framework. The framework should not generate code that would result in `bad_any_cast`. It's here just for completeness.

src/util/async/Operation.hpp
godexsoft: This is intentionally passed down to the erased execution context. Without it we would not be able to call user's callback with two args, second one being `bool isCancelled`.
It's handled by `isAborted = (ec == boost::asio::error::operation_aborted)` in `BasicExecutionContext.hpp`.

src/util/async/context/BasicExecutionContext.hpp
cindyyan317: I mean: Executor executor; I think the name is 👍

godexsoft: Ah, I see it now 🙏

benchmarks/ExecutionContextBenchmarks.cpp
```diff
@@ -0,0 +1,268 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
cindyyan317: 2024

benchmarks/ExecutionContextBenchmarks.cpp
```diff
@@ -0,0 +1,268 @@
...
+using namespace util;
+using namespace util::async;
+
+class TestThread {
+    std::vector<std::thread> threads_;
+    etl::ThreadSafeQueue<std::optional<uint64_t>> q_;
+    etl::ThreadSafeQueue<uint64_t> res_;
+
+public:
+    TestThread(std::vector<uint64_t> const& data) : q_(data.size() + 1), res_(data.size() + 1)
```
cindyyan317: I am trying to figure out why +1 here 🤔 .

src/util/async/Concepts.hpp
```diff
@@ -0,0 +1,127 @@
...
+    {
+        std::invoke(fn, std::declval<Args>()...)
+    };
+};
+
+template <typename T>
+concept SomeStdDuration = requires {
+    // Thank you Ed Catmur for this trick.
+    // See https://stackoverflow.com/questions/74383254/concept-that-models-only-the-stdchrono-duration-types
+    []<typename Rep, typename Period>( //
```
cindyyan317, line 119: 👍

benchmarks/ExecutionContextBenchmarks.cpp
godexsoft: I think it was needed because `push_back` on the queue would hang since capacity is reached or something like that. I believe the code of the implementation changed since then and it's no longer needed 👍

benchmarks/ExecutionContextBenchmarks.cpp
godexsoft: This file was added in 2023 😃 I will update 👍

cindyyan317:
Looks good to me. I did not pull it to local to try it out. (got some error when gh it). Look forward to using it in our code.

cindyyan317 (dismissed):

src/etl/ETLHelpers.hpp
```diff
@@ -206,6 +206,12 @@ class ThreadSafeQueue {
         cv_.notify_all();
         return ret;
     }
+
+    auto
```
kuznetsss: Why `auto`? I think putting `size_t` here would improve readability

benchmarks/Playground.cpp
```diff
@@ -0,0 +1,45 @@
...
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+/*
```
kuznetsss, line 20: Maybe it would be better to create a `README.md` here and put it there?

benchmarks/ExecutionContextBenchmarks.cpp
kuznetsss, line 1: I think it would be better to keep the same structure as in `src` and `unittests`

src/util/async/impl/Any.hpp
```diff
@@ -0,0 +1,51 @@
...
+ */
+class Any {
+    std::any value_;
+
+public:
+    Any() = default;
+    Any(Any const&) = default;
+    Any(Any&&) = default;
+
+    // note: this needs to be `auto` instead of `std::any` because of a bug in gcc 11.4
```
kuznetsss, line 42: So the whole class is needed only because of the bug in gcc 11.4? Maybe we could switch to a newer one then?

src/util/async/impl/ErasedOperation.hpp
```diff
@@ -0,0 +1,145 @@
...
+            return operation.get();
+        }
+
+        void
+        requestStop() override
+        {
+            if constexpr (SomeStoppableOperation<OpType>) {
+                operation.requestStop();
+            } else {
+                throw std::logic_error("Stop requested on non-stoppable operation");
```
kuznetsss, line 126: This could be moved to compile time. `requires` allow to enable methods by a concept.
https://godbolt.org/z/Kasf4o7hs

src/util/async/context/impl/Cancellation.hpp
```diff
@@ -0,0 +1,133 @@
...
+    SharedStopState shared_ = std::make_shared<StopState>();
+
+public:
+    class Token {
+        friend class YieldContextStopSource;
+        SharedStopState shared_;
+        boost::asio::yield_context yield_;
+
+        Token(YieldContextStopSource* source, boost::asio::yield_context yield)
+            : shared_{source->shared_}, yield_{std::move(yield)}
```
kuznetsss, line 60: You could pass here `SharedStopState` as first argument and then you will not need to make friend class. But that's a minor thing, feel free to ignore if don't want to change.

src/util/async/context/impl/Timer.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+template <typename ExecutorType>
+class SteadyTimer {
+    boost::asio::steady_timer timer_;
+
+public:
+    SteadyTimer(ExecutorType& executor, auto delay, auto&& fn) : timer_{executor}
+    {
+        timer_.expires_after(delay);
+        timer_.async_wait(std::forward<decltype(fn)>(fn));
```
kuznetsss, line 34: Shouldn't timer wrap `fn`? Something like
```cpp
[](auto ec) {
  if (ec != boost::error_code::operation_aborted) {
    fn();
};
```
I saw such code in some other place, but I'm thinking isn't it a timer's responsibility?

src/util/async/context/impl/Utils.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace util::async::impl {
+
+inline constexpr struct AssociatedExecutorExtractor {
+    template <typename CtxType>
+    [[nodiscard]] typename CtxType::ExecutorType&
+    operator()(CtxType& ctx) const noexcept
+    {
+        return ctx.context_.executor;
+    }
+} extractAssociatedExecutor;
```
kuznetsss, line 42: Couldn't it be simply a function?
I saw you're using `AssociatedExecutorExtractor` to make it `friend` but it could be done with a function as well.

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+    template <typename T>
+    using OperationType = ::testing::NiceMock<MockOperation<T>>;
+
+    template <typename T>
+    using StoppableOperationType = ::testing::NiceMock<MockStoppableOperation<T>>;
+
+    template <typename T>
+    using ScheduledOperationType = ::testing::NiceMock<MockScheduledOperation<T>>;
+
+    ::testing::NaggyMock<MockExecutionContext> mockExecutionContext;
```
kuznetsss: Maybe `StrictMock` or `NaggyMock` here because there is no point having warnings about uninteresting calls? Probably strict because I assume there is no warnings now.

kuznetsss: Probably you should delete `::testing::` here.

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+    EXPECT_CALL(mockExecutionContext, execute(An<std::function<impl::Any(AnyStopToken)>>(), _))
+        .WillOnce([](auto&&, auto) -> StoppableOperationType<impl::Any> const& { throw 0; });
+
+    EXPECT_ANY_THROW([[maybe_unused]] auto unused = ctx.execute([](auto) { return 42; }));
+}
+
+TEST_F(AnyExecutionContextTests, TimerCancellation)
+{
+    auto mockScheduledOp = ScheduledOperationType<impl::Any>{};
+    EXPECT_CALL(mockScheduledOp, cancel()).Times(1);
```
kuznetsss: `Times(1)` is by default for `EXPET_CALL()`.

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+}
+
+TEST_F(AnyExecutionContextTests, ExecuteWithStopTokenAndReturnValue)
+{
+    auto mockOp = StoppableOperationType<impl::Any>{};
+    EXPECT_CALL(mockOp, get()).WillOnce(Return(std::make_any<int>(42)));
+    EXPECT_CALL(mockExecutionContext, execute(An<std::function<impl::Any(AnyStopToken)>>(), _))
+        .WillOnce(ReturnRef(mockOp));
+
+    auto op = ctx.execute([](auto) { return 42; });
```
kuznetsss: If I understand this correctly the lambda is never called. I think to make it more clear to understand it would be to use a lambda without 42 here `[](auto) -> int { return {}; }`.

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+    template <typename T>
+    using ScheduledOperationType = ::testing::NiceMock<MockScheduledOperation<T>>;
+
+    ::testing::NaggyMock<MockExecutionContext> mockExecutionContext;
+    AnyExecutionContext ctx{static_cast<MockExecutionContext&>(mockExecutionContext)};
+};
+
+TEST_F(AnyExecutionContextTests, ExecuteWithoutTokenAndVoid)
+{
+    auto mockOp = OperationType<impl::Any>{};
```
kuznetsss, line 57: Shouldn't there be an expectation that `mockOp.get()` will be called?

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+    ASSERT_TRUE(op.get());
+}
+
+TEST_F(AnyExecutionContextTests, ExecuteWithStopTokenAndVoidThrowsException)
+{
+    EXPECT_CALL(mockExecutionContext, execute(An<std::function<impl::Any(AnyStopToken)>>(), _))
+        .WillOnce([](auto&&, auto) -> StoppableOperationType<impl::Any> const& { throw 0; });
+
+    EXPECT_ANY_THROW([[maybe_unused]] auto unused = ctx.execute([](auto) {}));
```
kuznetsss: Why creation `unused` is needed here?

unittests/util/async/AnyExecutionContextTests.cpp
```diff
@@ -0,0 +1,301 @@
...
+        scheduleAfter(An<std::chrono::milliseconds>(), An<std::function<impl::Any(AnyStopToken)>>())
+    )
+        .WillOnce([&mockScheduledOp](auto, auto&&) -> ScheduledOperationType<impl::Any> const& {
+            return mockScheduledOp;
+        });
+
+    auto timer = ctx.scheduleAfter(std::chrono::milliseconds{42}, [](auto) { return 42; });
+
+    static_assert(std::is_same_v<decltype(timer), AnyOperation<int>>);
+    EXPECT_EQ(timer.get().value(), 42);
```
kuznetsss: It seems you like the number 42 😄

I think here it makes the test a bit confusing whether timer returns milliseconds or the value from lambda or the value from `mockScheduledOp`.

unittests/util/async/AnyOperationTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    EXPECT_CALL(mockOp, get()).WillOnce(Return(std::make_any<int>(42)));
+    auto res = intOp.get();
+    EXPECT_EQ(res.value(), 42);
+}
+
+TEST_F(AnyOperationTests, WaitCallPropagated)
+{
+    auto called = false;
+    EXPECT_CALL(mockOp, wait()).WillOnce([&] { called = true; });
+    ;
```
kuznetsss: Extra `;`

unittests/util/async/AnyOperationTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    voidOp.wait();
+    EXPECT_TRUE(called);
+}
+
+TEST_F(AnyOperationTests, CancelCallPropagated)
+{
+    auto called = false;
+    EXPECT_CALL(mockScheduledOp, cancel()).WillOnce([&] { called = true; });
+    ;
```
kuznetsss: And here `;`

unittests/util/async/AnyOperationTests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    scheduledVoidOp.cancel();
+    EXPECT_TRUE(called);
+}
+
+TEST_F(AnyOperationTests, RequestStopCallPropagated)
+{
+    auto called = false;
```
kuznetsss: Better
```cpp
StrictMock<MockFunction<void()>> callback;
EXPECT_CALL(callback, Call());
EXPECT_CALL(mockScheduledOp, requestStop()).WillOnce([&] { callback.Call(); });
```

unittests/util/async/AnyStopTokenTests.cpp
```diff
@@ -0,0 +1,56 @@
...
+        EXPECT_EQ(stopToken.isStopRequested(), true);
+    }
+}
+
+TEST(AnyStopTokenTests, CanCopy)
+{
+    AnyStopToken stopToken{FakeStopToken{true}};
+    AnyStopToken token = stopToken;
+
+    EXPECT_EQ(token, stopToken);
```
kuznetsss: Maybe this should be checked with `false` as well. To make sure there is no default state on copying.
If you would do this, maybe it makes sense to create a parametrised test group with `true` and `false` for both tests.

unittests/util/async/AsyncExecutionContextTests.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+using ExecutionContextTypes = Types<CoroExecutionContext, PoolExecutionContext, SyncExecutionContext>;
+
+template <typename T>
+struct ExecutionContextTests : public ::testing::Test {
+    using ExecutionContextType = T;
+    ExecutionContextType ctx{2};
+};
+
+TYPED_TEST_CASE(ExecutionContextTests, ExecutionContextTypes);
```
kuznetsss, line 42: Cool 👍

kuznetsss (changes requested):
Very good overall! Just a few things that could be improved.

src/util/async/context/impl/Timer.hpp
godexsoft: No because we want `fn` to receive the `ec` in this case

src/util/async/context/impl/Utils.hpp
godexsoft: Can't be a function because `ctx.context_` is private and we need this "function" to be called with the outermost execution context

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: I think i forgot to change it to Nice after verifying it was not showing any warnings

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: Would it then warn if it was called NOT once by default?

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: Yea, it's there to infer the correct return type. I agree and will change

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: Because execute returns a nodiscard type or is marked as nodiscard itself. 
Alternative is `(void)ctx.execute...` or `_` in C++26. Currently we can't even use `_` because it clashes with gtest.

benchmarks/Playground.cpp
godexsoft: I'm not sure if that's better or not 🤔 It's nice to be able to uncomment the lines you need and start hacking. That's what playgrounds are for.. and the first part belongs to `Playground`, not to benchmarks in general.

src/util/async/impl/Any.hpp
godexsoft: No, this is just another bug. The Any wrapper is needed because `util::Expected` is using Boost.Outcome and there is an issue using it with `std::any` directly. `std::expected` from C++23 does not have this problem so we will be able to remove this wrapper in favour of `std::any` as well as remove `util::Expected` in favour of `std::expected`. 

Created issues: #1173 #1174 
Will add a note and reference the issues as comments to avoid future confusion

src/util/async/impl/ErasedOperation.hpp
godexsoft: Would be nice but it cannot in this context. The `ErasedOperation` is just one type, not a template. And the `Concept` part is an abstract base class while this code is inside of a member function we `override`. Unfortunately we can't SFINAE here.

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: I'll add it. Thanks

src/util/async/context/impl/Cancellation.hpp
godexsoft: The reason it's a `friend` is because the constructor of `Token` is private. This is by design as we don't want anyone to be able to create a `StopSource::Token` instance without going through `StopSource`. Even if we change the argument type we can't really get rid of `friend` here (and neither we should - that's what friends are for 👍).

unittests/util/async/AnyExecutionContextTests.cpp
godexsoft: The answer to my question is yes - I tried it. Basically by default it's expecting exactly one call and it does indeed fail the test if in reality we had more than 1 or no calls at all 👍

kuznetsss (approved):

PR #1013, opened 2023-12-01
Add nightly builds

### LINKED ISSUE #884: Nightly builds in CI
Nightly builds:
- Run clang-tidy only at night and create an issue if problems in code found
- Build without ccache
- Nightly releases
- Don't use clang-tidy in regular CI workflow
- Tests with asan

Other improvements:
- Custom docker for CI on Linux
- Maybe run clang-tidy on changed files

kuznetsss:
Partially fixes #884.
Adds:
- Docker image for CI on Linux
- Nightly builds without cache and releases
- Nightly clang-tidy checks
- Fix typos in .clang-tidy


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,51 @@
+name: Clang-tidy check
+on:
+  schedule:
+    - cron:  '0 6 * * 1-5'
```
godexsoft, line 4: Does clang-tidy need to be on cron too?

godexsoft (commented):
Looks like a big improvement. Only leaving one question where it looks like it may be a result of copy-paste, but i may be misunderstanding something 👍

.github/workflows/clang-tidy.yml
kuznetsss: Yeah, the idea is that it will check the code every night and if it finds something it will create an issue and a PR with  fixes if possible.

godexsoft: Got it, thanks!

godexsoft (approved):
LGTM! 🚀

PR #1054, opened 2023-12-15
Cleanup working directory for macos

kuznetsss:


godexsoft (approved):

godexsoft (approved):
I'm approving this but one thing worries me a little. What will happen if `${{ github.workspace }}` expands to an empty string? Can that happen?

Issue #1021, opened 2023-12-08
[Doc] nft_info returns uri in the response with burned NFT.


mounikakun:
NFT was created with uri during NFTokenMint and burned the NFT. While trying to retrieve nft_info, uri is still in the response.

Request:
```json
{
    "method": "nft_info",
    "params": [
      {
          "nft_id": "0000000079D053DCD0E60E1947E707B3B179CB09A315667FA85638C70023432C"
      }
    ]
}
```

Clio_response:
```json
{
    "result": {
        "nft_id": "0000000079D053DCD0E60E1947E707B3B179CB09A315667FA85638C70023432C",
        "ledger_index": 2310958,
        "owner": "rUfaLhqEUXETmwZgS3tvEegxNhvtV2US5u",
        "is_burned": true,
        "flags": 0,
        "transfer_fee": 0,
        "issuer": "rUfaLhqEUXETmwZgS3tvEegxNhvtV2US5u",
        "nft_taxon": 0,
        "nft_serial": 2310956,
        "validated": true,
        "uri": "697066733A2F2F62616679626569676479727A74357366703775646D37687537367568377932366E6633656675796C71616266336F636C67747179353566627A6469",
        "status": "success"
    },
    "warnings": [
        {
            "id": 2001,
            "message": "This is a clio server. clio only serves validated data. If you want to talk to rippled, include 'ledger_index':'current' in your request"
        }
    ]
}
```

cindyyan317:
There was a PR deliberately adding URI for burned NFT but forgot to update the document.  #634 
Let us update the document.

cindyyan317:
Created issue for document change: https://github.com/XRPLF/xrpl-dev-portal/issues/2328

PR #1056, opened 2023-12-18
Omit the uri when nft is burned

cindyyan317:
According to the doc, the uri should be omitted if nft has been burned. 

https://xrpl.org/nft_info.html#nft_info


shawnxie999:
I would double check before merging. This would be a breaking change and the doc is not necessarily up to date, as previously there has been a change to the URI encoding format.

cindyyan317:
The document needs to be updated. #1021

PR #1058, opened 2023-12-19
Add coverage_report target

Bronek:
Reusing the same `CodeCoverage.cmake` as in https://github.com/XRPLF/rippled/pull/4849 

**NOTE** This PR should be only merged by someone with administrative access to the repo, since it makes use of `secrets.CODECOV_TOKEN`. If I understand github workflows correctly, a commit (merge) made by someone _without_ access to repo secrets would result in this change _not_ having this access either, which will lead to [spurious failures of codecov integration](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954)


cindyyan317 (dismissed):
👍

.codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 50%
```
godexsoft, line 5: Is the plan to gradually update this value to a higher one as we add more coverage?

godexsoft (dismissed):
Very nice! Thank you for fixing codecov report 🚀 🥇

.codecov.yml
Bronek: Yes that's definitely the plan ! I will leave it to you :)

cindyyan317 (approved):

PR #1060, opened 2023-12-20
Implement amm_info handler

godexsoft:
As @arihantkothari could not continue working on this I'm taking it over to finish the missing/failing unit tests and finalize the handler.

Note: [Previous work on this code was here](https://github.com/XRPLF/clio/pull/857).


src/rpc/AMMHelpers.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "rpc/AMMHelpers.h"
+
+#include "data/BackendInterface.h"
+#include "util/log/Logger.h"
+
+#include <ripple/protocol/AMMCore.h>
+
+namespace rpc {
+
+std::pair<ripple::STAmount, ripple::STAmount>
```
kuznetsss, line 29: Maybe use a struct to improve readability?

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    if (output.asset1Frozen)
+        amm[JS(asset_frozen)] = *output.asset1Frozen;
+
+    if (output.asset2Frozen)
+        amm[JS(asset2_frozen)] = *output.asset2Frozen;
+
+    jv = {
+        {JS(amm), amm},
+        {JS(ledger_current_index), output.ledgerIndex},
```
cindyyan317: ledger_current_index is not for Clio, rippled returns ledger_current_index for not validated ledger. For Clio, validated will be always true, thus only ledger_index.

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,125 @@
...
+        boost::json::value amount2;
+        boost::json::value lpToken;
+        boost::json::array voteSlots;
+        boost::json::value auctionSlot;
+        std::uint16_t tradingFee = 0;
+        std::string ammAccount;
+        std::optional<bool> asset1Frozen;
+        std::optional<bool> asset2Frozen;
+
+        uint32_t ledgerIndex = 0;
```
cindyyan317: missing ledgerHash

src/rpc/common/Validators.cpp
```diff
@@ -212,4 +213,23 @@ CustomValidator SubscribeAccountsValidator =
         return MaybeError{};
     }};
 
+CustomValidator AMMAssetValidator =
+    CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+        ASSERT(value.is_object(), "{}: Value must be a JSON object", key);
```
kuznetsss: Now I'm thinking because it is a validator it is better to return `Error` here. Because input here comes from clients and clients shouldn't be able to kill clio with wrong input.

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,125 @@
...
+    };
+
+    using Result = HandlerReturnType<Output>;
+
+    AMMInfoHandler(std::shared_ptr<BackendInterface> const& sharedPtrBackend) : sharedPtrBackend_(sharedPtrBackend)
+    {
+    }
+
+    static RpcSpecConstRef
+    spec([[maybe_unused]] uint32_t apiVersion)
```
kuznetsss: Could you please move the implementation into cpp file?

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,125 @@
...
+             meta::IfType<std::string>{stringIssueValidator},
+             meta::IfType<boost::json::object>{
+                 meta::WithCustomError{validation::AMMAssetValidator, Status(RippledError::rpcISSUE_MALFORMED)},
+             }},
+            {JS(asset2),
+             meta::WithCustomError{
+                 validation::Type<std::string, boost::json::object>{}, Status(RippledError::rpcISSUE_MALFORMED)
+             },
+             meta::IfType<std::string>{stringIssueValidator},
+             meta::IfType<boost::json::object>{
```
cindyyan317, line 103: If we use meta::IfType<boost::json::value>, maybe we don't need the std::string validator?

unittests/rpc/handlers/AMMInfoTests.cpp
```diff
@@ -0,0 +1,1154 @@
...
+    });
+}
+
+TEST_F(RPCAMMInfoHandlerTest, HappyPathMinimalFirstXRPNoTrustline)
+{
+    backend->setRange(10, 30);
+
+    auto const account1 = GetAccountIDWithString(AMM_ACCOUNT);
+    auto const account2 = GetAccountIDWithString(AMM_ACCOUNT2);
+    auto const lgrInfo = CreateLedgerInfo(LEDGERHASH, SEQ);
```
kuznetsss, line 309: To make test a bit shorter you could move all the duplicated data into some new fixture inherited from `RPCAMMInfoHandlerTest`.

unittests/util/TestObject.h
```diff
@@ -284,8 +296,28 @@ CreateAMMObject(
...
+
+void
+AMMSetAuctionSlot(
+    ripple::STObject& amm,
+    ripple::AccountID const& accountId,
+    ripple::STAmount price,
+    uint16_t discountedFee,
+    uint32_t expiration,
+    std::vector<ripple::AccountID> const& authAccounts = {}
 );
```
kuznetsss, line 317: Probably at least for now we don't need these methods anywhere except AMMInfo tests. So maybe it would be better for now to place them into `AMMInfoTests.cpp`? We can always move them out when we need them somewhere else.

kuznetsss (changes requested):
Good job! Just a few comments.

src/rpc/handlers/AMMInfo.h
```diff
@@ -0,0 +1,125 @@
...
+                 meta::WithCustomError{validation::AMMAssetValidator, Status(RippledError::rpcISSUE_MALFORMED)},
+             }},
+            {JS(asset2),
+             meta::WithCustomError{
+                 validation::Type<std::string, boost::json::object>{}, Status(RippledError::rpcISSUE_MALFORMED)
```
cindyyan317: if we only allow boost::json::object here, the rpcISSUE_MALFORMED will raise for all string input. Then we don't need stringIssueValidator ?

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    auto ammKeylet = ammID != 0 ? keylet::amm(ammID) : keylet::amm(input.issue1, input.issue2);
+    auto const ammBlob = sharedPtrBackend_->fetchLedgerObject(ammKeylet.key, lgrInfo.seq, ctx.yield);
+
+    if (not ammBlob)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND}};
+
+    auto const amm = SLE{SerialIter{ammBlob->data(), ammBlob->size()}, ammKeylet.key};
+    auto const ammAccountID = amm.getAccountID(sfAccount);
+    auto const accBlob =
```
cindyyan317, line 105: We can skip this check if input.ammAccount has value. input.ammAccount should be the same with the sfAccount in amm object.

src/rpc/handlers/AMMInfo.cpp
godexsoft: Ok, i will update it back then. The docs are not specific about the difference and since `rippled` returns `ledger_current_index` i thought we should too because Mounika will definitely come to us with questions later 👍

src/rpc/common/Validators.cpp
godexsoft: Yea but you can't get here with non-object because we have `IfType` checks on call-site. But in general I agree that it may be that in the future someone uses it without `IfType` and then it may backfire..

src/rpc/handlers/AMMInfo.h
godexsoft: We don't want `rpcISSUE_MALFORMED` for all string input - we want to do exactly what `rippled` does which is to check the format. It just so happens that it results in `issueMalformed` atm.

src/rpc/AMMHelpers.cpp
```diff
@@ -0,0 +1,82 @@
...
+getAmmPoolHolds(
+    BackendInterface const& backend,
+    std::uint32_t sequence,
+    ripple::AccountID const& ammAccountID,
+    ripple::Issue const& issue1,
+    ripple::Issue const& issue2,
+    bool freezeHandling,
```
cindyyan317, line 36: Clio seems not need this flag while rippled uses it.

unittests/util/TestObject.h
godexsoft: Hmm. Not sure tbh. as TestObject files were exactly for this - to support all test stuff. 
It's like a collection of utils for all cases in our UT life 😃

unittests/rpc/handlers/AMMInfoTests.cpp
godexsoft: Yes but then you can't see exactly what each test uses and does. Sure there is a lot of duplication here but sometimes that might be better than hiding stuff in base-classes. WDYT?

src/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,245 @@
...
+            auction[JS(discounted_fee)] = auctionSlot[sfDiscountedFee];
+            auction[JS(account)] = to_string(auctionSlot.getAccountID(sfAccount));
+            auction[JS(expiration)] = toIso8601(NetClock::time_point{NetClock::duration{auctionSlot[sfExpiration]}});
+
+            if (auctionSlot.isFieldPresent(sfAuthAccounts)) {
+                boost::json::array auth;
+                for (auto const& acct : auctionSlot.getFieldArray(sfAuthAccounts)) {
+                    boost::json::object accountData;
+                    accountData[JS(account)] = to_string(acct.getAccountID(sfAccount));
+                    auth.push_back(accountData);
```
cindyyan317: std::move(accountData)

src/rpc/AMMHelpers.cpp
godexsoft: I think `pair` is actually not too bad, especially combined with structured bindings at call-site. If we would have 3 or more fields then 100% it should be a struct instead.

src/rpc/AMMHelpers.cpp
godexsoft: I think @arihantkothari just migrated the code as close as possible to the original implementation inside of `rippled`. It looks like all ends meet though?

src/rpc/handlers/AMMInfo.cpp
godexsoft: That may be the case but I'd rather keep the logic/code as close to `rippled` as possible. Let's not add extras for now.

src/rpc/AMMHelpers.cpp
cindyyan317: yea, i believe so

src/rpc/handlers/AMMInfo.h
cindyyan317: ha I see. I thought the error is the same.

cindyyan317 (approved):

kuznetsss (approved):

PR #1064, opened 2024-01-02
upload clio_server artifact

shichengripple001:
upload clio_server artifact for each build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: Linux
+          - os: heavy
             container:
               image: rippleci/clio_ci:latest
           - os: macOS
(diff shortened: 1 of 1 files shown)


kuznetsss (approved):
LGTM

.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: Linux
+          - os: heavy
```
kuznetsss, line 26: This should be changed everywhere not only in one place. I think it would be better to create another PR for that to not mix different changes in one PR.

kuznetsss (changes requested):

.github/workflows/build.yml
kuznetsss: I changed it in my PR #1078 for all workflows so don't worry about it.

kuznetsss (approved):

PR #1071, opened 2024-01-03
New subscription manager

cindyyan317:
SubscriptionManager is puffy now, containing all the subscribers management and feed generation work. It is hard to implement api_version up to it.
This PR will:
1 Separate each stream to the different feed class for better maintainability and testability, each feed focus on its own feed generation. 
2 SubscriptionManager is a dispatcher class now. The API keeps the same as before, except additional api_version parameter for transaction stream. 
   -RPC engine and web server remove the SubscriptionManager dependency.  
   -Executor is injected to SubscriptionManager, remove the waiting from unittests code.
3 Utilize boost/signal2 for sub/unsub/pub
4 Implement api_version
5 Remove cleanup from SubscriptionManager, the subscription will be removed as the web session gone.
6 Keep the original log + log when session removed to better track lifespan
7 Fix issue: Duplicated feed sent if multiple watched accounts get affected in same tx.


src/feed/Types.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+#include <string>
+
+namespace feed {
+using Subscriber = web::ConnectionBase;
+using SubscriberPtr = Subscriber*;
+using SubscriberSharedPtr = std::shared_ptr<Subscriber>;
+
+inline util::prometheus::GaugeInt&
+getSubscriptionsGaugeInt(std::string const& counterName)
```
kuznetsss: I think this function should be somewhere else because it is not related to types. Also it is used only in `feed/impl` so it seems it is a part of implementation.

And I think if possible it would be better to move as much as possible to cpp files to not slow down our compilation process.

godexsoft: Seems like Types.h may be not the best place for a function like this, wdyt?
I'd also try to avoid too many big includes, perhaps in favour of forward declarations.

src/feed/impl/BookChangesFeed.h
```diff
@@ -0,0 +1,56 @@
...
+#include "feed/impl/SingleFeedBase.h"
+#include "rpc/BookChangesHelper.h"
+
+#include <boost/asio/io_context.hpp>
+#include <boost/json/serialize.hpp>
+#include <ripple/protocol/LedgerHeader.h>
+
+#include <vector>
+
+namespace feed::impl {
```
godexsoft, line 32: We usually call the namespace `detail` but the folder is usually `impl`. It would be good to keep it in one style 👍

src/feed/impl/BookChangesFeed.h
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @brief Feed that publishes book changes. This feed will be published every ledger, even if there are no changes.
+ *  Example : {'type': 'bookChanges', 'ledger_index': 2647936, 'ledger_hash':
+ * '0A5010342D8AAFABDCA58A68F6F588E1C6E58C21B63ED6CA8DB2478F58F3ECD5', 'ledger_time': 756395682, 'changes': []}
+ */
+class BookChangesFeed : public SingleFeedBase {
+public:
```
godexsoft: Looks like this could be a `struct`?

src/feed/impl/ForwardFeed.h
```diff
@@ -0,0 +1,45 @@
...
+#include <boost/json/object.hpp>
+#include <boost/json/serialize.hpp>
+
+namespace feed::impl {
+
+/**
+ * @brief Feed that publishes the json object as it is.
+ */
+class ForwardFeed : public SingleFeedBase {
+public:
```
godexsoft: Also could be a `struct` and probably other feeds too

src/feed/impl/LedgerFeed.h
```diff
@@ -0,0 +1,87 @@
...
+#include <boost/asio/io_context.hpp>
+#include <boost/asio/spawn.hpp>
+#include <boost/json/object.hpp>
+#include <boost/json/serialize.hpp>
+#include <ripple/protocol/Fees.h>
+#include <ripple/protocol/LedgerHeader.h>
+
+#include <cstdint>
+#include <memory>
+#include <string>
```
godexsoft, line 35: Let's add a space line between includes and code :)

src/feed/impl/ProposedTransactionFeed.h
```diff
@@ -0,0 +1,124 @@
...
+        : strand_(boost::asio::make_strand(ioContext))
+        , subAllCount_(getSubscriptionsGaugeInt("tx_proposed"))
+        , subAccountCount_(getSubscriptionsGaugeInt("account_proposed"))
+
+    {
+    }
+
+    /**
+     * @brief Subscribe to the proposed transaction feed.
+     */
```
godexsoft, line 77: While `subscriber` is self-speaking, it needs at least `@param subscriber` in order to show up in the doc (as far as i remember). At least we have been adding those for self-explanatory parameters in other files.

src/feed/impl/ProposedTransactionFeed.h
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @brief Publishes the proposed transaction feed.
+     * @param receivedTxJson: The proposed transaction json.
+     */
+    void
+    pub(boost::json::object const& receivedTxJson);
+
+    /**
+     *@brief Get the number of subscribers of the proposed transaction feed.
```
godexsoft: Space 🔢  and below as well

src/main/Main.cpp
```diff
@@ -197,7 +198,7 @@ try {
     auto backend = data::make_Backend(config);
 
     // Manages clients subscribed to streams
-    auto subscriptions = feed::SubscriptionManager::make_SubscriptionManager(config, backend);
+    auto subscriptions = std::make_shared<feed::SubscriptionManagerRunner>(config, backend);
```
kuznetsss: Maybe this doesn't need to be `shared_ptr`?

src/feed/impl/SingleFeedBase.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    return subCount_.value();
+}
+
+void
+SingleFeedBase::unsubInternal(SubscriberPtr subscriber)
+{
+    if (signal_.disconnect(subscriber)) {
+        LOG(logger_.debug()) << subscriber->tag() << "Unsubscribed " << name_;
+        --subCount_;
```
godexsoft: Can't we reuse signal's `count()` somehow or is it better to directly increment/decrement `subCount_`?

src/web/interface/ConnectionBase.h
```diff
@@ -42,6 +43,8 @@ struct ConnectionBase : public util::Taggable {
     std::string const clientIp;
     bool upgraded = false;
     bool isAdmin_ = false;
+    boost::signals2::signal<void(ConnectionBase*)> onDisconnect;
+    std::uint32_t apiSubVersion = 0;
```
kuznetsss, line 47: It seems it is unused

src/web/interface/ConnectionBase.h
```diff
@@ -42,6 +43,8 @@ struct ConnectionBase : public util::Taggable {
     std::string const clientIp;
     bool upgraded = false;
     bool isAdmin_ = false;
+    boost::signals2::signal<void(ConnectionBase*)> onDisconnect;
```
kuznetsss, line 46: Maybe define this type in `feed/Types.h`?

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
-     * @param ioc The io_context to run on
-     */
-    explicit Subscription(boost::asio::io_context& ioc, std::string const& name)
-        : strand_(boost::asio::make_strand(ioc))
-        , subCount_(PrometheusService::gaugeInt(
-              "subscriptions_current_number",
-              util::prometheus::Labels({util::prometheus::Label{"stream", name}}),
-              fmt::format("Current subscribers number on the {} stream", name)
-          ))
+    SubscriptionManager(boost::asio::io_context& ioContext, std::shared_ptr<data::BackendInterface> const& backend)
```
kuznetsss: Probably const should be inside ptr? And maybe you don't need `&` so user could move backend here if they want?
```suggestion
    SubscriptionManager(boost::asio::io_context& ioContext, std::shared_ptr<data::BackendInterface const> backend)
```
And then 
```
, backend_(std::move(backend))
```

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
-
     /**
-     * @brief Adds the given session to the subscribers set.
-     *
-     * @param session The session to add
+     * @brief Subscribe to the book changes feed.
      */
     void
-    subscribe(SessionPtrType const& session);
+    subBookChanges(SubscriberSharedPtr const& subscriber);
```
kuznetsss, line 82: `shared_ptr` doesn't propagate `const` so if you want subscriber to unchanged it should be `std::shared_ptr<const Subscriber>&` (here and in methods below)

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
+        work_.emplace(ioContext_);
+        for (auto i = numThreads; i > 0; --i)
+            workers_.emplace_back([&] { ioContext_.run(); });
+    }
 
-    template <typename Key>
-    void
-    subscribeHelper(SessionPtrType const& session, Key const& k, SubscriptionMap<Key>& subs, CleanupFunction&& func);
+    std::shared_ptr<SubscriptionManager>
+    get()
```
kuznetsss: `get()` returns raw pointer in standard library, so maybe would be better to call it `getManager()`? But maybe it was a bit confusing only for me, so feel free to leave as it is if you want.

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
-    using CleanupFunction = std::function<void(SessionPtrType const)>;
+/**
+ * @brief The help class to run the subscription manager. The container of io_context which is used to publish the
+ * feeds.
+ */
+class SubscriptionManagerRunner {
+    boost::asio::io_context ioContext_;
+    std::shared_ptr<SubscriptionManager> SubscriptionManager_;
+    util::Logger logger_{"Subscriptions"};
+    std::optional<boost::asio::io_context::work> work_;
```
kuznetsss: It is deprecated in boost. Maybe change it to `executor_work_guard` if it is not complicated?

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
 
-private:
-    using CleanupFunction = std::function<void(SessionPtrType const)>;
+/**
+ * @brief The help class to run the subscription manager. The container of io_context which is used to publish the
+ * feeds.
+ */
+class SubscriptionManagerRunner {
+    boost::asio::io_context ioContext_;
+    std::shared_ptr<SubscriptionManager> SubscriptionManager_;
```
kuznetsss: Should be lowercase:
```suggestion
    std::shared_ptr<SubscriptionManager> subscriptionManager_;
```

src/feed/SubscriptionManager.h
```diff
@@ -20,602 +20,280 @@
...
-    // Each time we add a subscriber, we add the opposite lambda that unsubscribes that subscriber when cleanup is
-    // called with the session that closed.
-    std::mutex cleanupMtx_;
-    std::unordered_map<SessionPtrType, std::vector<CleanupFunction>> cleanupFuncs_ = {};
+    ~SubscriptionManagerRunner()
+    {
+        work_.reset();
+        for (auto& worker : workers_)
+            worker.join();
+        SubscriptionManager_.reset();
```
kuznetsss: It will be done automatically on destroying `SubscriptionManagerRunner`.

src/feed/impl/SingleFeedBase.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+void
+SingleFeedBase::sub(SubscriberSharedPtr const& subscriber)
+{
+    auto const weakPtr = std::weak_ptr(subscriber);
+    if (signal_.connectTrackableSlot(subscriber, [weakPtr](std::shared_ptr<std::string> const& msg) {
+            if (auto connectionPtr = weakPtr.lock())
+                connectionPtr->send(msg);
+        })) {
```
kuznetsss: Maybe move this out of `if ()` and put into variable for better readability?
```cpp
bool const connectionAdded = signal_.connectTrackableSlot(subscriber, [weakPtr = std::weak_ptr(subscriber)](std::shared_ptr<std::string> const& msg) {
            if (auto connectionPtr = weakPtr.lock())
                connectionPtr->send(msg);
        });
if (connectionAdded) {
    ...
}
```

src/feed/impl/SingleFeedBase.cpp
```diff
@@ -0,0 +1,86 @@
...
+        ++subCount_;
+        subscriber->onDisconnect.connect([this](SubscriberPtr connectionDisconnecting) {
```
kuznetsss, line 55: Is it possible that `subscriber` will live longer than `SignleFeedBase`?

src/feed/impl/SingleFeedBase.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+void
+SingleFeedBase::sub(SubscriberSharedPtr const& subscriber)
+{
+    auto const weakPtr = std::weak_ptr(subscriber);
+    if (signal_.connectTrackableSlot(subscriber, [weakPtr](std::shared_ptr<std::string> const& msg) {
+            if (auto connectionPtr = weakPtr.lock())
+                connectionPtr->send(msg);
+        })) {
+        ++subCount_;
```
kuznetsss: Maybe add debug log here that subscription happened?

src/feed/impl/LedgerFeed.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+
+namespace feed::impl {
+
+boost::json::object
+LedgerFeed::getLedgerPubMessage(
```
kuznetsss: Maybe better `makeLedgerPubMessage`? Because the method is constructing message, not getting it from somewhere.

src/feed/impl/ProposedTransactionFeed.h
```diff
@@ -0,0 +1,124 @@
...
+ * @brief Feed that publishes the Proposed Transactions.
+ * @note Be aware that the Clio only forwards this stream, not respect api_version.
+ */
+class ProposedTransactionFeed {
+    util::Logger logger_{"Subscriptions"};
+
+    std::unordered_set<SubscriberPtr>
+        notified_;  // Used by slots to prevent double notifications if tx contains multiple subscribed accounts
+    boost::asio::strand<boost::asio::io_context::executor_type> strand_;
+    util::prometheus::GaugeInt& subAllCount_;
```
kuznetsss: Maybe better to use `std::reference_wrapper` everywhere for prometheus counters?

src/feed/impl/ProposedTransactionFeed.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+void
+ProposedTransactionFeed::pub(boost::json::object const& receivedTxJson)
+{
+    auto pubMsg = std::make_shared<std::string>(boost::json::serialize(receivedTxJson));
+
+    auto const transaction = receivedTxJson.at("transaction").as_object();
+    auto const accounts = rpc::getAccountsFromTransaction(transaction);
+    auto affactedAccounts = std::unordered_set<ripple::AccountID>(accounts.cbegin(), accounts.cend());
```
kuznetsss: Typo `affectedAccounts`

src/feed/impl/ProposedTransactionFeed.cpp
```diff
@@ -0,0 +1,138 @@
...
+            .connectTrackableSlot(subscriber, account, [this, weakPtr](std::shared_ptr<std::string> const& msg) {
+                if (auto connectionPtr = weakPtr.lock()) {
+                    // Check if this connection already sent
+                    if (notified_.contains(connectionPtr.get()))
+                        return;
+
+                    notified_.insert(connectionPtr.get());
+                    connectionPtr->send(msg);
+                }
+            })) {
```
kuznetsss: Same here, maybe better to move this out of `if ()`

src/feed/impl/TransactionFeed.h
```diff
@@ -0,0 +1,187 @@
...
+
+                feed.get().notified_.insert(connection.get());
+
+                if (connection->apiSubVersion < 2u) {
+                    connection->send(allVersionMsgs[0]);
+                    return;
+                }
+                connection->send(allVersionMsgs[1]);
+            }
+        }
```
kuznetsss: Could you pease move implementation to cpp file?

unittests/feed/LedgerFeedTests.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+using namespace feed::impl;
+namespace json = boost::json;
+
+using FeedLedgerTest = FeedBaseTest<LedgerFeed>;
+
+TEST_F(FeedLedgerTest, SubPub)
+{
+    backend->setRange(10, 30);
+    auto const ledgerinfo = CreateLedgerInfo(LEDGERHASH, 30);
```
kuznetsss: `ledgerInfo`

unittests/feed/ProposedTransactionFeedTests.cpp
```diff
@@ -0,0 +1,325 @@
...
+struct ProposedTransactionFeedMockPrometheusTest : WithMockPrometheus, SyncAsioContextTest {
+protected:
+    util::TagDecoratorFactory tagDecoratorFactory{util::Config{}};
+    std::shared_ptr<web::ConnectionBase> sessionPtr;
+    std::shared_ptr<ProposedTransactionFeed> testFeedPtr;
+
+    void
+    SetUp() override
+    {
+        WithMockPrometheus::SetUp();
```
kuznetsss: You don't need to do this. `WithMockPrometheus` does everything constructor/destructor. I hope in the future we will refactor tests to not have this SetUp()/TearDown() pattern everywhere.

kuznetsss (changes requested):
It is a really good imrovement, thanks! Just a few comments.

kuznetsss (changes requested):
It is a really good imrovement, thanks! Just a few comments.

src/web/interface/ConnectionBase.h
cindyyan317: It is used , when connection is destroyed , the onDiscconect will be called to clean the subscription.

kuznetsss: I mean `apiSubVersion`.

src/web/interface/ConnectionBase.h
cindyyan317: onDisconnect can be used when other module has something to be cleanup , will not limit its usage.

src/feed/SubscriptionManager.h
cindyyan317: This part keeps the same with the previous interfaces. & is not to increase the ownership count. const is to protect

src/feed/impl/SingleFeedBase.cpp
cindyyan317: FeedClass will live as long as SubscriptionManager which should be alive while Clio process is alive. Individual web connection should not live longer than WebServer which also be alive while Clio process. So I think it is safe here. Good question.

src/feed/impl/TrackableSignalMap.h
```diff
@@ -0,0 +1,110 @@
...
+template <typename T>
+concept Hashable = requires(T a) {
+    {
+        std::hash<T>{}(a)
+    } -> std::convertible_to<std::size_t>;
+};
+
+/**
+ * @brief Class to manage a map of key and its associative signal.
+ * @tparam Key: The type of the key.
```
godexsoft: Let's not add `:` between `@tparam/@param` name and the description. Please also remove them in other places in this PR 👍

src/feed/impl/TrackableSignalMap.h
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param key The key to the signal.
+     * @param args The arguments to be passed to the slot.
+     */
+    void
+    emit(Key const& key, Args... args)
+    {
+        std::scoped_lock const lk(mutex_);
+        if (signalsMap_.contains(key))
+            signalsMap_[key].emit(args...);
```
godexsoft, line 107: Do we always want to copy args?

src/feed/impl/TransactionFeed.h
```diff
@@ -0,0 +1,187 @@
...
+#include <functional>
+#include <memory>
+#include <string>
+#include <unordered_set>
+
+namespace feed::impl {
+
+class TransactionFeed {
+    // Hold two versions of transaction messages
+    using AllVersionTransactionsType = std::array<std::shared_ptr<std::string>, 2>;
```
godexsoft, line 49: This is fine but you could also use `std::tuple` here as you are not really using the `array` for anything other than getting something by index.

src/main/Main.cpp
cindyyan317: For now , both work, because this main never ends. But i am thinking maybe we can control these service/managers lifespan more precious by shared_ptr in future?

src/web/interface/ConnectionBase.h
cindyyan317: used in TransactionFeed. Now TransactionFeed is different according to apiVersion.

src/feed/SubscriptionManager.h
cindyyan317: let us not move backend, it will be shared with other service later.

src/main/Main.cpp
kuznetsss: Maybe we don't need to do it now? Because reading this in the future we may have a question why it is shared_ptr. It is always easy to change it here to unique_ptr/shared_ptr later if we need more lifetime control.

src/feed/SubscriptionManager.h
godexsoft: I agree, i was a bit scared too when i saw main.cpp.

src/feed/impl/TrackableSignalMap.h
godexsoft: You can fix all of them at once in VSCode by using something like the following regex: `@param (\w+): (.*)` -> `@param $1 $2`. And the same thing for `@tparam`.

src/main/Main.cpp
```diff
@@ -206,24 +207,22 @@ try {
     // ETL uses the balancer to extract data.
     // The server uses the balancer to forward RPCs to a rippled node.
     // The balancer itself publishes to streams (transactions_proposed and accounts_proposed)
-    auto balancer = etl::LoadBalancer::make_LoadBalancer(config, ioc, backend, subscriptions, ledgers);
+    auto balancer = etl::LoadBalancer::make_LoadBalancer(config, ioc, backend, subscriptions->get(), ledgers);
```
godexsoft: Can we create a single variable above called `subscriptionManager` and do the `->get()` once?Also, `get()` is a scary name, consider renaming 👍

godexsoft (changes requested):
Great job - huge improvement over the old version 🎉 
I'm adding mostly minor comments 🔢 
Also, as usual, JSON formatting can be improved in tests 🥇

src/feed/impl/BookChangesFeed.h
kuznetsss: Maybe let's create an issue to change `detail` to `impl` everywhere? I think having the same namespaces as directories is more convenient.

src/feed/impl/SingleFeedBase.cpp
cindyyan317: I think subCount_ is more efficient possibly , because we use signal's tracking feature. When it gets count, it will check the weak_ptr one by one to see if connection still there.

src/feed/Types.h
```diff
@@ -19,28 +19,19 @@
...
-#include <fmt/core.h>
-
 #include <memory>
-#include <string>
+
+namespace web {
+
+struct ConnectionBase;
+
+}  // namespace web
```
kuznetsss: `ConnectionBase` is already included.

src/feed/impl/ForwardFeed.h
```diff
@@ -29,7 +29,7 @@ namespace feed::impl {
 /**
  * @brief Feed that publishes the json object as it is.
  */
-class ForwardFeed : public SingleFeedBase {
+struct ForwardFeed : public SingleFeedBase {
 public:
```
kuznetsss: Don't need public now.

src/feed/impl/BookChangesFeed.h
cindyyan317: Issue created https://github.com/XRPLF/clio/issues/1084

src/feed/impl/TrackableSignalMap.h
cindyyan317: Thanks for pointing it out 👍 
emit() as a forwarder does not need to copy the args.

kuznetsss (approved):
👍

src/feed/impl/BookChangesFeed.h
godexsoft: FWIW i like `detail` better but i don't mind if you two prefer `impl`.

src/feed/impl/TransactionFeed.h
cindyyan317: Indeed, tuple and array seem interchangeable here. The feeds for different versions maybe more than 2 in future. Array can save some typing.

godexsoft (approved):
Looks good, even though not all suggestions got implemented in the end, i think this is so much better than what we started with that we should merge it now and improve further later if we so desire.

PR #1079, opened 2024-01-04
Improve batching code

godexsoft:
Batching implemented in #1076 was not really fit for 'forward_range' (needed random access).
This is just a improved version of the same that actually works ok for containers other than `vector`.


kuznetsss (approved):
LGTM

PR #1102, opened 2024-01-12
Fix unstable unittest

cindyyan317:
We had a test failing at Debug build due to the data race happening in the MockSession. The MockSession was not properly mocked. We verified the "message" value from it. Because "message" is not thread safe, we had a SubscriptionManager test failing uncertainly. It has nothing to do with SubscriptionManager , it is an issue of MockSession. This PR changed MockSession to a proper mock. We verified the mock calls, make sure we only test the target we wanna test.

### DIFF
--- a/unittests/feed/FeedTestUtil.h
+++ b/unittests/feed/FeedTestUtil.h
@@ -21,22 +21,25 @@
 
 #include "util/Fixtures.h"
 #include "util/MockWsBase.h"
-#include "util/Taggable.h"
-#include "util/config/Config.h"
 #include "web/interface/ConnectionBase.h"
 
+#include <boost/json/parse.hpp>
 #include <gtest/gtest.h>
 
 #include <memory>
+#include <ostream>
 #include <string>
+#include <utility>
+
+namespace json = boost::json;
 
 // Base class for feed tests, providing easy way to access the received feed
 template <typename TestedFeed>
 class FeedBaseTest : public SyncAsioContextTest, public MockBackendTest {
 protected:
-    util::TagDecoratorFactory tagDecoratorFactory{util::Config{}};
     std::shared_ptr<web::ConnectionBase> sessionPtr;
     std::shared_ptr<TestedFeed> testFeedPtr;
+    MockSession* mockSessionPtr = nullptr;
 
     void
     SetUp() override
@@ -44,7 +47,8 @@ protected:
         SyncAsioContextTest::SetUp();
         MockBackendTest::SetUp();
         testFeedPtr = std::make_shared<TestedFeed>(ctx);
-        sessionPtr = std::make_shared<MockSession>(tagDecoratorFactory);
+        sessionPtr = std::make_shared<MockSession>();
+        mockSessionPtr = dynamic_cast<MockSession*>(sessionPtr.get());
     }
 
     void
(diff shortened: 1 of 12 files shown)


unittests/feed/FeedTestUtil.h
```diff
@@ -55,20 +59,39 @@ class FeedBaseTest : public SyncAsioContextTest, public MockBackendTest {
         MockBackendTest::TearDown();
         SyncAsioContextTest::TearDown();
     }
+};
+
+class SharedStringJsonEqMatcher {
```
cindyyan317, line 63: Custom matcher to verify the send function is called with the expected parameter.

unittests/feed/FeedTestUtil.h
```diff
@@ -21,30 +21,34 @@
...
 
+#include <boost/json/parse.hpp>
 #include <gtest/gtest.h>
 
 #include <memory>
+#include <ostream>
 #include <string>
+#include <utility>
+
+namespace json = boost::json;
```
godexsoft: This is no good in a header

unittests/feed/FeedTestUtil.h
```diff
@@ -55,20 +59,39 @@ class FeedBaseTest : public SyncAsioContextTest, public MockBackendTest {
...
 
-    std::string const&
-    receivedFeedMessage() const
+    void
+    DescribeTo(std::ostream* os) const
     {
-        auto const mockSession = dynamic_cast<MockSession*>(sessionPtr.get());
-        [&] { ASSERT_NE(mockSession, nullptr); }();
-        return mockSession->message;
+        *os << "contains json" << expected_;
```
godexsoft: I assume this is going to produce a message without a space between the label and the actual content.. let's make it readable if so 👍

godexsoft: Also capitalize "Contains" to match what's below 🔢

unittests/feed/FeedTestUtil.h
godexsoft: This is not very important in unit tests but this is a good candidate for `detail` namespace. You can then create it thru the non-detail `SharedStringJsonEq` function.

unittests/feed/LedgerFeedTests.cpp
```diff
@@ -85,20 +76,24 @@ TEST_F(FeedLedgerTest, SubPub)
             "validated_ledgers":"10-31",
             "txn_count":8
         })";
+    EXPECT_EQ(testFeedPtr->count(), 1);
+    EXPECT_CALL(*mockSessionPtr, send(SharedStringJsonEq(ledgerPub))).Times(1);
+    // test publish
```
godexsoft, line 82: Why did the space get removed? 😃

unittests/feed/LedgerFeedTests.cpp
```diff
@@ -127,10 +122,10 @@ TEST_F(FeedLedgerTest, AutoDisconnect)
     ctx.run();
     EXPECT_EQ(testFeedPtr->count(), 1);
 
+    EXPECT_CALL(*mockSessionPtr, send(_)).Times(0);
```
godexsoft, line 125: Space missing. Also the comment is kinda useless - isn't the code self-speaking here?

unittests/feed/LedgerFeedTests.cpp
```diff
@@ -127,10 +122,10 @@ TEST_F(FeedLedgerTest, AutoDisconnect)
...
     // destroy the session
     sessionPtr.reset();
     EXPECT_EQ(testFeedPtr->count(), 0);
-
```
godexsoft, line 133: I disagree. Spaces are good 🤣

godexsoft (changes requested):
Leaving a few small comments 👍

godexsoft (approved):
LGTM 👍

PR #1103, opened 2024-01-12
Enable DB tests via ScyllaDB service

godexsoft:
This adds backend tests to CI. Currently only for the one place where we upload code coverage stats. 
This could be extended to Nightly builds with similar changes but i'm not sure if we need it there.
The way we pass the hostname for ScyllaDB is by using a custom arg to `clio_tests` called `--backend_host`. This gets parsed into `TestGlobals` and is later propagated to all places where we need this host. The host itself is set to `scylladb` because the database is running as a `service` (see [the doc](https://docs.github.com/en/actions/using-containerized-services/about-service-containers)) and so we can leverage **bridge networks** instead of port forwarding.

### DIFF
--- a/.github/actions/code_coverage/action.yml
+++ b/.github/actions/code_coverage/action.yml
@@ -6,7 +6,7 @@ runs:
     - name: Run tests
       shell: bash
       run: |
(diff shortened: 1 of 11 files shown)


cindyyan317 (approved):
LGTM

PR #1108, opened 2024-01-15
Account tx type improvement

cindyyan317:
When new feature adds the new tx_type, Clio always needs to do according changes. The PR removes the hardcoded type. As long as xrpl supports the type, Clio can do filter with it.

### DIFF
--- a/src/rpc/handlers/AccountTx.h
+++ b/src/rpc/handlers/AccountTx.h
@@ -20,13 +20,32 @@
 #pragma once
 
 #include "data/BackendInterface.h"
-#include "rpc/RPCHelpers.h"
+#include "rpc/Errors.h"
+#include "rpc/JS.h"
 #include "rpc/common/JsonBool.h"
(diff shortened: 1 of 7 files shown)


src/rpc/handlers/AccountTx.h
```diff
@@ -131,6 +148,23 @@ class AccountTxHandler {
     process(Input input, Context const& ctx) const;
 
 private:
+    static std::unordered_set<std::string> const&
```
godexsoft: Would this not be useful as a public utility of sorts? That way not only we could reuse it if we need to but also we could unit-test it outside of the handler entirely. WDYT?

godexsoft (dismissed):
This is pretty great. If i understand correctly this would automatically support new types 🔥 I'm adding one suggestion, feel free to ignore if you think this is not needed for this PR (we can always refactor later).

godexsoft (approved):
Nice! 👍

PR #1116, opened 2024-01-16
Ledger entry type filter for account_objects and ledger_data

cindyyan317:
Refactor the ledger entry type filter for account_objects and ledger_data.
Both handlers will use the LedgerUtils functions. 
Unlike transaction type, the ledger entry string which is used by filter is not exported from xrpl.
To support new ledger entry type , we need to update the LEDGER_TYPES_MAP.


src/util/LedgerUtils.cpp
```diff
@@ -0,0 +1,115 @@
...
+        return ripple::ltANY;
+
+    return detail::LEDGER_TYPES_MAP.at(entryName).type;
+}
+
+std::vector<ripple::LedgerEntryType> const&
+getDeletionBlockerLedgerTypes()
+{
+    static std::vector<ripple::LedgerEntryType> const deletionBlockerLedgerTypes = []() {
+        // Move to std::ranges::views::filter when move to higher clang
```
godexsoft: Comments like this should be marked with TODO or FIXME and preferably with a github issue in the comment too. Otherwise we will likely not find them later 😀

unittests/util/TestGlobals.h
```diff
@@ -25,6 +25,13 @@
  * Contains global variables for use in tests.
  */
 struct TestGlobals {
+    TestGlobals(TestGlobals const&) = delete;
```
godexsoft: This is not a great change imo: all CTOR and operators were together before, now they are not

godexsoft: Ah i see what happened there. You may want to merge latest develop - i already have fixed this yesterday 👍

godexsoft (changes requested):
Few small things, otherwise it's good 👍

godexsoft (approved):
Looks good! 👍

PR #1128, opened 2024-01-22
Fixes for gcc 13

Bronek:
This PR fixes compilation errors in gcc-13 by adding `#include <cstdint>` where needed and explains how to avoid compilation errors in `rocksdb` from conan with this compiler.

There's also a drive-by addition of `.devcontainer` to `.gitignore`

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ conan remote add --insert 0 conan-non-prod http://18.143.149.228:8081/artifactor
 ```
 Now you should be able to download prebuilt `xrpl` package on some platforms.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -85,6 +85,8 @@ conan remote add --insert 0 conan-non-prod http://18.143.149.228:8081/artifactor
 ```
 Now you should be able to download prebuilt `xrpl` package on some platforms.
 
+You might need to edit the `~/.conan/remotes.json` file to ensure that this newly added artifactory is listed last. Otherwise you might see compilation errors when building the project with gcc version 13 (or newer).
```
Bronek, line 88: BTW both "might" are intentional - these things will change if we get the rocksdb patch into "our" artifactory.

godexsoft (approved):
Thanks 🙇

PR #1132, opened 2024-01-24
Fix ETL race condition problem

cindyyan317:
During the transition from a non-writer to a writer, if the previous publish has not finished ,Clio will encounter
1  the race condition here->https://github.com/XRPLF/clio/blob/develop/src/etl/impl/LedgerPublisher.h#L140 
Or 
2  the ledger from grpc arrives before the publish finishes, thus the cache not update. Get the exception here ->https://github.com/XRPLF/clio/blob/28c8fa2a9a492bd1fac1f36ecc9ef59d4e9e4f33/src/etl/impl/Transformer.h#L240

The PR adds wait before the transition.
Transition only happens when previous ledgers from DB have publish (update to cache).

### DIFF
--- a/src/etl/impl/LedgerFetcher.h
+++ b/src/etl/impl/LedgerFetcher.h
@@ -91,8 +91,11 @@ public:
 
         auto const isCacheFull = backend_->cache().isFull();
         auto const isLedgerCached = backend_->cache().latestLedgerSequence() >= sequence;
-        if (isLedgerCached)
-            LOG(log_.info()) << "Fetching a ledger which has been cached before." << sequence << " is already cached";
+        if (isLedgerCached) {
+            LOG(log_.info()) << sequence << " is already cached, the current latest seq in cache is "
+                             << backend_->cache().latestLedgerSequence() << " and the cache is "
+                             << (isCacheFull ? "full" : "not full");
+        }
 
         auto response = loadBalancer_->fetchLedger(sequence, true, !isCacheFull || isLedgerCached);
         if (response)
--- a/src/etl/impl/LedgerFetcher.h
+++ b/src/etl/impl/LedgerFetcher.h
@@ -89,9 +89,12 @@ public:
     {
         LOG(log_.debug()) << "Attempting to fetch ledger with sequence = " << sequence;
 
-        auto response = loadBalancer_->fetchLedger(
-            sequence, true, !backend_->cache().isFull() || backend_->cache().latestLedgerSequence() + 1 < sequence
-        );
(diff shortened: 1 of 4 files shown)


kuznetsss (commented):
👍

src/etl/impl/LedgerFetcher.h
```diff
@@ -89,9 +89,15 @@ class LedgerFetcher {
     {
         LOG(log_.debug()) << "Attempting to fetch ledger with sequence = " << sequence;
 
-        auto response = loadBalancer_->fetchLedger(
-            sequence, true, !backend_->cache().isFull() || backend_->cache().latestLedgerSequence() >= sequence
-        );
+        auto const isCacheFull = backend_->cache().isFull();
+        auto const isLedgerCached = backend_->cache().latestLedgerSequence() >= sequence;
```
cjcobb23, line 93: Hmm, the naming of this variable is a bit weird. What this is testing is whether or not the object neighbors can be correctly computed for the desired ledger with the current state of the cache. That computation can only be done correctly if the cached ledger is one less than the desired ledger. So this is actually testing if the ledger is **_not_** cached.

The original computation is a bit weird (probably my fault), because it should really be an inequality computation (`backend_->cache().latestLedgerSequence() != sequence - 1`). I feel like there is some hidden assumption that if the cache is full, then it can't be less than sequence - 1, so the logic amounts to the same thing. But I do feel it's confusing written like this. Double check me though, because I haven't looked at this code in a while. Wanted to chime in though, because I know this part of the code is tricky.

cjcobb23 (commented):
good catch!

src/etl/impl/LedgerFetcher.h
cindyyan317: Thanks for commenting it! Yes, we are not sure about this part, the second condition seems impossible to trigger.  Since it is not the root cause of this issue , we just put more log here to help us investigate. 
`backend_->cache().latestLedgerSequence() != sequence - 1` is not quite right too. Imagine if Clio has many ledgers to catch up, we will fetch these ledgers in parallel. The cache maybe not available for these ledgers when clio is fetching , but the cache will be available when these ledger are  deserialised (because deserialisition happens synchronously).  We don't wanna request more data than we need.

kuznetsss (approved):

src/etl/impl/LedgerFetcher.h
cindyyan317: I am going to resolve the conversation to get this merged. Please do comment in other thread.

Issue #51, opened 2021-11-15
Redesign proxy architecture


cjcobb23:
The proxy architecture is very simply, but may not be the best design. The way it works now is if a request needs to be proxied (either is one of the requests that is always proxied, like `submit`, or specifies `ledger_index:current`), clio picks an etl source at random and forwards the request there. This works, but there are a number of improvements that could be made.

1. Consider disabling proxy for certain requests. For instance, `ledger_data` probably should never be proxied. The call is iterative, and iterating the whole ledger takes minutes, during which time the open ledger changes many times over. The list of RPCs that actually need access to the open ledger is quite small, so we could probably disable proxying on quite a bit.
2. Think about how to provide a consistent view to the open ledger. clio uses a collection of etl sources, and each of those has a different view of the open ledger. Transaction propagation results in a sort of eventual consistency, though there is no guarantee on consistency prior to consensus, at which point the data is validated and clio writes it to the database. When we pick a random p2p node to proxy a request to, subsequent requests are likely to go to a different p2p node. So if a user calls `server_info` and uses the `load_factor` to calculate fees, and then calls `submit` to submit the transaction, `submit` will probably go to a different p2p node than `server_info` did, which could cause incorrect fee calculation. A similar situation can occur with `account_info`. 
3. Have clio be smarter about which node to proxy to. clio could track the load factor of each p2p node, and forward requests to the node with the lowest load factor, or some other measure of idleness. At a certain point, clio should reject proxy requests if the p2p nodes are in danger of going out of sync.
4. Consider caching data in clio. We proxy `server_info` to get the load factor, but we could probably just query each p2p node once per second to get their load factor, and then cache that in clio. We might be able to do something like this for `account_info` as well.
5. Enable separating ETL sources and proxy endpoints. Proxy endpoints could get bombarded with traffic and go out of sync, and then can no longer be ETL sources, and the database will stop getting updates.
6. Make the proxy asynchronous, so we are not sitting on a thread until the response comes back.

Lastly, what we really need is some way to handle a lot of requests that require a p2p node without causing the nodes to desync. Reporting nodes simply slow down when there are too many requests, forcing requests to wait in a queue and eventually returning an error `rpcTOO_BUSY`. p2p nodes on the other hand lose sync, at which point they can't handle any requests until they resync, which can take several seconds or longer.

natenichols:
A couple preliminary thoughts
[...]

PR #1140, opened 2024-01-26
Requests library

kuznetsss:
For #51.

First part of improving forwarding - library for easy async requests.


src/util/requests/WsConnection.cpp
```diff
@@ -0,0 +1,180 @@
...
+template <class StreamDataType>
+Expected<WsConnectionPtr, RequestError>
+WsConnectionBuilder::connectImpl(StreamDataType&& streamData, asio::yield_context yield) const
+{
+    auto context = asio::get_associated_executor(yield);
+    beast::error_code errorCode;
+
+    // These objects perform our I/O
+    asio::ip::tcp::resolver resolver(context);
+    auto& ws = streamData.stream;
```
cindyyan317: maybe we can call this after error check

src/util/requests/WsConnection.cpp
```diff
@@ -0,0 +1,180 @@
...
+    // Set a timeout on the operation
+    beast::get_lowest_layer(ws).expires_after(timeout_);
+
+    // Make the connection on the IP address we get from a lookup
+    auto endpoint = beast::get_lowest_layer(ws).async_connect(results, yield[errorCode]);
+    if (errorCode)
+        return Unexpected{RequestError{"Connect error", errorCode}};
+
+    if constexpr (StreamDataType::sslEnabled) {
+        beast::get_lowest_layer(ws).expires_never();
```
cindyyan317: duplicate with line 156?

src/util/requests/WsConnection.cpp
```diff
@@ -0,0 +1,180 @@
...
+        if (errorCode)
+            return Unexpected{RequestError{"SSL handshake error", errorCode}};
+    }
+
+    // Turn off the timeout on the tcp_stream, because
+    // the websocket stream has its own timeout system.
+    beast::get_lowest_layer(ws).expires_never();
+
+    // Set suggested timeout settings for the websocket
+    ws.set_option(websocket::stream_base::timeout::suggested(beast::role_type::client));
```
cindyyan317, line 154: looks like the timeout_ is timeout for connect only? here we set the timeout to suggested?

src/util/requests/impl/WsConnectionImpl.h
```diff
@@ -0,0 +1,92 @@
...
+#include <boost/beast/websocket/rfc6455.hpp>
+#include <boost/beast/websocket/stream.hpp>
+
+#include <optional>
+#include <string>
+#include <utility>
+
+namespace util::requests::impl {
+
+template <typename StreamType>
```
cindyyan317, line 42: delete the cpp file

src/util/requests/RequestBuilder.h
```diff
@@ -0,0 +1,155 @@
...
+     * @return expected response or error
+     */
+    Expected<std::string, RequestError>
+    get(boost::asio::yield_context yield);
+
+    /**
+     * @brief Perform a POST request asynchronously
+     *
+     * @note It is not thread-safe to call get() and post() of the same RequestBuilder from multiple threads. But it is
+     * fine to call only get() or only post() of the same RequestBuilder from multiple threads.
```
cindyyan317, line 136: curious why it is okay to call get() or post() in multiple threads but not get() and post()?

src/util/requests/RequestBuilder.cpp
```diff
@@ -0,0 +1,205 @@
...
+    beast::get_lowest_layer(stream).socket().shutdown(tcp::socket::shutdown_both, errorCode);
+
+    // not_connected happens sometimes
+    // so don't bother reporting it.
+    //
+    if (errorCode && errorCode != beast::errc::not_connected)
+        return Unexpected{RequestError{"Shutdown socket error", errorCode}};
+
+    if (response.result() != http::status::ok)
+        return Unexpected{RequestError{"Response status not OK"}};
```
cindyyan317: So we report the response error only if we close the connection successfully? Maybe it is okay. Does the shutdown error will shadow the response error?

src/util/requests/RequestBuilder.cpp
```diff
@@ -0,0 +1,205 @@
...
+        return doRequestImpl(std::move(streamData.value()), yield, method);
+    }
+
+    auto streamData = impl::TcpStreamData{yield};
+    return doRequestImpl(std::move(streamData), yield, method);
+}
+
+template <typename StreamDataType>
+Expected<std::string, RequestError>
+RequestBuilder::doRequestImpl(StreamDataType&& streamData, asio::yield_context yield, http::verb const method)
```
cindyyan317, line 147: Do we need && or just & here.

cindyyan317 (dismissed):
We need to think how to test SSL part. Nice to have the clients library. 
Just left some nit comments.

src/util/requests/WsConnection.cpp
kuznetsss: It is not duplicate. It sets timeout for only one logical operation according to the documentation. But in should be `expires_after(timeout_)`.

src/util/requests/WsConnection.cpp
kuznetsss: Yeah, probably I should change the name of the method setting timeout_. Suggested timeout for client role is:
- 30 seconds for handshake
- none for idle
- false for keep alive pings (as I understand server should sends pings and boost replies automatically)

I think it looks fine.

src/util/requests/RequestBuilder.h
kuznetsss: Because inside get() or post() we modify underling request setting method for it. So the same modifications from different threads are fine, but setting method get and post from different threads at the same time is a data race.

src/util/requests/RequestBuilder.cpp
kuznetsss: I can't imagine what could go wrong during connection closing, but let me move status check higher because status error seems to be more important than shutdown error.

src/util/requests/RequestBuilder.cpp
kuznetsss: Hmm, it will work fine with `&`. I thought moving stream is a more safe design for future changes, but maybe there is no difference. I'm ok with both options. Do you think I should change it?

cindyyan317: I personally will change it to &. Because the implementation does not need it being right value for now. But maybe future.

cindyyan317 (dismissed):

src/util/requests/RequestBuilder.cpp
godexsoft: This is a forwarding reference, not an rvalue. I don't think it matters too much which one we use here. What does matter is that below, stream is taken by reference and we need to make sure that async operations don't end up with a dangling reference there. If streamData is moved in, we need to be extra careful i think.

src/util/requests/RequestBuilder.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return doRequest(yield, http::verb::post);
+}
+
+Expected<std::string, RequestError>
+RequestBuilder::doRequest(asio::yield_context yield, beast::http::verb method)
+{
+    if (sslEnabled_) {
+        auto streamData = impl::SslTcpStreamData::create(yield);
+        if (not streamData.has_value())
+            return Unexpected{std::move(streamData.error())};
```
godexsoft: I think this should be `std::move(streamData).error()`. wdyt?

src/util/requests/RequestBuilder.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+        if (!SSL_set_tlsext_host_name(streamData->stream.native_handle(), host_.c_str())) {
+            beast::error_code errorCode;
+            errorCode.assign(static_cast<int>(::ERR_get_error()), asio::error::get_ssl_category());
+            return Unexpected{RequestError{"SSL setup failed", errorCode}};
+        }
+        return doRequestImpl(std::move(streamData.value()), yield, method);
```
godexsoft: Same here and probably in other places where similar `move` is used.

src/util/requests/RequestBuilder.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+    auto& stream = streamData.stream;
+
+    // Make the connection on the IP address we get from a lookup
+    beast::get_lowest_layer(stream).expires_after(timeout_);
+    beast::get_lowest_layer(stream).async_connect(resolverResult, yield[errorCode]);
+    if (errorCode)
+        return Unexpected{RequestError{"Connection error", errorCode}};
+
+    // Set up HTTP method
```
godexsoft: Comments like this one could perhaps be removed 🥇

src/util/requests/WsConnection.cpp
```diff
@@ -0,0 +1,184 @@
...
+            errorCode.assign(static_cast<int>(::ERR_get_error()), beast::net::error::get_ssl_category());
+            return Unexpected{RequestError{"SSL setup failed", errorCode}};
+        }
+        return connectImpl(std::move(streamData.value()), yield);
+    }
+
+    return connectImpl(impl::WsStreamData{yield}, yield);
+}
+
+template <class StreamDataType>
```
godexsoft: Let's try to use `typename` instead of `class` for this. I know we have multiple places where this is not the case but let's at least not add more 🙇

godexsoft (dismissed):
I left just a few nits. I know some of this code is already changed so feel free to ignore anything that already got updated.

godexsoft (approved):
Let's merge this 👍

PR #1142, opened 2024-01-30
Cassandra/Scylla data removal tool

godexsoft:
This tool can be used to delete from most tables in Clio from a specified ledger index and till the latest ledger index in DB.
The ledger specified will become the new latest ledger post deletion.

To build the tool use `cd tools/cassandra_delete_range && go build`.
Try it out with `./cassandra_delete_range --help`.


tools/cassandra_delete_range/main.go
```diff
@@ -0,0 +1,502 @@
...
+
+					var key []byte
+					var seq uint64
+					var rowsRetrieved uint64
+
+					for iter.Scan(&key, &seq) {
+						rowsRetrieved++
+
+						// only grab the rows that are in the correct range of sequence numbers
+						if fromLedgerIdx <= seq && seq <= toLedgerIdx {
```
cindyyan317: We actually can not guarantee the table does not contain the toLedgerIdx+1 . If the Clio stops in the middle of writing , the range table is not updated, but other tables might be wroten.

tools/cassandra_delete_range/main.go
```diff
@@ -0,0 +1,502 @@
...
+	log.Println("Generating delete queries for transactions table")
+	info, traversedRowsCount = prepareDeleteQueries(cluster,
+		fromLedgerIdx, toLedgerIdx,
+		"SELECT hash, ledger_sequence FROM transactions WHERE token(hash) >= %s AND token(hash) <= %s",
+		"DELETE FROM transactions WHERE hash = ?")
+	log.Printf("Total delete queries: %d\n", len(info.Data))
+	log.Printf("Total traversed rows: %d\n\n", traversedRowsCount)
+	performDeleteQueries(cluster, &info, columnSettings{UseBlob: true, UseSeq: false})
+
+	// ledger_transactions queries
```
cindyyan317, line 344: ledger_transactions partition key seems be ledger_sequence , maybe no need whole table scan.

tools/cassandra_delete_range/main.go
```diff
@@ -0,0 +1,502 @@
...
+	log.Println("Generating delete queries for ledger_transactions table")
+	info, traversedRowsCount = prepareDeleteQueries(cluster,
+		fromLedgerIdx, toLedgerIdx,
+		"SELECT hash, ledger_sequence FROM ledger_transactions WHERE token(ledger_sequence) >= %s AND token(ledger_sequence) <= %s",
+		"DELETE FROM ledger_transactions WHERE hash = ? AND ledger_sequence = ?")
+	log.Printf("Total delete queries: %d\n", len(info.Data))
+	log.Printf("Total traversed rows: %d\n\n", traversedRowsCount)
+	performDeleteQueries(cluster, &info, columnSettings{UseBlob: true, UseSeq: true})
+
+	// ledgers queries
```
cindyyan317, line 353: No need to scan the whole ledgers too.

tools/cassandra_delete_range/main.go
godexsoft: Yea, there is little we can do there i think, there is no transactions or anything like that in cassandra.. any suggestions?

tools/cassandra_delete_range/main.go
godexsoft: There are a couple of tables like that. Perhaps we don't need to scan for those, i'll check

tools/cassandra_delete_range/main.go
godexsoft: I misread your comment at first. I think we can just take everything that is newer than `fromLedgerIndex`. 
Originally i thought i tcould be useful to remove any range but i think that may only make more problems 👍

cindyyan317 (dismissed):
LGTM

tools/cassandra_delete_range/main.go
```diff
@@ -0,0 +1,544 @@
...
+						} else if bindCount == 1 {
+							if colSettings.UseSeq {
+								preparedQuery.Bind(r.Seq)
+							} else if colSettings.UseBlob {
+								preparedQuery.Bind(r.Blob)
+							}
+						}
+
+						if err := preparedQuery.Exec(); err != nil {
+							log.Printf("DELETE ERROR: %s\n", err)
```
kuznetsss, line 514: Could you please add more details to the log about error, for example failed query?

And also I think we should log errors to stderr.

kuznetsss (approved):
👍

Issue #953, opened 2023-10-26
Fix json string value conversion


kuznetsss:
There are 187 places in the code where string is getting from json as

```cpp
std::string s = json_value.as_string().c_str();
```

It potentially could cause data loss if string in json contains `'\0'`.
A better way to do it is:
```cpp
std::string s = boost::json::value_to<std::string>(json_value);
```

PR #1144, opened 2024-01-30
Side chain ledgerentry

cindyyan317:
ledger_entry supports :
bridge,
xchain_owned_claim_id,
xchain_owned_create_account_claim_id
More info about the objects: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0038d-cross-chain-bridge

### DIFF
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1355,9 +1355,9 @@ ripple::Issue
 parseIssue(boost::json::object const& issue)
 {
     Json::Value jv;
-    if (issue.contains(JS(issuer)))
+    if (issue.contains(JS(issuer)) && issue.at(JS(issuer)).is_string())
         jv["issuer"] = issue.at(JS(issuer)).as_string().c_str();
-    if (issue.contains(JS(currency)))
+    if (issue.contains(JS(currency)) && issue.at(JS(currency)).is_string())
         jv["currency"] = issue.at(JS(currency)).as_string().c_str();
 
     return ripple::issueFromJson(jv);
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1350,6 +1350,19 @@ parseTaker(boost::json::value const& taker)
         return Status{RippledError::rpcBAD_ISSUER, "invalidTakerAccount"};
     return *takerID;
 }
+
+ripple::Issue
+parseIssue(boost::json::value const& issue)
+{
+    Json::Value jvAsset;
+    if (issue.as_object().contains(JS(issuer)))
+        jvAsset["issuer"] = issue.at(JS(issuer)).as_string().c_str();
+    if (issue.as_object().contains(JS(currency)))
+        jvAsset["currency"] = issue.at(JS(currency)).as_string().c_str();
+
+    return ripple::issueFromJson(jvAsset);
+}
+
 bool
 specifiesCurrentOrClosedLedger(boost::json::object const& request)
 {
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1354,13 +1354,13 @@ parseTaker(boost::json::value const& taker)
 ripple::Issue
 parseIssue(boost::json::value const& issue)
 {
-    Json::Value jvAsset;
+    Json::Value jv;
     if (issue.as_object().contains(JS(issuer)))
-        jvAsset["issuer"] = issue.at(JS(issuer)).as_string().c_str();
+        jv["issuer"] = issue.at(JS(issuer)).as_string().c_str();
     if (issue.as_object().contains(JS(currency)))
-        jvAsset["currency"] = issue.at(JS(currency)).as_string().c_str();
+        jv["currency"] = issue.at(JS(currency)).as_string().c_str();
 
-    return ripple::issueFromJson(jvAsset);
+    return ripple::issueFromJson(jv);
 }
 
 bool
--- a/src/rpc/RPCHelpers.cpp
+++ b/src/rpc/RPCHelpers.cpp
@@ -1352,12 +1352,12 @@ parseTaker(boost::json::value const& taker)
 }
 
 ripple::Issue
-parseIssue(boost::json::value const& issue)
+parseIssue(boost::json::object const& issue)
 {
     Json::Value jv;
-    if (issue.as_object().contains(JS(issuer)))
+    if (issue.contains(JS(issuer)))
         jv["issuer"] = issue.at(JS(issuer)).as_string().c_str();
-    if (issue.as_object().contains(JS(currency)))
+    if (issue.contains(JS(currency)))
         jv["currency"] = issue.at(JS(currency)).as_string().c_str();
 
     return ripple::issueFromJson(jv);
--- a/src/rpc/handlers/LedgerEntry.cpp
+++ b/src/rpc/handlers/LedgerEntry.cpp
@@ -120,8 +120,7 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
 
         if (input.bridgeAccount) {
             auto const bridgeAccount = ripple::parseBase58<ripple::AccountID>(*(input.bridgeAccount));
-            ripple::STXChainBridge::ChainType const chainType =
-                ripple::STXChainBridge::srcChain(bridgeAccount == input.bridge->lockingChainDoor());
+            auto const chainType = ripple::STXChainBridge::srcChain(bridgeAccount == input.bridge->lockingChainDoor());
 
             if (bridgeAccount != input.bridge->door(chainType))
                 return Error{Status{ClioError::rpcMALFORMED_REQUEST}};
--- a/src/rpc/handlers/LedgerEntry.cpp
+++ b/src/rpc/handlers/LedgerEntry.cpp
@@ -30,13 +30,16 @@
 #include <boost/json/value_to.hpp>
 #include <ripple/basics/base_uint.h>
 #include <ripple/basics/strHex.h>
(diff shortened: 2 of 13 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1350,6 +1350,19 @@ parseTaker(boost::json::value const& taker)
         return Status{RippledError::rpcBAD_ISSUER, "invalidTakerAccount"};
     return *takerID;
 }
+
+ripple::Issue
+parseIssue(boost::json::value const& issue)
+{
+    Json::Value jv;
+    if (issue.as_object().contains(JS(issuer)))
```
godexsoft: It may be worth adding an ASSERT that `issue` is indeed an `object`. wdyt?

godexsoft: Or, alternatively, `parseIssue` could expect an `object` to start with.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -111,6 +114,25 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
                   getIssuerFromJson(input.amm->at(JS(asset))), getIssuerFromJson(input.amm->at(JS(asset2)))
         )
                   .key;
+    } else if (input.bridge) {
+        if (!input.bridgeAccount && !input.chainClaimId && !input.createAccountClaimId)
+            return Error{Status{ClioError::rpcMALFORMED_REQUEST}};
+
+        if (input.bridgeAccount) {
+            auto const bridgeAccount = ripple::parseBase58<ripple::AccountID>(*(input.bridgeAccount));
+            ripple::STXChainBridge::ChainType const chainType =
```
godexsoft: Could this be `auto`?

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -502,3 +503,21 @@ TEST_F(RPCHelpersTest, TransactionAndMetadataBinaryJsonV2)
     EXPECT_TRUE(json.contains(JS(tx_blob)));
     EXPECT_TRUE(json.contains(JS(meta_blob)));
 }
+
+TEST_F(RPCHelpersTest, ParseIssue)
+{
+    auto issue = parseIssue(boost::json::parse(
+        R"({
```
godexsoft: Formatting please 🙇

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -502,3 +503,21 @@ TEST_F(RPCHelpersTest, TransactionAndMetadataBinaryJsonV2)
...
+                "issuer": "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun",
+                "currency": "JPY"
+            })"
+    ));
+    EXPECT_TRUE(issue.account == GetAccountIDWithString(ACCOUNT2));
+
+    issue = parseIssue(boost::json::parse(
+        R"({
```
godexsoft: Formatting please 🙇

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -502,3 +503,21 @@ TEST_F(RPCHelpersTest, TransactionAndMetadataBinaryJsonV2)
...
+                "currency": "XRP"
+            })"
+    ));
+    EXPECT_TRUE(ripple::isXRP((issue.currency)));
```
godexsoft: Extra parens are not needed 🔢

godexsoft (changes requested):
Leaving a few nits. Overall looks good! 👍 🥇

src/rpc/RPCHelpers.cpp
cindyyan317: "issue" could be a value instead of the object. Then jv will be empty, the ripple::issueFromJson will handle from here: https://github.com/XRPLF/rippled/blob/90d463b925c377878d3abdb27eb3a1530782a668/src/ripple/protocol/impl/Issue.cpp#L80 I was thinking to handle it ourselves , but my second thought is: since I have called rippled's method to do the validation, why not delegate all jobs to it, including the format check.

godexsoft: i think what will happen is `issue.as_object()` will result in a crash.

godexsoft (changes requested):

src/rpc/RPCHelpers.cpp
cindyyan317: You are right. because the validation check it first, so here we can guarantee it is object.

godexsoft: Yea, the problem is that before we could not call this function/lambda from a new place but now in the future someone could call it without checking it's object first.

cindyyan317: Yea . Have fixed it.

src/rpc/RPCHelpers.cpp
```diff
@@ -1350,6 +1350,19 @@ parseTaker(boost::json::value const& taker)
         return Status{RippledError::rpcBAD_ISSUER, "invalidTakerAccount"};
     return *takerID;
 }
+
+ripple::Issue
+parseIssue(boost::json::object const& issue)
+{
+    Json::Value jv;
+    if (issue.contains(JS(issuer)))
+        jv["issuer"] = issue.at(JS(issuer)).as_string().c_str();
```
godexsoft, line 1359: I don't see anything in validation that checks there are these fields and that they are strings.. in this case i believe if i send an object where the `issuer` is a number or anything but string it will crash same way here.. or at least it will throw an exception and hopefully we catch that on handler level. wdyt?

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -225,7 +246,24 @@ tag_invoke(boost::json::value_to_tag<LedgerEntryHandler::Input>, boost::json::va
...
+        auto const lockingDoor = *ripple::parseBase58<ripple::AccountID>(
+            bridgeJson.at(ripple::sfLockingChainDoor.getJsonName().c_str()).as_string().c_str()
+        );
+        auto const issuingDoor = *ripple::parseBase58<ripple::AccountID>(
+            bridgeJson.at(ripple::sfIssuingChainDoor.getJsonName().c_str()).as_string().c_str()
+        );
+        auto const lockingIssue =
+            parseIssue(bridgeJson.at(ripple::sfLockingChainIssue.getJsonName().c_str()).as_object());
+        auto const issuingIssue =
+            parseIssue(bridgeJson.at(ripple::sfIssuingChainIssue.getJsonName().c_str()).as_object());
```
kuznetsss, line 264: `as_string()` and `as_object()` here could throw exception could it cause a problem?

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -502,3 +504,36 @@ TEST_F(RPCHelpersTest, TransactionAndMetadataBinaryJsonV2)
     EXPECT_TRUE(json.contains(JS(tx_blob)));
     EXPECT_TRUE(json.contains(JS(meta_blob)));
 }
+
+TEST_F(RPCHelpersTest, ParseIssue)
+{
+    auto issue = parseIssue(boost::json::parse(
+                                R"({
```
godexsoft, line 511: Formatting please

unittests/rpc/RPCHelpersTests.cpp
```diff
@@ -502,3 +504,36 @@ TEST_F(RPCHelpersTest, TransactionAndMetadataBinaryJsonV2)
...
+    EXPECT_TRUE(ripple::isXRP(issue.currency));
+
+    EXPECT_THROW(parseIssue(boost::json::parse(R"({"currency": "XRP2"})").as_object()), std::runtime_error);
+
+    EXPECT_THROW(
+        parseIssue(boost::json::parse(
+                       R"({
+                            "issuer": "abcd",
+                            "currency": "JPY"
+                            })"
```
godexsoft, line 531: Formatting please

godexsoft (changes requested):

src/rpc/RPCHelpers.cpp
kuznetsss, line 1359: String conversion could be done better see #953

cindyyan317: This might be an old issue we had. I have add the check here.

kuznetsss (changes requested):

src/rpc/handlers/LedgerEntry.cpp
cindyyan317: The validator will make sure the format is correct here.

src/rpc/RPCHelpers.cpp
cindyyan317: The issuer should be an account, which will be validated. 
Since we have #953 , How about we leave this to a dedicated task to fix them in one place?

src/rpc/handlers/LedgerEntry.cpp
kuznetsss: Does the validator check the internal structure?

cindyyan317: yea , i think it does. Please check CurrencyIssueValidator . And the check happens in rippled::issueFromJson

src/rpc/RPCHelpers.cpp
kuznetsss: I was thinking it is better to note create more work for the future. But maybe it would be solved by creating a new function in utils so maybe it is actually better to be solved all at once. So maybe it is fine to leave it as it is if you don't want to change it.

cindyyan317: Let us leave it another PR. I assume we need to touch lots of places

godexsoft (approved):
👍

kuznetsss (approved):

PR #1147, opened 2024-02-01
feat: Implement MPT changes

shawnxie999:
Spec is here https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033d-multi-purpose-tokens

Builds rippled on https://github.com/gregtatcam/rippled/tree/mpt-direct-pay

- refactor `Issue` and `STAmount` related changes (will need to change if rippled has further changes)
- introduces new table `mp_token_holders`
- ETL that extracts holders into `mp_token_holders` table whenever a new `MPTokenAuthorize` tx is observed
- new `mpt_holders` API
- inject synthetic `mpt_issuance_id` field into `MPTokenIssuanceCreate` transaction metadata for APIs (eg, `tx`, `ledger`, `subscribe`, `account_tx`)
- support new `MPTokenIssuance` and `MPToken` objects types for `ledger_entry` 
- inject synthetic `mpt_issuance_id` field into `MPTokenIssuance` object for `ledger_data` and `account_objects` APIs

### DIFF
--- a/src/data/BackendInterface.hpp
+++ b/src/data/BackendInterface.hpp
@@ -365,6 +365,25 @@ public:
     ) const = 0;
 
     /**
+     * @brief Fetches all holders' balances for a MPTIssuanceID
+     *
+     * @param mptID MPTIssuanceID you wish you query.
+     * @param limit Paging limit.
+     * @param cursorIn Optional cursor to allow us to pick up from where we last left off.
+     * @param ledgerSequence The ledger sequence to fetch for
+     * @param yield Currently executing coroutine.
+     * @return std::vector<Blob> of MPToken balances and an optional marker
+     */
+    virtual MPTHoldersAndCursor
+    fetchMPTHolders(
+        ripple::uint192 const& mptID,
+        std::uint32_t const limit,
+        std::optional<ripple::AccountID> const& cursorIn,
+        std::uint32_t const ledgerSequence,
+        boost::asio::yield_context yield
+    ) const = 0;
+
+    /**
      * @brief Fetches a specific ledger object.
      *
      * Currently the real fetch happens in doFetchLedgerObject and fetchLedgerObject attempts to fetch from Cache first
@@ -616,6 +635,14 @@ public:
      */
     virtual void
     writeNFTTransactions(std::vector<NFTTransactionsData> const& data) = 0;
+
+    /**
+     * @brief Write accounts that started holding onto a MPT.
+     *
+     * @param data A vector of MPT ID and account pairs
+     */
+    virtual void
+    writeMPTHolders(std::vector<MPTHolderData> const& data) = 0;
 
     /**
      * @brief Write a new successor.
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -547,6 +547,45 @@ public:
         return ret;
     }
 
+    MPTHoldersAndCursor
+    fetchMPTHolders(
+        ripple::uint192 const& mptID,
+        std::uint32_t const limit,
+        std::optional<ripple::AccountID> const& cursorIn,
+        std::uint32_t const ledgerSequence,
+        boost::asio::yield_context yield
+    ) const override
+    {
+        auto const holderEntries = executor_.read(
+            yield, schema_->selectMPTHolders, mptID, cursorIn.value_or(ripple::AccountID(0)), Limit{limit}
+        );
+
+        auto const& holderResults = holderEntries.value();
+        if (not holderResults.hasRows()) {
+            LOG(log_.debug()) << "No rows returned";
+            return {};
+        }
+
+        std::vector<ripple::uint256> mptKeys;
+        std::optional<ripple::AccountID> cursor;
+        for (auto const [holder] : extract<ripple::AccountID>(holderResults)) {
+            mptKeys.push_back(ripple::keylet::mptoken(mptID, holder).key);
+            cursor = holder;
+        }
+
+        auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);
+
+        auto it = std::remove_if(mptObjects.begin(), mptObjects.end(), [](Blob const& mpt) { return mpt.size() == 0; });
+
+        mptObjects.erase(it, mptObjects.end());
+
+        ASSERT(mptKeys.size() <= limit, "Number of keys can't exceed the limit");
+        if (mptKeys.size() == limit)
+            return {mptObjects, cursor};
+
+        return {mptObjects, {}};
+    }
+
     std::optional<Blob>
     doFetchLedgerObject(ripple::uint256 const& key, std::uint32_t const sequence, boost::asio::yield_context yield)
         const override
@@ -901,6 +940,16 @@ public:
                 );
             }
         }
+
+        executor_.write(std::move(statements));
+    }
+
+    void
+    writeMPTHolders(std::vector<MPTHolderData> const& data) override
+    {
+        std::vector<Statement> statements;
+        for (auto [mptId, holder] : data)
+            statements.push_back(schema_->insertMPTHolder.bind(std::move(mptId), std::move(holder)));
 
         executor_.write(std::move(statements));
     }
--- a/src/data/Types.hpp
+++ b/src/data/Types.hpp
@@ -234,6 +234,14 @@ struct NFTsAndCursor {
 };
 
 /**
+ * @brief Represents an array of MPTokens
+ */
+struct MPTHoldersAndCursor {
+    std::vector<Blob> mptokens;
+    std::optional<ripple::AccountID> cursor;
+};
+
+/**
  * @brief Stores a range of sequences as a min and max pair.
  */
 struct LedgerRange {
--- a/src/data/cassandra/Schema.hpp
+++ b/src/data/cassandra/Schema.hpp
@@ -257,6 +257,19 @@ public:
             qualifiedTableName(settingsProvider_.get(), "nf_token_transactions")
         ));
 
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                    mpt_id blob,
+                    holder blob,
+                   PRIMARY KEY (mpt_id, holder)
+                  ) 
+             WITH CLUSTERING ORDER BY (holder ASC)
+            )",
+            qualifiedTableName(settingsProvider_.get(), "mp_token_holders")
+        ));
+
         return statements;
     }();
 
@@ -390,6 +403,17 @@ public:
                 VALUES (?, ?, ?)
                 )",
                 qualifiedTableName(settingsProvider_.get(), "nf_token_transactions")
+            ));
+        }();
+
+        PreparedStatement insertMPTHolder = [this]() {
+            return handle_.get().prepare(fmt::format(
+                R"(
+                INSERT INTO {} 
+                       (mpt_id, holder)
+                VALUES (?, ?)
+                )",
+                qualifiedTableName(settingsProvider_.get(), "mp_token_holders")
             ));
         }();
 
@@ -684,6 +708,20 @@ public:
                  LIMIT ?
                 )",
                 qualifiedTableName(settingsProvider_.get(), "issuer_nf_tokens_v2")
+            ));
+        }();
+
+        PreparedStatement selectMPTHolders = [this]() {
+            return handle_.get().prepare(fmt::format(
+                R"(
+                SELECT holder
+                  FROM {}    
+                 WHERE mpt_id = ?
+                   AND holder > ?
+              ORDER BY holder ASC
+                 LIMIT ?
+                )",
+                qualifiedTableName(settingsProvider_.get(), "mp_token_holders")
             ));
         }();
 
--- a/src/etl/MPTHelpers.cpp
+++ b/src/etl/MPTHelpers.cpp
@@ -1,0 +1,78 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include "data/BackendInterface.hpp"
+#include "data/DBHelpers.hpp"
+#include "data/Types.hpp"
+
+#include <fmt/core.h>
+#include <ripple/protocol/STBase.h>
+#include <ripple/protocol/STTx.h>
+#include <ripple/protocol/TxMeta.h>
+
+#include <vector>
+
+namespace etl {
+
+/**
+ * @brief Get the MPToken created from a transaction
+ *
+ * @param txMeta Transaction metadata
+ * @return MPT and holder account pair
+ */
+static std::optional<MPTHolderData>
+getMPTokenAuthorize(ripple::TxMeta const& txMeta)
+{
+    for (ripple::STObject const& node : txMeta.getNodes()) {
+        if (node.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltMPTOKEN)
+            continue;
+
+        if (node.getFName() == ripple::sfCreatedNode) {
+            auto const& newMPT = node.peekAtField(ripple::sfNewFields).downcast<ripple::STObject>();
+            return MPTHolderData{newMPT[ripple::sfMPTokenIssuanceID], newMPT.getAccountID(ripple::sfAccount)};
+        }
+    }
+    return {};
+}
+
+std::optional<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
+{
+    if (txMeta.getResultTER() != ripple::tesSUCCESS || sttx.getTxnType() != ripple::TxType::ttMPTOKEN_AUTHORIZE)
+        return {};
+
+    return getMPTokenAuthorize(txMeta);
+}
+
+std::optional<MPTHolderData>
+getMPTHolderFromObj(std::string const& key, std::string const& blob)
+{
+    ripple::STLedgerEntry const sle =
+        ripple::STLedgerEntry(ripple::SerialIter{blob.data(), blob.size()}, ripple::uint256::fromVoid(key.data()));
+
+    if (sle.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltMPTOKEN)
+        return {};
+
+    auto const mptIssuanceID = sle[ripple::sfMPTokenIssuanceID];
+    auto const holder = sle.getAccountID(ripple::sfAccount);
+
+    return MPTHolderData{mptIssuanceID, holder};
+}
+
+}  // namespace etl
--- a/src/etl/MPTHelpers.hpp
+++ b/src/etl/MPTHelpers.hpp
@@ -1,0 +1,50 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+/** @file */
+#pragma once
+
+#include "data/DBHelpers.hpp"
+
+#include <ripple/protocol/STTx.h>
+#include <ripple/protocol/TxMeta.h>
+
+namespace etl {
+
+/**
+ * @brief Pull MPT data from TX via ETLService.
+ *
+ * @param txMeta Transaction metadata
+ * @param sttx The transaction
+ * @return The MPTIssuanceID and holder pair as a optional
+ */
+std::optional<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx);
+
+/**
+ * @brief Pull MPT data from ledger object via loadInitialLedger.
+ *
+ * @param key The owner key
+ * @param blob Object data as blob
+ * @return The MPTIssuanceID and holder pair as a optional
+ */
+std::optional<MPTHolderData>
+getMPTHolderFromObj(std::string const& key, std::string const& blob);
+
+}  // namespace etl
--- a/src/etl/impl/LedgerLoader.hpp
+++ b/src/etl/impl/LedgerLoader.hpp
@@ -22,6 +22,7 @@
 #include "data/BackendInterface.hpp"
 #include "data/DBHelpers.hpp"
 #include "data/Types.hpp"
+#include "etl/MPTHelpers.hpp"
 #include "etl/NFTHelpers.hpp"
 #include "etl/SystemState.hpp"
 #include "etl/impl/LedgerFetcher.hpp"
@@ -55,6 +56,7 @@ struct FormattedTransactionsData {
     std::vector<AccountTransactionsData> accountTxData;
     std::vector<NFTTransactionsData> nfTokenTxData;
     std::vector<NFTsData> nfTokensData;
+    std::vector<MPTHolderData> mptHoldersData;
 };
 
 namespace etl::impl {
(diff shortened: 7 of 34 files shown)


godexsoft:
@shawnxie999 you need to setup `clangd` locally in your VSCode, or at least some modern version of `clang-format` via an extension. Let me know if you need help to set that up 👍 
Another thing you are probably missing is our githook for checking format on commit. Take a look at our [guide](https://github.com/XRPLF/clio/blob/develop/CONTRIBUTING.md#install-git-hooks).

src/etl/MPTHelpers.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * @brief Pull MPT data from ledger object via loadInitialLedger.
+ *
+ * @param key The owner key
+ * @param blob Object data as blob
+ * @return The MPTIssuanceID and holder pair as a optional
+ */
+std::optional<std::pair<ripple::uint192, ripple::AccountID>>
+getMPTHolderFromObj(std::string const& key, std::string const& blob);
+
+}  // namespace etl
```
godexsoft: In your IDE you can setup to add new line at the end of the file automatically. This will make github happy 👍

src/data/Types.hpp
```diff
@@ -169,6 +169,11 @@ struct NFTsAndCursor {
     std::optional<ripple::uint256> cursor;
 };
 
+struct MPTHoldersAndCursor {
```
godexsoft, line 239: Doxygen comments missing. 
It would also be great if you want to add one for the above NFT stuff that is missing them too but that's up to you 👍

src/data/cassandra/Schema.hpp
```diff
@@ -268,6 +268,21 @@ class Schema {
             settingsProvider_.get().getTtl()
         ));
 
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                    mpt_id blob,
+                    holder  blob,
```
godexsoft: Formatting is not the same as with other existing create statements.. it's very minor but will not pass review 🔢

godexsoft (commented):
Adding a few nits in advance. This is not a real review, just a few things that are immediately visible by glancing thru the code 🔢

src/data/cassandra/Schema.hpp
shawnxie999: Im already on clang format version 17. And it gives me this format when I run `clang-format -i ` on this file

src/data/Types.hpp
shawnxie999: added

src/data/cassandra/Schema.hpp
godexsoft: Clang-format does nothing to user-provided strings. I'm referring to the formatting of the CQL inside the string.

src/data/CassandraBackend.hpp
```diff
@@ -849,6 +894,20 @@ class BasicCassandraBackend : public BackendInterface {
         executor_.write(std::move(statements));
     }
 
+    void
+    writeMPTHolders(std::vector<std::pair<ripple::uint192, ripple::AccountID>>&& data) override
```
cindyyan317: Do we really need r value here? seems no difference if we just pass a const&

shawnxie999: Will change

src/data/CassandraBackend.hpp
```diff
@@ -849,6 +890,20 @@ class BasicCassandraBackend : public BackendInterface {
         executor_.write(std::move(statements));
     }
 
+    void
+    writeMPTHolders(std::vector<std::pair<ripple::uint192, ripple::AccountID>> const& data) override
+    {
+        std::vector<Statement> statements;
+        for (auto const& record : data) {
+            auto const mpt_id = record.first;
```
cindyyan317: we don't use underscores name such variable

src/rpc/RPCHelpers.cpp
```diff
@@ -314,6 +315,65 @@ insertDeliveredAmount(
...
+    std::shared_ptr<ripple::TxMeta const> const& meta
+)
+{
+    if (!canHaveMPTIssuanceID(txn, meta))
+        return false;
+
+    if (auto const amt = getMPTIssuanceID(meta))
+        metaJson[JS(mpt_issuance_id)] = ripple::to_string(*amt);
+
+    return true;
```
cindyyan317: Do we want to return true if amt is emtpy ?

shawnxie999: good catch. will fix

src/rpc/handlers/AccountObjects.cpp
```diff
@@ -124,7 +124,16 @@ tag_invoke(boost::json::value_from_tag, boost::json::value& jv, AccountObjectsHa
         std::cbegin(output.accountObjects),
         std::cend(output.accountObjects),
         std::back_inserter(objects),
-        [](auto const& sle) { return toJson(sle); }
+        [](auto const& sle) {
+            auto sleJson = toJson(sle);
+
+            // if object type if mpt issuance, inject synthetic mpt id
+            if (sle.getType() == ripple::ltMPTOKEN_ISSUANCE)
+                sleJson[JS(mpt_issuance_id)] =
```
cindyyan317: Maybe move it to toJson(ripple::SLE) function if there is no exception 🤔

cindyyan317:
👍  LGTM. Just a few of small things. The CI seems fail to build xrpl , let's fix it and stamp.

godexsoft:
@shawnxie999 please merge develop again when you have a chance. We fixed the detection mechanism for undocumented code and it may show something new

src/data/BackendInterface.hpp
```diff
@@ -337,6 +337,26 @@ class BackendInterface {
         boost::asio::yield_context yield
     ) const = 0;
 
+    /**
+     * @brief Fetches all holders' balances for a MPTIssuanceID
+     *
+     * @param mptID MPTIssuanceID you wish you query.
+     * @param limit Paging limit.
+     * @param cursorIn Optional cursor to allow us to pick up from where we
+     * last left off.
```
godexsoft: This probably does not need to be on a separate line?

src/data/CassandraBackend.hpp
```diff
@@ -552,6 +552,47 @@ class BasicCassandraBackend : public BackendInterface {
...
+            mptKeys.push_back(ripple::keylet::mptoken(mptID, holder).key);
+            cursor = holder;
+        }
+
+        auto const mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);
+
+        // need to filter out the objs that don't exist at the ledger seq because these MPT are in no particular time
+        // order. The MPToken at the specified ledger index may not exist.
+        std::vector<Blob> filteredMpt;
+        std::copy_if(mptObjects.begin(), mptObjects.end(), std::back_inserter(filteredMpt), [](Blob mpt) {
```
godexsoft: Probably no need to copy the Blob in the predicate? 
Also, since we are copying anyway, consider using `std::cbegin(mptObjects)` and `cend` (or the member functions). Alternatively, it might be better to `move` them into `filteredMpt` using `std::move_iterator` (see examples in the codebase).

src/data/CassandraBackend.hpp
```diff
@@ -552,6 +552,47 @@ class BasicCassandraBackend : public BackendInterface {
...
+            return mpt.size() != 0;
+        });
+
+        if (mptKeys.size() == limit)
```
godexsoft, line 583: Can the size ever reach over the limit? Is the code correct in such case?
If it can't (or should not) consider adding an `ASSERT` for it just above.

src/data/CassandraBackend.hpp
```diff
@@ -849,6 +890,20 @@ class BasicCassandraBackend : public BackendInterface {
         executor_.write(std::move(statements));
     }
 
+    void
+    writeMPTHolders(std::vector<std::pair<ripple::uint192, ripple::AccountID>> const& data) override
+    {
+        std::vector<Statement> statements;
+        for (auto const& record : data) {
+            auto const mptId = record.first;
```
godexsoft: Consider using structured bindings in the `for` and then move the copies into `bind`:
```cpp
for (auto [mptId, holder] : data) { ... }
```

src/data/cassandra/Schema.hpp
```diff
@@ -281,6 +281,21 @@ class Schema {
             settingsProvider_.get().getTtl()
         ));
 
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                    mpt_id blob,
+                    holder blob,
+                    PRIMARY KEY (mpt_id, holder)
```
godexsoft: This one is still misaligned by one space compared to others 🔢

src/etl/impl/LedgerLoader.hpp
```diff
@@ -55,6 +56,7 @@ struct FormattedTransactionsData {
     std::vector<AccountTransactionsData> accountTxData;
     std::vector<NFTTransactionsData> nfTokenTxData;
     std::vector<NFTsData> nfTokensData;
+    std::vector<std::pair<ripple::uint192, ripple::AccountID>> mptHoldersData;
```
godexsoft: This pair seems to appear quite a bit. Consider making it a struct perhaps.

unittests/util/TestObject.cpp
```diff
@@ -1001,3 +1001,48 @@ CreateLPTCurrency(std::string_view assetCurrency, std::string_view asset2Currenc
...
+    mptoken.setFieldU32(ripple::sfPreviousTxnLgrSeq, 0);
+
+    if (mptAmount)
+        mptoken.setFieldU64(ripple::sfMPTAmount, mptAmount);
+
+    if (lockedAmount)
+        mptoken.setFieldU64(ripple::sfLockedAmount, lockedAmount);
+
+    return mptoken;
+}
```
godexsoft: Missing final `\n` here.

src/etl/impl/LedgerLoader.hpp
```diff
@@ -256,6 +262,7 @@ class LedgerLoader {
                 backend_->writeAccountTransactions(std::move(insertTxResult.accountTxData));
                 backend_->writeNFTs(insertTxResult.nfTokensData);
                 backend_->writeNFTTransactions(insertTxResult.nfTokenTxData);
+                backend_->writeMPTHolders(std::move(insertTxResult.mptHoldersData));
```
godexsoft: This move is not working as you expected it because your `writeMPTHolders` takes a const ref.

src/etl/impl/Transformer.hpp
```diff
@@ -213,6 +213,7 @@ class Transformer {
         backend_->writeAccountTransactions(std::move(insertTxResultOp->accountTxData));
         backend_->writeNFTs(insertTxResultOp->nfTokensData);
         backend_->writeNFTTransactions(insertTxResultOp->nfTokenTxData);
+        backend_->writeMPTHolders(std::move(insertTxResultOp->mptHoldersData));
```
godexsoft: Same move issue.

src/feed/impl/TransactionFeed.cpp
```diff
@@ -178,6 +178,7 @@ TransactionFeed::pub(
         pubObj[JS(meta)] = rpc::toJson(*meta);
         rpc::insertDeliveredAmount(pubObj[JS(meta)].as_object(), tx, meta, txMeta.date);
         rpc::insertDeliverMaxAlias(pubObj[txKey].as_object(), version);
+        rpc::insertMPTIssuanceID(pubObj[JS(meta)].as_object(), tx, meta);
```
godexsoft, line 206: Maybe take a ref to the `meta` object and reuse? Seems like we already have exactly the same above.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -139,6 +139,14 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
             key = ripple::keylet::xChainCreateAccountClaimID(input.bridge->value(), input.createAccountClaimId.value())
                       .key;
         }
+    } else if (input.mptIssuance) {
+        auto const mptIssuanceID = ripple::uint192{std::string_view(*(input.mptIssuance))};
+        key = ripple::keylet::mptIssuance(mptIssuanceID).key;
+    } else if (input.mptoken) {
+        auto const holder = ripple::parseBase58<ripple::AccountID>(input.mptoken->at(JS(account)).as_string().c_str());
```
godexsoft: Let's try to avoid `.c_str` if possible. 
Use `boost::json::value_to<std::string>()` if you need a string.

godexsoft: And the same for any other places you have `.c_str()` atm. 👍

src/rpc/handlers/MPTHolders.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+    auto const range = sharedPtrBackend_->fetchLedgerRange();
+    auto const lgrInfoOrStatus = getLedgerInfoFromHashOrSeq(
+        *sharedPtrBackend_, ctx.yield, input.ledgerHash, input.ledgerIndex, range->maxSequence
+    );
+    if (auto const status = std::get_if<Status>(&lgrInfoOrStatus))
+        return Error{*status};
+
+    auto const lgrInfo = std::get<LedgerInfo>(lgrInfoOrStatus);
+
```
godexsoft: We probably don't need that much space (between each constant). Maybe try to group them somehow.

src/rpc/handlers/MPTHolders.cpp
```diff
@@ -0,0 +1,147 @@
...
+    output.ledgerIndex = lgrInfo.seq;
+
+    boost::json::array mpts;
+    for (auto const& mpt : dbResponse.mptokens) {
+        ripple::STLedgerEntry const sle{ripple::SerialIter{mpt.data(), mpt.size()}, keylet::mptIssuance(mptID).key};
+        boost::json::object mptJson;
+
+        mptJson[JS(account)] = toBase58(sle[ripple::sfAccount]);
+        mptJson[JS(flags)] = sle.getFlags();
+        mptJson["mpt_amount"] = toBoostJson(ripple::STUInt64{sle[ripple::sfMPTAmount]}.getJson(JsonOptions::none));
```
godexsoft: There are no JS for these atm. i assume. Leaving it here for when they are added - please update here and below 👍

unittests/rpc/handlers/AccountObjectsTests.cpp
```diff
@@ -1629,3 +1629,95 @@ TEST_F(RPCAccountObjectsHandlerTest, LimitMoreThanMax)
...
+
+    auto const handler = AnyHandler{AccountObjectsHandler{backend}};
+    runSpawn([&](auto yield) {
+        auto const output = handler.process(input, Context{yield});
+        ASSERT_TRUE(output);
+        auto const& accountObjects = output->as_object().at("account_objects").as_array();
+        ASSERT_EQ(accountObjects.size(), 1);
+        EXPECT_EQ(accountObjects.front().at("LedgerEntryType").as_string(), "MPToken");
+    });
+}
```
godexsoft: `\n`

unittests/rpc/handlers/LedgerDataTests.cpp
```diff
@@ -726,3 +726,83 @@ TEST_F(RPCLedgerDataHandlerTest, JsonLimitMoreThanMax)
...
+        EXPECT_TRUE(output->as_object().contains("ledger"));
+        EXPECT_EQ(output->as_object().at("state").as_array().size(), 1);
+        EXPECT_EQ(output->as_object().at("marker").as_string(), INDEX2);
+        EXPECT_EQ(output->as_object().at("ledger_hash").as_string(), LEDGERHASH);
+        EXPECT_EQ(output->as_object().at("ledger_index").as_uint64(), RANGEMAX);
+
+        auto const& objects = output->as_object().at("state").as_array();
+        EXPECT_EQ(objects.front().at("LedgerEntryType").as_string(), "MPToken");
+    });
+}
```
godexsoft: And again :)

godexsoft (changes requested):
Found some things to improve 👍 
Another thing is that your test coverage is slightly negative - please take a look at [codecov for your branch](https://app.codecov.io/gh/XRPLF/clio/pull/1147?dropdown=coverage&src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) to see where it's lacking tests. Anything with red `change%` needs a bit more love.

src/data/CassandraBackend.hpp
shawnxie999: i think i'll use std::remove_if instead and mutate mptObjects directly

src/feed/impl/TransactionFeed.cpp
shawnxie999: could you elaborate? `meta` is already a shared pointer.

shawnxie999:
@godexsoft as for test coverage, i believe most of the parts that lack coverage are ETL related

src/data/CassandraBackend.hpp
```diff
@@ -578,19 +578,17 @@ class BasicCassandraBackend : public BackendInterface {
...
-        auto const mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);
+        auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);
 
-        // need to filter out the objs that don't exist at the ledger seq because these MPT are in no particular time
-        // order. The MPToken at the specified ledger index may not exist.
-        std::vector<Blob> filteredMpt;
-        std::copy_if(mptObjects.begin(), mptObjects.end(), std::back_inserter(filteredMpt), [](Blob mpt) {
-            return mpt.size() != 0;
-        });
+        auto it = std::remove_if(mptObjects.begin(), mptObjects.end(), [](Blob mpt) { return mpt.size() == 0; });
```
godexsoft: I think you can avoid a copy in the predicate still 👍

src/feed/impl/TransactionFeed.cpp
godexsoft: Sorry i missed this comment and only seeing it now. I think i meant `pubObj[JS(meta)].as_object()` that you are looking up a second time here

src/data/CassandraBackend.hpp
```diff
@@ -905,6 +944,16 @@ class BasicCassandraBackend : public BackendInterface {
         executor_.write(std::move(statements));
     }
 
+    void
+    writeMPTHolders(std::vector<MPTHolderData> const& data) override
```
cindyyan317, line 948: An edge case, I don't know if it is possible.  If this MPT have 2 payments in one ledger, it will generate 2 MPTHolderData, right?  The executor_.write will call batch write operation which does not guarantee the execution order.  We need to make sure the final one is written.

src/data/CassandraBackend.hpp
```diff
@@ -547,6 +547,45 @@ class BasicCassandraBackend : public BackendInterface {
...
+        }
+
+        auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);
+
+        auto it = std::remove_if(mptObjects.begin(), mptObjects.end(), [](Blob const& mpt) { return mpt.size() == 0; });
+
+        mptObjects.erase(it, mptObjects.end());
+
+        ASSERT(mptKeys.size() <= limit, "Number of keys can't exceed the limit");
+        if (mptKeys.size() == limit)
```
cindyyan317, line 583: The limit is for the all the holders. After filter out the ones not exist based on the given ledger sequence, it is possible that the size is not limit, but we still need a cursor.

src/etl/MPTHelpers.cpp
```diff
@@ -0,0 +1,78 @@
...
+    return {};
+}
+
+std::optional<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
+{
+    if (txMeta.getResultTER() != ripple::tesSUCCESS || sttx.getTxnType() != ripple::TxType::ttMPTOKEN_AUTHORIZE)
+        return {};
+
+    return getMPTokenAuthorize(txMeta);
```
cindyyan317, line 60: Looks like MPTId and new holder are available in tx info. We don't need to loop the meta data.

src/etl/MPTHelpers.cpp
```diff
@@ -0,0 +1,78 @@
...
+            return MPTHolderData{newMPT[ripple::sfMPTokenIssuanceID], newMPT.getAccountID(ripple::sfAccount)};
+        }
+    }
+    return {};
+}
+
+std::optional<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
+{
+    if (txMeta.getResultTER() != ripple::tesSUCCESS || sttx.getTxnType() != ripple::TxType::ttMPTOKEN_AUTHORIZE)
```
cindyyan317, line 57: Payment is not supported yet?

src/rpc/RPCHelpers.cpp
```diff
@@ -312,6 +314,67 @@ insertDeliveredAmount(
...
+ *
+ * @param txn The transaction
+ * @param meta The metadata
+ * @return true if the transaction can have a mpt_issuance_id
+ */
+static bool
+canHaveMPTIssuanceID(std::shared_ptr<ripple::STTx const> const& txn, std::shared_ptr<ripple::TxMeta const> const& meta)
+{
+    ripple::TxType const tt{txn->getTxnType()};
+    if (tt != ripple::ttMPTOKEN_ISSUANCE_CREATE)
```
cindyyan317: We can use txn->getTxnType() directly in if condition?

src/rpc/RPCHelpers.cpp
```diff
@@ -312,6 +314,67 @@ insertDeliveredAmount(
...
+{
+    if (!canHaveMPTIssuanceID(txn, meta))
+        return false;
+
+    if (auto const id = getMPTIssuanceID(meta)) {
+        metaJson[JS(mpt_issuance_id)] = ripple::to_string(*id);
+        return true;
+    }
+
+    return false;
```
cindyyan317, line 376: I think if we reach this line, means we have logic error in **canHaveMPTIssuanceID**. Maybe assert it.
Do we really use the return ?

src/data/CassandraBackend.hpp
```diff
@@ -547,6 +547,45 @@ class BasicCassandraBackend : public BackendInterface {
...
+    MPTHoldersAndCursor
+    fetchMPTHolders(
+        ripple::uint192 const& mptID,
+        std::uint32_t const limit,
+        std::optional<ripple::AccountID> const& cursorIn,
+        std::uint32_t const ledgerSequence,
+        boost::asio::yield_context yield
+    ) const override
+    {
+        auto const holderEntries = executor_.read(
```
cindyyan317, line 559: Actually , Only this part needs to be virtual. The filter logic is same for different DB.  By reducing the scope of virtual functions, you can unittest more of the logic without real DB.

src/etl/MPTHelpers.cpp
shawnxie999: That would not quite work. Because both issuer or holder submit this transaction. When the issuer submits this transaction, it means it could be either authorizing or unauthorizing a holder. When a holder submits this tx, it could either be creating or destroying a MPToken. In our case, we want only want to know the case where _holder creates a new MPToken_, so extracting the MPToken from the metadata is still the most robust approach. Although I know we still could avoid looping through the metadata by writing additional logic, I would prefer using the most robust approach because transactions like MPTokenAuthorize could potential have new features in the future, which could break the ETL logic if it was written in another way.

src/etl/MPTHelpers.cpp
shawnxie999: The only transaction type we need to is extract is MPTokenAuthorize because it is the only one that can create a new MPToken object on the ledger. Payment tx does support MPT transfer between accounts, however that has nothing to do with the mpt_holders API

src/data/CassandraBackend.hpp
shawnxie999: could you clarify what you mean? inside this function, there is another call to the database     `auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);`

cindyyan317: If we create a function in CassandraBackend.hpp to just select MPT holder, then all the filter logic can move to BackendInterface. If we want to add a new database, you don't need to copy all the logic to new database implementation. How to filter the valid holder is database unrelated. Instead, new database only needs to provide how to select MPT holder.

shawnxie999: But aren't existing functions in CassandraBackend.hpp also do something similar? BTW `fetchMPTHolders` doesn't only have filter logic, it does make another call to the `objects` table  on line 576:

`auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);`

cindyyan317: I understood the `fetchMPTHolders` does not only have filter logic. So I suggested that we can add the pure virtual function `doFetchMPTHolders` in BackendInterface, which only obtains the data from `selectMPTHolders` table. Then the filer logic can move to BackendInterface normal function. 

override doFetchfetchMPTHolders in CassandraBackend.hpp

```
    std::pair<std::vector<ripple::uint256>, std::optional<ripple::AccountID>>
    doFetchfetchMPTHolders(
        ripple::uint192 const& mptID,
        std::uint32_t const limit,
        std::optional<ripple::AccountID> const& cursorIn,
    ) const override
    {
        auto const holderEntries = executor_.read(
            yield, schema_->selectMPTHolders, mptID, cursorIn.value_or(ripple::AccountID(0)), Limit{limit}
        );

        auto const& holderResults = holderEntries.value();
        if (not holderResults.hasRows()) {
            LOG(log_.debug()) << "No rows returned";
            return {};
        }

        std::vector<ripple::uint256> mptKeys;
        std::optional<ripple::AccountID> cursor;
        for (auto const [holder] : extract<ripple::AccountID>(holderResults)) {
            mptKeys.push_back(ripple::keylet::mptoken(mptID, holder).key);
            cursor = holder;
        }
        return {mptKeys, cursor};
    }
```

Then your `fetchMPTHolders` will be like:
```
MPTHoldersAndCursor
BackendInterface::fetchMPTHolders(
    ripple::uint192 const& mptID,
    std::uint32_t const limit,
    std::optional<ripple::AccountID> const& cursorIn,
) const
{
    auto [mptKeys, cursor] = doFetchfetchMPTHolders(mptID, limit, cursorIn, yield);

    auto mptObjects = doFetchLedgerObjects(mptKeys, ledgerSequence, yield);

    auto it = std::remove_if(mptObjects.begin(), mptObjects.end(), [](Blob const& mpt) { return mpt.empty(); });

    mptObjects.erase(it, mptObjects.end());

    ASSERT(mptKeys.size() <= limit, "Number of keys can't exceed the limit");
    if (mptKeys.size() == limit)
        return {mptObjects, cursor};

    return {mptObjects, {}};
}
```


I admit that current CassandraBackend.hpp has similar implementation which is not correct. We will refactor them eventually. Feel free to leave it. We can refactor it in future.

cindyyan317 (dismissed):
💯  All comments are addressed. Looks like ready to go. 👍

cmake/deps/libxrpl.cmake
```diff
@@ -1 +1 @@
-find_package(xrpl REQUIRED CONFIG)
+find_package(xrpl-mpt REQUIRED CONFIG)
```
kuznetsss: I think this should be reverted before merging

shawnxie999: Yes these will be changed when it's ready to merge. We'd need wait for the rippled one to merge first

shawnxie999:
updated to rippled v2.3.0 in commit  https://github.com/XRPLF/clio/pull/1147/commits/4863a9fc1b873d1855488a46d0b7ffcd2d92701f

@cindyyan317 @godexsoft current build passes. i think would be ready to merge unless you guys have more feedbacks

src/feed/impl/TransactionFeed.cpp
godexsoft: It's minor so let's ignore it.

godexsoft (approved):
I have not spent too much time reviewing this PR but @cindyyan317 did. Once the new ETL is developed, we will have to move the logic into Extensions anyway and we can improve any nits then. Good job on the implementation @shawnxie999 👍

godexsoft:
@shawnxie999 can you squash the above commits and make sure the resulting one is signed by you? then you can force-push the branch.

PR #1150, opened 2024-02-02
Fix range deletion tool

godexsoft:
Seems like the code that we copied from the example was not dealing with paging very well. The fixes in this PR appear to have helped the situation. Now after re-running the tool it's always 0 deletes across the board except for the ones generated by simply computing for each sequence (see last few tables).

### DIFF
--- a/tools/cassandra_delete_range/main.go
+++ b/tools/cassandra_delete_range/main.go
@@ -42,6 +42,15 @@ var (
 	userName = kingpin.Flag("username", "Username to use when connecting to the cluster").String()
 	password = kingpin.Flag("password", "Password to use when connecting to the cluster").String()
 
+	skipSuccessorTable          = kingpin.Flag("skip-successor", "Whether to skip deletion from successor table").Default("false").Bool()
+	skipObjectsTable            = kingpin.Flag("skip-objects", "Whether to skip deletion from objects table").Default("false").Bool()
+	skipLedgerHashesTable       = kingpin.Flag("skip-ledger-hashes", "Whether to skip deletion from ledger_hashes table").Default("false").Bool()
+	skipTransactionsTable       = kingpin.Flag("skip-transactions", "Whether to skip deletion from transactions table").Default("false").Bool()
+	skipDiffTable               = kingpin.Flag("skip-diff", "Whether to skip deletion from diff table").Default("false").Bool()
+	skipLedgerTransactionsTable = kingpin.Flag("skip-ledger-transactions", "Whether to skip deletion from ledger_transactions table").Default("false").Bool()
+	skipLedgersTable            = kingpin.Flag("skip-ledgers", "Whether to skip deletion from ledgers table").Default("false").Bool()
+	skipWriteLatestLedger       = kingpin.Flag("skip-write-latest-ledger", "Whether to skip writing the latest ledger index").Default("false").Bool()
+
 	numberOfParallelClientThreads = 1           // the calculated number of parallel threads the client should run
 	ranges                        []*tokenRange // the calculated ranges to be executed in parallel
 )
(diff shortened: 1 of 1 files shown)


cindyyan317 (dismissed):
Nice !

tools/cassandra_delete_range/main.go
```diff
@@ -480,11 +546,14 @@ func performDeleteQueries(cluster *gocql.ClusterConfig, info *deleteInfo, colSet
...
-	sessionCreationWaitGroup.Add(len(chunks))
+	wg.Add(numberOfParallelClientThreads)
+	sessionCreationWaitGroup.Add(numberOfParallelClientThreads)
+
+	query := info.Query
+	bindCount := strings.Count(query, "?")
 
-	for i := 0; i < len(chunks); i++ {
-		go func() {
+	for i := 0; i < numberOfParallelClientThreads; i++ {
```
Tapanito: I'm not sure whether these are threads, I think the loop is launching go routines.  It's very small thing, but perhaps renaming the variable to `workerCount` or something similar might avoid confusion.

tools/cassandra_delete_range/main.go
```diff
@@ -424,30 +477,43 @@ func prepareDeleteQueries(cluster *gocql.ClusterConfig, fromLedgerIdx uint64, qu
...
 
-					for iter.Scan(&key, &seq) {
-						rowsRetrieved++
+					for {
+						iter := preparedQuery.PageSize(*clusterPageSize).PageState(pageState).Iter()
+						nextPageState := iter.PageState()
+						scanner := iter.Scanner()
+
+						for scanner.Next() {
+							var key []byte
```
Tapanito: If, for the same query, the keys are guaranteed to be the same size, the it should be safe to move `key` and `seq` outside of the loop stack.

tools/cassandra_delete_range/main.go
```diff
@@ -277,94 +316,108 @@ func deleteLedgerData(cluster *gocql.ClusterConfig, fromLedgerIdx uint64, toLedg
...
-		"SELECT key, seq FROM diff WHERE token(seq) >= %s AND token(seq) <= %s",
-		"DELETE FROM diff WHERE key = ? AND seq = ?")
-	log.Printf("Total delete queries: %d\n", len(info.Data))
-	log.Printf("Total traversed rows: %d\n\n", rowsCount)
-	totalErrors += errCount
-	totalRows += rowsCount
-	deleteCount, errCount = performDeleteQueries(cluster, &info, columnSettings{UseBlob: true, UseSeq: true})
-	totalErrors += errCount
-	totalDeletes += deleteCount
+	if !*skipSuccessorTable {
```
Tapanito, line 319: If run time is important, then it should be possible to run each table in it's own go-routine reporting the metrics via a channel with a struct such as the one below:
```
struct OperationResult {
errCount int64
rowsAffected int64
deleteCount int64
}
```

Tapanito (commented):
Looks good to me

tools/cassandra_delete_range/main.go
```diff
@@ -424,30 +477,43 @@ func prepareDeleteQueries(cluster *gocql.ClusterConfig, fromLedgerIdx uint64, qu
...
 					var rowsRetrieved uint64
 
-					for iter.Scan(&key, &seq) {
-						rowsRetrieved++
+					for {
+						iter := preparedQuery.PageSize(*clusterPageSize).PageState(pageState).Iter()
+						nextPageState := iter.PageState()
+						scanner := iter.Scanner()
+
+						for scanner.Next() {
```
Tapanito, line 495: Perhaps one way to improve the performance of the tool is to also move this internal loop as a separate goroutine. This would allow you to read the table concurrently. Though, that being said, it depends on whether `preparedQuery` supports creating multiple iterators at the same time.

tools/cassandra_delete_range/main.go
godexsoft: Yea, could do. It's what it was in the original example from Scylla (link at the top).

tools/cassandra_delete_range/main.go
godexsoft: Sounds like a nice idea for another day 👍 🙇

tools/cassandra_delete_range/main.go
godexsoft: Not too sure either.. but it's out of scope of this PR. Will consider in the future 👍 🙇

Tapanito (approved):
Looks good to me!

cindyyan317 (approved):

PR #1166, opened 2024-02-08
Upload codecov in separate workflow

### LINKED ISSUE #1165: Upload code coverage report in a separate job
Currently we are uploading xml with code coverage report to artefacts and to codecov in the same job. If the uploading to codecov fails (after all the retries) we have to restart the whole job including compiling clio again. This could be avoided if we move the uploading step to a separate job and use report file from artefacts.

godexsoft:
Fixes #1165


kuznetsss (approved):
👍

PR #1171, opened 2024-02-13
Refactor source

kuznetsss:
For #51


src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+SubscriptionSource::~SubscriptionSource()
+{
+    stop();
+
+    if (wsConnection_) {
+        boost::asio::spawn(strand_, [this](boost::asio::yield_context yield) { wsConnection_->close(yield); });
+    }
+    // To be sure that all the async operations are completed
+    std::future<void> future = boost::asio::post(strand_, boost::asio::use_future);
+    future.get();
```
kuznetsss: This could hang if there is no `io_context` running when the source is being destroyed. This is done to avoid having scheduled async operations related to the source when the source has been already destroyed.

Currently it doesn't affect anything because the source dies only on clio stop.

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+        static constexpr char const* const JS_ValidatedLedgers = "validated_ledgers";
+        static constexpr char const* const JS_LedgerClosed = "ledgerClosed";
+        static constexpr char const* const JS_Type = "type";
+        static constexpr char const* const JS_Transaction = "transaction";
+        static constexpr char const* const JS_ValidationReceived = "validationReceived";
+        static constexpr char const* const JS_ManifestReceived = "manifestReceived";
+
+        if (object.contains(JS_Result)) {
+            auto const& result = object.at(JS_Result).as_object();
+            if (result.contains("ledger_index"))
```
cindyyan317: JS_LedgerIndex

src/etl/impl/ForwardingSource.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "util/log/Logger.hpp"
+#include "util/requests/WsConnection.hpp"
+
+#include <boost/asio/spawn.hpp>
+#include <boost/json/object.hpp>
+
+#include <chrono>
+#include <optional>
+#include <string>
```
godexsoft, line 30: nit: Let's add a space please

src/etl/impl/GrpcSource.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <memory>
+#include <sstream>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace etl::impl {
+
+GrpcSource::GrpcSource(std::string const& ip, std::string const& grpcPort, std::shared_ptr<BackendInterface> backend)
+    : log_(fmt::format("ETL_Grpc[{}:{}]", ip, grpcPort)), backend_(std::move(backend))
```
godexsoft, line 49: The channel of the logger is not meant to be used like that but if this still shows up in logs then that's ok.
Normally you'd have to register the channel somewhere in the Logger code.

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+        onDisconnect_();
+        isForwarding_ = false;
+    }
+
+    if (wsConnection_ != nullptr) {
+        auto const error = wsConnection_->close(yield);
+        if (error) {
+            LOG(log_.error()) << "Error closing websocket connection: " << error->message();
+        }
+        wsConnection_.reset();
```
cindyyan317, line 246: Does here have data race problem with the reset in destructor?

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+        auto const writeErrorOpt = wsConnection_->write(subscribeCommand, yield);
+        if (writeErrorOpt) {
+            handleError(writeErrorOpt.value(), yield);
+            return;
+        }
+
+        retry_.reset();
+
+        while (!stop_) {
+            auto message = wsConnection_->read(yield);
```
cindyyan317: can it be const ?

src/etl/impl/SubscriptionSource.cpp
kuznetsss: This happens in coroutine and in the destructor we wait for all async operations to complete before doing reset, so I think there is no data race. But if you see some other dangerous case please describe it.

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+SubscriptionSource::handleMessage(std::string const& message)
+{
+    setLastMessageTime();
+
+    try {
+        auto const raw = boost::json::parse(message);
+        auto const object = raw.as_object();
+        uint32_t ledgerIndex = 0;
+
+        static constexpr char const* const JS_Result = "result";
```
godexsoft: Can't we use JS() for these or they don't exist in libxrpl?

src/etl/impl/SubscriptionSource.cpp
godexsoft: Stray one left here

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+        while (!stop_) {
+            auto message = wsConnection_->read(yield);
+            if (not message) {
+                handleError(message.error(), yield);
+                return;
+            }
+
+            auto handleErrorOpt = handleMessage(message.value());
+            if (handleErrorOpt) {
+                handleError(handleErrorOpt.value(), yield);
```
cindyyan317: I am thinking if handleMessage raises error , do we need to retry connect? How does retry help?

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -0,0 +1,336 @@
...
+        auto& errorCode = error.errorCode().value();
+        std::string errorString = fmt::format(
+            "({},{}) ",
+            boost::lexical_cast<std::string>(ERR_GET_LIB(errorCode.value())),
+            boost::lexical_cast<std::string>(ERR_GET_LIB(errorCode.value()))
+        );
+
+        static constexpr size_t BUFFER_SIZE = 128;
+        char buf[BUFFER_SIZE];
+        ::ERR_error_string_n(errorCode.value(), buf, sizeof(buf));
```
godexsoft: Is there really no C++ API for this? this block of code looks horrible 🤣

src/etl/impl/SubscriptionSource.hpp
```diff
@@ -0,0 +1,206 @@
...
+        boost::asio::io_context& ioContext,
+        std::string const& ip,
+        std::string const& wsPort,
+        std::shared_ptr<NetworkValidatedLedgersType> validatedLedgers,
+        std::shared_ptr<SubscriptionManagerType> subscriptions,
+        OnDisconnectHook onDisconnect,
+        std::chrono::steady_clock::duration const connectionTimeout = CONNECTION_TIMEOUT,
+        std::chrono::steady_clock::duration const retryDelay = RETRY_DELAY
+    )
+        : log_(fmt::format("GrpcSource[{}:{}]", ip, wsPort))
```
godexsoft, line 115: Channel again

cindyyan317, line 115: Do we enable this log?

src/main/Main.cpp
```diff
@@ -160,6 +160,8 @@ start(io_context& ioc, std::uint32_t numThreads)
         v.emplace_back([&ioc] { ioc.run(); });
 
     ioc.run();
+    for (auto& t : v)
```
godexsoft, line 163: This might be unnecessary but i'm not opposed to adding it just for completeness sake. 
I hope `jthread` will be available soon.

src/etl/LoadBalancer.cpp
```diff
@@ -128,11 +112,18 @@ LoadBalancer::LoadBalancer(
         }
 
         sources_.push_back(std::move(source));
-        LOG(log_.info()) << "Added etl source - " << sources_.back()->toString();
+        chooseForwardingSource();
```
cindyyan317: This line does nothing. Because the sources all disconnected at this moment . They start to connect at line 124.

godexsoft (commented):
Commenting on a few small things here and there. Huge improvement overall and very much looking forward to merging this 👍

src/etl/impl/SubscriptionSource.cpp
kuznetsss: I was thinking if there is an error on any stage we should start from scratch. There is a hope that if we started to receive wrong messages it could be better with a fresh connection. I think the same logic was in the old code.

src/etl/LoadBalancer.cpp
kuznetsss: Yeah, thanks!

src/etl/impl/GrpcSource.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        LOG(log_.trace()) << "Marker prefix = " << ptr->getMarkerPrefix();
+
+        auto result = ptr->process(stub_, cq, *backend_, abort, cacheOnly);
+        if (result != etl::impl::AsyncCallData::CallStatus::MORE) {
+            ++numFinished;
+            LOG(log_.debug()) << "Finished a marker. "
+                              << "Current number of finished = " << numFinished;
+
+            std::string const lastKey = ptr->getLastKey();
```
cindyyan317: auto can be used here , I think it has been like this for long. Maybe we can put this line inside the if.

src/etl/LoadBalancer.cpp
```diff
@@ -293,4 +269,15 @@ LoadBalancer::getETLState() noexcept
...
 
+void
+LoadBalancer::chooseForwardingSource()
+{
+    for (auto& source : sources_) {
+        if (source.isConnected()) {
+            source.setForwarding(true);
+            break;
+        }
+    }
```
kuznetsss, line 290: There could be a situation when all the sources got disconnected and I can't see how one of them could become forwarding again. Probably there should be `onConnect` callback as well.

src/etl/impl/GrpcSource.cpp
```diff
@@ -0,0 +1,152 @@
...
+        if (result == etl::impl::AsyncCallData::CallStatus::ERRORED)
+            abort = true;
+
+        if (backend_->cache().size() > progress) {
+            LOG(log_.info()) << "Downloaded " << backend_->cache().size() << " records from rippled";
+            progress += incr;
+        }
+    }
+
+    LOG(log_.info()) << "Finished loadInitialLedger. cache size = " << backend_->cache().size();
```
cindyyan317: Not the issue of this PR, but since u have been there, can we update this log to something can signal "abort"?

src/etl/LoadBalancer.cpp
cindyyan317: I agree. Now the forwarding is set when disconnect , right?

CMakeLists.txt
```diff
@@ -96,13 +96,14 @@ target_sources (clio PRIVATE
   src/data/cassandra/Handle.cpp
   src/data/cassandra/SettingsProvider.cpp
   ## ETL
-  src/etl/Source.cpp
-  src/etl/ProbingSource.cpp
   src/etl/NFTHelpers.cpp
   src/etl/ETLService.cpp
   src/etl/ETLState.cpp
   src/etl/LoadBalancer.cpp
-  src/etl/impl/ForwardCache.cpp
```
cindyyan317, line 106: Remove from folder

src/etl/impl/GrpcSource.cpp
kuznetsss: Sure, do you mean to add logging `abort` value?

src/etl/impl/ForwardingSource.cpp
```diff
@@ -0,0 +1,98 @@
...
+    : log_(fmt::format("ForwardingSource[{}:{}]", ip, wsPort)), connectionBuilder_(std::move(ip), std::move(wsPort))
+{
+    connectionBuilder_.setConnectionTimeout(connectionTimeout)
+        .addHeader(
+            {boost::beast::http::field::user_agent, fmt::format("{} websocket-client-coro", BOOST_BEAST_VERSION_STRING)}
+        );
+}
+
+std::optional<boost::json::object>
+ForwardingSource::forwardToRippled(
```
cindyyan317, line 55: We don't have forward cache now?

src/etl/impl/SubscriptionSource.cpp
cindyyan317: The coroutine posted at line 66 can be executed after wsConnection_. reset in handlerError,  right?  Both are coroutine. The one in handlerError gets executed first.

src/etl/impl/ForwardingSource.cpp
kuznetsss: Here we don't. I will add cache in the next PR. The old one was not working anyway.

src/etl/impl/SubscriptionSource.cpp
kuznetsss: No it can't because at the lines 69-70 we are posting future to the same strand and waiting for it to be ready.
Also because above we call `stop()` subscribe loop will just end without retrying. 

But now I'm not really sure that `io_context` will take the future as the last task. I will try to reproduce this in tests to check.

kuznetsss: After a test I confirmed that the future is not the last task when using boost coroutines. I will change it to use `boost::asio::use_future`.

src/etl/impl/SubscriptionSource.hpp
kuznetsss: I added the support for it to Logger.

godexsoft (approved):
Nice work! 🚀

PR #1172, opened 2024-02-14
Use json value_to<string> to do the string convert

### LINKED ISSUE #953: Fix json string value conversion

cindyyan317:
Fix #953 

Replace c_str() with value to, except the places where require a C string, like EXPECT_STREQ

### DIFF
--- a/src/etl/impl/CacheLoader.hpp
+++ b/src/etl/impl/CacheLoader.hpp
@@ -41,6 +41,7 @@
 #include <boost/json/object.hpp>
 #include <boost/json/parse.hpp>
 #include <boost/json/value.hpp>
+#include <boost/json/value_to.hpp>
 #include <grpcpp/grpcpp.h>
 #include <ripple/basics/base_uint.h>
 #include <ripple/basics/strHex.h>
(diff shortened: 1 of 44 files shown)


kuznetsss:
`EXPECT_STREQ()` was used because of `c_str()`. For example https://github.com/XRPLF/clio/blob/cce695c570057c497198f61d790e7a5c795a02d9/unittests/rpc/handlers/NFTBuyOffersTests.cpp#L487-L489

Could be replaced by 
```cpp
 EXPECT_EQ( 
     boost::json::value_to<std::string>(output->at("marker")), 
     "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC353" 
 ); 
```

cindyyan317:
> EXPECT_STREQ

I hesitated to change this one , because the case itself will ensure there is no '\0'. I don't bother to make it more unified. Will add this

kuznetsss (dismissed):
👍

godexsoft (dismissed):
👍

kuznetsss (approved):

PR #1185, opened 2024-02-15
Turn on macOS runners in CI

kuznetsss:


godexsoft (dismissed):
Changes look good.. let's hope the runner can be setup correctly soon 👍

godexsoft (approved):

Issue #1059, opened 2023-12-19
Split the README


godexsoft:
Currently our README.md is giant. It contains lots of useful information but it may be easier to navigate if we split it into separate files with links to them from the main README. Feel free to update any of the wording too.

maria-robobug:
I will start looking into this, not sure if I can assign the issue to myself - but just putting a note here so you're aware I'm working on it.

PR #1191, opened 2024-02-16
[1059] - [DOCS] Split README into separate files

maria-robobug:
https://github.com/XRPLF/clio/issues/1059

This change splits the `README.md` into different files, under the `/docs` directory.


maria-robobug:
@kuznetsss / @godexsoft here is the PR for splitting the README.

I can't seem to add any reviewers to the PR, maybe I don't have permission or something?

kuznetsss:
Thanks! Readme looks much better now!

A few things could be improved:
- I think we should have two sections:
  - for operators: how to configure, run and use Clio
  - for developers: how to build Clio and use developers tools (like static analysis or code coverage)
- Maybe there could be emojis in readme to make it more fresh and friendly, for example like I did in [my repo](https://github.com/kuznetsss/meadow.nvim) 
- Maybe small pages should be merged with some others, for example [database.md](https://github.com/maria-robobug/clio/blob/1059-split-readme/docs/database.md) is only a few sentences
- I'm not sure whether it should be done in this PR but probably we should add a link to the [API reference](https://xrpl.org/http-websocket-apis.html) in our readme

maria-robobug:
@kuznetsss thanks for the review! Great points, I'll update with the feedback changes for this PR (including API link).

godexsoft (approved):
Looking great! 🔥 🚀 Huge improvement. 
Love the emojis too 👍

PR #1193, opened 2024-02-20
Rewrite cache loader to async framework

godexsoft:
First rewrite with the async framework. :rage3:


src/etl/CacheLoader.hpp
```diff
@@ -0,0 +1,127 @@
...
+        ASSERT(not cache_.get().isFull(), "Cache must not be full. seq = {}", seq);
+
+        if (settings_.isDisabled()) {
+            cache_.get().setDisabled();
+            LOG(log_.warn()) << "Cache is disabled. Not loading";
+            return;
+        }
+
+        auto const provider = CursorProviderType{backend_, settings_.numCacheDiffs};
+        loader_ = std::make_unique<CacheLoaderType>(
```
kuznetsss, line 92: Is it ok that if `load()` is called 2 times with different sequences `loader_` will be recreated?

src/etl/CacheLoaderSettings.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <cstddef>
+
+namespace etl {
+
+/**
+ * @brief Settings for the cache loader
+ */
+struct CacheLoaderSettings {
+    enum class LoadStyle { ASYNC, SYNC, NOT_AT_ALL };
```
kuznetsss: Maybe `NONE` instead of `NOT_AT_ALL`?

src/etl/CacheLoaderSettings.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+    size_t numCacheDiffs = 32;       /**< number of diffs to use to generate cursors */
+    size_t numCacheMarkers = 48;     /**< number of markers to use at one time to traverse the ledger */
+    size_t cachePageFetchSize = 512; /**< number of ledger objects to fetch concurrently per marker */
+    size_t numThreads = 2;           /**< number of threads to use for loading cache */
+
+    LoadStyle loadStyle = LoadStyle::ASYNC; /**< how to load the cache */
+
+    int
```
kuznetsss: Why `int`?

src/etl/CacheLoaderSettings.hpp
```diff
@@ -0,0 +1,60 @@
...
+    operator<=>(CacheLoaderSettings const&) const = default;
```
kuznetsss, line 42: Maybe `operator==() const = default` would be enough? Do you need operators `<`,`>` and others?

src/etl/impl/CacheLoader.hpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2023, the clio developers.
+    Copyright (c) 2024, the clio developers.
```
kuznetsss: Are you sure? I thought this year is the year of the first appearance of this file.

src/etl/impl/CacheLoader.hpp
```diff
@@ -20,423 +20,133 @@
...
-            LOG(log_.info()) << "Finished downloading ledger from clio node. ip = " << ip;
-
-            cache_.get().setFull();
-            return true;
-        } catch (std::exception const& e) {
-            LOG(log_.error()) << "Encountered exception : " << e.what() << " - ip = " << ip;
-            return false;
-        }
+        LOG(log_.info()) << "Loading cache. Num cursors = " << queue_.size();
+        for ([[maybe_unused]] auto taskId : vs::iota(0u, numCacheMarkers))
```
kuznetsss, line 108: `tasks_.reserve(numCacheMarkers)`

src/etl/impl/CursorProvider.hpp
```diff
@@ -0,0 +1,113 @@
...
+            std::unique(
+                std::begin(diffs), std::end(diffs), [](auto const& a, auto const& b) { return a.key == b.key; }
+            ),
+            std::end(diffs)
+        );
+
+        std::vector<ripple::uint256> cursors{data::firstKey};
+        rg::copy(
+            diffs                                                                   //
+                | vs::filter([](auto const& obj) { return not obj.blob.empty(); })  //
```
kuznetsss, line 89: `//` without a comment

src/util/async/impl/ErasedOperation.hpp
```diff
@@ -61,8 +61,8 @@ class ErasedOperation {
     }
 
     /**
-     * @brief Request the operation to be stopped as soon as possible.
-     * @throw std::logic_error if the erased operation is non-stoppable
+     * @brief Request the operation to be stopped as soon as possible
+     * Note: ASSERTs if the operation is not stoppable
```
kuznetsss: `@note`

unittests/etl/CacheLoaderTests.cpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2023, the clio developers.
+    Copyright (c) 2024, the clio developers.
```
kuznetsss: Also I think year shouldn't be changed.

unittests/etl/CursorProviderTests.cpp
```diff
@@ -0,0 +1,87 @@
...
+    auto const provider = etl::impl::CursorProvider{backend, 0};
+
+    ON_CALL(*backend, fetchLedgerDiff(_, _)).WillByDefault(Return(diffs));
+    EXPECT_CALL(*backend, fetchLedgerDiff(_, _)).Times(0);
+
+    auto const cursors = provider.getCursors(SEQ);
+
+    EXPECT_EQ(cursors.front().start, firstKey);
+    EXPECT_EQ(cursors.back().end, lastKey);
+    EXPECT_EQ(cursors.size(), 1);
```
kuznetsss: Better to check the size using `ASSERT_EQ()` before calling `front()` and `back()`

unittests/etl/CursorProviderTests.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+    auto const numDiffs = GetParam();
+    auto const diffs = diffProvider.getLatestDiff();
+    auto const provider = etl::impl::CursorProvider{backend, numDiffs};
+
+    ON_CALL(*backend, fetchLedgerDiff(_, _)).WillByDefault(Return(diffs));
+    EXPECT_CALL(*backend, fetchLedgerDiff(_, _)).Times(numDiffs);
+
+    auto const cursors = provider.getCursors(SEQ);
+    EXPECT_EQ(cursors.size(), diffs.size() + 1);
```
kuznetsss: Better `ASSERT_EQ()`

kuznetsss (commented):
The code looks much better now! Just small comments

src/etl/CacheLoader.hpp
godexsoft: Yes, that's the idea. The previous one would get cancelled. But tbh., at that point we already have some data in the cache. As long as it's not `full` it should be ok to start over.

src/etl/CacheLoaderSettings.hpp
godexsoft: My bad, should be `auto` or one of the custom comparison categories, will change

src/etl/CacheLoaderSettings.hpp
godexsoft: Meh, why use `==` if you have spaceship :)

src/etl/impl/CursorProvider.hpp
godexsoft: Yea, that's to align the pipes properly. The alternative is to use `// clang-format off` etc. - too much extra comments then

kuznetsss (approved):

PR #1195, opened 2024-02-20
Cleanup location of docs and example files

maria-robobug:
- Deletes `RELEASENOTES.md` as it is not needed.
- Moves example configs to the `/examples/config` directory.

I noticed that `example-config.json` is used in `install.cmake` and the `Dockerfile`.
I have updated the paths, but don't want to break anything, so please let me know if I should steer clear 😄 !


godexsoft (changes requested):
This is a good start 🔥 
I think maybe we can move the entire `examples` folder into `docs`, wdyt?

maria-robobug:
> I think maybe we can move the entire examples folder into docs, wdyt?

Yep that sounds good, I'll move them in there and update any references 👍🏾.

godexsoft (approved):
Looks nice! Thanks 🥇

PR #1202, opened 2024-02-22
Update issue templates

godexsoft:
Adding a few templates for issues - mostly copying the format of `rippled` to be consistent and align with github's project guidelines.


kuznetsss (approved):
Looks fine, it's hard to say whether it is the best format, but it is definitely a good improvement 👍

PR #1204, opened 2024-02-22
Add forwarding cache

### LINKED ISSUE #51: Redesign proxy architecture

kuznetsss:
Fixes #51.


docs/examples/config/example-config.json
```diff
@@ -32,7 +32,8 @@
         {
             "ip": "127.0.0.1",
             "ws_port": "6006",
-            "grpc_port": "50051"
+            "grpc_port": "50051",
+            "cache_timeout_ms": 250 // could be 0, which means no cache
```
godexsoft: Should this be documented somewhere in .md?

src/etl/Source.cpp
```diff
@@ -59,7 +62,12 @@ make_Source(
         std::move(onConnect),
         std::move(onDisconnect)
     );
-    impl::ForwardingSource forwardingSource{ip, wsPort};
+
+    auto const cacheEntryTimeout = config.valueOr<uint64_t>("cache_entry_timeout", 0);
```
godexsoft: Was this meant to be `cache_timeout_ms`? Or the example config is wrong - must be one of the two i guess :)

src/etl/impl/ForwardingCache.cpp
```diff
@@ -0,0 +1,127 @@
...
+}
+
+void
+ForwardingCache::put(boost::json::object const& request, boost::json::object const& response)
+{
+    auto const command = getCommand(request);
+    if (not command.has_value()) {
+        return;
+    }
+    auto entry = cache_[*command].lock<std::unique_lock>();
```
godexsoft, line 121: Potentially modifying `cache_` without protection here?

src/etl/impl/ForwardingSource.cpp
```diff
@@ -92,6 +103,10 @@ ForwardingSource::forwardToRippled(
 
     auto responseObject = parsedResponse.as_object();
     responseObject["forwarded"] = true;
+
+    if (cache_ and ForwardingCache::shouldCache(request) and not responseObject.contains("error"))
```
godexsoft: Perhaps check for "error" could be first?

src/etl/impl/ForwardingSource.hpp
```diff
@@ -34,12 +35,15 @@ namespace etl::impl {
 class ForwardingSource {
     util::Logger log_;
     util::requests::WsConnectionBuilder connectionBuilder_;
+    mutable std::optional<ForwardingCache> cache_;
```
godexsoft: Why mutable? Since we are actually modifying the cache, should we not remove the `const` promise of the forwarding function instead?

unittests/etl/ForwardingSourceTests.cpp
```diff
@@ -118,3 +119,90 @@ TEST_F(ForwardingSourceOperationsTests, Success)
...
+            expectedReply["forwarded"] = true;
+            EXPECT_EQ(*result, expectedReply) << *result;
+        }
+    });
+}
+
+TEST_F(ForwardingSourceCacheTests, ResponseWithErrorNotCached)
+{
+    boost::json::object const request = {{"command", "server_state"}};
+    auto const errorResponse = R"({"reply":"some_reply","error":"some_error"})";
```
godexsoft: nit: Add some spaces to JSONs? 🏅

godexsoft (changes requested):
Leaving a few questions. Also, sprinkle some `noexcept` and `[[nodiscard]]` where applicable 👍

unittests/etl/ForwardingSourceTests.cpp
kuznetsss: I can't because they are being compared with JSONs generated automatically by `boost::json`.

godexsoft: I don't insist but the way we usually do this is by creating a `json::object` from the string and then comparing the objects instead :)

docs/examples/config/example-config.json
```diff
@@ -33,7 +33,7 @@
             "ip": "127.0.0.1",
             "ws_port": "6006",
             "grpc_port": "50051",
-            "cache_timeout_ms": 250 // could be 0, which means no cache
+            "forwarding_cache_timeout_ms": 250 // could be 0, which means no cache
```
godexsoft: Perhaps we should use double instead to keep it the same as we do in `IntervalSweepHandler`, wdyt?

src/etl/Source.hpp
```diff
@@ -48,40 +48,53 @@ namespace etl {
...
+    SourceImpl(SourceImpl const&) = delete;
+    SourceImpl(SourceImpl&&) = default;
+    SourceImpl&
+    operator=(SourceImpl const&) = delete;
+    SourceImpl&
+    operator=(SourceImpl&&) = default;
+
+    ~SourceImpl()
+    {
+        // subscriptionSource_ must die before forwardingSource_ because forwardingSource_ is used in callback
```
godexsoft: Is there some way to avoid it? like for example use `shared_ptr` and have shared ownership of the `forwardingSource` maybe?

godexsoft (dismissed):
I'm happy with how it is now. But leaving a few questions/suggestions for you to take a look.

docs/configure-clio.md
```diff
@@ -83,3 +83,22 @@ By default Clio checks admin privileges by IP address from requests (only `127.0
 
 If the password is presented in the config, Clio will check the Authorization header (if any) in each request for the password. The Authorization header should contain the type `Password`, and the password from the config (e.g. `Password secret`).
 Exactly equal password gains admin rights for the request or a websocket connection.
+
+## ETL sources forwarding cache
+
+By default Clio does not caches forwarding requests to ETL sources.
```
cindyyan317: worth to mention we only cache the response for certain commands.

cindyyan317:
I think this forwarding cache has two ways to invalidate the cache ,right?
1. when the timeout reaches
2. when ledger closed

Since we already invalidate all old cache when new ledger coming, I wonder why we need timeout option? If timeout > ledger gap, the cache will has been invalidated when timeout reaches, if timeout < ledger gap, the response will not change between ledgers, why bother to change? 

I observe every etl source will have separate forward cache, which can be configured differently. I am thinking if it is better to have a forward cache for the all sources?

kuznetsss:
> I wonder why we need timeout option?

Some of the cached requests have constantly changing data, e.g. transaction fee. Users may want to have fresh date more often than every ledger.


> I am thinking if it is better to have a forward cache for the all sources?

Good idea, let me change it.

cindyyan317:
> Some of the cached requests have constantly changing data, e.g. transaction fee. Users may want to have fresh date more often than every ledger.

That reminds me one thing: the fee response differs according to different etl source. Does it cause any problem? for example: user calls fee, we send the fee (X) of etl node 1, user calls submit with fee X, but we forward it to etl node 2.

kuznetsss:
> Does it cause any problem?

There is a fee limit in transaction which should save a user from unexpected expenses.

src/etl/LoadBalancer.cpp
```diff
@@ -193,22 +201,34 @@ LoadBalancer::forwardToRippled(
...
         sourceIdx = util::Random::uniform(0ul, sources_.size() - 1);
 
     auto numAttempts = 0u;
 
+    std::optional<boost::json::object> response;
     while (numAttempts < sources_.size()) {
-        if (auto res = sources_[sourceIdx].forwardToRippled(request, clientIp, yield))
-            return res;
+        if (auto res = sources_[sourceIdx].forwardToRippled(request, clientIp, yield)) {
+            response = res;
```
cindyyan317: Can we use move here?

src/etl/impl/ForwardingSource.cpp
```diff
@@ -92,6 +92,10 @@ ForwardingSource::forwardToRippled(
 
     auto responseObject = parsedResponse.as_object();
     responseObject["forwarded"] = true;
+
+    // if (cache_ and not responseObject.contains("error"))
```
cindyyan317: u may want to remove it

src/etl/impl/ForwardingSource.hpp
```diff
@@ -19,6 +19,7 @@
 
 #pragma once
 
+#include "etl/impl/ForwardingCache.hpp"
```
cindyyan317: remove?

src/etl/impl/ForwardingSource.hpp
```diff
@@ -56,7 +58,7 @@ class ForwardingSource {
         boost::json::object const& request,
         std::optional<std::string> const& forwardToRippledClientIp,
         boost::asio::yield_context yield
-    ) const;
+    );
```
cindyyan317: can bring const back 🤔

.gitignore
```diff
@@ -5,6 +5,7 @@
 .cache
 .vscode
 .python-version
+.DS_Store
```
cindyyan317, line 8: thanks 🙇

docs/examples/config/example-config.json
```diff
@@ -32,9 +32,10 @@
         {
             "ip": "127.0.0.1",
             "ws_port": "6006",
-            "grpc_port": "50051"
+            "grpc_port": "50051",
```
cindyyan317: remove ,

src/etl/ETLState.hpp
```diff
@@ -44,7 +44,7 @@ struct ETLState {
      */
     template <typename Forward>
     static std::optional<ETLState>
-    fetchETLStateFromSource(Forward const& source) noexcept
+    fetchETLStateFromSource(Forward& source) noexcept
```
cindyyan317, line 47: can const add back?

kuznetsss: We can't here because `source` here is either `Source` or `LoadBalancer` and for `LoadBalancer` `forwardToRippled()` is not const.

cindyyan317 (dismissed):
👍

docs/configure-clio.md
```diff
@@ -86,19 +86,14 @@ Exactly equal password gains admin rights for the request or a websocket connect
 
 ## ETL sources forwarding cache
 
-By default Clio does not caches forwarding requests to ETL sources.
-To enable the caching for a source, `forwarding_cache_timeout_ms` value should be added to the source entry in `etl_sources`:
+Clio can cache requests to ETL sources to reduce the load on the ETL source.
+Only following commands are cached: `server_info`, `server_state`, `server_definitions`, `fee`, `ledger_closed`.
+By default the forwarding cache is off.
```
godexsoft, line 91: Are you sure to change previous default like this? 
I think this way existing operators might just miss the change and have no cache on their prod system as a consequence. WDYT?

cindyyan317 (approved):

docs/configure-clio.md
kuznetsss: As I understand previous default state was also off.
https://github.com/XRPLF/clio/blob/97a63db51d487b50aaf1944c7cc0c2406d93979e/src/etl/impl/ForwardCache.hpp#L62-L76

My logic was that the cache brings some minor delay in data updates from user perspective so it should be turned on explicitly.

godexsoft: If you are certain that before it was off by default then i see no issue. Other than maybe us renaming the config keys. But in general this is a pretty different cache compared to before so i don't mind.

godexsoft (approved):
🚀

PR #1206, opened 2024-02-23
Add doxygen comments to Random and TerminationHandler

### LINKED ISSUE #1010: Add doxygen to public functions
Let's add doxygen comments for `Random` and `TerminationHandler` (as well as any other missing places on public interfaces and functions that we can find) 👍

kuznetsss:
Fixes #1010.

### DIFF
--- a/src/util/Random.hpp
+++ b/src/util/Random.hpp
@@ -25,7 +25,7 @@
 namespace util {
 
 /**
- * Random number generator
+ * @brief Random number generator
  */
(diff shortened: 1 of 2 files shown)


godexsoft (dismissed):
nit: We usually add `@brief` everywhere, even for oneliners. But the effect is the same so i'm still approving the PR 👍 Up to you if you want to change it.

godexsoft (approved):

PR #1210, opened 2024-02-26
Configure git-lfs

godexsoft:
This simply configures git lfs so that we don't store upcoming image files (png) in git itself.
To use you will need to install it locally: `brew install git-lfs` or read  [this doc](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage).


kuznetsss (approved):
Let's try to use it 👍

cindyyan317 (approved):

Issue #1067, opened 2024-01-02
Use cmake-format


kuznetsss:
We could try to use [cmake-format](https://github.com/cheshirekow/cmake_format) to automatically format cmake files.

PR #1213, opened 2024-02-26
Add cmake-format to docker and mac runners

kuznetsss:
For #1067.


kuznetsss:
CI is failing here but after merge and automatic docker image update it should be fine.

godexsoft (approved):
Looks like it should work. Only one way to find out 👍

Issue #889, opened 2023-10-04
Add doxygen to CI


kuznetsss:
- Automatically generate and upload docs to github pages (or similar). Ideally have docs for develop and releases branches
- Use this theme in our doxygen results https://jothepro.github.io/doxygen-awesome-css/

sublimator:
https://mrdox.com

godexsoft:
This is now working for `develop` docs. Next up in #1239 we should split release docs and developer docs (if we decide it's useful after all).

PR #1214, opened 2024-02-26
Use doxygen-awesome theme

godexsoft:
For #889 

This makes the generated docs look better. See screenshots below.
Please note that the version string 2.1.0 would only appear when generating the docs from `release/2.1.0`. In development builds the version string is quite long - we should address that by maybe always hardcoding it to `develop` or something like that when generating docs outside of a release branch.

Screenshot 2024-02-26 at 23 01 22

Screenshot 2024-02-26 at 23 01 31


src/web/interface/ConnectionBase.hpp
```diff
@@ -84,7 +84,7 @@ struct ConnectionBase : public util::Taggable {
      * @throws Not supported unless implemented in child classes. Will always throw std::logic_error.
      */
     virtual void
-    send(std::shared_ptr<std::string> /* msg */)
+    send([[maybe_unused]] std::shared_ptr<std::string> msg)
```
godexsoft, line 87: @kuznetsss this is needed because `doxygen` complains `msg` is not a known param while generating the doc. Let's use `[[maybe_unused]]` as a convention from now on 👍

Doxyfile
```diff
@@ -14,3 +17,18 @@ GENERATE_LATEX         = NO
...
+DISABLE_INDEX          = NO
+FULL_SIDEBAR           = NO
+HTML_HEADER            = ../docs/doxygen-awesome-theme/header.html
+HTML_EXTRA_STYLESHEET  = ../docs/doxygen-awesome-theme/doxygen-awesome.css \
+                         ../docs/doxygen-awesome-theme/doxygen-awesome-sidebar-only.css \
+                         ../docs/doxygen-awesome-theme/doxygen-awesome-sidebar-only-darkmode-toggle.css 
+HTML_EXTRA_FILES       = ../docs/doxygen-awesome-theme/doxygen-awesome-darkmode-toggle.js \
+                         ../docs/doxygen-awesome-theme/doxygen-awesome-interactive-toc.js
+
+HTML_COLORSTYLE        = LIGHT
```
kuznetsss, line 32: Is it light theme by default?

src/data/BackendInterface.cpp
```diff
@@ -48,6 +48,12 @@ namespace {
 util::Logger gLog{"Backend"};
 }  // namespace
 
+/**
+ * @brief This namespace implements the data access layer and related components.
+ *
+ * The data layer is responsible for fetching and storing data from the database.
+ * Cassandra and ScyllaDB are currently supported via the `CassandraBackend` implementation.
+ */
```
kuznetsss, line 56: Do we need to do it once somewhere?

src/data/LedgerCache.hpp
```diff
@@ -86,7 +86,7 @@ class LedgerCache {
     /**
      * @brief Update the cache with new ledger objects.
      *
-     * @param blobs The ledger objects to update cache with
+     * @param objs The ledger objects to update cache with
```
kuznetsss, line 89: Will the inconsistency between documentation and parameter name result to a warning or error in doxygen?
Is it possible to automatically detect such inconsistencies?

kuznetsss (commented):
Really good improvement, thanks! Just a few questions.

cindyyan317 (dismissed):
LGTM  ! Happy to see the names get consistent 👍

src/data/LedgerCache.hpp
godexsoft: Only in warnings. We could make the into errors if we wanted to. We have disabled some of that stuff in Doxyfile atm.

src/data/BackendInterface.cpp
godexsoft: Yes... :rage4: It needs to be somewhere on top of the namespace, otherwise the namespace does not even appear in the `namespaces` page 👎

Doxyfile
godexsoft: In the theme it actually says that this is required to be set to LIGHT :rage4: 
For me it shows dark theme by default. I think it takes your system settings tbh.

src/data/LedgerCache.hpp
kuznetsss: Shall we change it to have errors?

kuznetsss (dismissed):
👍

kuznetsss (approved):

PR #1217, opened 2024-02-27
remove _ from public member

cindyyan317:
I happen to see this public member has _ tail
isAdmin_ -> isAdmin

### DIFF
--- a/src/web/interface/ConnectionBase.hpp
+++ b/src/web/interface/ConnectionBase.hpp
@@ -44,11 +44,11 @@ namespace http = boost::beast::http;
 struct ConnectionBase : public util::Taggable {
 protected:
     boost::system::error_code ec_;
+    bool isAdmin_ = false;
 
 public:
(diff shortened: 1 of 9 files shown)


src/web/interface/ConnectionBase.hpp
```diff
@@ -48,7 +48,7 @@ struct ConnectionBase : public util::Taggable {
 public:
     std::string const clientIp;
     bool upgraded = false;
-    bool isAdmin_ = false;
```
kuznetsss, line 51: I'm thinking maybe we should move `isAdmin` to private and add `setAdmin()` to protect it from accidental change?

godexsoft: Or maybe make `isAdmin_` a protected member and keep the `isAdmin()`. We should not need to `setAdmin()` from outside of the child classes, right?

godexsoft (approved):

PR #1223, opened 2024-03-01
Add missing doxygen comments

### LINKED ISSUE #1218: Missing doxygen documentation (Version: 2.1.0)
## Issue Description
Some public classes and functions are missing documentation.

## Steps to Reproduce
Run `cmake --build . -j8 --target docs` to generate docs. Doxygen will display a big list of missing docs.

## Expected Result
Docs should be built without error. No missing documentation on public interfaces.

## Actual Result
Build fails with error.

## Environment
Any.

godexsoft:
Fixes #1218 

Adds missing docs for public interfaces, tweaks some existing comments, etc.
This is meant to be passing if we manually filter out uninteresting warnings such as 
- `(typedef)` - `using` declarations don't need documenting in Clio
- `(variable)` - member variables don't need documenting in Clio
- `=default` - undocumented trivial defaulted operators/constructors should not be warned about


src/rpc/Factories.hpp
```diff
@@ -54,6 +54,7 @@ namespace rpc {
  * @param range The ledger range that is available at request time
  * @param clientIp The IP address of the connected client
  * @param apiVersionParser A parser that is used to parse out the "api_version" field
+ * @return A Websocket context
```
kuznetsss: Maybe `A Websocket context or error Status` ? Here and below

src/rpc/RPCHelpers.hpp
```diff
@@ -217,15 +354,34 @@ read(
...
 );
 
-std::variant<Status, std::pair<ripple::PublicKey, ripple::SecretKey>>
-keypairFromRequst(boost::json::object const& request);
-
+/**
+ * @brief Get the account associated with a transaction
+ *
+ * @param transaction The transaction
+ * @return The account or an error status
```
kuznetsss: It returns a vector of accounts

cindyyan317 (dismissed):

src/rpc/common/JsonBool.hpp
```diff
@@ -34,12 +34,20 @@ namespace rpc {
 struct JsonBool {
```
kuznetsss, line 34: Maybe this type could be defined by `using`:
```cpp
using JsonBool = bool;
```

src/rpc/handlers/AccountTx.hpp
```diff
@@ -142,16 +162,41 @@ class AccountTxHandler {
...
+     * @param output The output to convert
+     */
     friend void
     tag_invoke(boost::json::value_from_tag, boost::json::value& jv, Output const& output);
 
+    /**
+     * @brief Convert a JSON object to Input type
+     *
+     * @param jv The JSON object to convert
+     * @return The input type
```
kuznetsss: Here and in similar places maybe better
```
@return Input parsed from the JSON object
```

src/rpc/handlers/LedgerData.hpp
```diff
@@ -73,9 +76,13 @@ class LedgerDataHandler {
         bool validated = true;
     };
 
-    // TODO: Clio does not implement "type" filter
-    // outOfOrder only for clio, there is no document, traverse via seq diff
-    // outOfOrder implementation is copied from old rpc handler
+    /**
+     * @brief A struct to hold the input data for the command
+     *
+     * @note TODO: Clio does not implement `type` filter
```
kuznetsss: Maybe add link to the related issue?

src/rpc/handlers/LedgerData.hpp
```diff
@@ -116,13 +134,32 @@ class LedgerDataHandler {
...
+     * @return The result of the operation
+     */
     Result
     process(Input input, Context const& ctx) const;
 
 private:
+    /**
+     * @brief Convert the Output to a JSON object
+     *
+     * @param jv The JSON object to convert to
```
kuznetsss: Maybe use `@param [out]` for such cases?

src/util/Retry.hpp
```diff
@@ -159,7 +163,7 @@ class ExponentialBackoffStrategy : public RetryStrategy {
  * @param delay The initial delay value
  * @param maxDelay The maximum delay value
  * @param strand The strand to use for async operations
- * @return Retry
+ * @return The retry mechanism
```
kuznetsss: Maybe `The retry object`?

src/util/async/context/BasicExecutionContext.hpp
```diff
@@ -84,12 +84,17 @@ template <
     typename ErrorHandlerType = impl::DefaultErrorHandler>
 class BasicExecutionContext {
     ContextType context_;
+
+    /** @cond */
```
kuznetsss, line 98: Maybe filter warnings about friends?

kuznetsss (commented):
Really good improvement, thanks! Just a few thoughts.

src/rpc/common/JsonBool.hpp
godexsoft: This leads to other issues with `tag_invoke` for `value_to<JsonBool>` and is out of scope of this PR

src/rpc/handlers/LedgerData.hpp
godexsoft: Actually, we support `type` filter. This looks like an old TODO that slipped through. Removing it 👍

src/util/async/context/BasicExecutionContext.hpp
godexsoft: This one is not because of `friend` but because it's in `impl` and since we disabled `impl` in `Doxyfile` it can not be found at all. So i'm basically disabling it here. It's a bit of an edge case.

kuznetsss (approved):
👍

PR #1226, opened 2024-03-02
Automatically detect missing doxygen comments

### LINKED ISSUE #1216: Check doxygen completnes on CI (Version: 2.1.0)
## Summary
It would be great to automatically find undocumented and/or missdocumented doxygen comments on public functions and classes in the codebase.

## Motivation
We want consistent and complete documentation for all public interfaces.

## Solution
Use doxygen with the correct mix of configuration flags and options in order to automatically detect missdocumentation.

## Paths Not Taken
- Leave documentation incomplete as it is
- Remove documentation entirely

godexsoft:
Fixes #1216


.githooks/check-docs
```diff
@@ -0,0 +1,61 @@
...
+    cat <<EOF
+
+                                   WARNING
+-----------------------------------------------------------------------------
+        'doxygen' is required to check documentation.
+        Please install it for next time. For the time being it's on CI.
+-----------------------------------------------------------------------------
+
+EOF
+fi
```
kuznetsss: Maybe check doxygen is available in the beginning of the script to not have if for the almost whole script?

.githooks/check-docs
```diff
@@ -0,0 +1,61 @@
...
+    pushd ${DOCDIR} > /dev/null
+
+    sed \
+        -e "s/\${LINT}/YES/" \
+        -e "s!\${SOURCE}!${ROOT}!" \
+        -e "s/\${USE_DOT}/NO/" \
+        -e "s/\${EXCLUDES}/impl/" \
+        ${ROOT}/Doxyfile > ./Doxyfile
+
+    ${DOXYGEN} ./Doxyfile 2> ${TMPFILE} 1> /dev/null
```
kuznetsss: Instead of creating a separate doxygen file you could do
```bash
cat ${ROOT} | sed -e "s/\${LINT}/YES/" \
        -e "s!\${SOURCE}!${ROOT}!" \
        -e "s/\${USE_DOT}/NO/" \
        -e "s/\${EXCLUDES}/impl/" | ${DOXYGEN} 2> ${TMPFILE} 1> /dev/null
```

.githooks/check-docs
```diff
@@ -0,0 +1,61 @@
...
+
+    # We don't want to check for default values and typedefs as well as for member variables
+    OUT=$(cat ${TMPFILE} \
+        | grep -v "=default" \
+        | grep -v "\(variable\)" \
+        | grep -v "\(typedef\)")
+
```
kuznetsss, line 46: Let's delete temporary file here?

.githooks/pre-commit
```diff
@@ -1,93 +1,6 @@
...
-  Please manually add any updated files and commit again.
------------------------------------------------------------------------------
-
-EOF
-    exit 1
-fi
+# This script is intended to be run from the root of the repository.
 
+source .githooks/check-format
+source .githooks/check-docs
```
kuznetsss, line 6: I think this should be only in CI. It is not convenient to not able to commit anything quickly into a local branch.

kuznetsss (changes requested):
Thanks, that's a really good automation. A few thoughts what could be improved.

.githooks/pre-commit
godexsoft: Hm, it's pretty fast tbh. 
On my mac it takes about the same time it takes us to do the `clang-format` job.
If `doxygen` is not installed, you will get a warning instead of an error so there is the option to ignore it if you really don't want to spend the extra few seconds on each commit.
It's only really slow to generate the docs with DOT enabled and with all the `impl` stuff. It's very fast without DOT and with only the public part of Clio being generated for.

kuznetsss: No, I mean imagine you are in progress of something and you decided to commit your current state. Currently unless your documentation is complete you can not commit and it is not convenient.

godexsoft: As we discussed on the call, you can always use `--no-verify` in such (rare) case. This should not be the norm either way, but more of an exception. Normally i'd expect all code pushed to the repo be at least compiling, let alone passing docs and format checks 👍

kuznetsss:
Please update the documentation about pre-commit hook.

godexsoft:
@cindyyan317 @kuznetsss it looks like there is some caching issue or something with clang-tidy. I suggest we merge it as is (after review and fixes if needed) and tomorrow we will find out if clang-tidy will be happy or not.

kuznetsss (approved):

.github/workflows/build.yml
```diff
@@ -17,12 +17,27 @@ jobs:
       - name: Run formatters
         id: run_formatters
         run: |
-          ./.githooks/pre-commit
+          ./.githooks/check-format
+        shell: bash
+  
+  check_docs:
```
cindyyan317, line 23: Why do we separate check-docs and check-format in two jobs?

kuznetsss: That's better because they could run in parallel and provide results independently.

cindyyan317 (approved):
Let's go 💯

PR #1231, opened 2024-03-05
Refactor cmake

### LINKED ISSUE #920: Refactor cmake
Currently cmake contains a huge list of all the source files and compiling all of them into one clio library. It would be better to make each directory in `src` as a library with it's own cmake file.

kuznetsss:
Fixes #920.


docker/develop/run
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+set +x
```
godexsoft: Do we need this still? Personally I prefer scripts to be silent other than logging through `echo`s.

src/util/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+         libbacktrace::libbacktrace
+         fmt::fmt
+         openssl::openssl
+         xrpl::libxrpl
+         Threads::Threads
+         clio_options
+)
+
+if (is_gcc)
+  # FIXME: needed on gcc for now For some reason cmake doesn't propagate the compile definitions from clio_options so we
```
godexsoft: Please split the comment into two lines "now\nFor some".

godexsoft (dismissed):
This is brilliant. Looks so much more modern 🥇 
Leaving just a couple questions/suggestions 👍

docker/develop/run
kuznetsss: Sorry, left from debug

src/util/CMakeLists.txt
kuznetsss: Hm, it was on different lines until cmake-format merged it 🤔

godexsoft: Perhaps it wants a `.` at the end of the sentence? :rage2:

godexsoft (approved):
🚀 :rage1:

PR #1232, opened 2024-03-05
Build docs and upload to gh-pages

godexsoft:
For #889 

This is still WIP.
The first step was to fix cmake so that generating documentation did not depend on Conan (and all Clio's deps).
Now we have a new workflow that we need to get to work but in this PR it will always fail because the permissions were missing and will only be given once the new workflow is merged to develop.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,45 @@
+name: Documentation
+on:
+  push:
+    branches: [release/*, develop]
+  pull_request:
+    branches: [release/*, develop]
```
kuznetsss: We don't want to update documentation on new PRs.

docs/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.16.3)
+project(docs)
```
kuznetsss, line 2: Maybe clio_docs?

godexsoft: I like the short `docs` better - it's obvious it's for Clio as it's part of our repo and folder structure

.github/workflows/docs.yml
godexsoft: Yep, true

kuznetsss (approved):
👍

PR #1244, opened 2024-03-07
Add clean up workdir

### LINKED ISSUE #1095: Clean up working directory on macOS
Github runner doesn't clean working directory, so there could be potentially overlapping between runs. We should add cleaning up working directory in the beginning of each workflow on mac. See #1054 for example how to do it. There should be aslo check that we are not deleting `/` and the code should be moved to an action to avoid duplication.

kuznetsss:
Fixes #1095.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,10 @@ jobs:
           --health-retries 5
 
     steps:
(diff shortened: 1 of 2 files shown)


godexsoft (approved):
Nice 👍

PR #1256, opened 2024-03-12
Export etl metrics to prometheus

### LINKED ISSUE #1248: Export ETL metrics to Prometheus
We should export all the data except `etl_sources` provided by `EtlService::getInfo()` to Prometheus.

kuznetsss:
Fixes #1248.

### DIFF
--- a/src/etl/SystemState.hpp
+++ b/src/etl/SystemState.hpp
@@ -38,7 +38,7 @@ struct SystemState {
      * as they are written to the database.
      */
     util::prometheus::Bool isReadOnly = PrometheusService::boolMetric(
-        "etl_read_only",
+        "read_only",
         util::prometheus::Labels{},
         "Whether the process is in strict read-only mode"
     );
--- a/src/etl/SystemState.hpp
+++ b/src/etl/SystemState.hpp
@@ -19,7 +19,9 @@
 
 #pragma once
 
-#include <atomic>
+#include "util/prometheus/Bool.hpp"
+#include "util/prometheus/Label.hpp"
+#include "util/prometheus/Prometheus.hpp"
 
 namespace etl {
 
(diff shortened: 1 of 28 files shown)


kuznetsss:
I also removed global `PrometheusService` instance from tests and added `WithPrometheus` where needed. Now if a test uses Prometheus counters (even somewhere under the hood) it will be failing by assertion. It is much better because forgetting about the global `PrometheusService` is leading to confusing situation when you think you just created a new variable but it has already not default value because the value is from some previous test with Prometheus counters.

src/etl/SystemState.hpp
```diff
@@ -33,9 +37,19 @@ struct SystemState {
      * In strict read-only mode, the process will never attempt to become the ETL writer, and will only publish ledgers
      * as they are written to the database.
      */
-    bool isReadOnly = false;
+    util::prometheus::Bool isReadOnly = PrometheusService::boolMetric(
+        "etl_read_only",
```
cindyyan317: better change it to read_only to distinguish the etl which is not writing at the moment

cindyyan317 (dismissed):
LGTM , just saw some name not fit

src/util/prometheus/Bool.hpp
```diff
@@ -0,0 +1,89 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: 2024 probably

unittests/util/prometheus/BoolTests.cpp
```diff
@@ -0,0 +1,63 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: 2024 also

godexsoft (dismissed):
Noticed few small things only - code looks good to me 👍

godexsoft (approved):

PR #1263, opened 2024-03-14
Update bug_report.md

mounikakun:
Adding a gentle warning to remind about not posting any credentials in github issues.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -8,6 +8,7 @@ assignees: ''
 ---
 
 <!-- Please search existing issues to avoid creating duplicates. -->
+<!-- DO NOT POST CREDENTIALS -->
```
godexsoft: Shall we make it a little more gentle and maybe add something about IPs and other sensitive information? 😃 
Also, it would not hurt to add it to the other template too. Chances are lower there but it does not hurt 👍

godexsoft (approved):
Thanks 🙇‍♂️

PR #1268, opened 2024-03-14
Remove mentions of release docs from workflow

### LINKED ISSUE #1239: Generate docs for different versions (Version: 2.2.0)
## Summary
In #889 we introduced generation and upload of doxygen docs to gh-pages.
However we are only generating and uploading `develop` docs atm.
There should be docs for each (new) release *and* for latest `develop`.
Release docs are generated and published for each `release/x.y.z` that is tagged (the tag is used as the docs version).

EDIT: see comment below for new behaviour 

## Motivation
It may be valuable to be able to see docs as they were for a specific released version. Not to mention having latest `develop` docs always up to date.

## Solution
Implement a workflow that can do both `develop` and `release` docs.

godexsoft:
Fixes #1239

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,7 +1,7 @@
 name: Documentation
 on:
   push:
-    branches: [release/*, develop]
+    branches: [develop]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


kuznetsss (approved):
👍

PR #1282, opened 2024-03-20
XRPFees - support new fields

### LINKED ISSUE #1280: Support amendment XRPFees 🔥
XRPFees amendment will replace some fields.
>Updates the FeeSettings ledger entry type. Replaces BaseFee, ReferenceFeeUnits, ReserveBase, and ReserveIncrement fields with BaseFeeDrops, ReserveBaseDrops, and ReserveIncrementDrops.

https://xrpl.org/resources/known-amendments/#xrpfees

The fields will be absent after XRPFees amendment is enable and the set_fees tx has been executed.
https://github.com/XRPLF/rippled/blob/0c32fc5f2adb5a1598dd7e97fc83ba0f64fdee20/src/ripple/app/tx/impl/Change.cpp#L365

Current Clio dependents on some fields -> https://github.com/XRPLF/clio/blob/release/2.2.0/src/data/BackendInterface.cpp#L353 

We shall update the way we construct the fee object. 

Reference : https://github.com/XRPLF/rippled/blob/0c32fc5f2adb5a1598dd7e97fc83ba0f64fdee20/src/ripple/app/ledger/Ledger.cpp#L632

godexsoft:
Fixes #1280 

Also adds tests for both new and old fees fields.

### DIFF
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -350,14 +350,37 @@ BackendInterface::fetchFees(std::uint32_t const seq, boost::asio::yield_context 
     ripple::SerialIter it(bytes->data(), bytes->size());
     ripple::SLE const sle{it, key};
 
-    if (sle.getFieldIndex(ripple::sfBaseFee) != -1)
-        fees.base = sle.getFieldU64(ripple::sfBaseFee);
+    // XRPFees amendment introduced new fields for fees calculations.
+    // New fields are set and the old fields are removed via `set_fees` tx.
+    // Fallback to old fields if `set_fees` was not yet used to update the fields on this tx.
+    auto hasNewFields = false;
 
-    if (sle.getFieldIndex(ripple::sfReserveBase) != -1)
-        fees.reserve = sle.getFieldU32(ripple::sfReserveBase);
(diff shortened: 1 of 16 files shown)


src/data/BackendInterface.cpp
```diff
@@ -350,14 +350,37 @@ BackendInterface::fetchFees(std::uint32_t const seq, boost::asio::yield_context
...
+        fees.reserve = reserveBaseXRP->xrp();
+
+    if (reserveIncrementXRP)
+        fees.increment = reserveIncrementXRP->xrp();
+
+    hasNewFields = baseFeeXRP || reserveBaseXRP || reserveIncrementXRP;
+
+    if (not hasNewFields) {
+        // Fallback to old fields
+        if (sle.getFieldIndex(ripple::sfBaseFee) != -1)
```
cindyyan317: Not important: We can unify the way of checking here.

cindyyan317 (dismissed):
Nice testcases

src/data/BackendInterface.cpp
godexsoft: Yes, would you prefer the optional way (new)? I don't mind, i just did not really want to mess with the working code much. But now with tests i think it would be safe to change.

cindyyan317 (approved):

PR #1287, opened 2024-03-22
Add support for Price Oracle in `ledger_entry`

### LINKED ISSUE #1277: `ledger_entry` should support “oracle“ object (Version: 2.2.0)
## Summary
Add support for Price Oracle (XLS-47d) (#974) in the `ledger_entry` RPC.
See `rippled` [implementation here](https://github.com/XRPLF/rippled/commit/e718378bdbd66bfba5b0c6778a1320ca89718a28#diff-618417ac1dcd9339ca0948e55060838b9146dd4012ee2afe8daefb3748920021).

## Motivation
API parity

## Solution
Add support for Price Oracle

godexsoft:
Fixes #1277 

Adds support for `oracle` in `ledger_entry` RPC.
Not a fan of the custom error but had to do it for API parity :rage4: 

Side rant: 
One day in the future, maybe we can decide to move to API 3 (or newer). I'd really like to do it by just removing all the `WithCustomError` from all specs in Clio and call it the new API. 
Just by removing the custom errors everywhere we would get so much closer to a clean and consistent API. 
The bigger task is then to port all that to `rippled` of course :rage3: 

cc @Bronek

### DIFF
--- a/src/rpc/handlers/LedgerEntry.cpp
+++ b/src/rpc/handlers/LedgerEntry.cpp
@@ -259,6 +259,7 @@ tag_invoke(boost::json::value_to_tag<LedgerEntryHandler::Input>, boost::json::va
         {JS(amm), ripple::ltAMM},
         {JS(xchain_owned_create_account_claim_id), ripple::ltXCHAIN_OWNED_CREATE_ACCOUNT_CLAIM_ID},
         {JS(xchain_owned_claim_id), ripple::ltXCHAIN_OWNED_CLAIM_ID},
+        {JS(oracle), ripple::ltORACLE},
     };
 
     auto const parseBridgeFromJson = [](boost::json::value const& bridgeJson) {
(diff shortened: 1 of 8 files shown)


src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -274,6 +276,23 @@ tag_invoke(boost::json::value_to_tag<LedgerEntryHandler::Input>, boost::json::va
         return ripple::STXChainBridge{lockingDoor, lockingIssue, issuingDoor, issuingIssue};
     };
 
+    auto const parseOracleFromJson = [](boost::json::value const& json) -> std::optional<ripple::uint256> {
+        ripple::uint256 key = beast::zero;
+
+        if (json.is_string()) {
```
cindyyan317: If we put the "oracle" in indexFieldTypeMap , we don't need to have string branch here. And it can prevent the scenario that the string index is not for an actual oracle.

cindyyan317 (changes requested):
Nice ! 
One thing need to be changed: add oracle to index lookup table. So the Clio will check the index actual type if matches the expected type.

src/rpc/handlers/LedgerEntry.cpp
godexsoft: Oh, interesting, thanks! 👍

cindyyan317 (approved):
LGTM 👍

kuznetsss (approved):
👍

Issue #1271, opened 2024-03-15
Move to c++23 compilers (Version: 2.2.0)


godexsoft:
## Summary
Let's move CI to clang-16 and gcc-12.
It would be nice to move to newer compilers and see CI pass before changing any code (like moving to `std::expected` etc.).

## Motivation
We want to move to c++23 (see #1173 for example).

## Solution
Upgrade compilers to versions that support c++23 features we need.

godexsoft:
Currently `rocksdb` dep is failing if conan profile is setup to prefer c++23 for its standard. with c++20 it compiles fine and we can use c++23 just on Clio level instead by setting 

```
target_compile_features(clio_options INTERFACE cxx_std_23)
```

godexsoft:
Moving to gcc-12 in current CI docker (ubuntu 20.04) is not easy as there is no official package.
It's fairly easy to move the CI docker image to ubuntu 22.04 which does provide gcc-12. 
By doing that though the final `clio_server` binary does not actually work on 20.04: 
```
./clio_server: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.35' not found (required by ./clio_server)
```

Additionally, gcc-12 detects a few new warnings (that become errors in Clio). They can be disabled but we should fix the code instead.
`-Wmissing-requires`, `-Wrestrict` and `-Wnull-dereference` are few that seem to be breaking the build on gcc-12.

godexsoft:
Current conclusions
- We need our docker image to stay on 20.04 as we can't run clio built with newer GLIBC on actual servers that are 20.04 still
- We will have to either find a package of gcc-12 for 20.04 or build from source
- The new warnings from gcc-12 needs fixing regardless

Issue #1174, opened 2024-02-14
Remove Any wrapper


godexsoft:
After migrating to `std::expected` (#1173) we can remove `util/async/impl/Any.hpp` in favour of `std::any`.

PR #1288, opened 2024-03-22
Move to `std::expected`

### LINKED ISSUE #1173: Move to std::expected and C++23
`std::expected` is now available for all compilers that we target. We just need to upgrade gcc to 12 and clang to 16 in CI/docker.
Let's get rid of `util::Expected`. Update docker build image and change std to c++23 in Cmake.

godexsoft:
Fixes #1173 

Note that we will have to cherry-pick some of this stuff for #1271 first, as planned before


godexsoft:
Currently this requires to update docker CI. Will be done in #1271

src/util/async/impl/Any.hpp
```diff
@@ -22,7 +22,7 @@
 #include <any>
 #include <type_traits>
 
-// Note: This is a workaround for util::Expected. This is not needed when using std::expected.
+// Note: This is a workaround for std::expected. This is not needed when using std::expected.
```
godexsoft: This will be entirely removed in #1174 but for the time being let's at least remove the comment 🔢

.clangd
```diff
@@ -1,3 +1,6 @@
+CompileFlags:
+  Add: [-D__cpp_concepts=202002]
```
godexsoft, line 2: Without this, `clangd` will not work correctly atm.
Maybe this will become redundant later when we change the way `gcc-12` and `clang` are added to our CI image.

cindyyan317 (approved):
LGTM

PR #1289, opened 2024-03-23
Upgrade CI docker image to gcc-12

### LINKED ISSUE #1271: Move to c++23 compilers (Version: 2.2.0)

godexsoft:
Fixes #1271 

Switches to `gcc-12` compiled from source. 
LLVM is left at 17 because that's what we are already using for tooling but ideally we should be using `clang-16` to compile and the tooling should use LLVM 18 as it fixes a bug where files that contain only concepts are mistakenly marked as unnecessary includes.

### DIFF
--- a/cmake/Settings.cmake
+++ b/cmake/Settings.cmake
@@ -23,6 +23,7 @@ set(COMPILER_FLAGS
     -Wno-null-dereference
     -Wno-maybe-uninitialized
     -Wno-unknown-warning-option # and this to work with clang
+    # TODO: Address these and others in https://github.com/XRPLF/clio/issues/1273
 )
 
 # TODO: reenable when we change CI #884 if (is_gcc AND NOT lint) list(APPEND COMPILER_FLAGS -Wduplicated-branches
--- a/cmake/Settings.cmake
+++ b/cmake/Settings.cmake
@@ -17,6 +17,12 @@ set(COMPILER_FLAGS
     -pedantic
     -Wpedantic
     -Wunused
+    # FIXME: The following bunch are needed for gcc12 atm.
+    -Wno-missing-requires
+    -Wno-restrict
+    -Wno-null-dereference
+    -Wno-maybe-uninitialized
+    -Wno-unknown-warning-option # and this to work with clang
 )
 
 # TODO: reenable when we change CI #884 if (is_gcc AND NOT lint) list(APPEND COMPILER_FLAGS -Wduplicated-branches
--- a/docker/ci/dockerfile
+++ b/docker/ci/dockerfile
@@ -7,7 +7,7 @@ USER root
 WORKDIR /root/
 
 ENV GCC_VERSION=11 \
-    GCC12_VERSION=12.2.0 \
+    GCC12_VERSION=12.3.0 \
     CCACHE_VERSION=4.8.3 \
     LLVM_TOOLS_VERSION=17 \
     GH_VERSION=2.40.0 \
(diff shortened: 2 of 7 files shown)


docker/ci/dockerfile
```diff
@@ -35,11 +36,21 @@ RUN apt update -qq \
...
 WORKDIR /tmp
 
+# Install gcc-12 from source
+RUN wget "http://mirrors.concertpass.com/gcc/releases/gcc-${GCC12_VERSION}/gcc-${GCC12_VERSION}.tar.gz" \
+    && tar xf "gcc-${GCC12_VERSION}.tar.gz" \
+    && cd "gcc-${GCC12_VERSION}" \
+    && mkdir build && cd build \
+    && apt-get install -y libmpc-dev libgmp3-dev libmpfr-dev \
+    && ../configure --enable-languages=c,c++ --disable-multilib \
+    && make -j$(($(nproc --all) * 2)) BOOT_CFLAGS='-O3 -g0' profiledbootstrap \
```
godexsoft: Not too sure about this `-j` setting. It works on my machine but it may kill our runner. Perhaps this is too aggressive and we need to make it somewhat smaller.

cmake/Settings.cmake
```diff
@@ -17,6 +17,12 @@ set(COMPILER_FLAGS
     -pedantic
     -Wpedantic
     -Wunused
+    # FIXME: The following bunch are needed for gcc12 atm.
```
kuznetsss, line 20: Let's maybe link an issue here?

docker/ci/dockerfile
```diff
@@ -7,6 +7,7 @@ USER root
 WORKDIR /root/
 
 ENV GCC_VERSION=11 \
+    GCC12_VERSION=12.2.0 \
```
kuznetsss: Maybe better to use gcc 12.3?

docker/ci/dockerfile
kuznetsss: I think `-j$(nproc)` would be fine.

docker/ci/dockerfile
```diff
@@ -35,11 +36,21 @@ RUN apt update -qq \
...
+    && pip3 install -q --no-cache-dir conan==1.62 gcovr cmake-format
 
 WORKDIR /tmp
 
+# Install gcc-12 from source
+RUN wget "http://mirrors.concertpass.com/gcc/releases/gcc-${GCC12_VERSION}/gcc-${GCC12_VERSION}.tar.gz" \
+    && tar xf "gcc-${GCC12_VERSION}.tar.gz" \
+    && cd "gcc-${GCC12_VERSION}" \
+    && mkdir build && cd build \
+    && apt-get install -y libmpc-dev libgmp3-dev libmpfr-dev \
```
kuznetsss: Maybe better to install these packages with all the other dependencies in the `apt-get` call above?

docker/ci/dockerfile
kuznetsss: I think compiling flags should be done via  `--with-build-config=‘bootstrap-O3’` or we could even try `boostrap-lto`.
https://gcc.gnu.org/install/build.html

kuznetsss (changes requested):

kuznetsss (approved):
👍

PR #1293, opened 2024-03-26
Round 3 of gcc-12 attempts

godexsoft:
According to @legleux `x-heavy` may be smaller that `heavy`, so let's try that. 
Also disabling `lto` which may be too expensive for our runner to be able to handle.


kuznetsss (approved):

legleux:
> I can't remember exactly what the changes were other than `heavy` was bumped up in some regard.
@shichengripple001 ?

PR #1298, opened 2024-03-27
Round 7 of gcc-12 attempts

godexsoft:
Ok, so our `self-hosted` runner was not able to finish the job. Either way, we were using a github runner for this workflow before and it was working every time so let's hope that will still be the case now that we don't need to build the actual gcc-12 inside of CI.


cindyyan317 (approved):
🚀

PR #1305, opened 2024-03-28
Round 9 of gcc-12 attempts

godexsoft:
Hopefully the final fix - add `gcov` alternatives to `/usr/bin` similar to the rest of `/opt/bin` things. This should fix the coverage step on CI.
The gcc-12 compiler is already pushed by hand into docker hub.


cindyyan317 (approved):
🚀

PR #1313, opened 2024-04-02
Implement get_aggregate_price

cindyyan317:
Migrate get_aggregate_price


src/rpc/handlers/GetAggregatePrice.cpp
```diff
@@ -0,0 +1,313 @@
...
+            return false;
+        });
+    }
+
+    if (timestampPricesBiMap.empty())
+        return Error{Status{ripple::rpcOBJECT_NOT_FOUND}};
+
+    auto const latestTime = timestampPricesBiMap.left.begin()->first;
+
+    Output out;
```
kuznetsss: Maybe it would be better to have constructor for `Output` to not forget to fill some field?

kuznetsss (dismissed):
LGTM 👍

kuznetsss (approved):

PR #1314, opened 2024-04-02
Warn about deprecated fields

### LINKED ISSUE #1251: Respond warnings if user uses deprecated field
## Summary
Inspired by this [issue](https://github.com/XRPLF/xrpl.js/issues/2611) 
Clio does not support 'type' field which is deprecated. It is [documented](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger/).
We shall respond warnings if user uses deprecated field.
 
## Motivation
User may expect different response. Sending the warning back can clarify the situation. 

## Solution
Notify user they are using deprecated fields.

kuznetsss:
Fixes #1251.


src/rpc/common/Checkers.hpp
```diff
@@ -0,0 +1,134 @@
...
+     * @param key The key to check
+     * @return A warning if the field is deprecated or std::nullopt otherwise
+     */
+    [[nodiscard]] std::optional<Warning>
+    check(boost::json::value const& value, std::string_view key) const
+    {
+        if (value.is_object() and value.as_object().contains(key) and
+            validation::checkType<T>(value.as_object().at(key))) {
+            using boost::json::value_to;
+            auto const res = value_to<T>(value.as_object().at(key));
```
cindyyan317, line 117: Does it throw error? If I have a check Deprecated(2), but field value is "ok"?

src/rpc/common/Checkers.hpp
```diff
@@ -0,0 +1,134 @@
...
+            if (value_ == res) {
+                return Warning{
+                    WarningCode::warnRPC_DEPRECATED, fmt::format("Value '{}' for field '{}' is deprecated", value_, key)
+                };
+            }
+        }
+        return {};
```
godexsoft: `std::nullopt` may be better here, wdyt?

src/rpc/handlers/Ledger.hpp
```diff
@@ -102,7 +103,9 @@ class LedgerHandler {
     {
         static auto const rpcSpec = RpcSpec{
             {JS(full), validation::Type<bool>{}, validation::NotSupported{true}},
```
cindyyan317, line 105: When full is true, Clio handler would not be called. We can remove the NotSupported part. Warning is only given when full is false, right?

cindyyan317 (commented):
'ident' in AccountInfo is deprecated. Please add that one too.
https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_info/

Some fields from subscribe handler are deprecated  too. https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/subscription-methods/subscribe/

src/rpc/common/Types.hpp
```diff
@@ -71,7 +88,29 @@ using HandlerReturnType = util::Expected<OutputType, Status>;
...
+        : result{std::move(result)}, warnings{std::move(warnings)}
+    {
+    }
+
+    /**
+     * @brief operator bool to check whether the ReturnType contains a response
+     */
+    operator bool() const
+    {
+        return result.operator bool();
```
godexsoft, line 106: Do you explicitly invoke the operator for readability? We usually just did `return result;` elsewhere.

godexsoft (dismissed):
Changes look good. Left a few nits that you can safely ignore if you prefer 👍 
suh-dude-ma-dud

src/rpc/common/Checkers.hpp
kuznetsss: There is a type check in the if above. If the field has a different type it will be ignored.

src/rpc/handlers/Ledger.hpp
kuznetsss: I think we should leave NotSupported, in case logic in forwarded part would change. 

Warning is given on any full value if Clio is processing the request. But because we are forwarding on `full: true`, yeah, only when full is false the warning will be given.

src/rpc/common/Types.hpp
kuznetsss: For some reason I didn't think `.operator bool()` could be avoided. But I think this is a bit more easy to read. But I don't mind to change it to `return result;` if you prefer it.

src/rpc/common/Checkers.hpp
kuznetsss: I don't mind about any, changed to `std::nullopt`.

src/rpc/common/Checkers.hpp
cindyyan317: Ha I see. For example, if there is a type restrict Type<std::string> ahead, but the field value is int. Does Clio send back the "deprecated" warning? or no warning?

src/rpc/handlers/AccountInfo.hpp
```diff
@@ -131,8 +132,11 @@ class AccountInfoHandler {
         static auto const rpcSpecV1 = RpcSpec{
             {JS(account), validation::AccountValidator},
             {JS(ident), validation::AccountValidator},
+            {JS(ident), check::Deprecated{}},
             {JS(ledger_hash), validation::Uint256HexStringValidator},
-            {JS(ledger_index), validation::LedgerIndexValidator}
+            {JS(ledger_index), validation::LedgerIndexValidator},
+            {JS(ledger), check::Deprecated{}},
+            {JS(strict), check::Deprecated{}}
```
cindyyan317, line 139: If we add warning for "strict" here, we shall add it for other handlers too. I remember a lot of handlers have this deprecated field.

src/rpc/common/Checkers.hpp
kuznetsss: If we specify `{"some_field", Deprecated{"some string"}}` but in the request it is `"some_field": 1` there will be no warning because deprecated value only string "some string".

src/rpc/handlers/AccountLines.hpp
```diff
@@ -136,6 +137,8 @@ class AccountLinesHandler {
              modifiers::Clamp<int32_t>{LIMIT_MIN, LIMIT_MAX}},
             {JS(ledger_index), validation::LedgerIndexValidator},
             {JS(marker), validation::AccountMarkerValidator},
+            {JS(ledger), check::Deprecated{}},
+            {"peer_index", check::Deprecated{}},
```
godexsoft, line 141: I assume there is no JS string for this?

godexsoft (approved):
Awesome! 🚀

src/rpc/handlers/AccountLines.hpp
kuznetsss: No

PR #1336, opened 2024-04-08
Remove Any wrapper

### LINKED ISSUE #1174: Remove Any wrapper

godexsoft:
Fixes #1174


unittests/util/async/AnyOperationTests.cpp
```diff
@@ -46,8 +46,7 @@ using AnyOperationDeathTest = AnyOperationTests;
 
 TEST_F(AnyOperationTests, VoidDataYieldsNoError)
 {
-    auto const noError = std::expected<impl::Any, ExecutionError>(impl::Any{});
-    EXPECT_CALL(mockOp, get()).WillOnce(Return(noError));
+    EXPECT_CALL(mockOp, get()).WillOnce(Return(std::any{}));
```
godexsoft, line 49: Gcc-12 was complaining about the copy. Not quite sure why. Clang and AppleClang are fine with it.
Clangd is still complaining about a few places in the file but I'm hoping this is yet another issue that will get ironed out by moving the tooling to llvm 18 as well as better gcc-12 and clang integration in our CI docker image.

src/util/async/AnyExecutionContext.hpp
```diff
@@ -66,9 +65,9 @@ class AnyExecutionContext {
     execute(SomeHandlerWithoutStopToken auto&& fn)
     {
         using RetType = std::decay_t<decltype(fn())>;
-        static_assert(not std::is_same_v<RetType, impl::Any>);
+        static_assert(not std::is_same_v<RetType, std::any>);
```
kuznetsss, line 68: Why not add this to the concept?

src/util/async/AnyExecutionContext.hpp
```diff
@@ -211,14 +208,14 @@ class AnyExecutionContext {
 
         virtual impl::ErasedOperation
         execute(
-            std::function<impl::Any(AnyStopToken)>,
+            std::function<std::any(AnyStopToken)>,
```
kuznetsss, line 211: Not related to the PR but I'm curious why did you need to use any with type erased stop token?

cindyyan317 (approved):
LGTM

src/util/async/AnyExecutionContext.hpp
godexsoft: Because we need to operate on arbitrary result type.. unfortunately it does not work very well with virtual functions so you need some type that can represent any result type. AnyStopToken only wraps the Token, not the result type in this case

src/util/async/AnyExecutionContext.hpp
godexsoft: No particular reason.. it was just coded up this way. We can change it but i'd prefer a separate issue for those.
Let's do it in #1346

PR #1348, opened 2024-04-11
Support clang 16 in docker CI

### LINKED ISSUE #1175: Build clio with clang
Currently in our CI we are building on Linux using gcc and on macOS using apple-clang. We should add (or switch one of the current jobs on Linux) a job using clang.

godexsoft:
Fixes #1175 

- Adds clang-16 as a secondary compiler to the docker CI image
- Slightly improves how gcc-12 is integrated
- Sets up the CI to make use of clang and run tests compiled by it alongside gcc's build
- These are not the robots you are looking for

Currently clang-tidy is producing a lot of interesting output. But i think this can be handled outside of this PR now.


cmake/CheckCompiler.cmake
```diff
@@ -1,20 +1,20 @@
 if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 14)
-    message(FATAL_ERROR "Clang 14+ required for building clio")
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 16)
```
kuznetsss, line 2: Need to update documentation about minimum compiler version supported.

docker/ci/dockerfile
```diff
@@ -67,8 +78,25 @@ RUN conan profile new default --detect \
...
-    && conan profile update env.CXX=/opt/bin/g++ default \
-    && conan profile update "conf.tools.build:compiler_executables={\"c\": \"/opt/bin/gcc\", \"cpp\": \"/opt/bin/g++\"}" default \
+    && conan profile update settings.compiler.libcxx=libstdc++11 default \    
     && conan remote add --insert 0 conan-non-prod http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod
+
+RUN conan profile new gcc --detect \
+    && conan profile update settings.compiler=gcc gcc \
+    && conan profile update settings.compiler.version=12 gcc \
+    && conan profile update settings.compiler.cppstd=20 gcc \
+    && conan profile update settings.compiler.libcxx=libstdc++11 gcc
```
kuznetsss: Probably we need to specify `conf.tools.build:compiler_executables` here as well

docker/ci/dockerfile
```diff
@@ -67,8 +78,25 @@ RUN conan profile new default --detect \
     && conan profile update settings.compiler=gcc default \
     && conan profile update settings.compiler.version=12 default \
     && conan profile update settings.compiler.cppstd=20 default \
-    && conan profile update settings.compiler.libcxx=libstdc++11 default \
-    && conan profile update env.CC=/opt/bin/gcc default \
-    && conan profile update env.CXX=/opt/bin/g++ default \
-    && conan profile update "conf.tools.build:compiler_executables={\"c\": \"/opt/bin/gcc\", \"cpp\": \"/opt/bin/g++\"}" default \
+    && conan profile update settings.compiler.libcxx=libstdc++11 default \
```
kuznetsss: I think we don't need default profile anymore

src/util/CMakeLists.txt
```diff
@@ -36,8 +36,8 @@ target_link_libraries(
          clio_options
 )
 
-if (is_gcc)
-  # FIXME: needed on gcc for now
+if (is_gcc OR is_clang)
```
kuznetsss: Could it be something else? msvc exists but we don't support it.

src/main/CMakeLists.txt
```diff
@@ -3,13 +3,14 @@ target_sources(clio PRIVATE impl/Build.cpp)
...
+target_compile_features(clio PUBLIC cxx_std_23)
 
 # Clio server
 add_executable(clio_server)
 target_sources(clio_server PRIVATE Main.cpp)
 target_link_libraries(clio_server PRIVATE clio)
 target_link_options(
-  clio_server PRIVATE $<$<AND:$<NOT:$<BOOL:${APPLE}>>,$<NOT:$<BOOL:${san}>>>:-static-libstdc++ -static-libgcc>
+  # For now let's assume that we only using libstdc++ under gcc
+  clio_server PRIVATE $<$<AND:$<BOOL:${is_gcc}>,$<NOT:$<BOOL:${san}>>>:-static-libstdc++ -static-libgcc>
```
kuznetsss, line 16: If possible maybe lets replace it with a regular if?

kuznetsss (changes requested):
Really good improvements 👍 Just a few comments.

cmake/CheckCompiler.cmake
godexsoft: Do you know where else do we have it? We already updated `build-clio.md` in a previous PR - we just forgot to update here previously. Let me know if there are other places that i forgot about 👍

docker/ci/dockerfile
godexsoft: We actually don't - i checked locally that it works with `gcc` profile as it is now. I'd avoid specifying extra paths when possible. For clang-16 it was still needed however.

cmake/CheckCompiler.cmake
kuznetsss: I forgot that we have already updated `build-clio.md`. I think it was the only place to update.

docker/ci/dockerfile
kuznetsss: Do clangd and clang-tidy work without it? The last time I was fixing this because clangd and clang-tidy were not able to find the right toolchain by the link.

docker/ci/dockerfile
godexsoft: I'm not sure about this, I could try to remove it. I vaguely remember that i had issues before if default did not exist but can't quite remember at which stage. 
In any case, do you think it hurts to have it? If anything, it allows anyone using the docker image to build Clio without specifying a profile.

src/util/CMakeLists.txt
godexsoft: The other option we have for this `if` is `is_appleclang` which is the one that we try to dodge here. As for MSVC and Windows - I'd not bother adding any extra checks for it as we don't support Windows anyway. Wdyt?

docker/ci/dockerfile
kuznetsss: Maybe you are right. I was thinking that from CI perspective it is better to get an error if we forget to specify profile. From user perspective less options is better of course.
Also could default be a link, to not duplicate the code of profile creation?

src/main/CMakeLists.txt
godexsoft: There is a difference between this syntax and regular `if`. This one evaluates at `make` stage while the regular `if` evaluates at configuration stage. Having that said, I don't really see why this needed to be the strange syntax to start with. I think normal `if` could work here just as well. I was not sure however so decided to leave it as it was to avoid accidentally breaking something that i don't fully understand.

src/util/CMakeLists.txt
kuznetsss: I think apple clang also requires this flag. About windows I agree that we don't need to be bothered.

docker/ci/dockerfile
godexsoft: I have not tried them and frankly did not think about that part - I'll add the paths back just in case 👍

docker/ci/dockerfile
godexsoft: Hm, a symlink is actually a decent idea i think 👍

kuznetsss (approved):
Looks good 👍

PR #1360, opened 2024-04-15
Remove `full` from ledger RPC forwarding

### LINKED ISSUE #777: Rippled response causes failed forward
UPDATE: we should remove `full` entirely as it's now deprecated. let's also remove `accounts` and any other deprecated fields. See the [doc here](https://xrpl.org/ledger.html#request-format).

-------
Rippled responds slowly for the request 
```
{
    "method": "ledger",
    "params": [
        {
            "ledger_index": "validated",
            "accounts": false,
            "full": true,
            "transactions": false,
            "expand": false,
            "owner_funds": false
        }
    ]
}
```
which causes failed forward. 
Do we need to increase the timeout or give warning for this scenario? @godexsoft

godexsoft:
Fixes #777 

Removes deprecated fields from forwarding proxy.

### DIFF
--- a/src/rpc/handlers/Ledger.hpp
+++ b/src/rpc/handlers/Ledger.hpp
@@ -62,13 +62,16 @@ public:
      * @brief A struct to hold the input data for the command
      *
      * Clio does not support:
-     * - accounts
-     * - full
-     * - owner_finds
      * - queue
+     *
+     * And the following are deprecated altogether:
+     * - full
+     * - accounts
+     * - ledger
      * - type
      *
-     * Clio will throw an error when any of `accounts`/`full`/`owner_funds`/`queue` are set to `true`
+     * Clio will throw an error when `queue` is set to `true`
+     * or if `full` or `accounts` are used.
      * @see https://github.com/XRPLF/clio/issues/603
      */
     struct Input {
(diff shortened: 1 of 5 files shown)


kuznetsss (dismissed):
LGTM 👍

src/rpc/handlers/Ledger.hpp
```diff
@@ -102,9 +105,9 @@ class LedgerHandler {
...
     {
         static auto const rpcSpec = RpcSpec{
-            {JS(full), validation::Type<bool>{}, validation::NotSupported{true}},
+            {JS(full), validation::NotSupported{}},
             {JS(full), check::Deprecated{}},
-            {JS(accounts), validation::Type<bool>{}, validation::NotSupported{true}},
+            {JS(accounts), validation::NotSupported{}},
             {JS(accounts), check::Deprecated{}},
             {JS(owner_funds), validation::Type<bool>{}},
             {JS(queue), validation::Type<bool>{}, validation::NotSupported{true}},
```
kuznetsss, line 113: What about `queue`? You removed it from ForwardingProxy but left value here.

cindyyan317 (dismissed):

src/rpc/handlers/Ledger.hpp
godexsoft: Thanks for spotting my mistake - i guess i was too enthusiastic about removing stuff :rage4:

kuznetsss (approved):

Issue #442, opened 2022-12-09
Handle SIGHUP and friends gracefully


godexsoft:
Currently sending any signal to clio will stop it as there is no handling of signals at all atm.
Let's define which signals should do what and implement the handlers. The goal is to shutdown gracefully on a TBD set of signals.

kuznetsss:
Take into account the new `Source` design or maybe change it to something better.
https://github.com/XRPLF/clio/pull/1171#discussion_r1489192405

PR #1365, opened 2024-04-18
Add signals handler

kuznetsss:
For #442.


src/util/SignalsHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+        ASSERT(handler_ != nullptr, "SignalsHandler is not initialized");
+        handler_->secondSignalHandler_(signal);
+    }
+};
+
+SignalsHandler* SignalsHandlerStatic::handler_ = nullptr;
+
+}  // namespace impl
+
+SignalsHandler::SignalsHandler(Config const& config, std::function<void(std::string)> forceExitHandler)
```
cindyyan317: I am thinking the type of forceExitHandler std::function<void(std::string)> . Do we really need std::string here? It seems only for logging. How about log in lambda ?

src/util/SignalsHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+    })
+    , secondSignalHandler_([forceExitHandler = std::move(forceExitHandler)](int) {
+        forceExitHandler("Force exit on second signal.");
+        setHandler();
+    })
+{
+    impl::SignalsHandlerStatic::registerHandler(*this);
+
+    auto const gracefulPeriod =
+        std::round(config.valueOr("graceful_period", 10.f) * static_cast<float>(util::MILLISECONDS_PER_SECOND));
```
cindyyan317, line 109: Do we need to add this to docs and example config?

src/util/SignalsHandler.cpp
kuznetsss: Good idea, thanks!

src/util/SignalsHandler.cpp
kuznetsss: This PR doesn't implement graceful period feature, it is only a part of it. But I added docs and config example about it.

cindyyan317 (approved):
😆  Unittests have been fixed ~

godexsoft (approved):
Very nice 🚀

Issue #1354, opened 2024-04-12
Stop Clio when cache has error


cindyyan317:
If successor table contains incorrect data:
1. contains deleted object in list
2. does not contain existing object in list

We should detect this ASAP and act before wrong data gets written to the DB. 

For case 1, we can detect error [here]( 
https://github.com/XRPLF/clio/blob/e744a5a8a90b06a7cb83f623f2a3b74eea7e8cf8/src/data/BackendInterface.cpp#L322). 
 
We shall stop Clio or disable cache to avoid writing more wrong data to successor table.

For case 2, we need shamap library to detect the corruption.

kennyzlei:
After we have shamap library in libxrpl then we can work on addressing case 2.  @cindyyan317 can you create a github issue on rippled side for the shamap library in libxrpl? 🙏

cindyyan317:
Feature requested: https://github.com/XRPLF/rippled/issues/5038

PR #1368, opened 2024-04-22
Disable cache on missing data

godexsoft:
For #1354 

This is the first part of the issue. The second part will have to wait and be implemented once we have SHAMap available as a library in Clio.

### DIFF
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -326,7 +326,7 @@ BackendInterface::fetchLedgerPage(
                 msg << " - " << ripple::strHex(keys[j]);
             }
             LOG(gLog.error()) << msg.str();
-            LOG(gLog.error()) << "Disabling the cache to avoid corrupting the DB.";
+            LOG(gLog.error()) << "Disabling the cache to avoid corrupting the DB further. Please investigate.";
             cache_.setDisabled();
         }
     }
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -297,7 +297,7 @@ BackendInterface::fetchLedgerPage(
     std::uint32_t const limit,
     bool outOfOrder,
     boost::asio::yield_context yield
-) const
+)
 {
     LedgerPage page;
 
@@ -326,6 +326,8 @@ BackendInterface::fetchLedgerPage(
                 msg << " - " << ripple::strHex(keys[j]);
             }
             LOG(gLog.error()) << msg.str();
+            LOG(gLog.error()) << "Disabling the cache to avoid corrupting the DB.";
+            cache_.setDisabled();
         }
     }
     if (!keys.empty() && !reachedEnd)
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -326,8 +326,9 @@ BackendInterface::fetchLedgerPage(
                 msg << " - " << ripple::strHex(keys[j]);
             }
             LOG(gLog.error()) << msg.str();
-            LOG(gLog.error()) << "Disabling the cache to avoid corrupting the DB further. Please investigate.";
-            cache_.setDisabled();
+
+            if (corruptionDetector_.has_value())
+                corruptionDetector_->onCorruptionDetected();
         }
     }
     if (!keys.empty() && !reachedEnd)
--- a/src/etl/CorruptionDetector.hpp
+++ b/src/etl/CorruptionDetector.hpp
@@ -1,0 +1,67 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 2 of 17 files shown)


unittests/data/BackendInterfaceTests.cpp
```diff
@@ -72,3 +77,41 @@ TEST_F(BackendInterfaceTest, FetchFeesLegacySuccessPath)
...
+    using namespace ripple;
+
+    backend->setRange(MINSEQ, MAXSEQ);
+    EXPECT_FALSE(backend->cache().isDisabled());
+    EXPECT_CALL(*backend, doFetchSuccessorKey(_, _, _))
+        .Times(10)
+        .WillRepeatedly(Return(uint256{"1FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF1FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"}));
+    EXPECT_CALL(*backend, doFetchLedgerObjects(_, _, _))
+        .WillOnce(Return(std::vector<Blob>{
+            Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}, Blob{'s'}
```
kuznetsss: `std::vector<Blob>{10, Blob{'s'}}` would look a bit nicer

src/data/BackendInterface.cpp
```diff
@@ -326,6 +326,8 @@ BackendInterface::fetchLedgerPage(
                 msg << " - " << ripple::strHex(keys[j]);
             }
             LOG(gLog.error()) << msg.str();
+            LOG(gLog.error()) << "Disabling the cache to avoid corrupting the DB further. Please investigate.";
```
kuznetsss: Will not this significantly slow down clio?
If yes, I think we need to notify about it more loud than just in the log. I would suggest to add Prometheus metric here and also add a warning to server_info response. WDYT?

cindyyan317: Add a flag in server_info +1 . Adding a integrity flag is a good idea.  
And highlight the consequence of corruption is incorrect data for ledger_data and book_offers .

cindyyan317: Yes . It will slow down the Clio. Another solution will be only disable cache for ETL side.  
https://github.com/XRPLF/clio/blob/develop/src/etl/impl/LedgerFetcher.hpp#L99
The cache will be still workable for RPC side. But data is incorrect.

cindyyan317 (commented):
The implementation looks good to me. Just a few comment on the feature itself

unittests/data/BackendInterfaceTests.cpp
godexsoft: Agree 👍 copy-pasting is more fun though 😃

src/data/BackendInterface.cpp
godexsoft: Well, you are definitely right that we should highlight it more.. i'll cook something up.

src/etl/CorruptionDetector.hpp
```diff
@@ -0,0 +1,67 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
cindyyan317: 2024

src/etl/impl/CacheLoader.hpp
```diff
@@ -121,7 +122,7 @@ class CacheLoaderImpl {
                 auto [start, end] = cursor.value();
                 LOG(log_.debug()) << "Starting a cursor: " << ripple::strHex(start);
 
-                while (not token.isStopRequested()) {
+                while (not token.isStopRequested() and not cache_.get().isDisabled()) {
```
cindyyan317, line 125: Here is a edge case: If corruption happens at the last cursor, the setFull will still be called. Some cache functions still can be called : https://github.com/XRPLF/clio/blob/develop/src/data/LedgerCache.cpp#L92

cindyyan317:
Nice wrote !
Found an edge case, please recheck if it is valid.
I am still thinking we should disable all cache usage or just for ETL? 🤔  
If we disable all cache usage, it makes the problem visible to users. Maybe it is better choice. But it indeed affects the performance.

src/etl/impl/CacheLoader.hpp
godexsoft: Not sure how that would matter if both `update` and `setFull` are checking the `disabled_` state and short circuit? I have added more checks inside of cache member functions but i don't think anything needs changing in CacheLoader, wdyt?

godexsoft:
> Nice wrote ! Found an edge case, please recheck if it is valid. I am still thinking we should disable all cache usage or just for ETL? 🤔 If we disable all cache usage, it makes the problem visible to users. Maybe it is better choice. But it indeed affects the performance.

I'd think let's disable it entirely for now. If we find it become an issue we can always adjust it later.

src/etl/impl/CacheLoader.hpp
cindyyan317: I saw u added the disable check for member function. Unfortunately, it will break here https://github.com/XRPLF/clio/blob/82b8316978e2b9133cd0b57d123a7ca9e739ee9c/src/etl/impl/LedgerLoader.hpp#L204
In ledgerLoader, whether the cache is disable or not, it will use it.

godexsoft: As we discussed on slack - looks like it should be safe due to the disabled checks inside of update and setFull as well as how user-disabled cache is activated after initial load.

cindyyan317 (approved):
👍

PR #1377, opened 2024-04-25
Static linkage

### LINKED ISSUE #1300: Support static linkage of Clio
## Summary
Clio could benefit from supporting fully static linkage. 

Note: also need to fix `main/CMakeLists.txt` to link libc++ statically for clang.

## Motivation
With the move to gcc-12, a new library is dynamically linked (libatomic1). Stock Ubuntu 20.04 which is a supported platform to run Clio does not have it installed by default (the package can be installed with `apt install libatomic1`). This means that if Clio is built with the newer compiler it will fail to run on stock Ubuntu 20.04. This is of course undesired.

## Solution
Allow static linkage, if not for all libraries then at least for some or just atomic1.

godexsoft:
Fixes #1300 

Adds a `static` option for both conan and cmake. Uses the option on CI for all linux builds.
Note: static linkage not currently supported for AppleClang path.

I'm not too happy about adding flags like `-static` manually but that's the best way i found to achieve what we wanted. If you know a more `cmake`-y way of doing it please do let me know 👍

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,8 @@ class Clio(ConanFile):
     description = 'Clio RPC server'
     settings = 'os', 'compiler', 'build_type', 'arch'
     options = {
-        'fPIC': [True, False],
+        'static': [True, False],    # static linkage
+        'fPIC': [True, False],      # unused?
         'verbose': [True, False],
         'tests': [True, False],     # build unit tests; create `clio_tests` binary
         'benchmark': [True, False], # build benchmarks; create `clio_benchmarks` binary
@@ -31,6 +32,7 @@ class Clio(ConanFile):
     ]
 
     default_options = {
+        'static': False,
         'fPIC': True,
         'verbose': False,
         'tests': False,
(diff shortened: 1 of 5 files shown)


conanfile.py
```diff
@@ -9,7 +9,8 @@ class Clio(ConanFile):
     description = 'Clio RPC server'
     settings = 'os', 'compiler', 'build_type', 'arch'
     options = {
-        'fPIC': [True, False],
+        'static': [True, False],    # static linkage
+        'fPIC': [True, False],      # unused?
```
godexsoft, line 13: We could remove this but i'm not 100% sure whether this is used by some deps or what. 
Or we could add `-fpic` flag to the compiler in case this is set.

cindyyan317, line 13: Looks like not used

cindyyan317 (approved):
LGTM

conanfile.py
godexsoft: Yea but it could be used internally by some dep or something. Let's leave as is for now and try to remove it later.

PR #1379, opened 2024-04-26
Combine requestStop and cancel as abort

### LINKED ISSUE #1367: Async framework: combine cancel and stop
## Summary

We should provide only one way of canceling and stopping scheduled operations in async framework.

## Motivation

Currently there are two parameters on scheduled operation callback - `stopToken`  and `bool`. In most cases user doesn't need to know the reason of stopping. 

## Solution

We should simplify it to one parameter `stopToken`. Also in case user still need to get a reason of stopping we should add method `reason()` returning enum value `Canceled` or `Stopped`.

godexsoft:
Fixes #1367 

This does not implement the `stopToken.reason()` idea from the issue as we decided to leave that for a later day.

### DIFF
--- a/src/util/SignalsHandler.cpp
+++ b/src/util/SignalsHandler.cpp
@@ -119,10 +119,8 @@ SignalsHandler::~SignalsHandler()
 void
 SignalsHandler::cancelTimer()
 {
-    if (timer_.has_value()) {
-        timer_->cancel();
-        timer_->requestStop();
-    }
+    if (timer_.has_value())
+        timer_->abort();
 }
 
 void
--- a/src/util/SignalsHandler.cpp
+++ b/src/util/SignalsHandler.cpp
@@ -84,7 +84,7 @@ SignalsHandler::SignalsHandler(Config const& config, std::function<void()> force
             gracefulPeriod_,
             [forceExitHandler = std::move(forceExitHandler)](auto&& stopToken, bool canceled) {
                 // TODO: Update this after https://github.com/XRPLF/clio/issues/1380
-                if (not stopToken.isStopRequested() and not canceled) {
+                if (not stopToken.isStopRequested() or not canceled) {
                     LOG(LogService::warn()) << "Force exit at the end of graceful period.";
                     forceExitHandler();
                 }
--- a/src/util/SignalsHandler.cpp
+++ b/src/util/SignalsHandler.cpp
@@ -84,7 +84,7 @@ SignalsHandler::SignalsHandler(Config const& config, std::function<void()> force
             gracefulPeriod_,
             [forceExitHandler = std::move(forceExitHandler)](auto&& stopToken, bool canceled) {
                 // TODO: Update this after https://github.com/XRPLF/clio/issues/1380
-                if (not stopToken.isStopRequested() or not canceled) {
+                if (not stopToken.isStopRequested() and not canceled) {
                     LOG(LogService::warn()) << "Force exit at the end of graceful period.";
                     forceExitHandler();
                 }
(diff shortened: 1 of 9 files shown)


src/util/async/Operation.hpp
```diff
@@ -120,6 +120,16 @@ struct BasicScheduledOperation {
...
         state_->get().requestStop();
     }
+
+    void
+    abort() noexcept
+    {
+        if constexpr (SomeCancellableOperation<OpType>)
+            timer_.cancel();
+
+        if constexpr (SomeCancellableOperation<OpType>)
```
kuznetsss: Shouldn't be here a different condition?
If yes, why tests didn't catch it?

src/util/async/impl/ErasedOperation.hpp
```diff
@@ -119,22 +106,15 @@ class ErasedOperation {
...
-        }
-
-        void
-        cancel() override
+        abort() override
         {
-            if constexpr (SomeCancellableOperation<OpType>) {
-                operation.cancel();
+            if constexpr (not SomeCancellableOperation<OpType> and not SomeStoppableOperation<OpType>) {
+                ASSERT(false, "Called abort() on an operation that can't be cancelled nor stopped");
```
kuznetsss, line 112: Maybe you could move it to concept? Looks like it could be checked at compile time.

src/util/SignalsHandler.cpp
```diff
@@ -119,10 +119,8 @@ SignalsHandler::~SignalsHandler()
 void
```
kuznetsss, line 119: For some reason I can't leave the comment at any place in the file.

If this is fixing #1379, we need to create a new issue to move bool in callback into stop token and adjust this TODO.
https://github.com/XRPLF/clio/blob/0dcbbf9afa9bd8b95d173cd3b0a4b669fb3d52bb/src/util/SignalsHandler.cpp#L86

src/util/async/Operation.hpp
godexsoft: Good catch, thanks. 
The test is checking similar logic in AnyOperation, not this one. I'll see if a test can be added 👍

src/util/async/impl/ErasedOperation.hpp
godexsoft: This can't be checked by a concept afaik. It's because this code is inside of a virtual function override.

src/util/SignalsHandler.cpp
godexsoft: I'm having second thoughts about changing it this way. But i'll create the ticket and fix this todo. We can then decide later what to do with it 👍

src/util/async/impl/ErasedOperation.hpp
kuznetsss: Maybe it is possible to put concept to `ErasedOperation::abort()` and add `static_assert()` here?

src/util/async/README.md
```diff
@@ -0,0 +1,231 @@
+# Async framework
+
+## Introduction
+
+Clio uses threads intensively. Multiple parts of Clio were/are implemented by running a `std::thread` with some sort of loop inside. Every time this pattern is reimplemented in a slightly different way. State is managed using asynchronous queues, atomic flags, mutexes and other low level primitives.
+
+On the other hand, Clio also uses `Boost.Asio` for more complex tasks such as networking, scheduling RPC handlers, and even interacting with the database is done via Asio’s coroutines.
+
+There was a need for a simple yet powerful framework that will cover the following in a unified way:
```
cindyyan317, line 9: in an unified way

godexsoft: I believe `a` is correct here. Checked with https://www.grammarly.com/sentence-checker as well 👍

src/util/async/impl/ErasedOperation.hpp
godexsoft: We can't add a requirement on `abort` because in `ErasedOperation` we simply don't have the actual type of the operation we are erasing.. This info is only known in the `Model` hence the assert and the checks have to be there. Unfortunately.

kuznetsss (approved):
👍

PR #1382, opened 2024-04-29
Use LLVM 18 tooling

godexsoft:
Fixes #1381  

This is part two of moving to LLVM 18 tooling.


src/rpc/RPCHelpers.cpp
```diff
@@ -197,7 +197,7 @@ accountFromStringStrict(std::string const& account)
     }
 
     if (result) {
-        return result.value();
+        return result;
```
godexsoft: Seems like this can be simply `return result`

unittests/util/MockPrometheus.hpp
```diff
@@ -138,10 +138,10 @@ struct MockPrometheusImpl : PrometheusInterface {
     getMetric(std::string name, Labels labels)
     {
         auto const labelsString = labels.serialize();
-        auto const key = name + labels.serialize();
+        auto const key = name + labelsString;
         auto it = metrics.find(key);
         if (it == metrics.end()) {
-            return makeMetric<MetricType>(std::move(name), labels.serialize());
+            return makeMetric<MetricType>(std::move(name), labelsString);
```
godexsoft: Maybe we can make it non const and move here

kuznetsss (approved):
👍

PR #1390, opened 2024-05-02
Fix missing tx from account_tx

### LINKED ISSUE #1389: account_tx with forward true is skipping transactions
## Issue Description
if I use account_tx ws query with forward true subsequent request with marker could skip 1 transaction
I'm connecting to wss://s2.ripple.com

## Steps to Reproduce
request 1:
```
{
    "id": 1,
    "command": "account_tx",
    "account": "rf1NrYAsv92UPDd8nyCG4A3bez7dhYE61r",
    "ledger_index_min": 77384137,
    "forward": true,
    "limit": 5
}
```

response will be with:
```
TransactionIndexes: 58, 59, 60, 61, 62
"marker": {
    "ledger": 77384137,
    "seq": 63
}
```

request 2 with marker from previous response:
```
{
    "id": 2,
    "command": "account_tx",
    "account": "rf1NrYAsv92UPDd8nyCG4A3bez7dhYE61r",
    "limit": 5,
    "forward": true,
    "marker": {
        "ledger": 77384137,
        "seq": 63
    }
}
```

response will be with:
```
TransactionIndexes: 64, 65, 66, 67, 68
"marker": {
    "ledger": 77384137,
    "seq": 69
}
```

transaction with index 63 is skipped

## Expected Result
In correct response transaction with index 63 has to be included
I can get expected result from wss://xrplcluster.com, but not from s2.ripple.com
```
TransactionIndexes: 63, 64, 65, 66, 67
"marker": {
    "ledger": 77384137,
    "seq": 68
}
```

cindyyan317:
Since our selectAccountTxForward is no longer inclusive, we should adjust this place.
Fix #1389

### DIFF
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -172,11 +172,6 @@ public:
             if (--numRows == 0) {
                 LOG(log_.debug()) << "Setting cursor";
                 cursor = data;
-
-                // forward queries by ledger/tx sequence `>=`
-                // so we have to advance the index by one
-                if (forward)
-                    ++cursor->transactionIndex;
             }
         }
 


godexsoft (approved):
We really need some sort of tests for this stuff :rage4: 
I'm approving because it looks like the right fix and i know we can't unit test this code right now.

intelliot:
Ouch. The two most common mistakes in programming: syntax errors, error handling, and off-by-one errors.

Should this be removed as well? or is the logic somehow different for `fetchNFTTransactions`? https://github.com/XRPLF/clio/blob/c1f6c2f983a9ff8253cef9501f09513a1ac61cbb/src/data/CassandraBackend.hpp#L444-L448

cindyyan317:
> Ouch. The two most common mistakes in programming: syntax errors, error handling, and off-by-one errors.
> 
> Should this be removed as well? or is the logic somehow different for `fetchNFTTransactions`?
> 
> https://github.com/XRPLF/clio/blob/c1f6c2f983a9ff8253cef9501f09513a1ac61cbb/src/data/CassandraBackend.hpp#L444-L448

Thanks for pointing it out. This logic is okay. Because the [cql](https://github.com/XRPLF/clio/blob/develop/src/data/cassandra/Schema.hpp#L653) is using ">=". To avoid same transaction being returned, +1 is correct. https://github.com/XRPLF/clio/blob/develop/src/data/cassandra/Schema.hpp#L597 is using ">" , +1 is not necessary.

cindyyan317:
Test attached:
Result from current s2[s2.txt](https://github.com/XRPLF/clio/files/15198084/s2.txt)

Result from fixed local clio[fixed.txt](https://github.com/XRPLF/clio/files/15198085/fixed.txt)

PR #1393, opened 2024-05-03
Separate unit tests and integration tests

### LINKED ISSUE #1391: Separate unit tests and DB/integration tests
## Summary
Currently all tests are in one group. It would be better to keep them separate.

## Motivation
Unit tests should be blazingly fast. The DB tests are taking long and slowing down the entire CI.

## Solution
- Split the two types of tests into separate bundles/groups
- Move DB tests to nightly on CI

godexsoft:
Fixes #1391


CMakeLists.txt
```diff
@@ -8,7 +8,8 @@ project(clio VERSION ${CLIO_VERSION} HOMEPAGE_URL "https://github.com/XRPLF/clio
 
 # =========================== Options ====================================== #
 option(verbose "Verbose build" FALSE)
-option(tests "Build tests" FALSE)
+option(tests "Build unit tests" FALSE)
+option(integration_tests "Build database tests" FALSE)
```
kuznetsss: I think here and in conanfile these should be `true` by default.

tests/unit/CMakeLists.txt
```diff
@@ -174,7 +164,7 @@ if (coverage)
     --gtest_brief=1
     ${TESTS_ADDITIONAL_ARGS}
     EXCLUDE
-    "unittests"
+    "tests"
```
kuznetsss, line 167: You need to exclude here all the backend part you excluded in CI.

CMakeLists.txt
godexsoft: I don't think so.. why would it be? 
Normal use-case for non-clio-devs is to build Clio for running it.. why would they need to mess around with unit and integration tests? It was false all along for this reason.

cindyyan317 (approved):
LGTM

PR #1394, opened 2024-05-03
Cover LoadBalancer with tests

### LINKED ISSUE #680: For some subscribe transactions, Clio returns twice the response compared to rippled.

### LINKED ISSUE #1222: Cover LoadBalancer with tests
LoadBalancer is not covered with tests. We should refactor it and add unit tests.

kuznetsss:
Fixes #680. Fixes #1222.


src/etl/LoadBalancer.cpp
```diff
@@ -240,53 +249,51 @@ LoadBalancer::toJson() const
...
+        ret.push_back(src->toJson());
 
     return ret;
 }
 
 template <typename Func>
-bool
-LoadBalancer::execute(Func f, uint32_t ledgerSequence)
+void
+LoadBalancer::execute(Func f, uint32_t ledgerSequence, std::chrono::steady_clock::duration retryTime)
```
cindyyan317: I feel retryAfter might be better?

src/etl/impl/LedgerLoader.hpp
```diff
@@ -197,58 +197,55 @@ class LedgerLoader {
...
-                                ASSERT(succ.has_value(), "Book base {} must have a successor", ripple::strHex(base));
-                                if (succ->key == cur->key) {
-                                    LOG(log_.debug()) << "Writing book successor = " << ripple::strHex(base) << " - "
-                                                      << ripple::strHex(cur->key);
-
-                                    backend_->writeSuccessor(
-                                        uint256ToString(base), sequence, uint256ToString(cur->key)
-                                    );
-                                }
+            auto edgeKeys = loadBalancer_->loadInitialLedger(sequence);
```
cindyyan317, line 200: why we don't check 'success' any more? I think it must be checked before we write anything to DB.

src/etl/impl/SubscriptionSource.hpp
```diff
@@ -162,6 +149,14 @@ class SubscriptionSource {
     bool
     isConnected() const;
 
+    /**
+     * @brief Get whther the source is forwarding
```
cindyyan317: Typo-whether

src/etl/LoadBalancer.cpp
```diff
@@ -150,53 +156,57 @@ LoadBalancer::~LoadBalancer()
     sources_.clear();
 }
 
-std::pair<std::vector<std::string>, bool>
```
godexsoft, line 153: Do i understand correctly that this change was introduced because execute could actually never return false?

src/etl/ETLHelpers.hpp
```diff
@@ -35,6 +32,44 @@
 #include <vector>
 
 namespace etl {
+
+/**
+ * @brief An interface for NetworkValidatedLedgers
+ */
+class NetworkValidatedLedgersInterface {
```
godexsoft: Wouldn't it be better to have individual `.h/.cpp` files for this and the implementation?

src/etl/impl/LedgerLoader.hpp
kuznetsss: Because there is infinite loop inside `loadInitialLedger()` until the ledger is successfully loaded.

https://github.com/kuznetsss/clio/blob/5cfc816a178fc25ed13071db9e73882c10e9e005/src/etl/LoadBalancer.cpp#L266-L296

cindyyan317 (dismissed):

godexsoft (commented):
Refactoring looks good. Just a few small questions 👍

src/etl/LoadBalancer.cpp
kuznetsss: Yeah, `execute()` has infinite loop until provided lambda returns `true`.

godexsoft (approved):
doomguy-godmode

PR #1409, opened 2024-05-14
Date of commit in version string; No git hash in release version

legleux:
A release's version string should be a signed annotated tag which Clio has generally been following.

Uses the date of the commit since that seems like a more useful item to track to not have identical source builds have different version strings.


.githooks/check-format
```diff
@@ -73,7 +73,7 @@ else
     grep_code '#include ".*"' | xargs sed -i -E 's|#include "(.*)"|#include <\1>|g'
 
     # make local includes to be "..." style
-    main_src_dirs=$(find ./src -type d -maxdepth 1 -exec basename {} \; | paste -sd '|' | sed 's/|/\\|/g')
+    main_src_dirs=$(find ./src -maxdepth 1 -type d  -exec basename {} \; | paste -sd '|' | sed 's/|/\\|/g')
```
kuznetsss, line 76: I'm curious why does the order matter here?

kuznetsss (approved):
LGTM

.githooks/check-format
legleux: I'm not sure why I've never seen this before because I've never paid attention to the order at all but:

> find: warning: you have specified the global option -maxdepth after the argument -type, but global options are not positional, i.e., -maxdepth affects tests specified before it as well as those specified after it.  Please specify global options before other arguments.

My Linux bash:
```$ bash --version
GNU bash, version 5.2.26(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2022 Free Software Foundation, Inc.
```
My macOS bash:
```bash
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)
Copyright (C) 2007 Free Software Foundation, Inc.
```
2007 😲

PR #1415, opened 2024-05-20
Fix: Duplicate messages when subscribe both accounts and proposed_accounts

cindyyan317:
This PR is to fix the duplicate messages sent to the subscribers when accounts and proposed_accounts both subscribed by the same client. To align with rippled, Clio will filter out the same session for **some** cases:

Screenshot 2024-05-20 at 09 20 32

proposed_accounts subscribers will receive two json messages :
1. The transaction json when transaction gets proposed
2. The transaction + meta json when transaction gets validated.
To filter out the session has sent the validated json, we only extract the proposed tx json from forwarding source. Clio will send the validated part by itself, being able to do filter.

### DIFF
--- a/src/etl/impl/SubscriptionSource.cpp
+++ b/src/etl/impl/SubscriptionSource.cpp
@@ -240,7 +240,8 @@ SubscriptionSource::handleMessage(std::string const& message)
 
         } else {
             if (isForwarding_) {
-                if (object.contains(JS(transaction))) {
+                // Only forward proposed transactions, validated transactions are sent by Clio itself
+                if (object.contains(JS(transaction)) and !object.contains(JS(meta))) {
                     subscriptions_->forwardProposedTransaction(object);
                 } else if (object.contains(JS(type)) && object.at(JS(type)) == JS_ValidationReceived) {
                     subscriptions_->forwardValidation(object);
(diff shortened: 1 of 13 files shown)


src/etl/impl/SubscriptionSource.cpp
```diff
@@ -240,7 +240,11 @@ SubscriptionSource::handleMessage(std::string const& message)
 
         } else {
             if (isForwarding_) {
-                if (object.contains(JS(transaction))) {
+                // Clio as rippled's proposed_transactions subscirber, will receive two jsons for each transaction
+                // 1 - Proposed transaction
+                // 2 - Validated transaction
+                // Only forward proposed transaction, validated transactions are sent by Clio itself
+                if (object.contains(JS(transaction)) and !object.contains(JS(meta))) {
```
kuznetsss, line 247: Is `!object.contains(JS(meta))` the best way to check whether the transaction is proposed? Doesn't it have some explicit field for that?

kuznetsss (approved):
LGTM

src/etl/impl/SubscriptionSource.cpp
cindyyan317: Unfortunately , there is not. I have noticed the difference by observation. https://github.com/XRPLF/rippled/blob/d5e5c3c2206ea7790ec5fb140355345e46f6755f/src/ripple/app/misc/NetworkOPs.cpp#L2759C9-L2759C18

godexsoft (approved):

PR #1418, opened 2024-05-21
feat: AmendmentCenter

godexsoft:
A starting point for an AmendmentCenter - a central place in Clio to query everything related to amendments on the XRPL. Currently it allows conversion between ID and amendment name as well as querying whether a certain amendment is supported and/or enabled at a given ledger index.


src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+std::vector<Amendment> const&
+AmendmentCenter::getAll() const
+{
+    return all_;
+}
+
+bool
+AmendmentCenter::isEnabled(std::string name, uint32_t seq) const
+{
+    return data::synchronous([this, name, seq](auto yield) { return isEnabled(yield, name, seq); });
```
cindyyan317: I might get it wrong, can not see the call path clearly from diff view.  isEnabled (line 115) will call the overload isEnabled (line 121), then AmendmentKey's constructor will be called implicitly, then insert something to the SUPPORTED_AMENDMENTS.

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+bool
+AmendmentCenter::isEnabled(boost::asio::yield_context yield, AmendmentKey const& key, uint32_t seq) const
+{
+    namespace rg = std::ranges;
+
+    // the amendments should always be present in ledger
```
cindyyan317: It should be always exist. We can add ASSERT here.

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+}
+
+Amendment const&
+AmendmentCenter::operator[](AmendmentKey const& key) const
+{
+    return getAmendment(key);
+}
+
+ripple::uint256
+Amendment::GetAmendmentId(std::string_view const name)
```
cindyyan317: const is not needed?

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include "data/AmendmentCenter.hpp"
+
+#include "data/BackendInterface.hpp"
+#include "data/Types.hpp"
+#include "util/Assert.hpp"
+
+#include <boost/asio/spawn.hpp>
+#include <fmt/compile.h>
+#include <ripple/basics/Slice.h>
```
kuznetsss: I think it is `<xrpl/` now.

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+    return Amendment::GetAmendmentId(name);
+}
+
+AmendmentCenter::AmendmentCenter(std::shared_ptr<data::BackendInterface> const& backend) : backend_{backend}
+{
+    namespace rg = std::ranges;
+    namespace vs = std::views;
+
+    rg::copy(
+        ripple::allAmendments() | vs::transform([&](auto const& p) {
```
kuznetsss, line 87: Maybe it would worth to provide type here instead of `auto` to make it more readable.

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+    );
+
+    for (auto const& am : all_ | vs::filter([](auto const& am) { return am.isSupportedByClio; }))
+        supported_.insert_or_assign(am.name, am);
+}
+
+bool
+AmendmentCenter::isSupported(std::string name) const
+{
+    return supported_.contains(name) && supported_.at(name).isSupportedByClio;
```
cindyyan317: _supported_.at(name).isSupportedByClio_ check seems unnecessary

src/data/AmendmentCenter.cpp
kuznetsss: Probably to make compiler find such issues it is better to add `explicit` to AmendmentKey's constructors.

src/data/AmendmentCenter.cpp
```diff
@@ -0,0 +1,160 @@
...
+    namespace rg = std::ranges;
+
+    // the amendments should always be present in ledger
+    auto const& amendments = backend_->fetchLedgerObject(ripple::keylet::amendments().key, seq, yield);
+
+    ripple::SLE const amendmentsSLE{
+        ripple::SerialIter{amendments->data(), amendments->size()}, ripple::keylet::amendments().key
+    };
+
+    auto const listAmendments = amendmentsSLE.getFieldV256(ripple::sfAmendments);
```
cindyyan317, line 144: We'd better check if field present first. Normally, it should have something. However, it could be not present for new chain. Amendments field states _"Required-No"_ in the [document](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/amendments/#amendments)

src/feed/SubscriptionManagerInterface.hpp
```diff
@@ -234,7 +234,6 @@ class SubscriptionManagerInterface {
 
     /**
      * @brief Get the number of subscribers.
-     *
```
kuznetsss: Probably accidental change

src/rpc/common/impl/HandlerProvider.hpp
```diff
@@ -41,6 +42,9 @@ class Counters;
 namespace feed {
 class SubscriptionManager;
 }  // namespace feed
+namespace data {
+class AmendmentCenter;
```
kuznetsss: Do you need forwarding declaration here? It seems there is only `AmendmentCenterInterface` used in this file.

src/rpc/handlers/AccountInfo.hpp
```diff
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include "data/AmendmentCenter.hpp"
```
kuznetsss: Probably this include is not needed.

tests/unit/Playground.cpp
```diff
@@ -22,6 +22,11 @@
  * Note: Please don't push your temporary work to the repo.
  */
 
-// TEST(PlaygroundTest, MyTest)
+// #include <gmock/gmock.h>
```
kuznetsss, line 25: Probably forgot to delete.

src/data/AmendmentCenter.cpp
godexsoft: No, the purpose of this proxy object is to allow seamless conversions. I just forgot to split the one that writes to SUPPORTED_AMENDMENTS out to a separate subclass - my bad. Great find @cindyyan317

src/data/AmendmentCenter.cpp
godexsoft: I don't see how that would make anything more readable.. what would be good is https://open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0931r0.pdf but it's not in C++ atm

tests/unit/Playground.cpp
godexsoft: No, just restoring what this file should be on git.. sometime ago i think clang-tidy deleted my includes here so i'm adding them back as comments ;)

src/data/AmendmentCenter.cpp
godexsoft: Changing p.second/p.first few lines below would probably help too 🔢

src/feed/SubscriptionManagerInterface.hpp
godexsoft: Actually it wasn't apparently. But i'll bring it back and change the other doxy comment that did not have the space line instead :rage4:

cindyyan317 (dismissed):
Nice work 👍

src/data/AmendmentCenter.cpp
```diff
@@ -45,24 +45,22 @@
...
 namespace {
-std::vector<std::string_view> SUPPORTED_AMENDMENTS = {};
+std::vector<std::string> SUPPORTED_AMENDMENTS = {};
 }  // namespace
 
 namespace data {
+namespace impl {
 
-AmendmentKey::AmendmentKey(std::string_view amendmentName) : name(amendmentName)
+WritingAmendmentKey::WritingAmendmentKey(std::string const& amendmentName) : AmendmentKey{amendmentName}
```
kuznetsss: Probably just string, so you could move it once.

godexsoft: Yes, that should be better

kuznetsss (dismissed):
👍

kuznetsss (dismissed):

kuznetsss (approved):

PR #1423, opened 2024-05-22
Fix invalid syntax in example-config.json

PeterChen13579:
remove "." fixes syntax for example-config; runs with ./clio_server


docs/examples/config/example-config.json
```diff
@@ -67,7 +67,7 @@
         "local_amdin": false
     },
     // Time in seconds for graceful shutdown. Defaults to 10 seconds. Not fully implemented yet.
-    "graceful_period": 10.,
```
kuznetsss, line 70: It would be better to change it to `10.0` to show that it is floating point value

kuznetsss (approved):
Thanks for the fix 👍

PR #1425, opened 2024-05-24
Add option to set X-User header value for forwarded requests

### LINKED ISSUE #1422: Add X-User header in forwarded requests
## Summary

 Clio should be able optionally to pass through rippled's rate limiter.

## Motivation

Rippled is rate limiting requests forwarded by Clio. Sometimes it is not desired.

## Solution

Rippled has ratelimiting on unauthorised requests. Adding `X-User` header to request will turn off ratelimiting on rippled's side for the request. We should always add `X-User` to requests with admin permissions and an option to config to make it possible to set up Clio to add `X-User` header value for all the forwarded requests.

kuznetsss:
Fixes #1422.

### DIFF
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -37,7 +37,7 @@
     ],
     "forwarding": {
         "cache_timeout": 0.250, // in seconds, could be 0, which means no cache
-        "x_user_header": "" // X-User header to add to forwarded requests, won't be added if empty
+        "x_user_header": "" // value of X-User header to add to forwarded requests, won't be added if empty
     },
     "dos_guard": {
         // Comma-separated list of IPs to exclude from rate limiting
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -35,10 +35,7 @@
             "grpc_port": "50051"
         }
     ],
-    "forwarding": {
-        "cache_timeout": 0.250, // in seconds, could be 0, which means no cache
-        "x_user_header": "" // value of X-User header to add to forwarded requests, won't be added if empty
-    },
+    "forwarding_cache_timeout": 0.250, // in seconds, could be 0, which means no cache
     "dos_guard": {
         // Comma-separated list of IPs to exclude from rate limiting
         "whitelist": [
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -35,7 +35,10 @@
             "grpc_port": "50051"
         }
     ],
-    "forwarding_cache_timeout": 0.250, // in seconds, could be 0, which means no cache
+    "forwarding": {
+        "cache_timeout": 0.250, // in seconds, could be 0, which means no cache
+        "x_user_header": "" // X-User header to add to forwarded requests, won't be added if empty
+    },
     "dos_guard": {
         // Comma-separated list of IPs to exclude from rate limiting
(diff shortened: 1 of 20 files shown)


docs/examples/config/example-config.json
```diff
@@ -35,7 +35,10 @@
             "grpc_port": "50051"
         }
     ],
-    "forwarding_cache_timeout": 0.250, // in seconds, could be 0, which means no cache
```
godexsoft: This was not released to the public yet, right?

docs/examples/config/example-config.json
```diff
@@ -64,7 +67,7 @@
         "admin_password": "xrp",
         // If local_admin is true, Clio will consider requests come from 127.0.0.1 as admin requests
         // It's true by default unless admin_password is set,'local_admin' : true and 'admin_password' can not be set at the same time
-        "local_amdin": false
+        "local_admin": false
```
godexsoft, line 67: :rage4:

src/etl/LoadBalancer.cpp
```diff
@@ -77,13 +77,18 @@ LoadBalancer::LoadBalancer(
...
-    auto const forwardingCacheTimeout = config.valueOr<float>("forwarding_cache_timeout", 0.f);
+    auto const forwardingCacheTimeout = config.valueOr<float>("forwarding.cache_timeout", 0.f);
     if (forwardingCacheTimeout > 0.f) {
         forwardingCache_ = impl::ForwardingCache{std::chrono::milliseconds{
             std::lroundf(forwardingCacheTimeout * static_cast<float>(util::MILLISECONDS_PER_SECOND))
         }};
     }
 
+    auto const forwardingXUserValue = config.valueOr<std::string>("forwarding.x_user_value", {});
+    if (not forwardingXUserValue.empty()) {
```
godexsoft: These kind of ifs can be collapsed to `if (auto const value = ...; not value.empty()) ...` if you don't need `forwardingXUserValue` down the line.

godexsoft: or even directly assign `config.maybeValue` to the member if that produces the same effect.

godexsoft (dismissed):
Looks good 🥇  
Leaving a few nits. Feel free to ignore though.

`isAdmin` may become a `role` at some point in the future. But it looks like the change would not be all that big if we will need it.

docs/examples/config/example-config.json
kuznetsss: No, it was not.

docs/examples/config/example-config.json
```diff
@@ -35,7 +35,10 @@
             "grpc_port": "50051"
         }
     ],
-    "forwarding_cache_timeout": 0.250, // in seconds, could be 0, which means no cache
+    "forwarding": {
+        "cache_timeout": 0.250, // in seconds, could be 0, which means no cache
+        "x_user_header": "" // value of X-User header to add to forwarded requests, won't be added if empty
```
cindyyan317: Why we need this configurable? Clio should always add it as a proxy.

cindyyan317 (commented):
According to rippled's comment https://github.com/XRPLF/rippled/blob/develop/cfg/rippled-example.cfg#L249, we shall add instruction in run-clio.md to remind user to add secure_gateway of ws port session

cindyyan317 (approved):

Issue #876, opened 2023-09-28
Move NameGenerator into utils


kuznetsss:
`NameGenerator` is duplicated in many tests. It would be better to move it into `util` and reuse in every test.

kuznetsss:
Also name generator can be very simple:
```cpp
  [](auto const& info) { return info.param.testName; }
  ```

PR #1428, opened 2024-05-27
Move NameGenerator to util

PeterChen13579:
Fixes for #876


tests/common/util/NameGenerator.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+namespace tests::util {
+
+struct NameGenerator {
+    inline static auto
+    GenerateName()
+    {
+        return [](auto const& info) { return info.param.testName; };
+    }
+};
```
kuznetsss: You could do it shorter:
```cpp
static const auto NameGenerator = [](auto const& info) { return info.param.testName; };
```
And then you also will not need to put `()` when using it.

kuznetsss (dismissed):
👍

tests/common/util/NameGenerator.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include <string>
+
+namespace tests::util {
+
+static auto const NameGenerator = [](auto const& info) { return info.param.testName; };
+
+}  // namespace tests::util
```
godexsoft: Please set your IDE up to automatically add `\n` at end of files on save 👍

godexsoft (changes requested):
Looks good 💯 
Just one thing - let's set your IDE up so github is happy 🥇

tests/common/util/NameGenerator.hpp
kuznetsss: Why do we need to have an empty line at the end of each file?

If we think it is useful, we could force it on project level by adjusting clang-tidy's config to add new line on formatting.

godexsoft: Yeah, we probably should set it up then. It's not "useful" but imo it's always better to satisfy github so it does not display this red icon at the bottom of each file in the diffs.

kuznetsss: The red icon just indicates the end of file. 
https://stackoverflow.com/questions/77201426/what-does-a-red-icon-with-a-hyphen-mean-in-a-github-pr

kuznetsss: But I don't mind if we turn on this rule for clang-tidy. But it should be fixed separately and maybe even the file could be fixed automatically by clang-tidy.

godexsoft: That's just misinformation. Even your link says "no newline at end of file" in the accepted answer.
In my experience, the icon only ever appears if the newline is missing at the end of the file. I'm for consistency - the rest of the codebase (hopefully) follows the newline rule. And if we can enforce it with clang-format/tidy then we should do that as well 👍

kuznetsss: I created an issue #1434

godexsoft (approved):

Issue #1431, opened 2024-05-28
Update doxygen to 1.12


kuznetsss:
Doxygen 1.11 has been released. We need to upgrade it in our CI. Also 1.11 produces error on checking Clio, we need to fix it as well.

godexsoft:
Doxygen 1.11.0 breaks `@ref` generation for our mainpage. Reported issue: https://github.com/doxygen/doxygen/issues/10906

godexsoft:
Doxygen 1.12 is out and seem to have fixed our issue. Let's start using it

PR #1432, opened 2024-05-28
Comment out precommit hook for Doc

PeterChen13579:
Temporary Fix for #1431

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -3,4 +3,4 @@
 # This script is intended to be run from the root of the repository.
 
 source .githooks/check-format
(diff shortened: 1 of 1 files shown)


kuznetsss (dismissed):
👍

godexsoft (changes requested):
Can we add a TODO with a reference to the issue on github that will be fixing doxygen and bringing this back please 👍

godexsoft (approved):
We now also have [this issue on doxygen github](https://github.com/doxygen/doxygen/issues/10906). They actually already proposed a fix but it will take forever to get it in and release a new version, i bet.

PR #1433, opened 2024-05-28
Flow to check new libXRPL version

godexsoft:
This is the Clio side of things for https://github.com/XRPLF/rippled/pull/5028

The new flow is triggered remotely from `rippled`'s CI once a new version of `libXRPL` is proposed. 
If either building or test are broken an issue for investigation is automatically created.


.githooks/update-libxrpl-version
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
kuznetsss: I don't think it is a githook. Maybe put it into `tools` or `.github/scripts`?

.github/workflows/check_libxrpl.yml
```diff
@@ -0,0 +1,91 @@
+name: Check new libXRPL
+on:
+  repository_dispatch:
+    types: [check_libxrpl]
```
kuznetsss, line 4: Where is this type defined?

godexsoft: It's not necessarily defined anywhere afaik. It's the same string that we pass from `rippled` CI though so it matches this particular flow using it.

kuznetsss (approved):
👍

PR #1439, opened 2024-05-30
Create generate free port class to avoid conflicting ports

### LINKED ISSUE #1317: Use random ports in tests
Currently some of our unit tests which need to start some server are using hardcoded ports. This could lead to an issue if the hardcoded port is taken by some other application.

The solution here is to make a small library using `0` port to ask operating system for a random free port. And then use the library everywhere in tests.

PeterChen13579:
Fixes #1317

### DIFF
--- a/tests/common/util/AssignRandomPort.cpp
+++ b/tests/common/util/AssignRandomPort.cpp
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2023, the clio developers.
+    Copyright (c) 2024, the clio developers.
 
     Permission to use, copy, modify, and distribute this software for any
     purpose with or without fee is hereby granted, provided that the above
@@ -19,6 +19,13 @@
 
 #include "util/AssignRandomPort.hpp"
 
+#include <boost/asio/io_context.hpp>
+#include <boost/asio/ip/tcp.hpp>
+
+#include <cstdint>
+
+using tcp = boost::asio::ip::tcp;
+
 namespace tests::util {
 
 uint32_t
@@ -32,8 +39,8 @@ generateFreePort()
     acceptor.set_option(tcp::acceptor::reuse_address(true));
     acceptor.bind(endpoint);
 
-    tcp::endpoint endPoint = acceptor.local_endpoint();
-    return endPoint.port();
+    tcp::endpoint localEndPoint = acceptor.local_endpoint();
+    return localEndPoint.port();
 }
 
 }  // namespace tests::util
--- a/tests/common/util/AssignRandomPort.cpp
+++ b/tests/common/util/AssignRandomPort.cpp
@@ -1,0 +1,39 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include "util/AssignRandomPort.hpp"
+
+namespace tests::util {
+
+uint32_t
+generateFreePort()
+{
+    boost::asio::io_context io_context;
+    tcp::acceptor acceptor(io_context);
+    tcp::endpoint const endpoint(tcp::v4(), 0);
+
+    acceptor.open(endpoint.protocol());
+    acceptor.set_option(tcp::acceptor::reuse_address(true));
+    acceptor.bind(endpoint);
+
+    tcp::endpoint endPoint = acceptor.local_endpoint();
+    return endPoint.port();
+}
+
+}  // namespace tests::util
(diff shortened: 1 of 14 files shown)


kuznetsss:
There are also `tests/unit/web/ServerTests.cpp`, `tests/unit/etl/SubscriptionSourceTests.cpp` and `tests/unit/rpc/ForwardingProxyTests.cpp` where we should use random port.

tests/common/util/AssignRandomPort.cpp
```diff
@@ -0,0 +1,39 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: 2024

tests/common/util/AssignRandomPort.hpp
```diff
@@ -0,0 +1,32 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: also 2024

tests/common/util/AssignRandomPort.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+    boost::asio::io_context io_context;
+    tcp::acceptor acceptor(io_context);
+    tcp::endpoint const endpoint(tcp::v4(), 0);
+
+    acceptor.open(endpoint.protocol());
+    acceptor.set_option(tcp::acceptor::reuse_address(true));
+    acceptor.bind(endpoint);
+
+    tcp::endpoint endPoint = acceptor.local_endpoint();
```
godexsoft: nit: The names look a bit strange. `endpoint` and then `endPoint`. Consider renaming to something like `localEndpoint` or perhaps we don't even need the temporary there.

tests/common/util/AssignRandomPort.hpp
```diff
@@ -0,0 +1,32 @@
...
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include <boost/asio/ts/buffer.hpp>
+#include <boost/asio/ts/internet.hpp>
+
+using tcp = boost::asio::ip::tcp;
```
godexsoft: This is always a no-no. Never add using in header files, only in cpp files 👍

tests/common/util/MockXrpLedgerAPIService.hpp
```diff
@@ -94,11 +94,17 @@ struct WithMockXrpLedgerAPIService : virtual ::testing::Test {
         serverThread_.join();
     }
 
+    int
+    getXRPLMockPort()
```
godexsoft: This probably should be `const`-qualified.

tests/common/util/MockXrpLedgerAPIService.hpp
```diff
@@ -94,11 +94,17 @@ struct WithMockXrpLedgerAPIService : virtual ::testing::Test {
...
+    {
+        return port_;
+    }
     MockXrpLedgerAPIService mockXrpLedgerAPIService;
 
 private:
     std::unique_ptr<grpc::Server> server_;
     std::thread serverThread_;
+    int port_;
```
godexsoft: Is this the correct type?

godexsoft (changes requested):
Leaving a few comments and suggestions 👍

tests/common/util/AssignRandomPort.cpp
```diff
@@ -0,0 +1,46 @@
...
+    boost::asio::io_context io_context;
+    tcp::acceptor acceptor(io_context);
+    tcp::endpoint const endpoint(tcp::v4(), 0);
+
+    acceptor.open(endpoint.protocol());
+    acceptor.set_option(tcp::acceptor::reuse_address(true));
+    acceptor.bind(endpoint);
+
+    tcp::endpoint localEndPoint = acceptor.local_endpoint();
+    return localEndPoint.port();
```
kuznetsss: Minor thing, you could do
```cpp
return acceptor.local_endpoint().port();
```

tests/unit/etl/SubscriptionSourceTests.cpp
```diff
@@ -46,8 +48,8 @@ struct SubscriptionSourceConnectionTests : public NoLoggerFixture {
     }
 
     boost::asio::io_context ioContext_;
-
-    TestWsServer wsServer_{ioContext_, "0.0.0.0", 11113};
+    uint32_t port = tests::util::generateFreePort();
+    TestWsServer wsServer_{ioContext_, "0.0.0.0", port};
```
kuznetsss: I think instead it is better to remove port parameter from `TestWsServer`'s constructor and use always `0`.  Then add `port()` method for `TestWsServer` and use it below.

tests/unit/util/requests/RequestBuilderTests.cpp
```diff
@@ -50,8 +52,9 @@ struct RequestBuilderTestBundle {
 };
 
 struct RequestBuilderTestBase : SyncAsioContextTest {
-    TestHttpServer server{ctx, "0.0.0.0", 11111};
-    RequestBuilder builder{"localhost", "11111"};
+    uint32_t port{tests::util::generateFreePort()};
+    TestHttpServer server{ctx, "0.0.0.0", port};
+    RequestBuilder builder{"localhost", std::to_string(port)};
```
kuznetsss: And the same as for `TestWsServer` here.

tests/unit/web/ServerTests.cpp
```diff
@@ -166,14 +182,21 @@ class WebServerTest : public NoLoggerFixture {
...
+    std::string
+    portToString() const
+    {
+        return std::to_string(port);
+    }
+
     // this ctx is for dos timer
     boost::asio::io_context ctxSync;
-    Config cfg{boost::json::parse(JSONData)};
+    uint32_t port = tests::util::generateFreePort();
```
kuznetsss: Probably better to save it as `std::string`

tests/common/util/TestHttpServer.cpp
```diff
@@ -134,3 +134,9 @@ TestHttpServer::handleRequest(TestHttpServer::RequestHandler handler, bool const
         boost::asio::detached
     );
 }
+
+int
+TestHttpServer::port() const
```
kuznetsss, line 139: You also should remove `port` parameter form constructor for `TestHttpServer` and `TestWsServer` and use `0` instead.

kuznetsss (dismissed):
👍

tests/common/util/TestHttpServer.hpp
```diff
@@ -43,7 +43,7 @@ class TestHttpServer {
      * @param host host to bind to
      * @param port port to bind to
```
godexsoft: This will later trigger doxygen to error out.. best removed now since the param is already gone 👍

godexsoft (changes requested):
Looks good, just fix the doxy comment and let's ship it 🚀

godexsoft (dismissed):

kuznetsss (approved):

Issue #1131, opened 2024-01-23
feat: add user version of `feature` RPC


intelliot:
- https://github.com/XRPLF/rippled/pull/4781
- Users should be able to use `feature` to determine whether an amendment is `supported` by Clio
- Docs for admin version of the method: https://xrpl.org/feature.html

godexsoft:
Removed 2.2 milestone for now as we can't implement it until support is added to `libxrpl` and that is not happening until after 2.2.0.
Replacing with #1440 until then.

PR #1440, opened 2024-05-31
Forward feature RPC

### LINKED ISSUE #1436: Forward `feature` RPC
## Summary
In #1131 we are going to implement a Clio-native `feature` API. However, this can't be done until https://github.com/XRPLF/rippled/pull/5026 is merged and it does not look it's going to make it into `libXRPL 2.2.0`.
For the time being, let's forward `feature` instead.

godexsoft:
Fixes #1436 

This is a temporary implementation of the `feature` RPC that will always return `noPermission` iff `vetoed` is set. 
If `vetoed` isn't specified, Clio will always forward the request to `rippled` instead.

In the future, #1131 will implement a Clio-native `feature` RPC. This requires specific support from `libxrpl` side and that is not going to be available till at least 2.2.1, hence the temporary forwarding.

It would be great to review the error message and code so that we pick the right one from the start.

### DIFF
--- a/src/rpc/handlers/Feature.cpp
+++ b/src/rpc/handlers/Feature.cpp
@@ -1,0 +1,82 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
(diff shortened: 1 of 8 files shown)


cindyyan317 (approved):
LGTM~ The warning message looks good.

src/rpc/handlers/Feature.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <cstdint>
+#include <string>
+
+namespace rpc {
+
+FeatureHandler::Result
+FeatureHandler::process([[maybe_unused]] FeatureHandler::Input input, [[maybe_unused]] Context const& ctx) const
+{
+    // For now this handler only fires when "vetoed" is set in the request.
+    // This always leads to a `notSupported` error as we don't want anyone to be able to
```
kuznetsss, line 42: Should we add an assert here?
```cpp
ASSERT(false, "Not implemented");
```

kuznetsss (approved):

src/rpc/handlers/Feature.cpp
kuznetsss: I will do it as a separate PR.

PR #1441, opened 2024-06-03
Add ip to log

cindyyan317:
For make log analysis easier, add ip to log.


cindyyan317:
We decide to remove all ip in log

PR #1448, opened 2024-06-10
Support string type integer for oracle_document_id

cindyyan317:
Seems like rippled supports string-type int for integer field of input json, Clio will starts supporting it.

This PR changed the field oracle_document_id to support both integer type and integer but string-type. For example 1 and "1".

### DIFF
--- a/src/rpc/common/MetaProcessors.hpp
+++ b/src/rpc/common/MetaProcessors.hpp
@@ -173,7 +173,7 @@ public:
      * @brief Constructs a validator that calls the given validator `req` and returns a custom error `err` in case `req`
      * fails.
      *
-     * @param req The requirement to validate against
+     * @param reqOrModifier The requirement to validate against
      * @param err The custom error to return in case `req` fails
      */
     WithCustomError(RequirementOrModifierType reqOrModifier, Status err)
@@ -216,6 +216,13 @@ public:
         return {};
     }
 
+    /**
+     * @brief Runs the stored modifier and produces a custom error if the wrapped modifier fails.
+     *
+     * @param value The JSON value representing the outer object. This value can be modified by the modifier.
+     * @param key The key used to retrieve the element from the outer object
+     * @return Possibly an error
+     */
     MaybeError
     modify(boost::json::value& value, std::string_view key) const
         requires SomeModifier<RequirementOrModifierType>
--- a/src/rpc/common/MetaProcessors.hpp
+++ b/src/rpc/common/MetaProcessors.hpp
@@ -38,6 +38,9 @@
 
 namespace rpc::meta {
 
+template <typename>
+static constexpr bool unsupported_v = false;
+
 /**
  * @brief A meta-processor that acts as a spec for a sub-object/section.
  */
@@ -162,9 +165,9 @@ private:
 /**
  * @brief A meta-processor that wraps a validator and produces a custom error in case the wrapped validator fails.
  */
-template <typename SomeRequirement>
+template <typename SomeRequirementOrModifier>
 class WithCustomError final {
-    SomeRequirement requirement;
+    SomeRequirementOrModifier requirement;
     Status error;
 
 public:
@@ -175,7 +178,7 @@ public:
      * @param req The requirement to validate against
      * @param err The custom error to return in case `req` fails
      */
-    WithCustomError(SomeRequirement req, Status err) : requirement{std::move(req)}, error{std::move(err)}
+    WithCustomError(SomeRequirementOrModifier req, Status err) : requirement{std::move(req)}, error{std::move(err)}
     {
     }
 
@@ -189,9 +192,12 @@ public:
     [[nodiscard]] MaybeError
     verify(boost::json::value const& value, std::string_view key) const
     {
-        if (auto const res = requirement.verify(value, key); not res)
-            return Error{error};
-
+        if constexpr (SomeRequirement<decltype(requirement)>) {
+            if (auto const res = requirement.verify(value, key); not res)
+                return Error{error};
+        } else {
+            static_assert(unsupported_v<decltype(requirement)>);
+        }
         return {};
     }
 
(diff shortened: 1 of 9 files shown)


src/rpc/handlers/GetAggregatePrice.hpp
```diff
@@ -131,23 +132,26 @@ class GetAggregatePriceHandler {
         static auto constexpr ORACLES_MAX = 200;
 
         static auto const oraclesValidator =
-            validation::CustomValidator{[](boost::json::value const& value, std::string_view) -> MaybeError {
+            modifiers::CustomModifier{[](boost::json::value& value, std::string_view) -> MaybeError {
```
kuznetsss, line 135: Why did you remove `const`?

src/rpc/common/Modifiers.hpp
```diff
@@ -100,4 +104,71 @@ struct ToLower final {
...
+    std::function<MaybeError(boost::json::value&, std::string_view)> modifier_;
+
+public:
+    /**
+     * @brief Constructs a custom modifier from any supported callable.
+     *
+     * @tparam Fn The type of callable
+     * @param fn The callable/function object
+     */
+    template <typename Fn>
```
kuznetsss, line 155: Maybe it is better to add a concept for `Fn` to make compile errors nicer.

src/rpc/common/Modifiers.hpp
```diff
@@ -100,4 +104,71 @@ struct ToLower final {
...
+     * @brief Modify the JSON value according to the custom modifier function stored.
+     *
+     * @param value The JSON value representing the outer object
+     * @param key The key used to retrieve the tested value from the outer object
+     * @return Any compatible user-provided error if modify/verify failed; otherwise no error is returned
+     */
+    [[nodiscard]] MaybeError
+    modify(boost::json::value& value, std::string_view key) const
+    {
+        if (not value.is_object() or not value.as_object().contains(key.data()))
```
kuznetsss: I don't think you need `.data()` here and in other places

src/rpc/common/MetaProcessors.hpp
```diff
@@ -189,26 +194,36 @@ class WithCustomError final {
     [[nodiscard]] MaybeError
     verify(boost::json::value const& value, std::string_view key) const
     {
-        if (auto const res = requirement.verify(value, key); not res)
-            return Error{error};
-
+        if constexpr (SomeRequirement<decltype(reqOrModifier)>) {
```
kuznetsss: I think you could do
```cpp
requires SomeRequirement<SomeRequirementOrModifier>
```
then you wouldn't need `unsupported_v`.

src/rpc/common/MetaProcessors.hpp
```diff
@@ -160,22 +163,24 @@ class IfType final {
 };
 
 /**
- * @brief A meta-processor that wraps a validator and produces a custom error in case the wrapped validator fails.
+ * @brief A meta-processor that wraps a validator or modifier and produces a custom error in case the wrapped validator
+ * or modifier fails.
  */
-template <typename SomeRequirement>
+template <typename SomeRequirementOrModifier>
```
kuznetsss: Maybe better `ReuirementOrModifierType`? Because `Some...` we usually use for a concept names.

And probably better to add a concept here that it is either requirement or modifier.

src/rpc/common/MetaProcessors.hpp
```diff
@@ -189,26 +194,36 @@ class WithCustomError final {
...
     verify(boost::json::value& value, std::string_view key) const
     {
-        if (auto const res = requirement.verify(value, key); not res)
-            return Error{error};
-
+        if constexpr (SomeRequirement<decltype(reqOrModifier)>) {
+            if (auto const res = reqOrModifier.verify(value, key); not res)
+                return Error{error};
+        } else if constexpr (SomeModifier<decltype(reqOrModifier)>) {
+            if (auto const res = reqOrModifier.modify(value, key); not res)
```
kuznetsss: It is a bit weird that method `verify()` performs a modification. Is it not possible to separate verification and modification?

tests/unit/rpc/BaseTests.cpp
```diff
@@ -604,3 +604,55 @@ TEST_F(RPCBaseTest, ToLowerModifier)
...
 }
+
+TEST_F(RPCBaseTest, ToNumberModifier)
+{
+    auto spec = RpcSpec{
+        {"str", ToNumber{}},
+    };
+
+    auto passingInput = json::parse(R"({ "str": [] })");
+    ASSERT_TRUE(spec.process(passingInput));
```
kuznetsss, line 616: I think, it is better to check that `passingInput` didn't change after this.

Also probably better to split this test in parametrised tests, because currently `passingInput` is shared and multiple tests cases could affect each other.

tests/unit/rpc/BaseTests.cpp
```diff
@@ -604,3 +604,55 @@ TEST_F(RPCBaseTest, ToLowerModifier)
...
+
+    passingInput = json::parse(R"({ "str2": "TesT" })");
+    ASSERT_TRUE(spec.process(passingInput));
+
+    passingInput = json::parse(R"([])");
+    ASSERT_TRUE(spec.process(passingInput));
+
+    passingInput = json::parse(R"({ "str": "123" })");
```
kuznetsss, line 624: What if it would be `{ "str": "123.456" }`?

src/rpc/common/MetaProcessors.hpp
cindyyan317: Yes. We can discuss how to improve this. I believe it will need framework change.

tests/unit/rpc/BaseTests.cpp
```diff
@@ -604,3 +604,58 @@ TEST_F(RPCBaseTest, ToLowerModifier)
...
+
+TEST_F(RPCBaseTest, CustomModifier)
+{
+    auto const customModifier = CustomModifier{[](json::value& value, std::string_view /* key */) -> MaybeError {
+        if (value.is_string()) {
+            value = json::value("modified");
+            return MaybeError{};
+        }
+        return Error{rpc::Status{"Uh oh"}};
+    }};
```
kuznetsss: I think a better way to test it is to use a mock function:
```cpp
testing::StrictMock<testing::MockFunction<MaybeError(json::value& value, std::string_view)>> mockModifier;
auto const customModifier = CustomModifier{mockModifier.AsStdFunction()};
```

and then you will be able to check whether the mock was called when you call `spec.process()`.

kuznetsss (approved):
👍

PR #1462, opened 2024-06-14
Add forwarding timeout option

### LINKED ISSUE #1454: Add timeout option for forwarded request
Currently forwarded messages don’t have a timeout, so if rippled is slow Clio may have many hanging ws connections to rippled. We should add a forwarding timeout option to Clio to fix the issue.

kuznetsss:
Fixes #1454.

### DIFF
--- a/src/util/requests/impl/WsConnectionImpl.hpp
+++ b/src/util/requests/impl/WsConnectionImpl.hpp
@@ -22,8 +22,13 @@
 #include "util/requests/Types.hpp"
 #include "util/requests/WsConnection.hpp"
 
+#include <boost/asio/associated_executor.hpp>
+#include <boost/asio/bind_cancellation_slot.hpp>
 #include <boost/asio/buffer.hpp>
+#include <boost/asio/cancellation_signal.hpp>
+#include <boost/asio/cancellation_type.hpp>
 #include <boost/asio/spawn.hpp>
+#include <boost/asio/steady_timer.hpp>
 #include <boost/beast/core/buffers_to_string.hpp>
 #include <boost/beast/core/error.hpp>
 #include <boost/beast/core/flat_buffer.hpp>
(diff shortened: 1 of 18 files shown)


src/util/requests/impl/WsConnectionImpl.hpp
```diff
@@ -92,6 +117,31 @@ class WsConnectionImpl : public WsConnection {
...
+        boost::asio::steady_timer timer{boost::asio::get_associated_executor(cyield), timeout};
+        timer.async_wait([&cancellationSignal](boost::system::error_code errorCode) {
+            if (!errorCode)
+                cancellationSignal.emit(boost::asio::cancellation_type::terminal);
+        });
+        operation(cyield);
+    }
+
+    static boost::system::error_code
+    mapError(boost::system::error_code const ec)
```
cindyyan317, line 138: Do we add 'const' for value parameter?

kuznetsss: I think it is useful, showing that `ec` will not be changed.

src/util/requests/impl/WsConnectionImpl.hpp
```diff
@@ -51,27 +57,46 @@ class WsConnectionImpl : public WsConnection {
...
-        if (errorCode)
+        auto operation = [&](auto&& token) { ws_.async_write(boost::asio::buffer(message), token); };
+        if (timeout) {
+            withTimeout(operation, yield[errorCode], *timeout);
+        } else {
+            operation(yield[errorCode]);
+        }
+
+        if (errorCode) {
+            errorCode = mapError(errorCode);
```
cindyyan317, line 97: Does the map error only need to happen when withTimeout? My concern is if it shadows any other error?

cindyyan317 (approved):
LGTM ! 💯

src/util/requests/impl/WsConnectionImpl.hpp
kuznetsss: No, it only to convert `operation_canceled` into `timed_out`.  You can see it at the line 138 in this file.

PR #1465, opened 2024-06-17
Improve etl check

cindyyan317:
When there was an ETL not working and `allow_no_etl` is false, Clio would fail to start.
This PR improved the behavior. As long as there is one ETL working, Clio should be able to start.

### DIFF
--- a/tests/unit/etl/LoadBalancerTests.cpp
+++ b/tests/unit/etl/LoadBalancerTests.cpp
@@ -109,7 +109,7 @@ TEST_F(LoadBalancerConstructorTests, forwardingTimeoutPassedToSourceFactory)
 
 TEST_F(LoadBalancerConstructorTests, fetchETLState_SourceAllFail)
 {
-    EXPECT_CALL(sourceFactory_, makeSource).Times(1);
+    EXPECT_CALL(sourceFactory_, makeSource).Times(2);
     EXPECT_CALL(sourceFactory_.sourceAt(0), forwardToRippled).WillOnce(Return(std::nullopt));
     EXPECT_CALL(sourceFactory_.sourceAt(1), forwardToRippled).WillOnce(Return(std::nullopt));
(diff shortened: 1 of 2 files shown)


tests/unit/etl/LoadBalancerTests.cpp
```diff
@@ -107,30 +107,32 @@ TEST_F(LoadBalancerConstructorTests, forwardingTimeoutPassedToSourceFactory)
     makeLoadBalancer();
 }
 
-TEST_F(LoadBalancerConstructorTests, fetchETLState_Source0Fails)
+TEST_F(LoadBalancerConstructorTests, fetchETLState_SourceAllFail)
```
kuznetsss: Probably better to name tests a bit more natural `fetchETLState_AllSourcesFail`.

tests/unit/etl/LoadBalancerTests.cpp
```diff
@@ -107,30 +107,32 @@ TEST_F(LoadBalancerConstructorTests, forwardingTimeoutPassedToSourceFactory)
...
     EXPECT_CALL(sourceFactory_.sourceAt(0), forwardToRippled)
         .WillOnce(Return(boost::json::object{{"error", "some error"}}));
-    EXPECT_CALL(sourceFactory_.sourceAt(0), toString);
+    EXPECT_CALL(sourceFactory_.sourceAt(1), forwardToRippled)
+        .WillOnce(Return(boost::json::object{{"error", "some error"}}));
     EXPECT_THROW({ makeLoadBalancer(); }, std::logic_error);
 }
 
-TEST_F(LoadBalancerConstructorTests, fetchETLState_Source1Fails)
+TEST_F(LoadBalancerConstructorTests, fetchETLState_Source1Fails0OK)
```
kuznetsss, line 128: I think there should be also a test when source 0 fails and source 1 is fine.

kuznetsss (approved):
👍

PR #1468, opened 2024-06-18
Make sure release commit signed and tagged

legleux:
Best practices, good housekeeping and help me look stuff up.

Looks like the annotated tag train has derailed. This is an attempt to enforce it.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,3 +1,58 @@
 #!/bin/sh
+
+
+# git for-each-ref refs/tags  # see which tags are annotated and which are lightweight. Annotated tags are "tag" objects.
+# # Set these so your commits and tags are always signed
+# git config commit.gpgsign true
+# git config tag.gpgsign true
+
+verify_commit_signed() {
+    if git verify-commit HEAD &> /dev/null; then
+        :
+        # echo "HEAD commit seems signed..."
+    else
+        echo "HEAD commit isn't signed!"
+        exit 1
+    fi
+}
+
+verify_tag() {
+    if git describe --exact-match --tags HEAD &> /dev/null; then
+        : # You might be ok to push
+        # echo "Tag is annotated."
+        return 0
+    else
+        echo "Tag for [$version] not an annotated tag."
+        exit 1
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


.githooks/pre-push
```diff
@@ -1,3 +1,58 @@
...
+
+# git for-each-ref refs/tags  # see which tags are annotated and which are lightweight. Annotated tags are "tag" objects.
+# # Set these so your commits and tags are always signed
+# git config commit.gpgsign true
+# git config tag.gpgsign true
+
+verify_commit_signed() {
+    if git verify-commit HEAD &> /dev/null; then
+        :
+        # echo "HEAD commit seems signed..."
```
cindyyan317, line 12: do you want to remove the comments

kuznetsss: It looks like a useful comment.

.githooks/pre-push
```diff
@@ -1,3 +1,58 @@
...
+        # echo "Tag appears signed"
+        return 0
+    else
+        echo "$version tag isn't signed"
+        echo "Sign it with [git tag -ams\"$version\" $version]"
+        exit 1
+    fi
+}
+
+while read local_ref local_oid remote_ref remote_oid; do
```
kuznetsss, line 42: Is it predefined variables in git hooks?

.githooks/pre-push
legleux: my comments disappeared?
I had said I was on the fence and usually don't leave comments like this but I kept uncommenting while debugging so I decided to leave them in. :shrug:

.githooks/pre-push
legleux: Sort of? (from the sample hook)
```
# This hook is called with the following parameters:
#
# $1 -- Name of the remote to which the push is being done
# $2 -- URL to which the push is being done
#
# If pushing without using a named remote those arguments will be equal.
#
# Information about the commits which are being pushed is supplied as lines to
# the standard input in the form:
#
#   <local ref> <local sha1> <remote ref> <remote sha1>
```

kuznetsss (approved):

PR #1469, opened 2024-06-18
Ignore checking formatting if no formatted files are staged

legleux:
I'm constantly harassed to follow the rules when I'm not doing anything relevant.

This is an attempt to ameliorate that rather than `--no-verify` or just disabling hooks altogether. :mushroom: :cloud: 

Somebody might want to confirm `basename` is included with macOS.

### DIFF
--- a/.githooks/check-format
+++ b/.githooks/check-format
@@ -4,12 +4,6 @@
 #
 # This script checks the format of the code and cmake files.
 # In many cases it will automatically fix the issues and abort the commit.
-#!/bin/bash
-
-# Note: This script is intended to be run from the root of the repository.
-#
-# This script checks the format of the code and cmake files.
-# In many cases it will automatically fix the issues and abort the commit.
 
 no_formatted_directories_staged() {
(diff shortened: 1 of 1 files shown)


.githooks/check-format
```diff
@@ -5,6 +5,32 @@
 # This script checks the format of the code and cmake files.
 # In many cases it will automatically fix the issues and abort the commit.
 
+no_formatted_files_staged() {
+    staged_files=$(git diff-index --cached --name-only HEAD)
+    for file in $staged_files; do
+        extension=${file##*.}
+        if [[ "$extension" =~ (h|hpp|c|cpp)$ ]]; then
```
kuznetsss: Maybe instead check directories of changed files? I think we need to check format only for `benchmarks`, `src`, `tests` and `cmake`.

legleux: Cleans it up and _should_ suit the situation.

kuznetsss (approved):
👍

PR #1479, opened 2024-06-20
Fix issue: "Updating cache" prints in log when cache is disabled

### LINKED ISSUE #1461: Log shows "Updating cache" while cache is disabled
## Issue Description
We still see "Updating cache" in the log when cache is disabled by configuration.
It is confusing. Maybe we also can reduce a db call here :  https://github.com/XRPLF/clio/blob/437ea7bf98baa2f8242e681606fca85eaae383b8/src/etl/impl/LedgerPublisher.hpp#L171

cindyyan317:
Fix #1461

### DIFF
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -93,7 +93,6 @@ BackendInterface::fetchLedgerObject(
         return obj;
     }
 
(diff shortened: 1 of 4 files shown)


kuznetsss (approved):
LGTM

PR #1483, opened 2024-06-21
fix: Support deleted object in ledger_entry

### LINKED ISSUE #1306: Support deleted object in `ledger_entry`
## Summary

Modifying the `ledger_entry` API as follows:

### New Request Parameter
| Field Name           | Type    | Description |
|--------------------- |:-------:| ------------|
| `include_deleted`       | ️boolean   |  _(Optional)_ If set to true and the queried object has been deleted, return its complete data prior to its deletion. If set to false or not provided and the queried object has been deleted, return `objectNotFound` (current behavior) .|

### New Response Parameter
| Field Name           | Type    | Description |
|--------------------- |:-------:| ------------|
| `deleted_ledger_index`       |number   |  _(Optional)_ Indicates the ledger index at which the object was deleted. 

## Motivation

Currently,  `ledger_entry`  API returns `objectNotFound` if the object has been deleted. But perhaps someone may want to find the object data before its deletion, without the need to dig through transaction history to find the index at which this particular object was deleted.

I think clio can introduce an improved `ledger_entry`  API with the a change - rather than returning `objectNotFound` for a deleted object, it would return the complete object data prior to its deletion, as well as an additional field to indicated the ledger index at which the object wad deleted. 

This is useful it  for something like XRPL explorer. For example, if I create a MPT page that queries `ledger_entry`, it would work fine as long as the MPT is not destroyed. However, after the MPT is destroyed, the MPT page will no longer be available because `ledger_entry` will now return `objectNotFound`. This would result in suboptimal experience in the explorer. This would improve the user experience on explorers in the long term, as explorer adds new page for each object type.

Kassaking7:
Fixes #1306

### DIFF
--- a/src/data/BackendInterface.cpp
+++ b/src/data/BackendInterface.cpp
@@ -102,6 +102,19 @@ BackendInterface::fetchLedgerObject(
     return dbObj;
 }
 
+std::optional<std::uint32_t>
+BackendInterface::fetchLedgerObjectSeq(
+    ripple::uint256 const& key,
+    std::uint32_t const sequence,
+    boost::asio::yield_context yield
+) const
+{
+    auto seq = doFetchLedgerObjectSeq(key, sequence, yield);
+    if (!seq) 
+        LOG(gLog.trace()) << "Missed in db";
+    return seq;
+}
+
 std::vector<Blob>
 BackendInterface::fetchLedgerObjects(
     std::vector<ripple::uint256> const& keys,
--- a/src/data/BackendInterface.hpp
+++ b/src/data/BackendInterface.hpp
@@ -379,6 +379,19 @@ public:
     fetchLedgerObject(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const;
 
     /**
+     * @brief Fetches a specific ledger object sequence.
+     *
+     * Currently the real fetch happens in doFetchLedgerObjectSeq
+     *
+     * @param key The key of the object
+     * @param sequence The ledger sequence to fetch for
+     * @param yield The coroutine context
+     * @return The sequence in unit32_t on success; nullopt otherwise
+     */
+    std::optional<std::uint32_t>
+    fetchLedgerObjectSeq(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const;
+
+    /**
      * @brief Fetches all ledger objects by their keys.
      *
      * Currently the real fetch happens in doFetchLedgerObjects and fetchLedgerObjects attempts to fetch from Cache
@@ -406,6 +419,17 @@ public:
      */
     virtual std::optional<Blob>
     doFetchLedgerObject(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const = 0;
+
+    /**
+     * @brief The database-specific implementation for fetching a ledger object sequence.
+     *
+     * @param key The key to fetch for
+     * @param sequence The ledger sequence to fetch for
+     * @param yield The coroutine context
+     * @return The sequence in unit32_t on success; nullopt otherwise
+     */
+    virtual std::optional<std::uint32_t>
+    doFetchLedgerObjectSeq(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const = 0;
 
     /**
      * @brief The database-specific implementation for fetching ledger objects.
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -567,6 +567,25 @@ public:
         return std::nullopt;
     }
 
+    std::optional<std::uint32_t>
+    doFetchLedgerObjectSeq(ripple::uint256 const& key, std::uint32_t const sequence, boost::asio::yield_context yield)
+        const override
+    {
+        LOG(log_.debug()) << "Fetching ledger object for seq " << sequence << ", key = " << ripple::to_string(key);
+        if (auto const res = executor_.read(yield, schema_->selectObject, key, sequence); res) {
+            if (auto const result = res->template get<Blob, std::uint32_t>(); result) { 
+                auto [_ ,seq] = result.value();
+                return seq;
+            } else {
+                LOG(log_.debug()) << "Could not fetch ledger object sequence - no rows";
+            }
+        } else {
+            LOG(log_.error()) << "Could not fetch ledger object sequence: " << res.error();
+        }
+
+        return std::nullopt;
+    }
+
     std::optional<TransactionAndMetadata>
     fetchTransaction(ripple::uint256 const& hash, boost::asio::yield_context yield) const override
     {
--- a/src/rpc/handlers/LedgerEntry.cpp
+++ b/src/rpc/handlers/LedgerEntry.cpp
@@ -162,17 +162,26 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
         return Error{*status};
 
     auto const lgrInfo = std::get<ripple::LedgerHeader>(lgrInfoOrStatus);
-    auto const ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
+    auto output = LedgerEntryHandler::Output{};
+    auto ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
 
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (not input.includeDeleted) 
+            return Error{Status{"entryNotFound"}};
+        auto const deletedSeq = sharedPtrBackend_->fetchLedgerObjectSeq(key, lgrInfo.seq, ctx.yield);
+        if (!deletedSeq) 
+            return Error{Status{"entryNotFound"}};
+        ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, deletedSeq.value() - 1, ctx.yield);
+        if (!ledgerObject || ledgerObject->empty()) 
+            return Error{Status{"entryNotFound"}};
+        output.deletedLedgerIndex = deletedSeq.value();
+    }
 
     ripple::STLedgerEntry const sle{ripple::SerialIter{ledgerObject->data(), ledgerObject->size()}, key};
 
     if (input.expectedType != ripple::ltANY && sle.getType() != input.expectedType)
         return Error{Status{"unexpectedLedgerType"}};
 
-    auto output = LedgerEntryHandler::Output{};
     output.index = ripple::strHex(key);
     output.ledgerIndex = lgrInfo.seq;
     output.ledgerHash = ripple::strHex(lgrInfo.hash);
@@ -219,6 +228,9 @@ tag_invoke(boost::json::value_from_tag, boost::json::value& jv, LedgerEntryHandl
         {JS(validated), output.validated},
         {JS(index), output.index},
     };
+
+    if (output.deletedLedgerIndex)
+        object["deleted_ledger_index"] = *(output.deletedLedgerIndex);
 
     if (output.nodeBinary) {
         object[JS(node_binary)] = *(output.nodeBinary);
@@ -336,6 +348,9 @@ tag_invoke(boost::json::value_to_tag<LedgerEntryHandler::Input>, boost::json::va
     } else if (jsonObject.contains(JS(oracle))) {
         input.oracleNode = parseOracleFromJson(jv.at(JS(oracle)));
     }
+
+    if (jsonObject.contains("include_deleted"))
+        input.includeDeleted = jv.at("include_deleted").as_bool();
 
     return input;
 }
--- a/src/rpc/handlers/LedgerEntry.hpp
+++ b/src/rpc/handlers/LedgerEntry.hpp
@@ -71,6 +71,7 @@ public:
         std::string ledgerHash;
         std::optional<boost::json::object> node;
         std::optional<std::string> nodeBinary;
+        std::optional<uint32_t> deletedLedgerIndex;
         bool validated = true;
     };
 
@@ -103,6 +104,7 @@ public:
         std::optional<uint32_t> chainClaimId;
         std::optional<uint32_t> createAccountClaimId;
         std::optional<ripple::uint256> oracleNode;
+        bool includeDeleted = false;
     };
 
     using Result = HandlerReturnType<Output>;
@@ -314,6 +316,7 @@ public:
                   meta::WithCustomError{modifiers::ToNumber{}, Status(ClioError::rpcMALFORMED_ORACLE_DOCUMENT_ID)}},
              }}},
             {JS(ledger), check::Deprecated{}},
+            {"include_deleted", validation::Type<bool>{}},
         };
 
         return rpcSpec;
--- a/tests/unit/rpc/handlers/LedgerEntryTests.cpp
+++ b/tests/unit/rpc/handlers/LedgerEntryTests.cpp
@@ -61,6 +61,8 @@ constexpr static auto RANGEMIN = 10;
 constexpr static auto RANGEMAX = 30;
 constexpr static auto LEDGERHASH = "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652";
 constexpr static auto TOKENID = "000827103B94ECBB7BF0A0A6ED62B3607801A27B65F4679F4AD1D4850000C0EA";
+constexpr static auto NFTID = "00010000A7CAD27B688D14BA1A9FA5366554D6ADCF9CE0875B974D9F00000004";
+constexpr static auto TXNID = "05FB0EB4B899F056FA095537C5817163801F544BAFCEA39C995D76DB4D16F9DD";
 
 class RPCLedgerEntryTest : public HandlerBaseTest {};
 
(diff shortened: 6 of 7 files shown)


src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -101,6 +102,7 @@ class LedgerEntryHandler {
         std::optional<uint32_t> chainClaimId;
         std::optional<uint32_t> createAccountClaimId;
         std::optional<ripple::uint256> oracleNode;
+        std::optional<bool> include_deleted;
```
cindyyan317: 1 Maybe we can use bool directly? False by default? 
 2 Name convention: includeDeleted

src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -70,6 +70,7 @@ class LedgerEntryHandler {
         std::optional<boost::json::object> node;
         std::optional<std::string> nodeBinary;
         bool validated = true;
+        std::optional<uint32_t> deleted_ledger_index;
```
cindyyan317: Our established name convention  : deletedLedgerIndex

src/data/BackendInterface.hpp
```diff
@@ -378,6 +378,29 @@ class BackendInterface {
     std::optional<Blob>
     fetchLedgerObject(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const;
 
+    /**
+     * @brief Fetches Last two ledger objects.
```
cindyyan317: Last -> last

src/data/BackendInterface.hpp
```diff
@@ -378,6 +378,29 @@ class BackendInterface {
...
+     *
+     *
+     * @param key The key of the object
+     * @param sequence The ledger sequence to fetch for
+     * @param yield The coroutine context
+     * @return A vector containing two Blobs, representing last two ledger objects
```
cindyyan317: Maybe we can put more detail here.

cindyyan317 (changes requested):
Since the **include_deleted** is optional, I would suggest to try to utilize our cache as much as possible. Maybe we don't need to land on DB every time.

1 Keep using 'fetchLedgerObject' to get the  last object by seq (This method will use cache effectively)
2 If **include_deleted** is true and the object from 1 is empty , use **fetchLedgerObject** + the seq_from_object_1 -1 getting the second one.
I think it might be easier and no need extra CQL statement.  I am happy to discuss it offline if needed.
 
BTW, please assign the issue to yourself when you start doing that. In case other people may pick it as no one works on it.

src/data/BackendInterface.cpp
```diff
@@ -103,6 +103,7 @@ BackendInterface::fetchLedgerObject(
     return dbObj;
 }
 
+
```
godexsoft, line 117: Extra space line is not needed here

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,21 +157,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
     auto const lgrInfo = std::get<ripple::LedgerHeader>(lgrInfoOrStatus);
-    auto const ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
-
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
-
+    std::optional<data::Blob> ledgerObject;
+    auto output = LedgerEntryHandler::Output{};
+    output.deletedLedgerIndex = std::nullopt;
+    ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
```
godexsoft: `auto ledgerObject` here is fine, no need to create an empty optional first just to overwrite it two lines down 👍 
And then let's add a line of space between this and the if.
We also usually don't add curly braces for one line `if`s.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,21 +157,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (input.includeDeleted == false) {
```
godexsoft: Another nit: `if (not input.includeDeleted)` reads better and is shorter. 
When it's possible to make code read like plain English it is usually better to go for it

src/data/BackendInterface.hpp
```diff
@@ -377,7 +377,7 @@ class BackendInterface {
      */
     std::optional<Blob>
     fetchLedgerObject(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const;
-
+
```
godexsoft: I'm surprised clang-format is not screaming at this.. let's remove unnecessary whitespace leaving only `\n` as before

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,21 +157,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
     ripple::STLedgerEntry const sle{ripple::SerialIter{ledgerObject->data(), ledgerObject->size()}, key};
 
     if (input.expectedType != ripple::ltANY && sle.getType() != input.expectedType)
         return Error{Status{"unexpectedLedgerType"}};
 
-    auto output = LedgerEntryHandler::Output{};
     output.index = ripple::strHex(key);
     output.ledgerIndex = lgrInfo.seq;
     output.ledgerHash = ripple::strHex(lgrInfo.hash);
-
```
godexsoft: Please don't remove spaces that make sense - they separate blocks of variables from `if`s, unrelated `if`s from each other, etc. - let's add them back and also add some more around your 'if' statements 👍

src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -101,6 +102,7 @@ class LedgerEntryHandler {
         std::optional<uint32_t> chainClaimId;
         std::optional<uint32_t> createAccountClaimId;
         std::optional<ripple::uint256> oracleNode;
+        bool includeDeleted;
```
godexsoft: Let's add `= false;`

src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -70,6 +70,7 @@ class LedgerEntryHandler {
         std::optional<boost::json::object> node;
         std::optional<std::string> nodeBinary;
         bool validated = true;
+        std::optional<uint32_t> deletedLedgerIndex;
```
godexsoft, line 74: let's keep `validated` at the bottom of the struct as it was before 👍

tests/unit/rpc/handlers/LedgerEntryTests.cpp
```diff
@@ -2612,3 +2614,239 @@ TEST(RPCLedgerEntrySpecTest, DeprecatedFields)
     EXPECT_EQ(warning.at("id").as_int64(), static_cast<int64_t>(rpc::WarningCode::warnRPC_DEPRECATED));
     EXPECT_NE(warning.at("message").as_string().find("Field 'ledger' is deprecated."), std::string::npos) << warning;
 }
+
+// Same as BinaryFalse with include_deleted set to true
+// Expected Result: same as BinaryFalse
+TEST_F(RPCLedgerEntryTest, BinaryFalseIncludeDeleted)
+{
+    static auto constexpr OUT = R"({
+        "ledger_hash":"4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A652",
```
godexsoft: I'd appreciate if you format the json a bit more by adding spaces between `:` and values. I know not all jsons in our tests are already formatted correctly - "thanks" to everyone using copilot 😢 
But let's try not to add more to the mess 😃

tests/unit/rpc/handlers/LedgerEntryTests.cpp
```diff
@@ -2612,3 +2614,239 @@ TEST(RPCLedgerEntrySpecTest, DeprecatedFields)
...
+                "index": "{}",
+                "include_deleted": true
+            }})",
+            INDEX1
+        ));
+        auto const output = handler.process(req, Context{yield});
+        ASSERT_TRUE(output);
+        EXPECT_EQ(*output.result, json::parse(OUT));
+    });
+}
```
godexsoft: Ah, classic. Please configure your IDE to automatically insert last empty line on save 👍

godexsoft (changes requested):
Leaving a few nits 👍

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
         return Error{*status};
 
     auto const lgrInfo = std::get<ripple::LedgerHeader>(lgrInfoOrStatus);
-    auto const ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
-
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    auto output = LedgerEntryHandler::Output{};
+    output.deletedLedgerIndex = std::nullopt;
```
godexsoft: This line is redundant - the line above already creates `Output` with a `nullopt` for this field via `optional`'s default constructor.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -215,6 +222,9 @@ tag_invoke(boost::json::value_from_tag, boost::json::value& jv, LedgerEntryHandl
         {JS(index), output.index},
     };
 
+    if (output.deletedLedgerIndex) {
+        object["deleted_ledger_index"] = *(output.deletedLedgerIndex);
+    }
```
godexsoft: I'd add a line of space below this block and perhaps remove the curlies for this one line `if` 👍

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -332,6 +342,8 @@ tag_invoke(boost::json::value_to_tag<LedgerEntryHandler::Input>, boost::json::va
         input.oracleNode = parseOracleFromJson(jv.at(JS(oracle)));
     }
 
+    if (jsonObject.contains("include_deleted"))
+        input.includeDeleted = jv.at("include_deleted").as_bool();
```
godexsoft, line 353: I'd separate the final return from the `if` by a line of space.

src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -302,6 +304,7 @@ class LedgerEntryHandler {
                   meta::WithCustomError{modifiers::ToNumber{}, Status(ClioError::rpcMALFORMED_ORACLE_DOCUMENT_ID)}},
              }}},
             {JS(ledger), check::Deprecated{}},
+            {"include_deleted", validation::Type<bool>{}},
```
godexsoft, line 319: I assume you checked and there is no `JS(include_deleted)` but thought i'd ask ❓

godexsoft (changes requested):
Found a few more small improvements. This PR looks very close though - great job 🙇

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
         return Error{*status};
 
     auto const lgrInfo = std::get<ripple::LedgerHeader>(lgrInfoOrStatus);
-    auto const ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
-
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    auto output = LedgerEntryHandler::Output{};
+    output.deletedLedgerIndex = std::nullopt;
+    auto ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
```
cindyyan317, line 166: const?

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
+
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (not input.includeDeleted) 
+            return Error{Status{"entryNotFound"}};
+        ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq - 1, ctx.yield);
+        if (!ledgerObject || ledgerObject->empty())
```
cindyyan317, line 168: If an object is deleted at seq, it must exist for seq-1, right? We can use an ASSERT here.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    auto output = LedgerEntryHandler::Output{};
+    output.deletedLedgerIndex = std::nullopt;
+    auto ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
+
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (not input.includeDeleted) 
+            return Error{Status{"entryNotFound"}};
+        ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq - 1, ctx.yield);
```
cindyyan317: lgrInfo.seq is the seq specified by the input, right? It is not the seq for the ledgerObject.

for example if object A got deleted at seq 100 but created at seq 80
in db, objects table is like:
key 100 null
key 80 0x1234
If the seq from input request is 120: 
First, we got the null as ledgerObject
Second, we shall use fetchLedgerObject(key,**99**, ctx.yield); to get 0x1234. If we use lgrInfo.seq - 1 which is 119, we will still get null, right?

cindyyan317 (changes requested):
I think the seq you are using to query the second latest object is not right. You need the seq which the object is modified at.
You can find more info about 'objects' table from [here](https://github.com/XRPLF/clio/blob/b2eacf9868fa1005cec0dc010416cf611de36835/src/data/README.md#objects)

src/rpc/handlers/LedgerEntry.cpp
godexsoft: `ledgerObject` is modified again below. But maybe there is no need to do that and it would be ok to just create a new variable in the loop, maybe hide it like `if (auto const ledgerObject = ...; !ledgerObject.isEmpty() or ...)`.

Kassaking7:
I give a change on the logic:
1. Try to fetch the object with a given sequence
2. If failed, it means the sequence it deleted is at least <= current given sequence
3. Use a new created function to fetch the sequence that the object deleted (no need for new CQL statement)
4. Use the new sequence-1 to fetch the latest object that is not deleted

Hope this time it works good :) I created a unit test for this case and looks good to me.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
-
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    auto output = LedgerEntryHandler::Output{};
+    auto ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
+
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (not input.includeDeleted) 
+            return Error{Status{"entryNotFound"}};
+        auto deleted_seq = sharedPtrBackend_->fetchLedgerObjectSeq(key, lgrInfo.seq, ctx.yield);
```
cindyyan317: const ,name should be deletedSeq

cindyyan317 (commented):
Nice work! Thanks 🙇 
Just some nits to fix, then we are ready to go

src/rpc/handlers/LedgerEntry.hpp
Kassaking7: Yes there's no JS(include_deleted)

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,24 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
-
-    if (!ledgerObject || ledgerObject->empty())
-        return Error{Status{"entryNotFound"}};
+    auto output = LedgerEntryHandler::Output{};
+    auto ledgerObject = sharedPtrBackend_->fetchLedgerObject(key, lgrInfo.seq, ctx.yield);
+
+    if (!ledgerObject || ledgerObject->empty()) {
+        if (not input.includeDeleted) 
+            return Error{Status{"entryNotFound"}};
+        auto const deletedSeq = sharedPtrBackend_->fetchLedgerObjectSeq(key, lgrInfo.seq, ctx.yield);
```
cindyyan317, line 171: Do we need to check deletedSeq, it can be nullopt?

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -157,17 +157,26 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
+        if (!deletedSeq) 
+            return Error{Status{"entrySeqNotFound"}};
```
cindyyan317: Should also be entryNotFound, right?

cindyyan317 (dismissed):
LGTM

src/data/BackendInterface.hpp
```diff
@@ -407,6 +420,17 @@ class BackendInterface {
     virtual std::optional<Blob>
     doFetchLedgerObject(ripple::uint256 const& key, std::uint32_t sequence, boost::asio::yield_context yield) const = 0;
 
+    /**
+     * @brief The database-specific implementation for fetching a ledger object sequence.
+     *
+     * @param key The key to fetch for
+     * @param sequence The ledger sequence to fetch for
+     * @param yield The coroutine context
+     * @return The object as a sequence on success; nullopt otherwise
```
godexsoft: Confusing a bit: i think this returns the sequence if available, not an object as a sequence

src/data/CassandraBackend.hpp
```diff
@@ -567,6 +567,24 @@ class BasicCassandraBackend : public BackendInterface {
         return std::nullopt;
     }
 
+    std::optional<std::uint32_t>
+    doFetchLedgerObjectSeq(ripple::uint256 const& key, std::uint32_t const sequence, boost::asio::yield_context yield)
+        const override
+    {
+        LOG(log_.debug()) << "Fetching ledger object for seq " << sequence << ", key = " << ripple::to_string(key);
+        if (auto const res = executor_.read(yield, schema_->selectObject, key, sequence); res) {
+            if (auto const result = res->template get<std::uint32_t>(); result) {
```
godexsoft: I'm not sure how this works, you are trying to read a Blob as a uint32_t. Is that allowed and what was meant?

cindyyan317: It is true. it gets by order.

godexsoft: You need to either `auto [seq, _] = get<Blob, uint32_t>` and discard the blob or add a `get<T>(std::size_t position)` and use that instead (in `data/cassandra/impl/Result.hpp`).

Kassaking7:
Gave a test on it with real DB and it works good on my side:
1. Do ledger_accept 300 times
2. Create 2 users
3. Mint a NFT
4. Create a offer with that NFT id
5. Record that offer id and cancel it (now in DB the corresponding object is 0x, sequence is 307)
6. Create another 2 users so current sequence is far beyond the deleted sequence
7. Call ledger_entry with include_deleted true and return deleted_ledger_index 307

tests/unit/rpc/handlers/LedgerEntryTests.cpp
```diff
@@ -2620,3 +2622,321 @@ TEST(RPCLedgerEntrySpecTest, DeprecatedFields)
...
+
+// Test for object seq not exist in database
+// Expected Result: return entrySeqNotFound error
```
cindyyan317: Nice to add it. But the comment look off. Should be entryNotFound?

cindyyan317 (dismissed):
Good job!

cindyyan317 (approved):

godexsoft (approved):
Looks fine now 👍 🚀

Issue #1052, opened 2023-12-14
Ledger_index API


godexsoft:
Add API that given a date, returns the ledger sequence. Similar to [getLedgerIndex](https://data.xrplf.org/docs/static/index.html#/Ledgers/getLedgerIndex).
Provide 'last/latest' ledger index on or before the provided time. The default value is the current time.

Examples:
```
2023-12-14T07:53:30.000Z - 84580357
2023-12-14T07:53:27.000Z - 84580356
2023-12-14T07:53:23.000Z - 84580356
2023-12-14T07:53:21.000Z - 84580356
2023-12-14T07:53:20.000Z - 84580355
```

PR #1503, opened 2024-06-28
feat: Ledger index

cindyyan317:
#1052 
### Feature ledger_index
Return the latest closed ledger index before the given date.
Without any database change, this PR binary search from valid range.
For the full history node, the range is less than 9,000,000 now. The O(n) -> logN should be 27. Comparing with other handlers, it is acceptable. 

#### Example input:
```
{
 "method": "ledger_index",
 "params": [
     {
       "date":"2024-06-20T09:00:42.000Z"
     }
   ]
}

```

#### Example output:
```
{
  "result":
    {
      "ledger_index": 88816562,
      "ledger_hash": "63911F91F399EFC29E97041477D41B0454159B82EBF0C0223CBE021D89A81D70",
      "closed": "2024-06-20T09:00:42.000Z"
    }
}

```


src/rpc/common/Validators.hpp
```diff
@@ -288,6 +291,33 @@ class Max final {
...
+ * @brief Validate that value can be converted to time according to the given format.
+ */
+class TimeFormatValidator final {
+    std::string format_;
+
+public:
+    /**
+     * @brief Construct the validator storing format value.
+     *
+     * @param format
```
godexsoft: Even though it's self-explanatory, it will probably look weird in the generated HTML documentation.. let's add some text here 🙇

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "rpc/JS.hpp"
+#include "rpc/common/Types.hpp"
+#include "util/Assert.hpp"
+
+#include <boost/json/conversion.hpp>
+#include <boost/json/object.hpp>
+#include <boost/json/value.hpp>
+#include <fmt/core.h>
+#include <ripple/basics/chrono.h>
```
godexsoft: Isn't this xrpl/* now?

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <utility>
+#include <vector>
+
+namespace rpc {
+
+LedgerIndexHandler::Result
+LedgerIndexHandler::process(LedgerIndexHandler::Input input, Context const& ctx) const
+{
+    auto const range = sharedPtrBackend_->fetchLedgerRange();
+    auto [minIndex, maxIndex] = *range;
```
godexsoft: Perhaps could be const too

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+            .ledgerIndex = index,
+            .ledgerHash = ripple::strHex(ledger->hash),
+            .closeTimeIso = ripple::to_string_iso(ledger->closeTime)
+        };
+    };
+
+    if (!input.date)
+        return fillOutputByIndex(maxIndex);
+
+    std::vector<std::uint32_t> ledgerRange(maxIndex - minIndex + 1);
```
godexsoft: On a FH node this would allocate ~34mb each time. Probably not a big deal but worth taking into account.. especially because over time the number will only increase

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+        std::stringstream ss(isoTimeStr);
+        ss >> std::get_time(&time, DATE_FORMAT_);
+        return std::chrono::system_clock::from_time_t(std::mktime(&time)).time_since_epoch().count();
+    };
+
+    auto const ticks = convertISOTimeStrToTicks(*input.date);
+
+    auto const earlierThan = [&](std::uint32_t ledgerIndex) {
+        auto const header = sharedPtrBackend_->fetchLedgerBySequence(ledgerIndex, ctx.yield);
+        auto const ledgetTime =
```
godexsoft: Typo: legerTime

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+    // If the given date is earlier than the first valid ledger, return lgrNotFound
+    if (earlierThan(minIndex))
+        return Error{Status{RippledError::rpcLGR_NOT_FOUND, "ledgerNotInRange"}};
+
+    auto const greaterEqLedgerIter =
+        std::lower_bound(ledgerRange.begin(), ledgerRange.end(), ticks, [&](std::uint32_t ledgerIndex, std::int64_t) {
+            return not earlierThan(ledgerIndex);
+        });
+
+    if (greaterEqLedgerIter != ledgerRange.end()) {
```
godexsoft: nit: Perhaps no need for the curlies here

src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+void
+tag_invoke(boost::json::value_from_tag, boost::json::value& jv, LedgerIndexHandler::Output const& output)
+{
+    auto object = boost::json::object{
+        {JS(ledger_index), output.ledgerIndex},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(close_time_iso), output.closeTimeIso},
+        {JS(validated), true},
+    };
+    jv = std::move(object);
```
godexsoft: I think no need to go through temporary in this case, just directly assign to `jv`

src/rpc/handlers/LedgerIndex.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace rpc {
+
+/**
+ * @brief The ledger_index method fetches the lastest closed ledger before the given date.
+ *
+ */
+class LedgerIndexHandler {
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+    static constexpr char const* DATE_FORMAT_ = "%Y-%m-%dT%TZ";
```
godexsoft: `auto` should be fine here?

godexsoft: Also, `_` for a constant seems strange, even though it's private

src/rpc/handlers/LedgerIndex.hpp
```diff
@@ -0,0 +1,119 @@
...
+
+public:
+    /**
+     * @brief A struct to hold the output data of the command
+     */
+    struct Output {
+        uint32_t ledgerIndex{};
+        std::string ledgerHash;
+        std::string closeTimeIso;
```
godexsoft, line 53: No `validated` for this API?

tests/unit/rpc/BaseTests.cpp
```diff
@@ -353,6 +353,39 @@ TEST_F(RPCBaseTest, WithCustomError)
...
 }
 
+TEST_F(RPCBaseTest, TimeFormatValidator)
+{
+    auto const spec = RpcSpec{
+        {"date", TimeFormatValidator{"%Y-%m-%dT%H:%M:%SZ"}},
+    };
+
+    auto passingInput = json::parse(R"({ "date": "2023-01-01T00:00:00Z" })");
+    ASSERT_TRUE(spec.process(passingInput));
```
godexsoft: Keep in mind that unlike EXPECT_\*, ASSERT_\* immediately terminates the test and you only get the first error out of all. Sometimes it's desired but sometimes not

godexsoft (changes requested):
Nice. Leaving a few comments and nits 👍

src/rpc/handlers/LedgerIndex.hpp
cindyyan317: Since the "validated" only can be "true", I just add it in tag_invoke , instead of using a variable.

src/rpc/handlers/LedgerIndex.cpp
cindyyan317: 👍  Thanks for pointing it out. I was thinking this too. I failed to find a binary search function which supports just two integer range.  Maybe I can implement an iterator. It will be fun.

src/rpc/handlers/LedgerIndex.cpp
cindyyan317: Should be xrpl. I forgot to update this after rebasing. 👍

src/rpc/handlers/LedgerIndex.cpp
godexsoft: You might need to do one of:
- implement your own binary search
- have this vector cached at all times (would require periodic updates)
- implement a fake container (begin, end, iterator) that does not require 32mb but capable of using lower_bound

src/util/IotaIterator.hpp
```diff
@@ -0,0 +1,233 @@
...
+    using difference_type = std::ptrdiff_t;
+    using pointer = int*;
+    using reference = int&;
+    using iterator_category = std::random_access_iterator_tag;
+
+    /**
+     * @brief Construct a new Iota Iterator object
+     * @param v The value of the iterator
+     */
+    IotaIterator(int v) : value_(v)
```
godexsoft: Shouldn't this be some unsigned and/or bigger type like std::size_t?

src/util/IotaIterator.hpp
```diff
@@ -0,0 +1,233 @@
...
+};
+
+/**
+ * @brief Compare two IotaIterator objects, return true if they contains the same value_
+ * @param lhs The left hand side IotaIterator object
+ * @param rhs The right hand side IotaIterator object
+ * @return true If they contains the same value_
+ */
+bool
+operator==(IotaIterator const& lhs, IotaIterator const& rhs) noexcept;
```
godexsoft: This and below look like the 'friend' versions. Do we actually need them at all? would it not work with just member functions that take rhs and compare with *this? Otherwise shouldn't they be hidden friends instead?

godexsoft (dismissed):
Very cool! I'm leaving a few questions, feel free to ignore as it already works as is anyway 🚀

godexsoft (approved):
Awesome 🚀

PR #1519, opened 2024-07-05
fix: Fix extra brackets in warnings

### LINKED ISSUE #1518: Extra `[]` in warnings
## Issue Description

Clio returns extra `[]` in warnings array.

## Steps to Reproduce

Compile Clio on Linux with gcc 12.3. Start Clio and make `server_info` request. Response will have an extra `[]` in the array of warnings.

## Expected Result

No extra `[]` in warnings.

kuznetsss:
Fixes #1518


kuznetsss:
I added tests first to see that they will be failing without the fix.

PR #1522, opened 2024-07-08
fix: Fix the ledger_index timezone issue

cindyyan317:
The get_time does not respect the timezone. The std::mktime functions add the local timezone to it. 
The fix uses **strptime** which converts to UTC time.

We can change to utc_clock once it is available. https://en.cppreference.com/w/cpp/chrono/parse

### DIFF
--- a/src/rpc/handlers/LedgerIndex.cpp
+++ b/src/rpc/handlers/LedgerIndex.cpp
@@ -72,7 +72,6 @@ LedgerIndexHandler::process(LedgerIndexHandler::Input input, Context const& ctx)
         auto const header = sharedPtrBackend_->fetchLedgerBySequence(ledgerIndex, ctx.yield);
         auto const ledgerTime =
             std::chrono::system_clock::time_point{header->closeTime.time_since_epoch() + ripple::epoch_offset};
-        std::cout << "input" << ticks << "ledger" << ledgerTime.time_since_epoch().count() << std::endl;
         return ticks < ledgerTime.time_since_epoch().count();
     };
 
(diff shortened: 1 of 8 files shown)


src/rpc/handlers/LedgerIndex.cpp
```diff
@@ -61,10 +61,9 @@ LedgerIndexHandler::process(LedgerIndexHandler::Input input, Context const& ctx)
         return fillOutputByIndex(maxIndex);
 
     auto const convertISOTimeStrToTicks = [](std::string const& isoTimeStr) {
-        std::tm time = {};
-        std::stringstream ss(isoTimeStr);
-        ss >> std::get_time(&time, DATE_FORMAT);
-        return std::chrono::system_clock::from_time_t(std::mktime(&time)).time_since_epoch().count();
+        std::tm timeStruct;
+        strptime(isoTimeStr.c_str(), DATE_FORMAT, &timeStruct);
```
kuznetsss: I think `strptime()` is a POSIX function, not from C or C++ standard library. Maybe there is nothing wrong with using it.

I found this one in case you want to try to use something a bit more standard:
https://en.cppreference.com/w/cpp/chrono/c/strftime

cindyyan317: Thanks. I have checked the strftime, it is to convert the tm to the string.

godexsoft (dismissed):
Looks like a useful util that we might reuse later - want to move it out to the util namespace?

kuznetsss (dismissed):

godexsoft (approved):
Looks good 🚀

PR #1523, opened 2024-07-08
feat: More verbose forwarding errors

### LINKED ISSUE #1516: Update warning on forwarding failure
## Summary

Clio should provide a bit more verbose error if forwarding failed (requested in #1480). 

## Motivation

Currently it is not very clear what happened if forwarding failed, whether it was because of Clio internal error or something with rippled.

kuznetsss:
Fixes #1516.

### DIFF
--- a/src/etl/ETLState.hpp
+++ b/src/etl/ETLState.hpp
@@ -47,7 +47,10 @@ struct ETLState {
     fetchETLStateFromSource(Forward& source) noexcept
     {
         auto const serverInfoRippled = data::synchronous([&source](auto yield) {
-            return source.forwardToRippled({{"command", "server_info"}}, std::nullopt, {}, yield);
+            if (auto result = source.forwardToRippled({{"command", "server_info"}}, std::nullopt, {}, yield)) {
+                return std::move(result).value();
+            }
+            return std::nullopt;
         });
 
         if (serverInfoRippled)
--- a/src/etl/ETLState.hpp
+++ b/src/etl/ETLState.hpp
@@ -23,6 +23,7 @@
 
 #include <boost/json.hpp>
 #include <boost/json/conversion.hpp>
+#include <boost/json/object.hpp>
 #include <boost/json/value.hpp>
 #include <boost/json/value_to.hpp>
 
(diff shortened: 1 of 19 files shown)


src/etl/ETLState.hpp
```diff
@@ -46,8 +47,11 @@ struct ETLState {
     static std::optional<ETLState>
     fetchETLStateFromSource(Forward& source) noexcept
     {
-        auto const serverInfoRippled = data::synchronous([&source](auto yield) {
-            return source.forwardToRippled({{"command", "server_info"}}, std::nullopt, {}, yield);
+        auto const serverInfoRippled = data::synchronous([&source](auto yield) -> std::optional<boost::json::object> {
+            if (auto result = source.forwardToRippled({{"command", "server_info"}}, std::nullopt, {}, yield)) {
+                return std::move(result).value();
+            }
+            return std::nullopt;
```
godexsoft, line 54: Wasn't this NRVO'd before? I think the way you wrote it you have two different types and RVO might be disabled because of that.. worth checking

godexsoft (approved):
Looks fine to me. Please check RVO stuff though 👍

src/etl/LoadBalancer.cpp
```diff
@@ -233,20 +235,26 @@ LoadBalancer::forwardToRippled(
...
+    rpc::ClioError error = rpc::ClioError::etlCONNECTION_ERROR;
     while (numAttempts < sources_.size()) {
-        if (auto res = sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield)) {
-            response = std::move(res);
+        auto res = sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield);
+        if (res) {
+            response = std::move(res).value();
             break;
         }
+        error = std::max(error, res.error());  // Choose the best result between all sources
```
cindyyan317, line 245: Why the max one is the best one?

kuznetsss: The values of errors are defined so the higher the value the more Clio was able to do in forwarding:
connect < make request < wait for response < parse response.

cindyyan317 (approved):

src/etl/ETLState.hpp
kuznetsss: Checked, there is no difference https://godbolt.org/z/P4GTd9Yd1

PR #1525, opened 2024-07-09
feat: Support Dynamic NFT

### LINKED ISSUE #1471: Support DynamicNFT XLS-46
Clio has an index table nf_token_uris to store the uri. Since the URI is changeable for [XLS-46](https://github.com/XRPLF/rippled/pull/5048) , we need to update our logic to support.

cindyyan317:
Clio's changes for supporting DNFT https://github.com/XRPLF/rippled/pull/5048/files


src/etl/impl/LedgerLoader.hpp
```diff
@@ -136,19 +144,25 @@ class LedgerLoader {
...
 
-        // Now we can unique the NFTs by tokenID.
-        auto last = std::unique(
-            result.nfTokensData.begin(),
-            result.nfTokensData.end(),
-            [](NFTsData const& a, NFTsData const& b) { return a.tokenID == b.tokenID; }
-        );
-        result.nfTokensData.erase(last, result.nfTokensData.end());
+        // Put uri change at the end to ensure the uri not overwritten
+        result.nfTokensData.insert(result.nfTokensData.end(), nfTokenURIChanges.begin(), nfTokenURIChanges.end());
```
shawnxie999, line 158: what would happen if NFTokenMint and NFTokenModify occur in the same ledger? in that case, would the  uri of NFTokenModify overwrite the NFTokenMint in the nft_uri table, considering they would share the same token_id and sequence?

cindyyan317: I wonder if this is a valid scenario. NFTTokenModify before NFTokenMint will be considered as an invalid tx, right? The permission check will fail.

shawnxie999: Yes perm will fail if Modify occurs before Mint, but it's possible that Modify occur after a Mint in the same ledger, although incredibly rare

cindyyan317: Yea. Modify can occur after a mint. In this case, the uri changes insert behind the mint nftdatas, so the modify uri will get written after mint. so modified uri is the nft's final state for this ledger, right?

shawnxie999: yes the modified URI should be the final state. this would mean that the uri written by the Mint would be overwritten by the Modify, since both would have the same `token_id` and `sequence`. i just want to confirm the behavior

src/data/DBHelpers.hpp
```diff
@@ -107,6 +107,7 @@ struct NFTsData {
     ripple::AccountID owner;
     std::optional<ripple::Blob> uri;
     bool isBurned = false;
+    bool onlyUriChanged = false;  // Whether only the URI was changed
```
shawnxie999, line 110: why can't we just use the `owner` field? All other NFT transactions would the `owner` field

cindyyan317: Originally, I changed owner to optional type , so I can use it as a flag. However, I think onlyUriChanged is more readable.

conanfile.py
```diff
@@ -28,7 +28,8 @@ class Clio(ConanFile):
         'protobuf/3.21.9',
         'grpc/1.50.1',
         'openssl/1.1.1u',
-        'xrpl/2.3.0-b1',
```
cindyyan317: Please ignore this change. DNFT will be merged after libxrpl gets updated.

src/data/DBHelpers.hpp
cindyyan317, line 110: After we move to ETL plugins, this intermediate data structure will be replaced by the interface like `onNFTModifyTransaction` `onNFTMintTransaction`.

tests/unit/etl/NFTHelpersTests.cpp
```diff
@@ -17,45 +17,450 @@
...
+// Page index is a valid nft page for ACCOUNT
+constexpr static auto PAGE_INDEX = "E1CD8B79A8BCB52335CD40DE97401B2D60A82872FFFFFFFFFFFFFFFFFFFFFFFF";
+constexpr static auto OFFER_ID = "AA86CBF29770F72FA3FF4A5D9A9FA54D6F399A8E038F72393EF782224865E27F";
 
-struct NFTHelpersTests : public NoLoggerFixture {};
+struct NFTHelpersTest : NoLoggerFixture {};
 
-TEST_F(NFTHelpersTests, ConvertDataFromNFTCancelOfferTx)
+void
+VerifyNFTTransactionsData(
```
kuznetsss: It is better to put such functions into fixture to avoid name collision because all tests are linked into one binary. Also  function name should start by a lowercase letter.

src/data/CassandraBackend.hpp
```diff
@@ -573,11 +573,12 @@ class BasicCassandraBackend : public BackendInterface {
...
-            if (auto const result = res->template get<Blob, std::uint32_t>(); result) { 
-                auto [_ ,seq] = result.value();
+            if (auto const result = res->template get<Blob, std::uint32_t>(); result) {
+                auto [_, seq] = result.value();
                 return seq;
-            }                 LOG(log_.debug()) << "Could not fetch ledger object sequence - no rows";
-           
+            }
+            LOG(log_.debug()) << "Could not fetch ledger object sequence - no rows";
+
```
godexsoft: Looks like we can remove this space line here

src/data/CassandraBackend.hpp
```diff
@@ -881,21 +882,28 @@ class BasicCassandraBackend : public BackendInterface {
...
+                    ));
+                    statements.push_back(
+                        schema_->insertNFTURI.bind(record.tokenID, record.ledgerSequence, record.uri.value())
+                    );
+                }
+            } else {
+                // only uri changed, we update the uri table only
                 statements.push_back(
                     schema_->insertNFTURI.bind(record.tokenID, record.ledgerSequence, record.uri.value())
                 );
```
godexsoft, line 974: It looks like this insert happens on both paths, we can simplify the code to something like this i believe:

```suggestion
if (record.uri) {
    statements.push_back(schema_->insertIssuerNFT.bind(
        ripple::nft::getIssuer(record.tokenID),
        static_cast<uint32_t>(ripple::nft::getTaxon(record.tokenID)),
        record.tokenID
    ));
} 

statements.push_back(
    schema_->insertNFTURI.bind(record.tokenID, record.ledgerSequence, record.uri.value())
);
```

src/etl/NFTHelpers.cpp
```diff
@@ -47,6 +47,16 @@
...
 namespace etl {
 
+std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
+getNFTokenMofidyData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
+{
+    auto tx = NFTTransactionsData(sttx.getFieldH256(ripple::sfNFTokenID), txMeta, sttx.getTransactionID());
+
+    auto const tokenID = sttx.getFieldH256(ripple::sfNFTokenID);
+    // note: sfURI is optional, if it is absent, we will update the uri as empty string
+    return {{tx}, NFTsData(tokenID, txMeta, sttx.getFieldVL(ripple::sfURI))};
```
godexsoft: We probably want to `std::move` the `tx` here or even avoid the temporary at all, wdyt?

src/etl/NFTHelpers.cpp
```diff
@@ -165,7 +175,7 @@ getNFTokenBurnData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
                 node.peekAtField(ripple::sfPreviousFields).downcast<ripple::STObject>();
             if (previousFields.isFieldPresent(ripple::sfNFTokens))
                 prevNFTs = previousFields.getFieldArray(ripple::sfNFTokens);
-        } else if (!prevNFTs && node.getFName() == ripple::sfDeletedNode) {
+        } else if (node.getFName() == ripple::sfDeletedNode) {
```
godexsoft, line 180: I assume this logic is checked via unit tests and the change is correct but can you please explain what's going on here in a few words? 🙇

src/etl/impl/LedgerLoader.hpp
```diff
@@ -121,8 +123,14 @@ class LedgerLoader {
 
             auto const [nftTxs, maybeNFT] = getNFTDataFromTx(txMeta, sttx);
             result.nfTokenTxData.insert(result.nfTokenTxData.end(), nftTxs.begin(), nftTxs.end());
-            if (maybeNFT)
-                result.nfTokensData.push_back(*maybeNFT);
+            // We need to unique the URI changes separately, in case the URI changes are discarded
```
godexsoft, line 129: I'd appreciate a line of space before the comment 🙇

src/etl/impl/LedgerLoader.hpp
```diff
@@ -136,19 +144,25 @@ class LedgerLoader {
...
+            });
+
+            // Now we can unique the NFTs by tokenID.
+            auto last = std::unique(nfTokensData.begin(), nfTokensData.end(), [](NFTsData const& a, NFTsData const& b) {
+                return a.tokenID == b.tokenID;
+            });
+            nfTokensData.erase(last, nfTokensData.end());
+        };
+
+        uniqueNftData(result.nfTokensData);
```
godexsoft: Can any of these be empty? if yes, should we short circuit in the lambda?

godexsoft (changes requested):
Leaving a few nits 👍

src/data/CassandraBackend.hpp
godexsoft: Ah, nevermind, i see now that the `else` belongs to the other `if` 😃

src/etl/NFTHelpers.cpp
cindyyan317: I was thinking to move it , but got error from clang-tidy
>Std::move of the variable 'tx' of the trivially-copyable type 'NFTTransactionsData' has no effect; remove std::move() (fix available)clang-tidy[performance-move-const-arg](https://clang.llvm.org/extra/clang-tidy/checks/performance/move-const-arg.html)

I think I can just avoid the local copy.

src/etl/NFTHelpers.cpp
cindyyan317: The prevNFTS defines 170 line, it is impossible to be true here. So I removed the check. It must be false.

src/etl/impl/LedgerLoader.hpp
cindyyan317: Yes. They could be empty 👍

src/etl/NFTHelpers.cpp
shawnxie999: Nice catch! checking if prevNFTs is empty is unnecessary in this case since it will be always be empty if it reaches this line

src/etl/impl/LedgerLoader.hpp
```diff
@@ -121,8 +123,15 @@ class LedgerLoader {
...
             auto const [nftTxs, maybeNFT] = getNFTDataFromTx(txMeta, sttx);
             result.nfTokenTxData.insert(result.nfTokenTxData.end(), nftTxs.begin(), nftTxs.end());
-            if (maybeNFT)
-                result.nfTokensData.push_back(*maybeNFT);
+
+            // We need to unique the URI changes separately, in case the URI changes are discarded
+            if (maybeNFT) {
+                if (maybeNFT->onlyUriChanged) {
+                    nfTokenURIChanges.push_back(*maybeNFT);
+                } else {
```
shawnxie999, line 133: may not need braces

shawnxie999 (dismissed):
looks good

tests/unit/data/cassandra/ExecutionStrategyTests.cpp
```diff
@@ -405,6 +405,47 @@ TEST_F(BackendCassandraExecutionStrategyTest, WriteMultipleAndCallSyncSucceeds)
...
+    auto strat = makeStrategy();
+    auto const totalRequests = 1024u;
+    auto const numStatements = 16u;
+    auto callCount = std::atomic_uint{0u};
+
+    auto work = std::optional<boost::asio::io_context::work>{ctx};
+    auto thread = std::thread{[this]() { ctx.run(); }};
+
+    ON_CALL(handle, asyncExecute(A<FakeStatement const&>(), A<std::function<void(FakeResultOrError)>&&>()))
+        .WillByDefault([this, &callCount](auto const&, auto&& cb) {
```
kuznetsss: You could make it a bit shorter:
```cpp
EXPECT_CALL( ... ).Times( ... ).WillRepeteadely( ... );
```

tests/unit/data/cassandra/ExecutionStrategyTests.cpp
```diff
@@ -405,6 +405,47 @@ TEST_F(BackendCassandraExecutionStrategyTest, WriteMultipleAndCallSyncSucceeds)
...
 }
 
+TEST_F(BackendCassandraExecutionStrategyTest, WriteEachAndCallSyncSucceeds)
+{
+    auto strat = makeStrategy();
+    auto const totalRequests = 1024u;
+    auto const numStatements = 16u;
+    auto callCount = std::atomic_uint{0u};
+
+    auto work = std::optional<boost::asio::io_context::work>{ctx};
```
kuznetsss: `asio::work` is deprecated, please use `make_work_guard()` instead.

src/data/cassandra/impl/ExecutionStrategy.hpp
```diff
@@ -213,6 +228,24 @@ class DefaultExecutionStrategy {
...
+     * Retries forever with retry policy specified by @ref AsyncExecutor.
+     *
+     * @param statements Vector of statements to execute
+     * @throw DatabaseTimeout on timeout
+     */
+    void
+    writeEach(std::vector<StatementType>&& statements)
+    {
+        if (statements.empty())
+            return;
```
kuznetsss: Probably `for_each` will not do anything if the vector is empty so this is not needed.

tests/unit/data/cassandra/ExecutionStrategyTests.cpp
cindyyan317: I just copy these part from other test, will have a look

cindyyan317: Just check , willRepeatedly can not customise return value.

cindyyan317:
After https://github.com/XRPLF/rippled/pull/5048 is merged, we can merge this PR.

src/etl/impl/LedgerLoader.hpp
cindyyan317: Braces are needed.  clang-tidy will complain.

src/data/AmendmentCenter.hpp
```diff
@@ -132,6 +132,8 @@ struct Amendments {
     REGISTER(fixAMMv1_2);
     REGISTER(AMMClawback);
     REGISTER(Credentials);
+    REGISTER(DynamicNFT);
+    REGISTER(PermissionedDomains);
```
kuznetsss, line 137: I'm not sure whether this is correct, because you don't add support for this amendment to Clio. Maybe we should temporary allow amendment center test to fail instead.

conanfile.py
```diff
@@ -28,7 +28,7 @@ class Clio(ConanFile):
         'protobuf/3.21.9',
         'grpc/1.50.1',
         'openssl/1.1.1u',
-        'xrpl/2.4.0-b1',
+        'xrpl/2.4.0-b1@clio/permission_domains',
```
shawnxie999: does this need to change?

godexsoft: This PR will not be merged until it's in the next beta or something

src/etl/NFTHelpers.hpp
```diff
@@ -33,6 +33,16 @@
...
 
+/**
+ * @brief Get the NFT URI change data from a NFToken Modify transaction
+ *
+ * @param txMeta Transaction metadata
+ * @param sttx The transaction
+ * @return NFT URI change data as a pair of transactions and optional NFTsData
+ */
+std::pair<std::vector<NFTTransactionsData>, std::optional<NFTsData>>
+getNFTokenMofidyData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx);
```
shawnxie999:
```suggestion
getNFTokenModifyData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx);
```

src/data/AmendmentCenter.hpp
cindyyan317: Yes. Here is weird.

PeterChen13579 (dismissed):
I tested rippled/clio standalone with all the test cases I could think of and I didn't find any discrepancies. 💯 👍

src/data/AmendmentCenter.hpp
cindyyan317: Added TODO for now.

kuznetsss (dismissed):

tests/unit/rpc/handlers/TxTests.cpp
```diff
@@ -136,6 +136,7 @@ constexpr auto kDEFAULT_OUT2 = R"({
     "close_time_iso": "2000-01-01T00:00:00Z",
     "validated": true
 })";
+constexpr static auto kINDEX = "E6DBAFC99223B42257915A63DFC6B0C032D4070F9A574B255AD97466726FC322";
```
godexsoft: `static` is redundant here. Also, for future: we prefer `static constexpr auto` instead of `constexpr static auto`. Unfortunately there is no check that i know of to enforce it via clang-format or clang-tidy.

godexsoft (changes requested):
Spotted a nit. Otherwise looks good to me 👍

godexsoft (approved):

PR #1531, opened 2024-07-10
fix: Change the field name from "close_time_iso" to "closed

cindyyan317:
@mounikakun  spotted the field was "closed" from the original design. During the implementation, I changed it to "close_time_iso", because it is used to label time in "ledger" API. Since it mimics the existing API from XRPLF , using the same field name maybe more friendly to users.


kuznetsss (approved):
👍

PR #1533, opened 2024-07-10
docs: Use non-admin WS port

mDuo13:
Changes the default config file to match `rippled`'s defaults where port 6006 is admin WS API and 6005 (commented out by default) for non-admin WS.

Clio should use a non-admin connection to `rippled`. Users will have to modify their `rippled` server config to enable the non-admin WS port as part of setup though.

### DIFF
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -31,7 +31,7 @@
     "etl_sources": [
         {
             "ip": "127.0.0.1",
-            "ws_port": "6006",
+            "ws_port": "6005",
             "grpc_port": "50051"
         }
     ],


godexsoft (approved):
Should be fine 👍

intelliot (approved):
when not using an admin connection, Clio may need to consider rippled's [rate limiting](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/rate-limiting/) ?

cindyyan317 (approved):

cindyyan317:
> when not using an admin connection, Clio may need to consider rippled's [rate limiting](https://xrpl.org/docs/references/http-websocket-apis/api-conventions/rate-limiting/) ?

Yes, we have document to ask user to set Clio ip to secure_gateway.

intelliot (approved):

cindyyan317:
The change is submitted by #1592.

PR #1540, opened 2024-07-12
fix: Relax error when `full` or `accounts` set to `false`

### LINKED ISSUE #1537: Ledger RPC does not honor v1/v2 api spec (Version: 2.2.0)
## Issue Description
In https://github.com/XRPLF/clio/pull/1360/files we made `ledger` RPC throw an error if the `full`/`accounts` parameter is at all present in the request. By doing that we broke client libraries that still use `full`. 

The suggested way to fix this is by relaxing the NotSupported validators for `full` and `accounts` so that they only error out for `true` but just produce the deprecation warning for `false`.

godexsoft:
Fixes #1537


cindyyan317 (dismissed):
LGTM

Bronek (dismissed):
LGTM

cindyyan317 (approved):

Bronek (approved):
Thanks for extra test !

PR #1542, opened 2024-07-15
fix: Fail to deduplicate the same nfts in ttNFTOKEN_CANCEL_OFFER

cindyyan317:
The ttNFTOKEN_CANCEL_OFFER can cancel multiple NFT offers in one go. We need to deduplicate the same NFT if there are some. 
Previously it was 
```
std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
        return a.tokenID < b.tokenID && a.transactionIndex < b.transactionIndex;
    });
```
It fails to deduplicate. I think the intention is to sort by transactionIndex if tokenID is the same.
```
    (a.tokenID == b.tokenID) ? a.transactionIndex < b.transactionIndex : a.tokenID < b.tokenID;
```

In this case, the transactionIndex must be the same, because it comes from one transaction. Comparing the tokenID is enough.

### DIFF
--- a/src/etl/NFTHelpers.cpp
+++ b/src/etl/NFTHelpers.cpp
@@ -294,14 +294,12 @@ getNFTokenCancelOfferData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx
         txs.emplace_back(tokenID, txMeta, sttx.getTransactionID());
     }
 
-    // Deduplicate any transactions based on tokenID/txIdx combo. Can't just
-    // use txIdx because in this case one tx can cancel offers for several
-    // NFTs.
+    // Deduplicate any transactions based on tokenID
     std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
-        return a.tokenID < b.tokenID && a.transactionIndex < b.transactionIndex;
+        return a.tokenID < b.tokenID;
     });
     auto last = std::unique(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
-        return a.tokenID == b.tokenID && a.transactionIndex == b.transactionIndex;
+        return a.tokenID == b.tokenID;
     });
     txs.erase(last, txs.end());
     return {txs, {}};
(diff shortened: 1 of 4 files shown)


shawnxie999:
could you elaborate why the following doesnt work?
```
std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
        return a.tokenID < b.tokenID && a.transactionIndex < b.transactionIndex;
    });
```

godexsoft:
> could you elaborate why the following doesnt work?
> 
> ```
> std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
>         return a.tokenID < b.tokenID && a.transactionIndex < b.transactionIndex;
>     });
> ```

I think you can see the issue here: https://godbolt.org/z/c4sMheMxh

@cindyyan317 are you sure we should use just tokenID? seems like unique would not work if you see my example. But maybe there is some guarantee from the input data that i'm missing

cindyyan317:
> > could you elaborate why the following doesnt work?
> > ```
> > std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
> >         return a.tokenID < b.tokenID && a.transactionIndex < b.transactionIndex;
> >     });
> > ```
> 
> I think you can see the issue here: https://godbolt.org/z/c4sMheMxh
> 
> @cindyyan317 are you sure we should use just tokenID? seems like unique would not work if you see my example. But maybe there is some guarantee from the input data that i'm missing

I think we can confirm all NFTTransactionsData have the same transactionIndex, because it all generated from here -> https://github.com/XRPLF/clio/blob/develop/src/etl/NFTHelpers.cpp#L294

shawnxie999:
> I think we can confirm all NFTTransactionsData have the same transactionIndex, because it all generated from here -> https://github.com/XRPLF/clio/blob/develop/src/etl/NFTHelpers.cpp#L294

Yes you are right, but the current logic isn't broken, it's just unnecessary I guess.

godexsoft (approved):
With the confirmation that `transactionIndex` is always the same, it makes sense to only account for `tokenID` 👍

shawnxie999 (approved):
thanks for making this change! 👍

PR #1543, opened 2024-07-15
fix: Add more account check  

cindyyan317:
Make sure all char is alphanumeric for account

### DIFF
--- a/src/util/AccountUtils.hpp
+++ b/src/util/AccountUtils.hpp
@@ -28,8 +28,9 @@
 namespace util {
 
 /**
- * @brief A wrapper of parseBase58 function. It adds the check if all characters in the input string are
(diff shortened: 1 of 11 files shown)


src/util/AccountUtils.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <cctype>
+#include <optional>
+#include <string>
+
+namespace util {
+
+/**
+ * @brief A wrapper of parseBase58 function. It adds the check if all characters in the input string are
+ *alphanumeric. If not, it returns an empty optional, instead of calling the parseBase58 function.
```
godexsoft: Let's fix the missing space here and in other doxy comments 👍

godexsoft (changes requested):
Code looks good 👍 Need to fix the formatting on the doxygen and we are good to go 🚀

godexsoft (approved):

PR #1544, opened 2024-07-15
refactor: Clio Config

PeterChen13579:
Implementation of new config definition + methods + UT

Steps that still need to be implemented: 
-  Make ClioConfigDefinition and it's method to be as constexpr as possible
-  Getting User Config file and populating the values in ConfigDefinition while checking for constraints on user values
-  Replacing all the places where we fetch config values (by using config.valueOr/MaybeValue) to instead get it from Config Definition
- Generate markdown file using Clio Config Description


src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+namespace util::config {
+
+/**
+ * @brief Array definition for Json/Yaml config
+ *
+ * Used in ClioConfigDefinition to represent multiple potential values (like whitelist)
+ */
+class Array {
+public:
+    /** @brief Custom iterator class which returns ValueView of what's underneath the Array
```
godexsoft: If the brief does not fit on one line, we add a space after `/**` and before `@brief` (like above) - let's keep the style uniform 👍

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+        /**
+         * @brief Prefix increment operator
+         *
+         * @return Reference to the incremented ArrayIterator
+         */
+        ArrayIterator&
+        operator++()
+        {
+            m_ptr++;
```
godexsoft: While the compiler is good at optimizing these out, it's still a good idea to use the prefix form whenever you don't actually need a copy of the old value. So i'd replace with `++m_ptr` but also would replace `m_ptr` with `ptr_` to obey our overall style 👍

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    /**
+     * @brief Constructs an Array with the provided arguments
+     *
+     * @tparam Args Types of the arguments
+     * @param args Arguments to initialize the elements of the Array
+     */
+    template <typename... Args>
+    constexpr Array(Args... args) : elements_{args...}
```
godexsoft: This probably should be `Args&&... args` and `std::forward<Args>(args)...` or similar

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * @brief Returns an iterator to the end of the Array
+     *
+     * @return Iterator to the end of the Array
+     */
+    auto
+    end() const
+    {
+        return ArrayIterator{elements_.data() + elements_.size()};
```
godexsoft: I think these days this should be a `Sentinel type` which differs from the type of the iterator (an empty type usually) and needs to always report unequal to an instance of the ArrayIterator

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /** @brief Add ConfigValues to Array class
```
godexsoft: Same issue with missing space

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @param value The ConfigValue to add
+     */
+    void
+    emplace_back(ConfigValue value)
+    {
+        elements_.emplace_back(value);
```
godexsoft: You are not actually `emplacing` the value here. This calls the copy constructor under the hood i believe.
If you want to move the copy of a value that you have at line 158, you need to `elements_.push_back(std::move(value))`. Keep in mind that `emplacing` is a different thing though - it's using the provided value(s) and passes them as-is to the constructor of what you are emplacing (ConfigValue). So let's say your `ConfigValue` takes an `int` then you would `emplace_back(123)` and that constructs the `ConfigValue` inside of the vector, instead of copying or moving it in.

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <cstddef>
+#include <stdexcept>
+#include <string_view>
+
+namespace util::config {
+
+ArrayView::ArrayView(std::string_view prefix, ClioConfigDefinition const& configDef)
+    : prefix_{prefix}, clioConfig_{configDef}
+{
+    auto it = std::find_if(configDef.map_.begin(), configDef.map_.end(), [this](auto const& pair) {
```
godexsoft: If the reason to `friend` all the view classes inside of `ClioConfigDefinition` is to access the `map_` member, consider exposing the map via a readonly accessor or even add `begin()` and `end()` on the `ClioConfigDefinition` so that you can read the values.

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+size_t
+ArrayView::size() const
+{
+    for (auto const& pair : clioConfig_.map_) {
+        if (pair.first.starts_with(prefix_)) {
+            return std::get<Array>(pair.second).size();
+        }
+    }
+    throw std::logic_error("Arrayview is initialized with incorrect prefix.");
```
godexsoft: "ArrayView" 🔢

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace util::config {
+
+ArrayView::ArrayView(std::string_view prefix, ClioConfigDefinition const& configDef)
+    : prefix_{prefix}, clioConfig_{configDef}
+{
+    auto it = std::find_if(configDef.map_.begin(), configDef.map_.end(), [this](auto const& pair) {
+        return pair.first.starts_with(prefix_);
+    });
+    ASSERT(it != clioConfig_.map_.end(), "prefix does not exist in config definition.");
```
godexsoft: Let's keep the message capitalized like the rest of them 👍

godexsoft: It could also be useful to include the offending prefix in the assert message

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,110 @@
...
+     */
+    ArrayView(std::string_view prefix, ClioConfigDefinition const& configDef);
+
+    /**
+     * @brief Returns an ObjectView at the specified index.
+     *
+     * @param idx Index of the object to retrieve.
+     * @return ObjectView at the specified index.
+     */
+    ObjectView
```
godexsoft: You can add `[[nodiscard]]` to all member functions where it makes sense

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,110 @@
...
+    auto
+    endValues() const
+    {
+        ASSERT(clioConfig_.map_.contains(prefix_), "Current string is prefix, not full key");
+        return std::get<Array>(clioConfig_.map_.at(prefix_)).end();
+    }
+
+private:
+    std::string const prefix_;
+    ClioConfigDefinition const& clioConfig_;
```
godexsoft: This may not be important in this case but worth mentioning. It's best to avoid reference members as they implicitly delete assignment operators because assigning to a reference would not rewire the reference but  overwrite whatever object the reference is pointing to. In your case, probably you don't reassign objects of these view classes so you never notice that the assignment is broken/disabled.

Normally, if you need a reference as a member, it's much safer to use `std::reference_wrapper`

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        if (mapKey.starts_with(prefixWithDot) && std::holds_alternative<ConfigValue>(mapVal))
+            return ObjectView{prefix, *this};
+
+        if (mapKey.starts_with(prefixWithDot) && std::holds_alternative<Array>(mapVal)) {
+            ASSERT(std::get<Array>(mapVal).size() > idx, "index provided is out of scope");
+            return ObjectView{prefix, idx.value(), *this};
+        }
+    }
+    throw std::invalid_argument("Key is not found in config");
```
godexsoft: Again, would be useful to specify what key is not found

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+ArrayView
+ClioConfigDefinition::getArray(std::string_view prefix) const
+{
+    std::string key = std::string(prefix);
```
godexsoft: You can use `auto` a lot more often than you think 😁

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+    throw std::invalid_argument("no matching key");
+}
+
+/**
+ * @brief Description of each config key and what they mean. Used to generate markdown file
+ *
+ * Key-value pairs. Key is configKey, value is its matching description
+ * Maybe_unused will be removed when markdown file is generated
+ */
+[[maybe_unused]] static constexpr ClioConfigDescription const DESCRIPTIONS{
```
godexsoft: `constexpr` is implicitly `const` 👍

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,197 @@
...
+    /** @brief Struct to represent a key-value pair*/
+    struct KV {
+        constexpr KV() = default;
+
+        /** @brief Constructs a key value pair of config key and it's corresponding description
+         *
+         * @param k key of config
+         * @param v value of config
+         */
+        constexpr KV(std::string_view k, std::string_view v) : key(k), value(v)
```
godexsoft: I don't think you need either default constructor or this one - without them `KV` is an aggregate type and can be easily constructed like `KV{k, v}` or `KV{.key = k, .value = v}`

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,197 @@
...
+    /** @brief Constructs a new Clio Config Description based on pre-existing descriptions
+     *
+     * @param init initializer list with a predefined set of key-value pairs where key is
+     * config key and value is the description of the key
+     */
+    constexpr ClioConfigDescription(std::initializer_list<KV> init)
+    {
+        auto it = init.begin();
+        for (std::size_t i = 0; i < configDescription.size(); ++i) {
+            if (it != init.end()) {
```
godexsoft: This looks suspiciously like a `std::copy_n`?

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,197 @@
...
+    {
+        auto const itr = std::find_if(configDescription.begin(), configDescription.end(), [&key](auto const& v) {
+            return v.key == key;
+        });
+        ASSERT(itr != configDescription.end(), "key doesn't exist in config");
+        return itr->value;
+    }
+
+private:
+    std::array<KV, 53> configDescription;
```
godexsoft: I feel like this is going to be painful to have to change every time we add a new config option..

src/util/newconfig/ConfigFileInterface.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @brief The interface for configuration files.
+ *
+ * This class defines the interface for handling configuration files,
+ * which can be implemented for different formats such as JSON or YAML.
+ */
+class ConfigFileInterface {
+public:
+    virtual ~ConfigFileInterface() = default;
```
godexsoft, line 38: Space line needed

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <variant>
+
+namespace util::config {
+
+template <typename>
+constexpr bool always_false = false;
```
godexsoft: Let's not use snake case 👍

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+ * @brief Get the corresponding clio config type
+ *
+ * @tparam Type The type to get the corresponding ConfigType for
+ * @return The corresponding ConfigType
+ */
+template <typename Type>
+constexpr ConfigType
+getType()
+{
+    if constexpr (std::is_same_v<Type, int>) {
```
godexsoft: Do we only support signed `int` type here?

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+        return ConfigType::Integer;
+    } else if constexpr (std::is_same_v<Type, std::string>) {
+        return ConfigType::String;
+    } else if constexpr (std::is_same_v<Type, double>) {
```
godexsoft, line 51: Same question for `double` and `float`

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+ * Used in ClioConfigDefinition to indicate the required type of value and
+ * whether it is mandatory to specify in the configuration
+ */
+class ConfigValue {
+    friend class ValueView;
+
+public:
+    using Type = std::variant<int, std::string, bool, double>;
+
+    constexpr ConfigValue() = default;
```
godexsoft: Space line would be nice 👍

src/util/newconfig/Errors.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#pragma once
+
+#include <string>
+#include <string_view>
+
+namespace util::config {
+
+/** @brief todo: Will display the different errors when parsing config */
+struct Error {
```
godexsoft, line 28: This can be a simple aggregate i believe?
Also, members with `_` at the end of their name are usually for `private` members - use just `key` and `error` if they are public data members.

src/util/newconfig/ObjectView.cpp
```diff
@@ -0,0 +1,93 @@
...
+std::string
+ObjectView::getFullKey(std::string_view key) const
+{
+    return fmt::format("{}.{}", prefix_, key);
+}
+
+bool
+ObjectView::containsKey(std::string_view key) const
+{
+    std::string fullKey = getFullKey(key);
```
godexsoft: This temporary can be made `const`. Or simply turn this into a one-liner `clioConfig_.map_.contains(getFullKey(key))`. I'd still prefer ClioConfigDefinition to hide `map_` and introduce some accessors for it. Either the functions you need or at least a function that returns read-only reference to the map.

src/util/newconfig/ObjectView.cpp
```diff
@@ -0,0 +1,93 @@
...
+    return clioConfig_.map_.contains(fullKey);
+}
+
+bool
+ObjectView::startsWithKey(std::string_view key) const
+{
+    for (auto const& [valueKey, value] : clioConfig_.map_) {
```
godexsoft: Could be rewritten with `std::find_if` or similar

src/util/newconfig/ObjectView.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (valueKey.starts_with(key))
+            return true;
+    }
+    return false;
+}
+
+ValueView
+ObjectView::getValue(std::string_view key) const
+{
+    std::string fullKey = getFullKey(key);
```
godexsoft: Const

tests/unit/util/newconfig/ValueViewTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include "util/newconfig/ConfigDefinition.hpp"
+#include "util/newconfig/ConfigValue.hpp"
+#include "util/newconfig/ValueView.hpp"
+
+#include <../common/newconfig/FakeConfigData.hpp>
```
godexsoft: Looks like a strange include path. This should be in quotes instead and should not need to use `..`.

tests/unit/NewConfigTests.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+// TODO: parsing config file and populating into config will be here once implemented
+
+TEST(ConfigDescription, getValues)
+{
+    ClioConfigDescription definition{
+        {{"database.type", "Type of database."}, {"key", "value"}, {"array[].sub", "this is the subsection."}}
+    };
+
+    ASSERT_EQ(definition.get("key"), "value");
```
godexsoft: Here and in other tests: ASSERT vs. EXPECT is important to understand. Most of the time, you will want EXPECT_* instead of the ASSERT_* alternatives. ASSERT will immediately halt the test and error out while EXPECT will report the issue and continue trying. With EXPECT you may get more errors at once. With ASSERT you will likely only get the first one. Sometimes it makes sense to use ASSERT though - mostly when continuing with execution not be ok. Like when you check your vector to contain at least N elements with an ASSERT but later checks (within bounds) on the vector can be simple EXPECTED calls instead.

tests/unit/util/newconfig/ObjectViewTests.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+using namespace util::config;
+
+struct ObjectViewTest : testing::Test {
+    ClioConfigDefinition configData = generateConfig();
+};
+
+TEST_F(ObjectViewTest, ObjectValueTest)
+{
+    auto headerObj = configData.getObject("header");
```
godexsoft: Here and in other places - see if you can add `const` where suitable 👍

src/util/newconfig/ValueView.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace util::config {
+
+std::string_view
+ValueView::asString() const
+{
+    if (this->type() == ConfigType::String)
```
godexsoft: Any reason you check `has_value` in other accessors but not in this one?

godexsoft (changes requested):
Nice progress. Leaving some improvement ideas 👍

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,72 @@
...
+    ASSERT(prefix_.contains(".[]"), "Not an array");
+}
+
+ValueView
+ArrayView::valueAt(std::size_t idx) const
```
cindyyan317, line 40: I feel the interface is a bit weird here. The valueAt will cause error when prefix_ is just prefix, but size() is fine.

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * @brief View for array structure for config.
+ *
+ * This class provides a view into an array structure within ClioConfigDefinition.
+ * It allows accessing individual elements of the array as either values or objects, and
+ * is used within the ClioConfigDefinition to represent multiple potential values.
+ */
+class ArrayView {
+public:
+    ArrayView() = delete;
```
cindyyan317: Is this deleted automatically when we have defined our own constructor?

src/util/newconfig/ArrayView.hpp
cindyyan317: And if view can extend the underlying object's life, we might use shared_ptr. Or in your case, you can guarantee the clioConfig_ lives longer than all views.

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "util/newconfig/ValueView.hpp"
+
+#include <cstddef>
+#include <string>
+#include <string_view>
+
+namespace util::config {
+
+/**
+ * @brief View for array structure for config.
```
cindyyan317, line 36: I saw some methods in ArrayView require "prefix_" is full key. If ArrayView does not have any meaningful functions when prefix_ just a prefix. Let us just make it a full key , don't relax it.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <string_view>
+#include <variant>
+
+namespace util::config {
+
+/**
+ * @brief A class to enforce constraints on certain values within ClioConfigDefinition
+ * TODO: will be used when parsing user config
+ */
+class ConfigConstraints {
```
cindyyan317: not used for now?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+     {"database.cassandra.replication_factor", ConfigValue{ConfigType::Integer}.defaultValue(3)},
+     {"database.cassandra.table_prefix", ConfigValue{ConfigType::String}.defaultValue("table_prefix")},
+     {"database.cassandra.max_write_requests_outstanding", ConfigValue{ConfigType::Integer}.defaultValue(10'000)},
+     {"database.cassandra.max_read_requests_outstanding", ConfigValue{ConfigType::Integer}.defaultValue(100'000)},
+     {"database.cassandra.threads",
+      ConfigValue{ConfigType::Integer}.defaultValue(static_cast<int>(std::thread::hardware_concurrency()))},
+     {"database.cassandra.core_connections_per_host", ConfigValue{ConfigType::Integer}.defaultValue(1)},
+     {"database.cassandra.queue_size_io", ConfigValue{ConfigType::Integer}.optional()},
+     {"database.cassandra.write_batch_size", ConfigValue{ConfigType::Integer}.defaultValue(20)},
+     {"etl_source.[].ip", Array{ConfigValue{ConfigType::String}.optional()}},
```
cindyyan317, line 63: I am not sure the ip is optional?

cindyyan317: I am a bit confused about the [] definition . When we use a.[].b, is a an array or b?
For etl_source.[].ip, i guess etlsource is array.
but for dos_guard.[].whitelist, should the whitelist is array?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+     {"api_version.min", ConfigValue{ConfigType::Integer}},
+     {"api_version.max", ConfigValue{ConfigType::Integer}}}
+};
+
+ObjectView
+ClioConfigDefinition::getObject(std::string_view prefix, std::optional<std::size_t> idx) const
+{
+    std::string prefixWithDot = std::string(prefix) + ".";
+    for (auto const& [mapKey, mapVal] : map_) {
+        if (mapKey.starts_with(prefixWithDot))
```
cindyyan317: You might want to add idx is nullopt here? **if (mapKey.starts_with(prefixWithDot and idx==nullopt))** Because it actually allows to get object from array, right?

cindyyan317: I would suggest to split it to two  overload functions , one for object, one for array.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+    }
+    throw std::invalid_argument("Key is not found in config");
+}
+
+ArrayView
+ClioConfigDefinition::getArray(std::string_view prefix) const
+{
+    std::string key = std::string(prefix);
+    if (!prefix.contains(".[]"))
```
cindyyan317: nit: do we support ends_with?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+        key += ".[]";
+
+    for (auto const& [mapKey, mapVal] : map_) {
+        if (mapKey.starts_with(key))
+            return ArrayView{key, *this};
+        ASSERT(!mapKey.starts_with(prefix), "Trying to retrieve an Array when value is an object");
```
cindyyan317: if my config is 
{
   "server_ip" :[],
   "server_ips": www.ok.com
}
does this assert fire?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,213 @@
...
+    throw std::invalid_argument("Key is not found in config");
+}
+
+ValueView
+ClioConfigDefinition::getValue(std::string_view fullKey) const
+{
+    if (map_.contains(fullKey) && std::holds_alternative<ConfigValue>(map_.at(fullKey))) {
+        return ValueView{std::get<ConfigValue>(map_.at(fullKey))};
+    }
+    ASSERT(
```
cindyyan317: I think you only need to assert !map_.contains(fullKey) here .

src/util/newconfig/ConfigDefinition.cpp
cindyyan317: Can **ClioConfigDefinition** also be constexpr? If so, you can do the static check for their length, ensure their size are same. It will be nice to have such obligations.

src/util/newconfig/ConfigValue.hpp
cindyyan317: Is default ConfigValue a required intType?

src/util/newconfig/ConfigDefinition.hpp
cindyyan317: Can we make it deductible ?
something like
```
template<std::size_t T>
struct TT{
    TT(std::array<int, T> const& s){
        
    }
};

int main()
{
    TT t(std::array<int,3>{{1,2,3}});
    return 0;
}
```

src/util/newconfig/ConfigDefinition.cpp
godexsoft: yeah, this could be just `prefix.ends_with(..)`, even without the temporary. We already use string's .ends_with in the codebase. I think string_view support should be on par with string but who knows - need to try.

src/util/newconfig/ConfigDefinition.hpp
godexsoft: Either thru a helper or with CTAD and have it fully defined in the header: https://godbolt.org/z/csc8bMo4K

src/util/newconfig/ValueView.cpp
PeterChen13579: I forgot it here 🤦

src/util/newconfig/ArrayView.hpp
PeterChen13579: yea you're right 😄

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: In the case of a.[].b, both getArrayView(a) and getArrayView(a.[].b) is a correct ArrayView, but getArrayView(a) is restricted to certain methods. This is because consider : 
"etl_sources": [
    {
       "ip": "127.0.0.1",
       "ws_port": "6006",
       "grpc_port": "50051"
    },
 {
       "ip": "idk",
       "ws_port": "1234",
       "grpc_port": "1111"
    }
        ],

First etl_sources opens and closes with  "[]" so getArray(etl_sources) is supported.
Within each entry of etl_sources, there are "{}", which represents an object. Therefore, once you getArray(etl_sources), you can get the object at a certain index,  for example objectAt(0) will return the object with "ip": "127.0.0.1", "ws_port": "6006", etc., and objectAt(1) will return  "ip": "idk” …etc... 

In the other case of getArray(etl_sources.[].ip), it is just an arrayView in configDefinition, and so the .begin() can iterate through all the ip values, and valueAt(0) will fetch "127.0.0.1", and ivalueAt(1) will fetch "idk” because the 2nd ip has value of "idk"
so "dos_guard.[].whitelist" is an array.
Let me know if you have any questions or that there can be improvements. 😀.

PeterChen13579: > I am not sure the ip is optional?

Screenshot 2024-07-16 at 4 39 51 PM

I'm seeing in our codebase that it is optional. Let me know if I'm wrong 👍

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: > I would suggest to split it to two overload functions , one for object, one for array.

I agree!! thanks 🙇

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: Shouldn't be .ends_with() because we can get ArrayView by doing getArray(etl_sources.[].ip) 😅 👍

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: yea, It will be used next part of config when I take config values from user and check constraints on them. This PR is already beefy so I decided to leave it out for now 🙇

src/util/newconfig/ArrayView.hpp
PeterChen13579: See  https://github.com/XRPLF/clio/pull/1544#discussion_r1680301679 😄

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: You're right 👍. If you're searching for "server_ip" and "server_ips" is checked before it, the current assertion could lead to bugs. I'm going to change it to ASSERT the value is of type Array instead; Thank you for the catch!! 🚀

src/util/newconfig/ArrayView.cpp
PeterChen13579: This is because if there are 2 values for "dosguard.[].whitelist", if we do valueAt(2), it is going to error out because there isn't a 3rd value there. Check https://github.com/XRPLF/clio/pull/1544#discussion_r1680301679 😄

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: yea! The next step is to make it as constexpr as possible.

src/util/newconfig/ConfigValue.hpp
PeterChen13579: Don't think the default constructor is needed actually. 👍

PeterChen13579:
Thank you for taking the time to review @godexsoft @cindyyan317 🙇

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,68 @@
...
+    ASSERT(clioConfig_.get().contains(prefix_), "Current string {} is a prefix, not a key of config", prefix_);
+    ConfigValue const& val = clioConfig_.get().atArray(prefix_).at(idx);
+    return ValueView{val};
+}
+
+size_t
+ArrayView::size() const
+{
+    return clioConfig_.get().arraySize(prefix_);
+    throw std::logic_error("ArrayView is initialized with incorrect prefix. Shoudn't reach here");
```
kuznetsss: How could it reach here after `return`?

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,109 @@
...
+    [[nodiscard]] size_t
+    size() const;
+
+    /**
+     * @brief Returns an iterator to the beginning of the values.
+     *
+     * @return Iterator to the beginning of the values.
+     */
+    [[nodiscard]] auto
+    beginValues() const
```
kuznetsss: This implementation could be also moved to cpp file.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,190 @@
...
+     {"database.cassandra.max_read_requests_outstanding", ConfigValue{ConfigType::Integer}.defaultValue(100'000)},
+     {"database.cassandra.threads",
+      ConfigValue{ConfigType::Integer}.defaultValue(static_cast<uint32_t>(std::thread::hardware_concurrency()))},
+     {"database.cassandra.core_connections_per_host", ConfigValue{ConfigType::Integer}.defaultValue(1)},
+     {"database.cassandra.queue_size_io", ConfigValue{ConfigType::Integer}.optional()},
+     {"database.cassandra.write_batch_size", ConfigValue{ConfigType::Integer}.defaultValue(20)},
+     {"etl_source.[].ip", Array{ConfigValue{ConfigType::String}.optional()}},
+     {"etl_source.[].ws_port", Array{ConfigValue{ConfigType::String}.optional().min(1).max(65535)}},
+     {"etl_source.[].grpc_port", Array{ConfigValue{ConfigType::String}.optional().min(1).max(65535)}},
+     {"forwarding_cache_timeout", ConfigValue{ConfigType::Integer}},
```
kuznetsss: Instead of this we now have:
```json
    "forwarding": {
        "cache_timeout": 0.250, // in seconds, could be 0, which means no cache
        "request_timeout": 10.0 // time for Clio to wait for rippled to reply on a forwarded request (default is 10 seconds)
    },
```

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,200 @@
...
+#include <optional>
+#include <stdexcept>
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+#include <variant>
+
+namespace util::config {
+
+class ArrayView;
```
kuznetsss: I think it would be better (if it is possible) to make a forward declaration of `ClioConfigDefinition` in `ArrayView.hpp` but here include `ArrayView.hpp`.

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,200 @@
...
+    std::optional<Error>
+    validate(ConfigFileInterface const& config) const;
+
+    /**
+     * @brief Generate markdown file of all the clio config descriptions
+     *
+     * @param configDescription The configuration description object
+     * @return An optional Error if generating markdown fails
+     */
+    std::optional<Error>
```
kuznetsss: Would be better to return `std::expected<std::string, Error>`.

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * @brief Returns the ObjectView specified with the prefix from Array
+     *
+     * @param prefix The key prefix for the ObjectView
+     * @param idx Optional index for the ObjectView
+     * @return ObjectView with the given prefix
+     * @throws std::runtime_error if no valid prefix exists in the config
+     */
+    ObjectView
+    getObject(std::string_view prefix, std::size_t idx) const;
```
kuznetsss: You could merge `getObject()` methods into one:
```cpp
ObjectView
getObject(std::string_view prefix, std::optional<std::size_t> index = std::nullopt) const;
```

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,190 @@
...
+     {"api_version.max", ConfigValue{ConfigType::Integer}}}
+};
+
+ObjectView
+ClioConfigDefinition::getObject(std::string_view prefix) const
+{
+    auto const prefixWithDot = std::string(prefix) + ".";
+    for (auto const& [mapKey, mapVal] : map_) {
+        if (mapKey.starts_with(prefixWithDot))
+            ASSERT(!mapKey.ends_with(".[]"), "Trying to retrieve an object when value is an Array");
```
kuznetsss: What if the function called with`dosguard` prefix, but there is `dosguard.whitelist.[]` which will fail the assertion, while the call is correct.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+        if (mapKey.starts_with(prefixWithDot) && std::holds_alternative<ConfigValue>(mapVal)) {
+            ASSERT(std::holds_alternative<ConfigValue>(mapVal), "Trying to get object from Array but requires index");
+            return ObjectView{prefix, *this};
+        }
+    }
+    throw std::invalid_argument(fmt::format("Key {} is not found in config", prefixWithDot));
```
kuznetsss: I think this should be `ASSERT()`. Because we define all possible config keys in the definition. All the arguments of `getObject()` will be hardcoded as well. So we should check we never call it with a wrong key.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+ValueView
+ClioConfigDefinition::getValue(std::string_view fullKey) const
+{
+    if (map_.contains(fullKey) && std::holds_alternative<ConfigValue>(map_.at(fullKey))) {
+        return ValueView{std::get<ConfigValue>(map_.at(fullKey))};
+    }
+    ASSERT(
+        map_.contains(fullKey) && std::holds_alternative<Array>(map_.at(fullKey)),
```
kuznetsss: Probably not array here.

src/util/newconfig/ConfigFileInterface.hpp
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @brief Parses the provided path of user clio configuration data
+     *
+     * @param filePath The path to the Clio Config data
+     */
+    virtual void
+    parse(std::string_view filePath) = 0;
```
kuznetsss, line 45: Maybe better to move it to constructor which will throw if there is an error during file parsing.

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <variant>
+
+namespace util::config {
+
+template <typename>
+constexpr bool alwaysFalse = false;
```
kuznetsss: Maybe better to move into a separate file in utils. We already have at least one place with such code.

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * @brief Sets the default value for the config
+     *
+     * @param value The default value
+     * @return Reference to this ConfigValue
+     */
+    ConfigValue&
+    defaultValue(Type value)
+    {
+        required_ = false;
```
kuznetsss: I think you don't need `required_` at all. After parsing each key should have some value. If it was not provided by a default value, than it should be parsed from file. If there is no value in the file it is error.

src/util/newconfig/ObjectView.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+}
+
+ObjectView::ObjectView(std::string_view prefix, std::size_t arrayIndex, ClioConfigDefinition const& clioConfig)
+    : prefix_{prefix}, arrayIndex_{arrayIndex}, clioConfig_{clioConfig}
+{
+}
+
+std::string
+ObjectView::getFullKey(std::string_view key) const
```
kuznetsss: It is not strict, but it is better when the order of methods in cpp file is the same as in hpp file.

src/util/newconfig/ValueView.hpp
```diff
@@ -0,0 +1,126 @@
...
+class ValueView {
+public:
+    /**
+     * @brief Constructs a ValueView object
+     *
+     * @param configVal the config Value to view
+     */
+    ValueView(ConfigValue const& configVal) : configVal_{configVal}
+    {
+    }
```
kuznetsss: All implementations which are not templates should be in cpp file.

tests/common/newconfig/FakeConfigData.hpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/* The config definition above would look like this structure in config.json:
+"header": {
+   "text1": "value",
+   "port": 123,
+   "admin": true,
+   "sub": {
+       "sub2Value": "TSM"
+   }
```
kuznetsss: There should be a case of an array inside an object.

tests/unit/NewConfigTests.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+using namespace util::config;
+
+// TODO: parsing config file and populating into config will be here once implemented
+
+TEST(ConfigDescription, getValues)
+{
+    ClioConfigDescription definition{};
+
+    EXPECT_EQ(definition.get("database.type"), "Type of database to use.");
```
kuznetsss: There should be a negative test, checking `get()` behaviour for a not existing key.

tests/unit/util/newconfig/ArrayViewTests.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+struct ArrayViewTest : testing::Test {
+    ClioConfigDefinition const configData = generateConfig();
+};
+
+TEST_F(ArrayViewTest, ArrayValueTest)
+{
+    ArrayView const arrVals = configData.getArray("array.[].sub");
+    auto valIt = arrVals.beginValues();
+    EXPECT_EQ((*valIt++).asDouble(), 111.11);
```
kuznetsss: It is better to not compare doubles for exact equality.

kuznetsss:
There should be also tests for Array, ConfigConstraints (or delete them for now), ConfigDefinition, ConfigValue.

src/util/UnsupportedType.hpp
```diff
@@ -0,0 +1,27 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+namespace util {
+
+template <typename>
```
kuznetsss, line 25: There should be documentation for this type.

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,208 @@
...
+    struct Sentinel {};
+
+    /**
+     * @brief Custom iterator class which returns ValueView of what's underneath the Array
+     */
+    struct ArrayIterator {
+        using iterator_category = std::forward_iterator_tag;
+        using pointer = ConfigValue const*;
+        using reference = ConfigValue const&;
+        using valueType = ConfigValue;
```
kuznetsss: Should be `value_type` to follow the convention here.

tests/unit/util/newconfig/ArrayTests.cpp
```diff
@@ -0,0 +1,49 @@
...
+//==============================================================================
+
+#include "util/newconfig/Array.hpp"
+#include "util/newconfig/ConfigValue.hpp"
+
+#include <gtest/gtest.h>
+
+using namespace util::config;
+
+TEST(ArrayTest, testConfigArray)
```
kuznetsss, line 28: There should be tests for ArrayIterator as well.

src/util/newconfig/Array.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+        /**
+         * @brief Equality operator for Sentinel comparison
+         *
+         * @return true since it will always compare to end
+         */
+        bool
+        operator==(Sentinel const&) const
+        {
+            return true;
```
kuznetsss: Looks like this will be `true`:
```cpp
bool foo() {
  Array a{1, 2, 3};
  return a.begin() == a.end();
}
```

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return pair.first.starts_with(prefix_);
+    });
+    ASSERT(it != clioConfig_.get().end(), "Prefix {} does not exist in config definition.", prefix);
+    ASSERT(prefix_.contains(".[]"), "Not an array");
+}
+
+ValueView
+ArrayView::valueAt(std::size_t idx) const
+{
+    ASSERT(clioConfig_.get().contains(prefix_), "Current string {} is a prefix, not a key of config", prefix_);
```
kuznetsss, line 42: Why not to move this to constructor and replace the check there?

src/util/newconfig/ArrayView.cpp
```diff
@@ -0,0 +1,80 @@
...
+ArrayView::objectAt(std::size_t idx) const
+{
+    ASSERT(idx < this->size(), "Object index is out of scope");
+    return ObjectView{prefix_, idx, clioConfig_};
+}
+
+Array::ArrayIterator
+ArrayView::beginValues() const
+{
+    ASSERT(clioConfig_.get().contains(prefix_), "Current string {} is a prefix, not a key of config", prefix_);
```
kuznetsss: `prefix_` couldn't be different from the one was got in constructor, so there is no point of checking this.

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include "util/newconfig/ValueView.hpp"
+
+#include <cstddef>
+#include <functional>
+#include <string>
+#include <string_view>
+
+namespace util::config {
+
+class ClioConfigDefinition;
```
kuznetsss: There should be either forward declaration or include.

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,102 @@
...
+    [[nodiscard]] size_t
+    size() const;
+
+    /**
+     * @brief Returns an iterator to the beginning of the values.
+     *
+     * @return Iterator to the beginning of the values.
+     */
+    [[nodiscard]] Array::ArrayIterator
+    beginValues() const;
```
kuznetsss: I think this method could be called just `begin()`. Btw what if there are objects but not values inside an Array?

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @brief Returns an iterator to the end of the values.
+     *
+     * @return Iterator to the end of the values.
+     */
+    [[nodiscard]] Array::Sentinel
```
kuznetsss: I think this should still return `ArrayIterator` and `Sentinel` should be hidden inside `ArrayIterator`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,174 @@
...
+      ConfigValue{ConfigType::Integer}.defaultValue(static_cast<uint32_t>(std::thread::hardware_concurrency()))},
+     {"database.cassandra.core_connections_per_host", ConfigValue{ConfigType::Integer}.defaultValue(1)},
+     {"database.cassandra.queue_size_io", ConfigValue{ConfigType::Integer}.optional()},
+     {"database.cassandra.write_batch_size", ConfigValue{ConfigType::Integer}.defaultValue(20)},
+     {"etl_source.[].ip", Array{ConfigValue{ConfigType::String}.optional()}},
+     {"etl_source.[].ws_port", Array{ConfigValue{ConfigType::String}.optional().min(1).max(65535)}},
+     {"etl_source.[].grpc_port", Array{ConfigValue{ConfigType::String}.optional().min(1).max(65535)}},
+     {"forwarding.cache_timeout", ConfigValue{ConfigType::Double}.defaultValue(0.0)},
+     {"forwarding.request_timeout", ConfigValue{ConfigType::Double}.defaultValue(10.0)},
+     {"dos_guard.[].whitelist", Array{ConfigValue{ConfigType::String}}},
```
kuznetsss: Probably should be `dos_guard.whitelist.[]`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,174 @@
...
+     {"api_version.min", ConfigValue{ConfigType::Integer}},
+     {"api_version.max", ConfigValue{ConfigType::Integer}}}
+};
+
+ObjectView
+ClioConfigDefinition::getObject(std::string_view prefix, std::optional<std::size_t> idx) const
+{
+    auto const prefixWithDot = std::string(prefix) + ".";
+    for (auto const& [mapKey, mapVal] : map_) {
+        if (mapKey.starts_with(prefixWithDot) && (std::holds_alternative<ConfigValue>(mapVal) || !idx.has_value()))
```
kuznetsss: I think there should be `&&` instead of `||` here.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,174 @@
...
+            return ObjectView{prefix, *this};
+
+        if (mapKey.starts_with(prefixWithDot) && std::holds_alternative<Array>(mapVal)) {
+            ASSERT(std::get<Array>(mapVal).size() > idx, "index provided is out of scope");
```
kuznetsss: I think I saw this check multiple times before ObjectView creation. It is better to move it to ObjectView's constructor.

src/util/newconfig/ConfigDefinition.cpp
kuznetsss: I don't think we should be able to do `getArray(etl_sources.[].ip)`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,174 @@
...
+    }
+    ASSERT(false, "Value of key {} is an Array, not an object", fullKey);
+    return ValueView{ConfigValue{}};
+}
+
+Array const&
+ClioConfigDefinition::atArray(std::string_view key) const
+{
+    ASSERT(map_.contains(key), "Current string {} is a prefix, not a key of config", key);
+    ASSERT(std::holds_alternative<Array>(map_.at(key)), "Value of {} is not an array", key);
```
kuznetsss: Better to get an iterator first and then use it. It will save two searches in the map.

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,187 @@
...
+    validate(ConfigFileInterface const& config) const;
+
+    /**
+     * @brief Generate markdown file of all the clio config descriptions
+     *
+     * @param configDescription The configuration description object
+     * @return An optional Error if generating markdown fails
+     */
+    std::expected<std::string, Error>
+    getMarkDown(ClioConfigDescription const& configDescription) const;
```
kuznetsss: I think It should be `getMarkdown`

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,187 @@
...
+    ArrayView
+    getArray(std::string_view prefix) const;
+
+    /**
+     * @brief Returns an iterator to the beginning of the configuration map.
+     *
+     * @return A constant iterator to the beginning of the map.
+     */
+    auto
+    begin() const
```
kuznetsss, line 177: Are these `begin()` and `end()` used?

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,219 @@
...
+    if constexpr (std::is_same_v<Type, int>) {
+        return ConfigType::Integer;
+    } else if constexpr (std::is_same_v<Type, std::string>) {
+        return ConfigType::String;
+    } else if constexpr (std::is_same_v<Type, double>) {
+        return ConfigType::Double;
+    } else if constexpr (std::is_same_v<Type, bool>) {
+        return ConfigType::Boolean;
+    } else {
+        static_assert(util::unsupportedType<Type>, "Wrong config type");
```
kuznetsss: I think if you change `unsupportedType` to `Unsupported` it would be more natural to read.

src/util/newconfig/ConfigValue.hpp
kuznetsss: Let's delete it then? 🙂

src/util/newconfig/ConfigValue.hpp
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * @brief Sets the config value as optional, meaning the user doesn't have to provide the value in their config
+     *
+     * @return Reference to this ConfigValue
+     */
+    constexpr ConfigValue&
+    optional()
+    {
+        optional_ = true;
+        ASSERT(!value_.has_value(), "value must not exist in optional");
```
kuznetsss: Value still could exist in optional I think.

src/util/newconfig/ValueView.cpp
```diff
@@ -0,0 +1,66 @@
...
+        return std::get<std::string>(configVal_.get().getValue());
+    throw std::bad_variant_access();
+}
+
+bool
+ValueView::asBool() const
+{
+    if (type() == ConfigType::Boolean && configVal_.get().hasValue())
+        return std::get<bool>(configVal_.get().getValue());
+    throw std::bad_variant_access();
```
kuznetsss: Probably better to use `ASSERT()` here

tests/unit/util/newconfig/ArrayViewTests.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+struct ArrayViewTest : testing::Test {
+    ClioConfigDefinition const configData = generateConfig();
+};
+
+TEST_F(ArrayViewTest, ArrayValueTest)
+{
+    ArrayView const arrVals = configData.getArray("array.[].sub");
+    auto valIt = arrVals.beginValues();
+    EXPECT_NEAR((*valIt++).asDouble(), 111.11, 1e-9);
```
kuznetsss: Better to create a constant with precision.

tests/unit/util/newconfig/ValueViewTests.cpp
kuznetsss: This is marked as resolved, but it is not fixed.

src/util/newconfig/ArrayView.cpp
PeterChen13579: Need to be kept as fullkey must exist in ClioConfig to fetch config values

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: This assert occurs once but kept here as we don't want to find the map value again in Object's constructor

src/util/newconfig/ConfigDefinition.hpp
PeterChen13579: Used in ObjectView for method "startsWithKey"

kazh719cs (approved):
if (auto const res = executor_.read(yield, schema_->selectObject, key, sequence); res) {
            if (auto const result = res->template get<Blob, std::uint32_t>(); result) {
                auto [_, seq] = result.value();
                return seq;
            }
            LOG(log_.debug()) << "Could not fetch ledger object sequence - no rows";

        } else {
            LOG(log_.error()) << "Could not fetch ledger object sequence: " << res.error();
        }

        return std::nullopt;
    }

    std::optional<TransactionAndMetadata>
    fetchTransaction(ripple::uint256 const& hash, boost::asio::yield_context yield) const override

src/util/newconfig/ConfigValue.hpp
PeterChen13579: Used in file to return empty configValue now: https://github.com/XRPLF/clio/pull/1544/files#diff-4e8d0eeb9ae15e14a017e6d321f3987f1c1e411fd58bb6eea9540f715328818eR178

src/util/newconfig/ArrayView.hpp
```diff
@@ -0,0 +1,203 @@
...
+        /**
+         * @brief Equality operator
+         *
+         * @param other Another ArrayIterator to compare
+         * @return true if iterators are equal, otherwise false
+         */
+        bool
+        operator==(ArrayIterator const& other) const
+        {
+            return index_ == other.index_;
```
kuznetsss: Probably you should check that they are pointing to the same array.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,209 @@
...
+{
+    auto const prefixWithDot = std::string(prefix) + ".";
+    for (auto const& [mapKey, mapVal] : map_) {
+        auto const hasPrefix = mapKey.starts_with(prefixWithDot);
+        if (idx.has_value() && hasPrefix && std::holds_alternative<Array>(mapVal)) {
+            ASSERT(std::get<Array>(mapVal).size() > idx.value(), "Index provided is out of scope");
+            // we want to support getObject("array") and getObject("array.[]"), so we check if "[]" exists
+            if (!prefix.contains("[]"))
+                return ObjectView{prefixWithDot + "[]", idx.value(), *this};
+            return ObjectView{prefix, idx.value(), *this};
```
kuznetsss, line 132: How is this possible when prefix is not an array, but `idx` has value?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,209 @@
...
+        }
+        if (hasPrefix && !idx.has_value() && !mapKey.contains(prefixWithDot + "[]")) {
+            ASSERT(!mapKey.contains(prefixWithDot + "[]"), "Key {} is an array, not an object", mapKey);
```
kuznetsss: It is always `true` here, because you are checking the same condition in `if`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,209 @@
...
+}
+
+ArrayView
+ClioConfigDefinition::getArray(std::string_view prefix) const
+{
+    auto key = checkForBracketsInArray(prefix);
+
+    for (auto const& [mapKey, mapVal] : map_) {
+        if (mapKey.starts_with(key)) {
+            ASSERT(std::holds_alternative<Array>(mapVal), "Trying to retrieve an Array when value is an object ");
```
kuznetsss: Value here could also be a value, not an object.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -0,0 +1,209 @@
...
+bool
+ClioConfigDefinition::contains(std::string_view key) const
+{
+    return map_.contains(key);
+}
+
+bool
+ClioConfigDefinition::startsWith(std::string_view key) const
+{
+    auto it = std::find_if(map_.begin(), map_.end(), [&key](auto const& pair) { return pair.first.starts_with(key); });
```
kuznetsss: You could use `std::ranges::find_if`.

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: In this case, prefix is always an array, (i.e in Clio Config should be defined as array.[].something), but we want to support getObject("array") and getObject("array.[]"), so I did the explicit check to see if "[]" exists, as prefix passed into objectView should be  "array.[]" not "array"

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /**
+     * @brief Used for all Array API's; check to make sure "[]" exists, if not, append to end
+     *
+     * @param key key to check for
+     * @return the key with "[]" appended to the end
+     */
+    static std::string
+    checkForBracketsInArray(std::string_view key)
```
kuznetsss: This method is modifying `key`, so I think better would be `addBracketsForArrayKey()`.

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -0,0 +1,127 @@
...
+     * @param key The key to look up the description for
+     * @return The description associated with the key, or "Not Found" if the key does not exist
+     */
+    static constexpr std::string_view
+    get(std::string_view key)
+    {
+        auto const itr = std::find_if(configDescription.begin(), configDescription.end(), [&key](auto const& v) {
+            return v.key == key;
+        });
+        ASSERT(itr != configDescription.end(), "key doesn't exist in config");
```
kuznetsss: This could be `static_assert` because it is compile time.

src/util/newconfig/ConfigFileInterface.hpp
PeterChen13579: Hmm.... so Githook format checker auto makes it without space line 🤔

src/util/newconfig/ObjectView.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+ArrayView
+ObjectView::getArray(std::string_view key) const
+{
+    auto fullKey = getFullKey(key);
+    if (!fullKey.contains(".[]"))
+        fullKey = fullKey + ".[]";
+
+    ASSERT(clioConfig_.get().startsWith(fullKey), "Key {} does not exist in object", fullKey);
```
kuznetsss: I think here `startsWith()` makes a reader think that it is something about `clioConfig_`. Probably better to call it something like`hasItemsWithPrefix()`.

src/util/newconfig/ArrayView.hpp
```diff
@@ -117,6 +117,7 @@ class ArrayView {
         bool
         operator==(ArrayIterator const& other) const
         {
+            ASSERT(this->arr_ == other.arr_, "Trying to compare two different arrays");
```
kuznetsss: It is probably not so critical error to compare two iterators of different arrays. I think it should be just in return. And it should compare the address of `arr_`, because currently it is true if you compare an iterator of an array with its clone.

src/util/newconfig/ConfigDescription.hpp
kuznetsss: It seems it is not so easy to do. Let's leave it as it is for now.

tests/unit/NewConfigTests.cpp
kuznetsss, line 1: This file should be named `ClioConfigDefinitionTests.cpp` and it should be placed to `tests/unit/util/newconfig/`.

src/util/newconfig/ArrayView.hpp
```diff
@@ -112,13 +112,12 @@ class ArrayView {
...
          * @param other Another ArrayIterator to compare
-         * @return true if iterators are equal, otherwise false
+         * @return true if iterators are pointing to the same ArrayView underneath, otherwise false
          */
         bool
         operator==(ArrayIterator const& other) const
         {
-            ASSERT(this->arr_ == other.arr_, "Trying to compare two different arrays");
-            return index_ == other.index_;
+            return (&(this->arr_) == &(other.arr_)) ? index_ == other.index_ : false;
```
kuznetsss:
```cpp
return &arr_ == &(other.arr_) and index_ == other.index_;
```

src/util/newconfig/ArrayView.hpp
```diff
@@ -112,13 +112,12 @@ class ArrayView {
          * @brief Equality operator
          *
          * @param other Another ArrayIterator to compare
-         * @return true if iterators are equal, otherwise false
+         * @return true if iterators are pointing to the same ArrayView underneath, otherwise false
```
kuznetsss: True if iterators are pointing to the same element in the same ArrayView object.

tests/unit/util/newconfig/ArrayViewTests.cpp
```diff
@@ -78,6 +78,10 @@ TEST_F(ArrayViewTest, IterateArray)
     EXPECT_EQ((*it++).asString(), "125.5.5.2");
     EXPECT_EQ((*it++).asString(), "204.2.2.2");
     EXPECT_EQ((it), arr.end<ValueView>());
+
+    auto itArray = configData.getArray("array.[].sub");
+    auto itDosguard = configData.getArray("dosguard.whitelist.[]");
+    ASSERT_FALSE(itArray.begin<ValueView>() == itDosguard.begin<ValueView>());
```
kuznetsss: `ASSERT_EQ` - assert equal.

tests/unit/util/newconfig/ArrayViewTests.cpp
```diff
@@ -78,6 +78,10 @@ TEST_F(ArrayViewTest, IterateArray)
     EXPECT_EQ((*it++).asString(), "125.5.5.2");
     EXPECT_EQ((*it++).asString(), "204.2.2.2");
     EXPECT_EQ((it), arr.end<ValueView>());
+
+    auto itArray = configData.getArray("array.[].sub");
```
kuznetsss: Better to make it as a separate test and also check that they are not always equal.

kuznetsss (dismissed):
👍

src/util/newconfig/Array.hpp
```diff
@@ -36,128 +39,26 @@ namespace util::config {
...
-        return ArrayIterator{elements_.data() + elements_.size()};
-    }
-
-    /** @brief Add ConfigValues to Array class
+     * @brief Add ConfigValues to Array class
      *
      * @param value The ConfigValue to add
      */
     void
     emplace_back(ConfigValue value)
```
godexsoft: Are we trying to keep the `std` names, despite this being part of Clio? I don't mind personally for this particular case but the rest of the codebase is camelCase and it looks a little out of place.

src/util/newconfig/ArrayView.hpp
```diff
@@ -41,7 +41,125 @@ namespace util::config {
...
+        /**
+         * @brief Prefix increment operator
+         *
+         * @return Reference to the incremented ArrayIterator
+         */
+        ArrayIterator&
+        operator++()
+        {
+            if (index_ < arr_.size())
+                index_++;
```
godexsoft: I'd still use `++index_` here (the compiler will likely optimize this but it's best to get into the habit of using the prefix form when you don't actually need the copy of the original value).

src/util/newconfig/ArrayView.hpp
```diff
@@ -41,7 +41,125 @@ namespace util::config {
...
+         * @brief Postfix increment operator
+         *
+         * @return Copy of the ArrayIterator before increment
+         */
+        ArrayIterator
+        operator++(int)
+        {
+            ArrayIterator temp = *this;
+            if (index_ < arr_.size())
+                index_++;
```
godexsoft: Same as above.

src/util/newconfig/ArrayView.hpp
```diff
@@ -66,45 +184,20 @@ class ArrayView {
...
-     */
-    auto
-    endValues() const
-    {
-        ASSERT(clioConfig_.map_.contains(prefix_), "Current string is prefix, not full key");
-        return std::get<Array>(clioConfig_.map_.at(prefix_)).end();
-    }
-
 private:
     std::string const prefix_;
```
godexsoft: We should stop using const on member variables.. i know it's a bit controversial but the reasoning is the same as for reference member variables (more or less). 

We should probably also enable this check: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -101,113 +108,102 @@ static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
      {"api_version.max", ConfigValue{ConfigType::Integer}}}
 };
 
+ClioConfigDefinition::ClioConfigDefinition(std::initializer_list<KeyValuePair> pair)
+{
+    for (auto const& p : pair) {
```
godexsoft: You should be able to make this slightly nicer by using `auto const& [key, value] : pair` and use the two variables below instead of `first` and `second`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -101,113 +108,102 @@ static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
...
+bool
+ClioConfigDefinition::contains(std::string_view key) const
+{
+    return map_.contains(key);
+}
+
+bool
+ClioConfigDefinition::hasItemsWithPrefix(std::string_view key) const
+{
+    auto it = std::ranges::find_if(map_, [&key](auto const& pair) { return pair.first.starts_with(key); });
```
godexsoft: Do we really need a mutable (or any) temporary here? 

It also could be possible to use `std::ranges::any_of` (https://compiler-explorer.com/z/W3hTjMPzM).
Unfortunately for some reason `std::ranges::contains` doesn't provide an overload that takes a user predicate 🤦 Would be the most suitable algorithm otherwise imo.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -101,113 +108,102 @@ static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
...
-    if (map_.contains(fullKey) && std::holds_alternative<ConfigValue>(map_.at(fullKey))) {
+    ASSERT(map_.contains(fullKey), "key {} does not exist in config", fullKey);
+    if (std::holds_alternative<ConfigValue>(map_.at(fullKey))) {
         return ValueView{std::get<ConfigValue>(map_.at(fullKey))};
     }
-    ASSERT(
-        !map_.contains(fullKey) && std::holds_alternative<Array>(map_.at(fullKey)), "value of key is not Config Value."
-    );
+    ASSERT(false, "Value of key {} is an Array, not an object", fullKey);
+    return ValueView{ConfigValue{}};
```
godexsoft: This can never be reached, right? there is `std::unreachable` for that. (https://compiler-explorer.com/z/1EGsMTaz5)

godexsoft: Keep in mind that if for some reason the ASSERT does not work and `std::unreachable` is actually reached it will lead to UB and a SEGFAULT in practice.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -101,113 +108,102 @@ static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
...
+}
 
-    throw std::invalid_argument("no matching key");
+ValueView
+ClioConfigDefinition::getValueInArray(std::string_view fullKey, std::size_t index) const
+{
+    auto it = this->getArrayIterator(fullKey);
```
godexsoft: You may or may not need the `this->` here (i keep forgetting the rules for that).
Also, i don't think we need a mutable iterator here and in most other places.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -101,113 +108,102 @@ static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
...
+ClioConfigDefinition::atArray(std::string_view fullKey) const
+{
+    auto it = this->getArrayIterator(fullKey);
+    return std::get<Array>(it->second);
+}
+
+std::size_t
+ClioConfigDefinition::arraySize(std::string_view prefix) const
+{
+    auto key = addBracketsForArrayKey(prefix);
```
godexsoft: `const`?

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * @brief All the config description are stored and extracted from this class
+ *
+ * Represents all the possible config description
+ */
+struct ClioConfigDescription {
+public:
+    /** @brief Struct to represent a key-value pair*/
+    struct KV {
```
godexsoft, line 38: Hmmm... i think i have seen this type somewhere. perhaps `std::pair`? 😃 
I don't mind rolling our own but you can also just do `using KV = std::pair<std::string_view, std::string_view>;`. The only downside would be the more generic names `first` and `second`.

src/util/newconfig/ObjectView.cpp
```diff
@@ -44,50 +44,59 @@ ObjectView::ObjectView(std::string_view prefix, std::size_t arrayIndex, ClioConf
...
-
-bool
-ObjectView::startsWithKey(std::string_view key) const
-{
-    for (auto const& [valueKey, value] : clioConfig_.map_) {
-        if (valueKey.starts_with(key))
-            return true;
-    }
-    return false;
+    auto const fullKey = getFullKey(key);
```
godexsoft: Sometimes it's not worth making a constant/variable (even though generated code will be the same). I'd just write `return clioConfig_.get().contains(getFullKey(key));` here to make this a one-liner. This is always going to be a case by case thing.

src/util/newconfig/ObjectView.cpp
```diff
@@ -44,50 +44,59 @@ ObjectView::ObjectView(std::string_view prefix, std::size_t arrayIndex, ClioConf
...
+    ASSERT(false, "Key {} does not exist in object", fullKey);
+    return ObjectView{"", clioConfig_};
+}
+
+ArrayView
+ObjectView::getArray(std::string_view key) const
+{
+    auto fullKey = getFullKey(key);
+    if (!fullKey.contains(".[]"))
+        fullKey = fullKey + ".[]";
```
godexsoft: Isn't this `+=`?

src/util/newconfig/ObjectView.cpp
```diff
@@ -44,50 +44,59 @@ ObjectView::ObjectView(std::string_view prefix, std::size_t arrayIndex, ClioConf
...
+std::string
+ObjectView::getFullKey(std::string_view key) const
+{
+    return fmt::format("{}.{}", prefix_, key);
+}
+
+bool
+ObjectView::startsWithKey(std::string_view key) const
+{
+    auto it = std::ranges::find_if(clioConfig_.get(), [&key](auto const& pair) { return pair.first.starts_with(key); });
```
godexsoft: Looks like a job for `std::ranges::any_of` again.

src/util/newconfig/ValueView.cpp
```diff
@@ -19,44 +19,63 @@
...
 std::string_view
 ValueView::asString() const
 {
-    if (this->type() == ConfigType::String)
-        return std::get<std::string>(configVal_.value_.value());
-    throw std::bad_variant_access();
+    if (this->type() == ConfigType::String && configVal_.get().hasValue())
+        return std::get<std::string>(configVal_.get().getValue());
+    ASSERT(false, "Value view is not of String type");
+    return "";
```
godexsoft: Unreachable

src/util/newconfig/ValueView.hpp
```diff
@@ -58,40 +61,86 @@ class ValueView {
...
+     * @throws std::bad_variant_access if the value is not of type int
      */
-    int
-    asInt() const;
+    template <typename T>
+    [[nodiscard]] T
+    asIntType() const
+    {
+        if ((type() == ConfigType::Integer) && configVal_.get().hasValue()) {
+            auto val = std::get<int64_t>(configVal_.get().getValue());
```
godexsoft: `const`

godexsoft (changes requested):
Overall this is a pretty cool system and i like the implementation 👍 
This could be already be shipped but since this is a learning experience for you I'm leaving a few more suggestions and comments.

You can still use `[[nodiscard]]` more as well as `const` is missing in many places where it could be added. 
It's also good to look for other algorithms than `find_if` - many times there will be some algo that's a better choice for the job at hand.

src/util/newconfig/ConfigDescription.hpp
kuznetsss: I think using pair in most cases decreases readability. So it's fine to define own structures.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -125,20 +140,33 @@ ClioConfigDefinition::getObject(std::string_view prefix, std::optional<std::size
...
         if (mapKey.starts_with(key)) {
-            ASSERT(std::holds_alternative<Array>(mapVal), "Trying to retrieve an Array when value is an object ");
+            ASSERT(
+                std::holds_alternative<Array>(mapVal), "Trying to retrieve Object or ConfigValue, instead of an Array "
+            );
             return ArrayView{key, *this};
         }
     }
     ASSERT(false, "Key {} is not found in config", key);
     return ArrayView{"", *this};
```
cindyyan317: Can't reach.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -150,24 +178,31 @@ ClioConfigDefinition::getValue(std::string_view fullKey) const
...
+ValueView
+ClioConfigDefinition::getValueInArray(std::string_view fullKey, std::size_t index) const
+{
+    auto it = this->getArrayIterator(fullKey);
+    return ValueView{std::get<Array>(it->second).at(index)};
+}
+
 Array const&
-ClioConfigDefinition::atArray(std::string_view key) const
+ClioConfigDefinition::atArray(std::string_view fullKey) const
```
cindyyan317: I guess the logic is more like "asArray" not "atArray"

godexsoft (approved):
Changes since my last review look good 👍 🚀

PR #1550, opened 2024-07-23
fix: nftData unique bug

cindyyan317:
Similar with this one. https://github.com/XRPLF/clio/pull/1542 
It failed to unique the NFT data as we wish. https://godbolt.org/z/jP384P7rb
It will write incorrect data if 
1 mint and burn happen in the same ledger , wrong data write to nf_tokens and uri tables.
2 accept nft offer and burn happen in the same ledger, the isburn maybe incorrect.

I am separating the function to NFTHelper to unittest it.

### DIFF
--- a/src/etl/NFTHelpers.cpp
+++ b/src/etl/NFTHelpers.cpp
@@ -356,4 +356,22 @@ getNFTDataFromObj(std::uint32_t const seq, std::string const& key, std::string c
 
     return nfts;
 }
+
+std::vector<NFTsData>
+getUniqueNFTsDatas(std::vector<NFTsData> const& nfts)
+{
+    std::vector<NFTsData> results = nfts;
+    // Remove all but the last NFTsData for each id. unique removes all but the first of a group, so we want to
+    // reverse sort by transaction index
+    std::sort(results.begin(), results.end(), [](NFTsData const& a, NFTsData const& b) {
+        return a.tokenID == b.tokenID ? a.transactionIndex > b.transactionIndex : a.tokenID > b.tokenID;
+    });
+
+    auto const last = std::unique(results.begin(), results.end(), [](NFTsData const& a, NFTsData const& b) {
+        return a.tokenID == b.tokenID;
+    });
+    results.erase(last, results.end());
+    return results;
+}
+
 }  // namespace etl
--- a/src/etl/NFTHelpers.cpp
+++ b/src/etl/NFTHelpers.cpp
@@ -295,7 +295,7 @@ getNFTokenCancelOfferData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx
     }
 
     // Deduplicate any transactions based on tokenID
-    std::ranges::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
+    std::ranges::sort(txs, [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
         return a.tokenID < b.tokenID;
     });
     auto last = std::unique(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
@@ -362,7 +362,7 @@ getUniqueNFTsDatas(std::vector<NFTsData> const& nfts)
 {
     std::vector<NFTsData> results = nfts;
 
-    std::ranges::sort(results.begin(), results.end(), [](NFTsData const& a, NFTsData const& b) {
+    std::ranges::sort(results, [](NFTsData const& a, NFTsData const& b) {
         return a.tokenID == b.tokenID ? a.transactionIndex > b.transactionIndex : a.tokenID > b.tokenID;
     });
 
--- a/src/etl/NFTHelpers.cpp
+++ b/src/etl/NFTHelpers.cpp
@@ -116,8 +116,8 @@ getNFTokenMintData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
         }
     }
 
-    std::ranges::sort(finalIDs.begin(), finalIDs.end());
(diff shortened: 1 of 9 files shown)


shawnxie999:
> 1 mint and burn happen in the same ledger , wrong data write to nf_tokens and uri tables.

IMO it should not matter. A NFT that was minted and burned in the ledger could be regarded as "non-existent".

> 2 accept nft offer and burn happen in the same ledger, the isburn maybe incorrect.

How would it be incorrect? the data from the burn will have the correct owner and isBurn.

(I haven't looked at the code yet)

cindyyan317:
> How would it be incorrect? the data from the burn will have the correct owner and isBurn.

As the test code shows,  https://godbolt.org/z/jP384P7rb, the sorted result is not what we expected. 
The 1:1 is sorted after 1:2 sometimes.  So the burn tx actually can get erased, so only the nftDatas from accept offer tx gets stored in DB. 

>IMO it should not matter. A NFT that was minted and burned in the ledger could be regarded as "non-existent".

Again, the sorted result is not expected. So the nftDatas from burn tx might get removed. 

```
std::sort requires a comparison with strict weak ordering and A<B, B<C, C<A violates that.

This violation incurs undefined behavior, and in practice, results in some of the worst kinds of undefined behavior.
```

I agreed the token minted and burned in the same ledger should be considered as "non-existent", however, current implementation will save the data into the DB , right? Maybe this is another bug we shall fix.

src/etl/NFTHelpers.cpp
```diff
@@ -356,4 +356,22 @@ getNFTDataFromObj(std::uint32_t const seq, std::string const& key, std::string c
 
     return nfts;
 }
+
+std::vector<NFTsData>
+getUniqueNFTsDatas(std::vector<NFTsData> const& nfts)
```
kuznetsss, line 361: Probably there is no point to pass it as const reference since you copy it anyway.

src/etl/NFTHelpers.cpp
```diff
@@ -356,4 +356,22 @@ getNFTDataFromObj(std::uint32_t const seq, std::string const& key, std::string c
...
+{
+    std::vector<NFTsData> results = nfts;
+    // Remove all but the last NFTsData for each id. unique removes all but the first of a group, so we want to
+    // reverse sort by transaction index
+    std::sort(results.begin(), results.end(), [](NFTsData const& a, NFTsData const& b) {
```
kuznetsss: `std::ranges::sort` would look a bit nicer.

kuznetsss (dismissed):
Looks good, just a few nits, but feel free to ignore.

src/etl/NFTHelpers.hpp
```diff
@@ -104,4 +104,14 @@ getNFTDataFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx);
 std::vector<NFTsData>
 getNFTDataFromObj(std::uint32_t seq, std::string const& key, std::string const& blob);
 
+/**
+ * @brief Get the unique NFTs data from a vector of NFTsData happening in the same ledger. For example, if a NFT has
+ both accept offer and burn happening in the same ledger,we only keep the final state of the NFT.
+
```
godexsoft, line 110: Something went wrong with the format here

src/etl/NFTHelpers.cpp
```diff
@@ -356,4 +356,22 @@ getNFTDataFromObj(std::uint32_t const seq, std::string const& key, std::string c
 
     return nfts;
 }
+
+std::vector<NFTsData>
+getUniqueNFTsDatas(std::vector<NFTsData> const& nfts)
+{
+    std::vector<NFTsData> results = nfts;
+    // Remove all but the last NFTsData for each id. unique removes all but the first of a group, so we want to
```
godexsoft: This could be a doxygen note instead perhaps? I mean, the code speaks for itself tbh

godexsoft: At the very least, it needs a line of space before the comment here or move the comment all the way up

godexsoft (changes requested):
Few small things on top of what others already noticed 👍

src/etl/NFTHelpers.cpp
cindyyan317: I think we can keep it. Just in case, if it is modified in function, we can get error from compiler.

src/etl/NFTHelpers.cpp
```diff
@@ -116,8 +116,8 @@ getNFTokenMintData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
         }
     }
 
-    std::sort(finalIDs.begin(), finalIDs.end());
-    std::sort(prevIDs.begin(), prevIDs.end());
+    std::ranges::sort(finalIDs.begin(), finalIDs.end());
```
kuznetsss: I thought for ranges you could do 
```cpp
std::ranges::sort(finalIDs);
```

src/etl/NFTHelpers.cpp
```diff
@@ -295,7 +295,7 @@ getNFTokenCancelOfferData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx
     }
 
     // Deduplicate any transactions based on tokenID
-    std::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
+    std::ranges::sort(txs.begin(), txs.end(), [](NFTTransactionsData const& a, NFTTransactionsData const& b) {
```
kuznetsss: Also here and below begin and end could be avoided.

kuznetsss (approved):
👍

shawnxie999 (approved):
Looks good. Thanks for fixing this! 🙇

kuznetsss (approved):

godexsoft (approved):

PR #1551, opened 2024-07-23
fix: Static linkage

### LINKED ISSUE #1507: Statically linked Clio is fragile
## Issue Description
When Clio is built with the `static` option, `glibc` is also statically linked. This worked until we introduced hostname resolving for dockerization. Seems like it's not enough to just statically link anymore.

## Suggested fix
We should move to App image instead of plain old static linkage.

Alternatively try to remove `-static-libgcc` and see if that works.

## Steps to Reproduce
Just running statically linked Clio is enough to observe the crash.

## Environment
Ubuntu linux 20.04 or 22.04.

godexsoft:
Fixes #1507 

Researched AppImage but it seems like it's a bit of an overkill for our use case.
This PR works for both clang and gcc. The only dynamically linked libraries are glibc and its bits. This means that when Clio is statically linked on an older system it will work on a newer one (tested on newer Ubuntu while linked on 20.04).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,28 +52,28 @@ jobs:
             build_type: Release
             conan_profile: gcc
             code_coverage: false
-            static: false
+            static: true
           - os: heavy
             container:
               image: rippleci/clio_ci:latest
             build_type: Debug
             conan_profile: gcc
             code_coverage: true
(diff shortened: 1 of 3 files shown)


kuznetsss (approved):
🚀

PR #1555, opened 2024-07-24
refactor: Refactor main

kuznetsss:
For #442.


src/app/CliArgs.cpp
```diff
@@ -0,0 +1,70 @@
...
+    // clang-format on
+    po::positional_options_description positional;
+    positional.add("conf", 1);
+
+    po::variables_map parsed;
+    po::store(po::command_line_parser(argc, argv).options(description).positional(positional).run(), parsed);
+    po::notify(parsed);
+
+    if (parsed.count("help") != 0u) {
+        std::cout << "Clio server " << util::build::getClioFullVersionString() << "\n\n" << description;
```
cindyyan317, line 57: why two '\n' here?

kuznetsss: I just copied it from main.  I assume the empty line separates version from options:
```
  $ clio_server --help
  Clio server clio-20240724174205-442_Refactor_main-47b31ed5

  Options:
    -h [ --help ]                         print help message and exit
    -v [ --version ]                      print version and exit
    -c [ --conf ] arg (=/etc/opt/clio/config.json)
                                          configuration file
```
I don't mind removing it if you think it will be better without it.

src/web/Server.cpp
```diff
@@ -0,0 +1,50 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
cindyyan317: 2024

src/web/Server.cpp
```diff
@@ -0,0 +1,50 @@
...
+makeServerSslContext(util::Config const& config)
+{
+    bool const configHasCertFile = config.contains("ssl_cert_file");
+    bool const configHasKeyFile = config.contains("ssl_key_file");
+
+    if (configHasCertFile != configHasKeyFile) {
+        return std::unexpected{"Config entries 'ssl_cert_file' and 'ssl_key_file' must be set or unset together."};
+    }
+
+    if (not configHasCertFile) {
```
cindyyan317: nit, should we remove {} here? I thought it is forced by clang-format

src/app/CliArgs.cpp
godexsoft: Definitely not better without it 🔢

cindyyan317 (dismissed):
Code looks good to me. But ut is failing.

src/web/Server.cpp
kuznetsss: Maybe clang-tidy would found it. Fixed.

cindyyan317 (dismissed):

src/app/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_library(clio_app)
+target_sources(clio_app PRIVATE CliArgs.cpp ClioApplication.cpp)
+
+target_link_libraries(clio_app PUBLIC clio_etl clio_feed clio_web clio_rpc)
+
+# target_compile_features(clio_app PUBLIC cxx_std_23)
```
godexsoft: I assume we don't need this line?

src/app/ClioApplication.hpp
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @brief Construct a new ClioApplication object
+     *
+     * @param config The configuration of the application
+     */
+    ClioApplication(util::Config const& config);
+
+    /**
+     * @brief Run the application
+-    *
```
godexsoft: Looks like a stray `-`

src/util/requests/impl/SslContext.cpp
```diff
@@ -82,9 +82,9 @@ getRootCertificate()
 }  // namespace
 
 std::expected<boost::asio::ssl::context, RequestError>
-makeSslContext()
+makeClientSslContext()
 {
-    ssl::context context{ssl::context::sslv23_client};
+    ssl::context context{ssl::context::tls_client};
```
godexsoft, line 87: Seems unrelated to main refactoring.. just wanted to confirm that this is tested and works fine?

src/web/Server.hpp
```diff
@@ -201,7 +211,7 @@ class Server : public std::enable_shared_from_this<Server<PlainSessionType, SslS
 
     util::Logger log_{"WebServer"};
     std::reference_wrapper<boost::asio::io_context> ioc_;
-    std::optional<std::reference_wrapper<boost::asio::ssl::context>> ctx_;
+    std::optional<boost::asio::ssl::context> ctx_;
```
godexsoft, line 214: Are we sure this works as expected? i imagine it was previously a reference for a reason?

godexsoft (changes requested):
Found some small things. Overall i like the changes 👍

src/app/CMakeLists.txt
godexsoft: I don't see it added anywhere else. Please check we really don't need it

src/util/requests/impl/SslContext.cpp
kuznetsss: I checked, works fine.

src/web/Server.hpp
kuznetsss: It was a reference to be able to create a context outside of Server. It was created in main and passed by reference. Now Server owns the context.

src/app/CMakeLists.txt
kuznetsss: It is defined for `clio_options` library, so I think here it was duplicated.

godexsoft (approved):

PR #1556, opened 2024-07-25
ci: Setup dependabot for github-actions

### LINKED ISSUE #1502: Setup dependabot for github actions
## Summary

Automatically update third-party github actions with dependabot.

## Motivation

Some third-party github actions may become outdated and it is convenient to receive notiffications (or maybe even PRs) about it.

godexsoft:
Fixes #1502

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
     reviewers:
       - "cindyyan317"
       - "godexsoft"
-      - "keznetsss"
+      - "kuznetsss"
     commit-message:
       prefix: "[CI] "
     target-branch: "develop"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "04:00"
+      timezone: "Etc/GMT"
+    reviewers:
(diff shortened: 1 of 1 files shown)


cindyyan317 (dismissed):

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "04:00"
+      timezone: "Etc/GMT"
+    reviewers:
+      - "cindyyan317"
+      - "godexsoft"
+      - "keznetsss"
```
kuznetsss: `kuznetsss`

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    commit-message:
+      prefix: "[CI] "
```
kuznetsss, line 15: It should follow conventional commits, so `ci: ` probably.

.github/dependabot.yml
godexsoft: How do i mess that up 🔢 🤷

.github/dependabot.yml
godexsoft: Notice this is not the title, it's the commit message - this is the same as we have for our clang-tidy commit message. The title of the PR should already be something like `chore:` or `ci:` and we can update it to what we like so dependabot can remember it for later PRs it opens.

kuznetsss (approved):

PR #1557, opened 2024-07-25
fix: Remove cassandra from log

### LINKED ISSUE #1452: Remove 'Cassandra' from Backend log
## Summary
Remove '**Cassandra**' from Backend log

## Motivation
We have many logs still using Cassandra
_Cassandra: Error setting_

_Cassandra sync write error, retrying_
Since we support both Cassandra / ScyllaDB, ScyllaDB user may find them misleading.

## Solution
Remove hardcoded _Cassandra_

cindyyan317:
Fix [#1452 ](https://github.com/XRPLF/clio/issues/1452)

### DIFF
--- a/src/data/BackendFactory.hpp
+++ b/src/data/BackendFactory.hpp
@@ -51,8 +51,7 @@ make_Backend(util::Config const& config)
     auto const type = config.value<std::string>("database.type");
     std::shared_ptr<BackendInterface> backend = nullptr;
 
-    // TODO: retire `cassandra-new` by next release after 2.0
-    if (boost::iequals(type, "cassandra") or boost::iequals(type, "cassandra-new")) {
+    if (boost::iequals(type, "cassandra")) {
         auto cfg = config.section("database." + type);
         backend = std::make_shared<data::cassandra::CassandraBackend>(data::cassandra::SettingsProvider{cfg}, readOnly);
     }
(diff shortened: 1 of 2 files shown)


src/data/BackendFactory.hpp
```diff
@@ -51,8 +51,7 @@ make_Backend(util::Config const& config)
     auto const type = config.value<std::string>("database.type");
     std::shared_ptr<BackendInterface> backend = nullptr;
 
-    // TODO: retire `cassandra-new` by next release after 2.0
-    if (boost::iequals(type, "cassandra") or boost::iequals(type, "cassandra-new")) {
+    if (boost::iequals(type, "cassandra")) {
```
kuznetsss, line 54: This is a breaking change for config. Are you sure it is ok to do that?

kuznetsss (approved):
LGTM, just one question.

src/data/BackendFactory.hpp
cindyyan317: 👍 Good call. 
I think "cassandra-new" is only used internally, when we compared the performance between scylladb and cassandra. Will check.

cindyyan317: Checked with @godexsoft . It is never announced, should be safe to delete.

PR #1587, opened 2024-08-07
fix: Use doxygen 1.12

### LINKED ISSUE #1431: Update doxygen to 1.12

godexsoft:
Fixes #1431 

Also enables back checking docs in pre-commit hook. This means that Clio devs should upgrade to doxygen 1.12 (already on brew).


kuznetsss (approved):
👍

PR #1588, opened 2024-08-07
refactor: Move interval timer into a separate class

kuznetsss:
For #442.


src/util/Repeat.cpp
```diff
@@ -17,17 +17,22 @@
...
-    std::reference_wrapper<std::size_t> callCount;
+namespace util {
 
-    void
-    operator()() const
-    {
-        ++(callCount.get());
-    }
-};
+Repeat::Repeat(boost::asio::io_context& ioc) : ioc_{ioc}, timer_(ioc_)
```
godexsoft: nit: it's generally recommended to avoid using member variables to initialize other member variables in the init list. in this case it's fine but perhaps it's better to `timer_(ioc)` since it's possible

src/util/Repeat.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <semaphore>
+
+namespace util {
+
+/**
+ * @brief A class to repeat some action at a regular interval
+ * @note io_context must be stopped before the Repeat object is destroyed. Otherwise it is undefined behavior
+ */
+class Repeat {
+    boost::asio::io_context& ioc_;
```
godexsoft: :hidethepain:

tests/unit/util/RepeatTests.cpp
```diff
@@ -0,0 +1,82 @@
...
+using testing::AtLeast;
+
+struct RepeatTests : SyncAsioContextTest {
+    Repeat repeat{ctx};
+    testing::StrictMock<testing::MockFunction<void()>> handlerMock;
+
+    void
+    withRunningContext(std::function<void()> func)
+    {
+        tests::common::util::withTimeout(std::chrono::seconds{1000}, [this, func = std::move(func)]() {
```
godexsoft, line 45: 1000 seconds timeout? 😃

godexsoft (changes requested):
Check the nits please. It's almost there 👍

godexsoft (approved):
LGTM :rage4:

PR #1589, opened 2024-08-07
fix: AccountNFT with invalid marker

### LINKED ISSUE #1497: "account_nfts" with unassociated marker is returning empty account_nfts instead of error.
## Issue Description

"account_nfts" with unassociated marker(i.e., marker from another method "account_channels") is not returning error.

Not sure if the response should error out or return empty account_nfts. As rippled is fixing the issue, will have to replicate it in Clio.

## Steps to Reproduce

Sample request:
`{
  "method": "account_nfts",
  "params": [
    {
        "account":"ra1egD7jzyZWhcDaPD1eFPKzdSMVQCTsXM",
        "marker": "10394D3EA937A85A4482D2071412A33C8582A6EB4C3B7FD59C76012D83AC08C6"
    }
  ]
}`

## Expected Result

Current response matches with rippled. If there's a fix for [https://github.com/XRPLF/rippled/issues/4314](url) which returns error, Clio should replicate it.

## Actual Result

`{
    "result": {
        "ledger_hash": "41560CD72E0E9B68BC8D9A8E261695392C22B3E115E0796119DC40B28C26E242",
        "ledger_index": 1936194,
        "validated": true,
        "account": "ra1egD7jzyZWhcDaPD1eFPKzdSMVQCTsXM",
        "account_nfts": [],
        "limit": 100,
        "status": "success"
    },
    "warnings": [
        {
            "id": 2001,
        }
    ]
}`

PeterChen13579:
Fixes [#1497](https://github.com/XRPLF/clio/issues/1497)
Mimics the behavior of the [fix on Rippled side](https://github.com/XRPLF/rippled/pull/5045)

### DIFF
--- a/src/rpc/handlers/AccountNFTs.cpp
+++ b/src/rpc/handlers/AccountNFTs.cpp
@@ -86,6 +86,10 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
     }
 
     std::optional<ripple::SLE const> page{ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, pageKey}};
+
+    if (page->getType() != ripple::ltNFTOKEN_PAGE)
+        return Error{Status{RippledError::rpcINVALID_PARAMS, "Marker matches Page ID from another Account"}};
+
     auto numPages = 0u;
 
     while (page) {
--- a/src/rpc/handlers/AccountNFTs.cpp
+++ b/src/rpc/handlers/AccountNFTs.cpp
@@ -40,6 +40,7 @@
 #include <xrpl/protocol/Serializer.h>
 #include <xrpl/protocol/jss.h>
 #include <xrpl/protocol/nft.h>
+#include <xrpl/protocol/nftPageMask.h>
 
 #include <cstdint>
 #include <optional>
@@ -78,8 +79,16 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
         input.marker ? ripple::uint256{input.marker->c_str()} : ripple::keylet::nftpage_max(*accountID).key;
     auto const blob = sharedPtrBackend_->fetchLedgerObject(pageKey, lgrInfo.seq, ctx.yield);
 
-    if (!blob)
+    bool pastMarker = !static_cast<bool>(input.marker);
+    bool markerFound = false;
+    bool const markerSet = input.marker.has_value();
+    auto const maskedMarker = pageKey & ripple::nft::pageMask;
+
+    if (!blob) {
+        if (markerSet)
+            return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidMarker"}};
         return response;
+    }
 
     std::optional<ripple::SLE const> page{ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, pageKey}};
     auto numPages = 0u;
@@ -89,6 +98,25 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
 
         for (auto const& nft : arr) {
             auto const nftokenID = nft[ripple::sfNFTokenID];
+            auto const maskedNftokenID = nftokenID & ripple::nft::pageMask;
+
+            if (!pastMarker) {
+                if (maskedNftokenID < maskedMarker)
+                    continue;
+
+                if (maskedNftokenID == maskedMarker && nftokenID < pageKey)
+                    continue;
+
+                if (nftokenID == pageKey) {
+                    markerFound = true;
+                    continue;
+                }
+            }
+
+            if (markerSet && !markerFound)
+                return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidMarker"}};
+
+            pastMarker = true;
 
             response.nfts.push_back(toBoostJson(nft.getJson(ripple::JsonOptions::none)));
             auto& obj = response.nfts.back().as_object();
@@ -117,6 +145,9 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
             page.reset();
         }
     }
+
+    if (markerSet && !markerFound)
+        return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidMarker"}};
 
     return response;
 }
--- a/src/rpc/handlers/AccountNFTs.cpp
+++ b/src/rpc/handlers/AccountNFTs.cpp
@@ -79,7 +79,7 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
     auto const blob = sharedPtrBackend_->fetchLedgerObject(pageKey, lgrInfo.seq, ctx.yield);
 
     if (!blob)
-        return Error{Status{RippledError::rpcINVALID_PARAMS, "invalidMarker"}};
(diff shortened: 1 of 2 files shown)


src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -78,8 +79,16 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
         input.marker ? ripple::uint256{input.marker->c_str()} : ripple::keylet::nftpage_max(*accountID).key;
     auto const blob = sharedPtrBackend_->fetchLedgerObject(pageKey, lgrInfo.seq, ctx.yield);
 
-    if (!blob)
+    bool pastMarker = !static_cast<bool>(input.marker);
```
godexsoft: `not input.marker.has_value()` reads better, wdyt?

src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -78,8 +79,16 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
...
+    bool markerFound = false;
+    bool const markerSet = input.marker.has_value();
```
godexsoft: can use `auto` here and above

src/rpc/handlers/AccountNFTs.cpp
cindyyan317: is it input.marker.has_value()?

godexsoft (changes requested):
Nice, just a few nits 👍

src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -89,6 +98,25 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
...
+            auto const maskedNftokenID = nftokenID & ripple::nft::pageMask;
+
+            if (!pastMarker) {
+                if (maskedNftokenID < maskedMarker)
+                    continue;
+
+                if (maskedNftokenID == maskedMarker && nftokenID < pageKey)
+                    continue;
+
+                if (nftokenID == pageKey) {
```
cindyyan317: i understood that this part of code came from rippled's code. However, Clio's implementation is slightly different.
In rippled, the marker is the nft id, however in Clio, the marker is page id. 
So it doesn't make sense to compare nfttokenId with pageKey here. 
The 'limit' works differently. for rippled, limit is the number of nftokens, for clio is the page number. According to the document , It should be page number.

cindyyan317 (changes requested):
Good stuff 👍 

I think the marker is different in Clio, comparing with rippled. So rippled's approach might not work on Clio. 
Also please check the document https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_nfts#account_nfts, I think Clio's limit is also different.

shawnxie999:
@cindyyan317 rippled can directly fetch a page based on the NFTokenID marker. But clio can't do this, it has to traverse from the last page. So I assume that's why the marker is the ID of the page itself. I think we can still try to return a NFTokenID as a marker, but it would require clio to traverse from the last token page everytime a marker is provided, which would make the API less efficient

src/rpc/handlers/AccountNFTs.cpp
kuznetsss: I would say `bool` is completely fine and maybe for the line 84 is even better than `auto`.

godexsoft: i don't mind either way. my preference would be `auto` though.

cindyyan317:
> @cindyyan317 rippled can directly fetch a page based on the NFTokenID marker. But clio can't do this, it has to traverse from the last page. So I assume that's why the marker is the ID of the page itself. I think we can still try to return a NFTokenID as a marker, but it would require clio to traverse from the last token page everytime a marker is provided, which would make the API less efficient

@shawnxie999 Thanks for the suggestion. According to the [doc](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_nfts#account_nfts) , the marker should be page id. We will flag this mismatch to rippled and discuss the correct fix.

For this PR, I think we don't mimic rippled. The marker is considered valid if it is a valid page object of account .

tests/unit/rpc/handlers/AccountNFTsTests.cpp
```diff
@@ -402,6 +403,42 @@ TEST_F(RPCAccountNFTsHandlerTest, Marker)
...
+    auto const ledgerHeader = CreateLedgerHeader(LEDGERHASH, MAXSEQ);
+    EXPECT_CALL(*backend, fetchLedgerBySequence).Times(1);
+    ON_CALL(*backend, fetchLedgerBySequence).WillByDefault(Return(ledgerHeader));
+
+    auto const accountObject = CreateAccountRootObject(ACCOUNT, 0, 1, 10, 2, TXNID, 3);
+    auto const accountID = GetAccountIDWithString(ACCOUNT);
+    ON_CALL(*backend, doFetchLedgerObject(ripple::keylet::account(accountID).key, 30, _))
+        .WillByDefault(Return(accountObject.getSerializer().peekData()));
+
+    auto const pageObject =
```
cindyyan317, line 474: I think when marker is invalid, handler will return error. These mock will not be used.

src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -79,7 +79,7 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
     auto const blob = sharedPtrBackend_->fetchLedgerObject(pageKey, lgrInfo.seq, ctx.yield);
 
     if (!blob)
```
cindyyan317: If the blob is empty or does not exist , maybe because this account does not have any NFT. In this case, the marker is not given, we should not return error. 
If the blob is not empty ,maybe its type is not NFT page. In this case, we should return error. So I think we need to check the type somewhere.

src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -77,11 +77,19 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
     auto const pageKey =
         input.marker ? ripple::uint256{input.marker->c_str()} : ripple::keylet::nftpage_max(*accountID).key;
     auto const blob = sharedPtrBackend_->fetchLedgerObject(pageKey, lgrInfo.seq, ctx.yield);
+    auto const markerSet = input.marker.has_value();
 
-    if (!blob)
+    if (!blob) {
+        if (markerSet)
```
cindyyan317: Since markerSet is used only inside the if, we may move 80 into if branch, or just use input.marker.has_value() here

src/rpc/handlers/AccountNFTs.cpp
```diff
@@ -77,11 +77,19 @@ AccountNFTsHandler::process(AccountNFTsHandler::Input input, Context const& ctx)
...
+            return Error{Status{RippledError::rpcINVALID_PARAMS, "Marker field does not match any valid Page ID"}};
         return response;
+    }
 
     std::optional<ripple::SLE const> page{ripple::SLE{ripple::SerialIter{blob->data(), blob->size()}, pageKey}};
+
+    if (page->getType() != ripple::ltNFTOKEN_PAGE)
```
cindyyan317, line 89: Worth adding an unittest to cover this line.

cindyyan317 (approved):
LGTM

godexsoft (approved):
👍

PR #1593, opened 2024-08-12
refactor: Clio Config

PeterChen13579:
Add constraint + parse json into Config
Second part of refactoring Clio Config; First PR found [here](https://github.com/XRPLF/clio/pull/1544)

Steps that are left to implement:
- Replacing all the places where we fetch config values (by using config.valueOr/MaybeValue) to instead get it from Config Definition
- Generate markdown file using Clio Config Description


src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace util::config {
+
+std::optional<Error>
+PortConstraint::checkConstraint(Constraint::ValueType const& port) const
+{
+    uint32_t p = 0;
```
cindyyan317: We can move this line after checkType

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+        return Error{"Port must be an integer or string"};
+
+    if (std::holds_alternative<std::string>(port)) {
+        p = static_cast<uint32_t>(std::stoi(std::get<std::string>(port)));
+    } else {
+        p = static_cast<uint32_t>(std::get<int64_t>(port));
+    }
+    if (p >= portMin && p <= portMax)
+        return std::nullopt;
+    return Error{"Port does not satisfy the constraint bounds"};
```
cindyyan317, line 59: nit: we can format the port number in the error message.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+ChannelNameConstraint::checkConstraint(Constraint::ValueType const& channel) const
+{
+    if (!checkType(channel))
+        return Error{"Channel name must be a string"};
+
+    if (std::ranges::any_of(channels, [&channel](std::string_view name) {
+            return std::get<std::string>(channel) == name;
+        }))
+        return std::nullopt;
+    return Error{"Channel name must be one of General, WebServer, Backend, RPC, ETL, Subscriptions, Performance"};
```
cindyyan317: Maybe format the error message by **channels**

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+#include <variant>
+
+namespace util::config {
+class ValueView;
+class ConfigValue;
+
+/**
+ * @brief specific values that are accepted for channel names in config
+ */
+static constexpr std::array<char const*, 7> channels = {
```
cindyyan317: Can we reuse the Logger::CHANNELS?

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+std::optional<Error>
+LogLevelNameConstraint::checkConstraint(Constraint::ValueType const& logger) const
+{
+    if (!checkType(logger))
+        return Error{"log_level must be a string"};
```
cindyyan317: log_level is the key of this config?

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+    checkType(ValueType const& type) const override
+    {
+        return std::holds_alternative<std::string>(type);
+    }
+};
+
+/**
+ * @brief A constraint to ensure the Cassandra name is valid.
+ */
+class CassandraName final : public Constraint {
```
cindyyan317: For expansibility, we can use something like DBTypeConstraint to check if the type is one of the candidates. Or use more general constraint like "OneOf" to check the value must be one of the provided values.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+        return std::nullopt;
+    return Error{"ip is not a valid ip address"};
+}
+
+std::optional<Error>
+CassandraName::checkConstraint(ValueType const& name) const
+{
+    if (checkType(name) && std::get<std::string>(name) == "cassandra")
+        return std::nullopt;
+    return Error{"database.type must be string Cassandra"};
```
cindyyan317: The key is "database.type" in our case. But Constraints should not know this information, right? I am thinking maybe send the real key to the Constraint and make the error more descriptive.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+std::optional<Error>
+LoadConstraint::checkConstraint(ValueType const& name) const
+{
+    if (!checkType(name))
+        return Error{"cache.load must be a string"};
+
+    auto const load = std::get<std::string>(name);
+    if (load == "sync" || load == "async" || load == "none")
```
cindyyan317: I have seen some Constraints share the common logic which is checking the value is one of the provided items. Maybe we can simplify them to one Constraint "OneOf"?

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+    return Error{"log_tag_style must be string int, uint, null, none, or uuid"};
+}
+
+std::optional<Error>
+APIVersionConstraint::checkConstraint(ValueType const& apiVersion) const
+{
+    if (!checkType(apiVersion))
+        return Error{"api_version must be an integer"};
+
+    if (std::get<int64_t>(apiVersion) <= API_VERSION_MAX && std::get<int64_t>(apiVersion) >= API_VERSION_MIN)
```
cindyyan317: This one also share the same logic with constraint of port. They can abstract to one constraint, like "Between". It is more readable when we use it .  
like 
`ConfigValue{ConfigType::Integer}.defaultValue(1u).withConstraint(Between(1,3))`

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+    "uuid",
+};
+
+/**
+ * @brief Minimum API version supported by this build
+ *
+ * Note: Clio does not natively support v1 and only supports v2 or newer.
+ * However, Clio will forward all v1 requests to rippled for backward compatibility.
+ */
+static constexpr uint32_t API_VERSION_MIN = 1u;
```
cindyyan317: For maintainability, we can somehow reuse API_VERSION_MIN in APIVersion.hpp?

src/util/newconfig/Array.hpp
```diff
@@ -21,33 +21,32 @@
...
      * @tparam Args Types of the arguments
-     * @param args Arguments to initialize the elements of the Array
+     * @param arg Argument to set the type and constraint of ConfigValues in Array
      */
-    template <typename... Args>
-    constexpr Array(Args&&... args) : elements_{std::forward<Args>(args)...}
+    template <typename Arg>
+    constexpr Array(Arg&& arg) : elements_{std::forward<Arg>(arg)}
     {
+        ASSERT(!elements_.at(0).hasValue(), "Array does not include default values");
```
kuznetsss: You should also check that there is at least one element in `elements_`.

src/util/newconfig/Array.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "util/newconfig/ConfigValue.hpp"
+
+#include <cstddef>
+#include <utility>
+#include <vector>
+
+namespace util::config {
+
+void
+Array::emplaceBack(ConfigValue value)
```
kuznetsss: Maybe it would be better to get here `ConfigValue::type` and put it into existing empty `ConfigValue` in `elements_` and then create there another empty `ConfigValue`?

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+std::optional<Error>
+PortConstraint::checkConstraint(Constraint::ValueType const& port) const
+{
+    uint32_t p = 0;
+    if (!checkType(port))
+        return Error{"Port must be an integer or string"};
+
+    if (std::holds_alternative<std::string>(port)) {
+        p = static_cast<uint32_t>(std::stoi(std::get<std::string>(port)));
```
kuznetsss: I think `std::stoi()` may throw an exception if string is invalid. You should handle this as well.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+    "RPC",
+    "ETL",
+    "Subscriptions",
+    "Performance",
+};
+
+/**
+ * @brief specific values that are accepted for logger levels in config
+ */
+static constexpr std::array<char const*, 7> logLevels = {
```
kuznetsss: Do we use this values somewhere in logger? Maybe we could reuse them?

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+ * @brief Maximum API version supported by this build
+ */
+static constexpr uint32_t API_VERSION_MAX = 2u;
+
+/**
+ * @brief A interface to enforce constraints on certain values within ClioConfigDefinition.
+ */
+class Constraint {
+public:
+    using ValueType = std::variant<int64_t, std::string, bool, double>;
```
kuznetsss: This duplicates `Type` from `ConfigValue`. Probably it would be better to move this definition to a separate header file and reuse everywhere.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+
+/**
+ * @brief Maximum API version supported by this build
+ */
+static constexpr uint32_t API_VERSION_MAX = 2u;
+
+/**
+ * @brief A interface to enforce constraints on certain values within ClioConfigDefinition.
+ */
+class Constraint {
```
kuznetsss, line 85: Every constraint should have a compile time description.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,401 @@
...
+     * @return true if the type is correct, false otherwise.
+     */
+    [[nodiscard]] bool
+    checkType(ValueType const& type) const override
+    {
+        return std::holds_alternative<int64_t>(type) || std::holds_alternative<std::string>(type);
+    }
+
+    uint32_t const portMin = 1;
+    uint32_t const portMax = 65535;
```
kuznetsss: This could be `static constexpr`.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <variant>
+
+namespace util::config {
+
+std::optional<Error>
+PortConstraint::checkConstraint(Constraint::ValueType const& port) const
+{
+    uint32_t p = 0;
+    if (!checkType(port))
+        return Error{"Port must be an integer or string"};
```
kuznetsss: To not duplicate everywhere this you could change a bit interface of `Constraint`:
```cpp
class Constraint {
public:
  std::optional<Error> check(value) {
     if (auto  maybeError = checkTypeImpl(value); maybeError.has_value())
        return std::move(maybeError).value();
     return checkValueImpl(value);
  }
protected:
  virtual std::optional<Error> checkTypeImpl(value) = 0;
  virtual std::optional<Error> checkValueImpl(value) = 0;
};
```

cindyyan317 (commented):
Maybe we can just use one general  constraint to constraint the port range, api range, instead of creating different constraint for each usage. For logger channels and levels, we can use one constraint (like OneOf) . I saw the similarity between the config constraint and the input validators, worth to check if anything we can borrow from validators.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -47,62 +53,69 @@ namespace util::config {
  * without default values must be present in the user's config file.
  */
 static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
-    {{"database.type", ConfigValue{ConfigType::String}.defaultValue("cassandra")},
+    {{"database.type", ConfigValue{ConfigType::String}.defaultValue("cassandra").withConstraint(validateCassandraName)},
      {"database.cassandra.contact_points", ConfigValue{ConfigType::String}.defaultValue("localhost")},
-     {"database.cassandra.port", ConfigValue{ConfigType::Integer}},
+     {"database.cassandra.port", ConfigValue{ConfigType::Integer}.withConstraint(validatePort)},
      {"database.cassandra.keyspace", ConfigValue{ConfigType::String}.defaultValue("clio")},
      {"database.cassandra.replication_factor", ConfigValue{ConfigType::Integer}.defaultValue(3u)},
```
kuznetsss, line 60: This should have a constraint checking it is a positive integer.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -47,62 +53,69 @@ namespace util::config {
...
      {"database.cassandra.table_prefix", ConfigValue{ConfigType::String}.defaultValue("table_prefix")},
      {"database.cassandra.max_write_requests_outstanding", ConfigValue{ConfigType::Integer}.defaultValue(10'000)},
      {"database.cassandra.max_read_requests_outstanding", ConfigValue{ConfigType::Integer}.defaultValue(100'000)},
      {"database.cassandra.threads",
       ConfigValue{ConfigType::Integer}.defaultValue(static_cast<uint32_t>(std::thread::hardware_concurrency()))},
```
kuznetsss: Also should be a positive integer.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -47,62 +53,69 @@ namespace util::config {
...
      {"forwarding.cache_timeout", ConfigValue{ConfigType::Double}.defaultValue(0.0)},
      {"forwarding.request_timeout", ConfigValue{ConfigType::Double}.defaultValue(10.0)},
      {"dos_guard.whitelist.[]", Array{ConfigValue{ConfigType::String}}},
-     {"dos_guard.max_fetches", ConfigValue{ConfigType::Integer}.defaultValue(1000'000)},
-     {"dos_guard.max_connections", ConfigValue{ConfigType::Integer}.defaultValue(20)},
-     {"dos_guard.max_requests", ConfigValue{ConfigType::Integer}.defaultValue(20)},
+     {"dos_guard.max_fetches", ConfigValue{ConfigType::Integer}.defaultValue(1000'000).withConstraint(ValidateUint32)},
+     {"dos_guard.max_connections", ConfigValue{ConfigType::Integer}.defaultValue(20).withConstraint(ValidateUint32)},
+     {"dos_guard.max_requests", ConfigValue{ConfigType::Integer}.defaultValue(20).withConstraint(ValidateUint32)},
      {"dos_guard.sweep_interval", ConfigValue{ConfigType::Double}.defaultValue(1.0)},
```
kuznetsss: Should check for a positive double with min value 0.1.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+std::optional<Error>
+ValidIPConstraint::checkConstraint(Constraint::ValueType const& ip) const
+{
+    if (!checkType(ip))
+        return Error{"ip must be a string"};
+
+    static std::regex const ipv4Regex(
+        R"(^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)"
+    );
+    if (std::regex_match(std::get<std::string>(ip), ipv4Regex))
```
kuznetsss: IP could also be a URL. URL couldn't start from a number https://stackoverflow.com/a/11749096.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -206,4 +219,62 @@ ClioConfigDefinition::arraySize(std::string_view prefix) const
...
+                    if (!(configVal.isOptional() || configVal.hasValue()))
+                        listOfErrors.emplace_back(key, " key is required in user Config");
+                }
+            }
+            continue;
+        }
+        ASSERT(
+            std::holds_alternative<ConfigValue>(value) || std::holds_alternative<Array>(value),
+            "Value must be of type ConfigValue or Array"
+        );
```
kuznetsss, line 249: Could it be something else here?

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -206,4 +219,62 @@ ClioConfigDefinition::arraySize(std::string_view prefix) const
...
+        std::visit(
+            util::OverloadSet{
+                [&key, &config, &listOfErrors](ConfigValue& val) {
+                    if (auto const setVal = val.setValue(config.getValue(key), key); setVal.has_value())
```
kuznetsss: Probably something like `error` or `maybeError` would be a better name for `setVal`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -206,4 +219,62 @@ ClioConfigDefinition::arraySize(std::string_view prefix) const
...
+                        auto const resultOfSettingValue = constraint.has_value()
+                            ? configVal.withConstraint(constraint.value()).setValue(val, key)
+                            : configVal.setValue(val, key);
+
+                        if (resultOfSettingValue.has_value()) {
+                            listOfErrors.emplace_back(resultOfSettingValue.value());
+                        } else {
+                            arr.emplaceBack(std::move(configVal));
+                        }
+                    }
```
kuznetsss: I think all of this should be hidden inside `Array`. Here it should be something similar to how you set value for a `ConfigValue`.

src/util/newconfig/ConfigFileInterface.hpp
```diff
@@ -42,25 +44,34 @@ class ConfigFileInterface {
      * @param filePath The path to the Clio Config data
      */
     virtual void
-    parse(std::string_view filePath) = 0;
+    parse(boost::filesystem::path filePath) = 0;
```
kuznetsss: Why `boost::filesystem`?

src/util/newconfig/ConfigFileInterface.hpp
```diff
@@ -42,25 +44,34 @@ class ConfigFileInterface {
...
-     * @brief Retrieves a configuration value.
+     * @brief Retrieves the value of configValue.
      *
-     * @param key The key of the configuration value.
-     * @return An optional containing the configuration value if found, otherwise std::nullopt.
+     * @param key The key of configuration.
+     * @return the value assosiated with key.
      */
-    virtual std::optional<ConfigValue>
+    virtual std::variant<int64_t, std::string, bool, double>
```
kuznetsss: Should be `ConfigValue::Type` to no duplicate the same type.

src/util/newconfig/ConfigFileInterface.hpp
```diff
@@ -42,25 +44,34 @@ class ConfigFileInterface {
...
 
     /**
      * @brief Retrieves an array of configuration values.
      *
      * @param key The key of the configuration array.
-     * @return An optional containing a vector of configuration values if found, otherwise std::nullopt.
+     * @return A vector of configuration values if found, otherwise std::nullopt.
      */
-    virtual std::optional<std::vector<ConfigValue>>
+    virtual std::vector<std::variant<int64_t, std::string, bool, double>>
```
kuznetsss: Duplication of type here as well.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,132 @@
...
+        if (in) {
+            std::stringstream contents;
+            contents << in.rdbuf();
+            auto opts = boost::json::parse_options{};
+            opts.allow_comments = true;
+            auto const tempObj = boost::json::parse(contents.str(), {}, opts).as_object();
+            flattenJson(tempObj, "");
+        }
+    } catch (std::exception const& e) {
+        LOG(util::LogService::error()) << "Could not read configuration file from '" << configFilePath.string()
```
kuznetsss: I think it should return an `Error` instead. Because a user couldn't know now whether parsing was successful or not.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+bool
+ConfigFileJson::containsKey(std::string_view key) const
+{
+    return jsonObject_.contains(key);
+}
+
+void
+ConfigFileJson::flattenJson(boost::json::object const& obj, std::string const& prefix)
```
kuznetsss, line 133: Does this method create additional json values inside parsed json with keys from config definition?

src/util/newconfig/ConfigFileJson.hpp
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @brief Extracts the value from a JSON object and converts it into the corresponding type.
+     *
+     * @param jsonValue The JSON value to extract.
+     * @return A variant containing the same type corresponding to the extracted value.
+     */
+    [[nodiscard]] static std::variant<int64_t, std::string, bool, double>
+    extractJsonValue(boost::json::value const& jsonValue)
+    {
+        std::variant<int64_t, std::string, bool, double> variantValue;
```
kuznetsss: Implementation could be moved to cpp file.

src/util/newconfig/ConfigValue.hpp
```diff
@@ -83,47 +63,101 @@ class ConfigValue {
...
+     * @param key The Config key associated with the value. Optional to include; Used for debugging message to user.
+     * @return optional Error if user tries to set a value of wrong type or not within a constraint
      */
-    [[nodiscard]] constexpr ConfigType
-    type() const
+    [[nodiscard]] std::optional<Error>
+    setValue(ValueType value, std::optional<std::string_view> key = std::nullopt)
     {
-        return type_;
+        auto err = checkTypeConsistency(type_, value);
```
kuznetsss, line 82: Do you need to check type in constraints while there is the check here?

src/util/newconfig/ConfigValue.hpp
```diff
@@ -83,47 +63,101 @@ class ConfigValue {
...
+     * @brief Retrieves the constraint associated with this ConfigValue, if any.
      *
-     * @param max The maximum value
-     * @return Reference to this ConfigValue
+     * @return An optional reference to the associated Constraint.
      */
-    [[nodiscard]] constexpr ConfigValue&
-    max(std::uint32_t max)
+    [[nodiscard]] std::optional<std::reference_wrapper<Constraint const>>
+    getConstraint() const
```
kuznetsss, line 145: Why do you need to this method?

src/util/newconfig/ConfigValue.hpp
```diff
@@ -178,39 +212,28 @@ class ConfigValue {
...
     {
-        if (std::holds_alternative<std::string>(value)) {
-            ASSERT(type == ConfigType::String, "Value does not match type string");
-        } else if (std::holds_alternative<bool>(value)) {
-            ASSERT(type == ConfigType::Boolean, "Value does not match type boolean");
-        } else if (std::holds_alternative<double>(value)) {
-            ASSERT(type == ConfigType::Double, "Value does not match type double");
-        } else if (std::holds_alternative<int64_t>(value)) {
-            ASSERT(type == ConfigType::Integer, "Value does not match type integer");
+        if (type == ConfigType::String && !std::holds_alternative<std::string>(value)) {
```
kuznetsss, line 216: Why is it only for string now?

src/util/newconfig/Errors.hpp
```diff
@@ -24,10 +24,28 @@
...
+    {
+    }
+
+    /**
+     * @brief Constructs an Error with a custom error message.
+     *
+     * @param key the key associated with the error.
+     * @param err the error message to display to users.
+     */
+    Error(std::string_view key, std::string_view err) : error{std::string(key) + std::string(err)}
```
kuznetsss: Better to use `fmt::format()` instead of `+` to not create 2 extra strings.

src/util/newconfig/Array.cpp
godexsoft: Yea, this code is dangerous.. we should try and avoid relying on `front()` existing in the array. Also this is not an `emplace` :)

src/util/newconfig/ConfigConstraints.cpp
godexsoft: We should definitely avoid hardcoding the list here. If we add a new channel, this message will have to be updated too. Not very future proof as it's super easy to forget (or to not even know this error message exists)

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,137 @@
...
+LogLevelNameConstraint::checkConstraint(Constraint::ValueType const& logger) const
+{
+    if (!checkType(logger))
+        return Error{"log_level must be a string"};
+
+    if (std::ranges::any_of(logLevels, [&logger](std::string_view name) {
+            return std::get<std::string>(logger) == name;
+        }))
+        return std::nullopt;
+    return Error{"log_level must be one of trace, debug, info, warning, error, fatal, count"};
```
godexsoft: This is slightly less of an issue but still the same idea of avoiding hardcoding the list applies.

src/util/newconfig/ConfigFileInterface.hpp
godexsoft: @kuznetsss means that there is `std::filesystem` already and we should use that instead 👍

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: validatePort itself checks if value is a positive integer 👍

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: this one is actually not used anywhere in logger. Only CHANNELS are and I fixed it 👍

src/util/newconfig/ConfigConstraints.cpp
PeterChen13579: yea, I'll explicitly state that it's key 👍

src/util/newconfig/ConfigValue.hpp
PeterChen13579: This check is still there for all other supported types. I think github is being weird

src/util/newconfig/ConfigValue.hpp
PeterChen13579: I think so, because constraints applies to only a specific type, so if I do ConfigValue{Boolean}.withConstraint{port}
would be invalid

src/util/newconfig/ConfigFileJson.cpp
PeterChen13579: It only creates key's of what the user has in their own local config. If some keys do not match clioConfig, parse will catch it and error out.

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: I think I'm going to leave it for next PR, as that'll include generating the compile time description 👍

src/util/newconfig/ConfigValue.hpp
PeterChen13579: Used for arrays to addValues. https://github.com/XRPLF/clio/pull/1593/files#diff-f36b5e571fae890137e9f29ef7b22765535aebb946ae552c43e36bf344bf3a22R46

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: I don't think so, unless in the future we decide to support something other than ConfigValue or Arrays

src/util/newconfig/ConfigConstraints.cpp
PeterChen13579: Thanks for the suggestions!! After some discussion, I think I'm going to stick with current impl for constraints. The `Between(1,3)` looks fine, but if it is for port, or uint16, it's going to look like `withConstraint(Between(0, 65535)), `which is hard to decipher what the constraint means. 
I think for `oneOf()` It's going to look like `oneOf("trace", "debug", "info"....)` which makes Clio Config look quite long and if there another config value that require `log_levels` constraint, going to be hard to maintain. 
Let me know what you think!

src/util/newconfig/ConfigFileJson.cpp
PeterChen13579: I don't think returning error here makes sense as this function is called in constructor. WDYT about a "Throw" instead?

src/util/newconfig/Array.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    auto const firstVal = elements_.at(0);
+    auto const constraint = firstVal.getConstraint();
+
+    auto newElem = constraint.has_value() ? ConfigValue{firstVal.type()}.withConstraint(constraint->get())
+                                          : ConfigValue{firstVal.type()};
+    if (auto const maybeError = newElem.setValue(value, key); maybeError.has_value())
+        return maybeError;
+    elements_.emplace_back(std::move(newElem));
+    return std::nullopt;
```
kuznetsss: Maybe what you can do here to simplify the logic is to decide that each `Array` has an element without a value in the end at any time (it should be covered by an assert). Then inserting algorithm will be:
- Insert into `elements_` a copy of the last element
- Set `value` to the second from back element

But it is up to you because it may make other methods more complicated.

kuznetsss: Or another probably better idea is to put the value from constructor not into `elements_` but in a separate field `itemPattern_`. Then when inserting you just would need to copy `itemPattern_` into elements and set value for it.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+     *
+     * @param val The value to be checked
+     * @return An Error object if the constraint is not met, nullopt otherwise
+     */
+    [[nodiscard]]
+    std::optional<Error>
+    checkConstraint(Value const& val) const
+    {
+        if (auto maybeError = checkTypeImpl(val); maybeError.has_value())
+            return std::move(maybeError).value();
```
kuznetsss: I think you can do here just `return maybeError;` because `maybeError` has the same type as return value.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+    checkTypeImpl(Value const& val) const = 0;
+
+    /**
+     * @brief Check if the value is within the constraint.
+     *
+     * @param val The value type to be checked
+     * @return An Error object if the constraint is not met, nullopt otherwise
+     */
+    virtual std::optional<Error>
+    checkValueImpl(Value const& val) const = 0;
```
kuznetsss, line 150: This should be `protected` I think. Doesn't make much difference but I think it would additionally reflect that these methods are available for inherited classes. While in inherited classes they could stay private.

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+static constexpr LogLevelNameConstraint const validateLogLevelName{};
+static constexpr ValidIPConstraint const validateIP{};
+static constexpr CassandraName const validateCassandraName{};
+static constexpr LoadConstraint const validateLoadMode{};
+static constexpr LogTagStyle const validateLogTag{};
+static constexpr APIVersionConstraint const validateApiVersion{};
+static constexpr PositiveNumConstraint<uint16_t> const ValidateUint16{};
+static constexpr PositiveNumConstraint<uint32_t> const ValidateUint32{};
+static constexpr PositiveNumConstraint<uint64_t> const ValidateUint64{};
+static constexpr PositiveDouble const ValidatePositiveDouble{};
```
kuznetsss: You could remove number of classes a bit by having a `NumberValueConstraint`:
```cpp
template <typename T>
class NumberValueConstraint {
public:
    NumberValueConstraint(T min, T max);
   ...
};
```

And then using this class you could define any constraint checking a number value:
```cpp
static constexpr NumberValueConstraint<uint16_t> const validateUint16{std::numeric_limits<uint16_t>::min(), std::numeric_limits<uint16_t>::max()};
```

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+
+static constexpr PortConstraint const validatePort{};
+static constexpr ChannelNameConstraint const validateChannelName{};
+static constexpr LogLevelNameConstraint const validateLogLevelName{};
+static constexpr ValidIPConstraint const validateIP{};
+static constexpr CassandraName const validateCassandraName{};
+static constexpr LoadConstraint const validateLoadMode{};
+static constexpr LogTagStyle const validateLogTag{};
+static constexpr APIVersionConstraint const validateApiVersion{};
+static constexpr PositiveNumConstraint<uint16_t> const ValidateUint16{};
```
kuznetsss: All variables should start with a lower case letter

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+
+/**
+ * @brief An interface to enforce constraints on certain values within ClioConfigDefinition.
+ */
+class Constraint {
+public:
+    // using "{}" instead of = default because of gcc bug.
+    // see here for more info:
+    // https://stackoverflow.com/questions/72835571/constexpr-c-error-destructor-used-before-its-definition
+    constexpr virtual ~Constraint() noexcept {};
```
kuznetsss, line 91: Should the destructor be `constexpr`? Anyway all of them are static objects.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+std::optional<Error>
+ChannelNameConstraint::checkValueImpl(Value const& channelName) const
+{
+    if (std::ranges::any_of(Logger::CHANNELS, [&channelName](std::string_view name) {
+            return std::get<std::string>(channelName) == name;
+        }))
+        return std::nullopt;
+    return Error{makeErrorMsg("channel", Logger::CHANNELS)};
```
kuznetsss: I think it would make a bit more obvious for a user what is wrong if you include incorrect value in error message.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -47,62 +53,76 @@ namespace util::config {
...
+          .defaultValue(static_cast<uint32_t>(std::thread::hardware_concurrency()))
+          .withConstraint(ValidateUint32)},
+     {"database.cassandra.core_connections_per_host",
+      ConfigValue{ConfigType::Integer}.defaultValue(1).withConstraint(ValidateUint16)},
+     {"database.cassandra.queue_size_io", ConfigValue{ConfigType::Integer}.optional().withConstraint(ValidateUint16)},
+     {"database.cassandra.write_batch_size",
+      ConfigValue{ConfigType::Integer}.defaultValue(20).withConstraint(ValidateUint16)},
+     {"etl_source.[].ip", Array{ConfigValue{ConfigType::String}.optional().withConstraint(validateIP)}},
+     {"etl_source.[].ws_port", Array{ConfigValue{ConfigType::String}.optional().withConstraint(validatePort)}},
+     {"etl_source.[].grpc_port", Array{ConfigValue{ConfigType::String}.optional().withConstraint(validatePort)}},
```
kuznetsss: Why are they optional? Is an entry of etl source config with only one of them valid?

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -90,7 +90,9 @@ struct ClioConfigDescription {
         KV{"server.ip", "IP address of the Clio HTTP server."},
         KV{"server.port", "Port number of the Clio HTTP server."},
         KV{"server.max_queue_size", "Maximum size of the server's request queue."},
+        KV{"server.workers", "Maximum number of threads for server to run with."},
         KV{"server.local_admin", "Indicates if the server should run with admin privileges."},
+        KV{"server.admin_password", "Password required to run Clio Server with admin privileges"},
```
kuznetsss: Better `Password for Clio admin-only APIs`.

src/util/newconfig/ConfigFileJson.hpp
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * @brief Construct a new ConfigJson object and stores the values from
+     * user's config in the object.
+     *
+     * @param filePath the path to the Json file to parse.
+     */
+    ConfigFileJson(std::string_view filePath)
+    {
+        parse(filePath);
+    }
```
kuznetsss: Move to cpp please.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <string_view>
+#include <vector>
+
+namespace util::config {
+
+void
+ConfigFileJson::parse(boost::filesystem::path configFilePath)
+{
+    try {
+        std::ifstream const in(configFilePath.string(), std::ios::in | std::ios::binary);
```
kuznetsss: Does boost::json require to read files as binaries?

src/util/newconfig/ConfigFileJson.cpp
kuznetsss: You could change constructor of `ConfigFileJson` to accept `boost::json::object` and also create a separate function:
```cpp
std::expected<ConfigFileJson, Error> make_ConfigFileJson(filename);
```

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+/**
+ * @brief Extracts the value from a JSON object and converts it into the corresponding type.
+ *
+ * @param jsonValue The JSON value to extract.
+ * @return A variant containing the same type corresponding to the extracted value.
+ */
+[[nodiscard]] static Value
```
kuznetsss: Instead of making this function `static` you could create an anonymous namespace `namespace {` on top of this file and move this function there. Putting a function into anonymous namespace in cpp file makes it static. But maybe it is just my habit. @godexsoft, WDYT?

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,157 @@
...
+        variantValue = jsonValue.as_string().c_str();
+    } else if (jsonValue.is_bool()) {
+        variantValue = jsonValue.as_bool();
+    } else if (jsonValue.is_double()) {
+        variantValue = jsonValue.as_double();
+    }
+    return variantValue;
+}
+
+Value
```
kuznetsss, line 103: Maybe make this returning `std::expected<Value, Error>` instead of checking whether key exists each time? And the same with `getArray`. With such case it is not possible to forget to check whether a key exists.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,157 @@
...
+ConfigFileJson::containsKey(std::string_view key) const
+{
+    return jsonObject_.contains(key);
+}
+
+void
+ConfigFileJson::flattenJson(boost::json::object const& obj, std::string const& prefix)
+{
+    for (auto const& [key, value] : obj) {
+        std::string fullKey = prefix.empty() ? std::string(key) : prefix + "." + std::string(key);
```
kuznetsss: `fmt::format()` would be better for string concatenation here.

src/util/newconfig/ConfigValue.hpp
kuznetsss: > ConfigValue{Boolean}.withConstraint{port}

I think it is possible to make a compilation error for such case.

What I mean is you could call `checkConstraint()` and it will return an error if the type is wrong.

src/util/newconfig/ConfigValue.hpp
```diff
@@ -83,47 +61,101 @@ class ConfigValue {
...
      */
     [[nodiscard]] ConfigValue&
-    defaultValue(Type value)
+    defaultValue(Value value)
     {
-        setValue(value);
+        auto const err = checkTypeConsistency(type_, value);
+        ASSERT(!err.has_value(), "{}", err->error);
+        value_ = value;
         return *this;
```
kuznetsss, line 69: Why? I thought `setValue()` was enough.

src/util/newconfig/ConfigValue.hpp
kuznetsss: I see the other checks are below.

src/util/newconfig/Errors.hpp
```diff
@@ -19,15 +19,39 @@
```
kuznetsss, line 19: This file should be called `Error.hpp`.

tests/unit/util/newconfig/ArrayTests.cpp
```diff
@@ -18,33 +18,57 @@
...
 
 #include <gtest/gtest.h>
 
+#include <algorithm>
+#include <cstdint>
+#include <vector>
+
 using namespace util::config;
 
 TEST(ArrayTest, testConfigArray)
```
kuznetsss: Probably better to separate this test to some specific test cases. Ideally you should test each method separately (of course it is not always possible).

tests/unit/util/newconfig/ConfigValueTests.cpp
```diff
@@ -17,24 +17,188 @@
...
+        "Key \"log_level\"'s value must be one of the following: trace, debug, info, warning, error, fatal, count"
+    );
+    EXPECT_FALSE(logLevelChannel.setValue("count"));
+}
+
+TEST(ConfigValue, ipConstraint)
+{
+    auto ip = ConfigValue{ConfigType::String}.defaultValue("127.0.0.1").withConstraint(validateIP);
+    EXPECT_FALSE(ip.setValue("http://127.0.0.1").has_value());
+    EXPECT_FALSE(ip.setValue("http://127.0.0.1.com").has_value());
```
kuznetsss, line 112: We should support URL (e.g. `some_server.local`) as well. But maybe you could change it as a separate PR.

tests/unit/util/newconfig/ConfigValueTests.cpp
```diff
@@ -17,24 +17,188 @@
...
+    auto const err4 = ip.setValue("123.44");
+    EXPECT_TRUE(err4.has_value());
+    EXPECT_EQ(err4->error, "ip is not a valid ip address");
+    EXPECT_FALSE(ip.setValue("126.0.0.2"));
```
kuznetsss, line 116: There is no test for:
- more than 4 `.` in ip
- numbers are higher that 255
- empty string

tests/unit/util/newconfig/ConfigValueTests.cpp
```diff
@@ -17,24 +17,188 @@
...
+}
+
+TEST(ConfigValue, databaseTypeConstraint)
+{
+    auto cassandra = ConfigValue{ConfigType::String}.defaultValue("cassandra").withConstraint(validateCassandraName);
```
kuznetsss: You should test constraints by directly calling their methods.

tests/unit/util/newconfig/ConfigValueTests.cpp
```diff
@@ -17,24 +17,188 @@
...
+    testConstraintDeath(validateIP);
+    testConstraintDeath(validateCassandraName);
+    testConstraintDeath(validateLoadMode);
+    testConstraintDeath(validateLogTag);
+    testConstraintDeath(validateApiVersion);
+    testConstraintDeath(ValidateUint16);
+    testConstraintDeath(ValidateUint32);
+    testConstraintDeath(ValidateUint64);
+    testConstraintDeath(ValidatePositiveDouble);
+}
```
kuznetsss: You could create a parameterized test instead.
https://github.com/google/googletest/blob/main/docs/advanced.md#value-parameterized-tests

tests/unit/util/newconfig/JsonFileTests.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
```
kuznetsss: File name should be the same as the file you are testing with `Tests` suffix.

tests/unit/util/newconfig/JsonFileTests.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <cstdint>
+#include <string>
+#include <variant>
+#include <vector>
+
+struct ParseJson : testing::Test {
+    ParseJson() : jsonFileObj{TmpFile(JSONData).path}
```
kuznetsss: It is not a good idea to create a temporary file, because if some test crashes the temporary file will stay somewhere.

src/util/newconfig/Array.cpp
PeterChen13579: oh yea, I definitely think the second idea is good 👍

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: Based on the stack-overflow, I think so, but I'll try a commit without it and see if it passes the gcc build

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: Yea. Currently in the codebase, having any one of them, (or combinations of them) is valid. (Tested it running Clio locally). Perhaps we can change the code such that all 3 fields must be present (or all 3 not present)? WDYT?
Screenshot 2024-09-06 at 10 04 21 AM

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,470 @@
...
+     * @param arr The array with hard-coded values to add to error message
+     * @return The error message specifying what the value of key must be
+     */
+    template <std::size_t arrSize>
+    constexpr std::string
+    makeErrorMsg(std::string_view key, std::array<char const*, arrSize> arr) const
+    {
+        auto errorMsg = fmt::format("Key \"{}\"'s value must be one of the following: ", key);
+        for (auto const elem : arr)
+            errorMsg += fmt::format("{}, ", elem);
```
cindyyan317: Maybe ` fmt::join(arr, ",")` does the trick.

src/util/newconfig/Array.cpp
cindyyan317: Yes. Second option makes more sense. It is kinda hard to understand why size() is not 0 when nothing gets inserted.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <string_view>
+#include <variant>
+
+namespace util::config {
+
+std::optional<Error>
+PortConstraint::checkTypeImpl(Value const& port) const
+{
+    if (!(std::holds_alternative<int64_t>(port) || std::holds_alternative<std::string>(port)))
+        return Error{"Port must be a string or Integer"};
```
cindyyan317: nit: Integer 's I

src/util/newconfig/ConfigFileJson.cpp
PeterChen13579: This is what I am finding, but @godexsoft if you can confirm that'll be appreciated 🙇 
Screenshot 2024-09-06 at 2 13 36 PM

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: Consensus is to make these fields mandatory in the next [PR](https://github.com/XRPLF/clio/pull/1627)

src/util/newconfig/ConfigValue.hpp
PeterChen13579: Ah, after testing I remember the problem now. Imagine we have ConfigValue{Integer}.withConstraint{port} without a default value. Then the user must provide a value. Then, for example the user provides 123. We have to check to make sure 123 is of type Integer before then checking with constraint.

src/util/newconfig/ConfigFileJson.cpp
PeterChen13579: There's actually no point on checking whether key exists because there is a [separate check before the call to these get functions](https://github.com/XRPLF/clio/pull/1593/files#diff-4e8d0eeb9ae15e14a017e6d321f3987f1c1e411fd58bb6eea9540f715328818eR236). I removed the asserts.

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (p >= portMin && p <= portMax)
+        return std::nullopt;
+    return Error{"Port does not satisfy the constraint bounds"};
+}
+
+std::optional<Error>
+ChannelNameConstraint::checkTypeImpl(Value const& channelName) const
+{
+    if (!std::holds_alternative<std::string>(channelName))
+        return Error{"Key \"channel\"'s value must be a string"};
```
godexsoft: nit: You can avoid escaping the double quotes if you use raw string literal like we do in our tests for json. Maybe not important here but good to remember it's possible: 
```cpp
R"(Key "channel"'s value must be a string)";
```

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,197 @@
...
+    return std::nullopt;
+}
+
+std::optional<Error>
+ChannelNameConstraint::checkValueImpl(Value const& channelName) const
+{
+    if (std::ranges::any_of(Logger::CHANNELS, [&channelName](std::string_view name) {
```
godexsoft: nit: For readability you could extract the lambda into a local constant. Then the `if statement` is more obvious. Something like this:
```cpp
namespace rg = std::ranges;

auto const check = [](std::string_view name) { return ...; }
if (rg::any_of(Logger.CHANNELS, check))
    return std::nullopt;

return Error{...};
```

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,455 @@
...
+    constexpr std::string
+    makeErrorMsg(std::string_view key, Value const& value, std::array<char const*, arrSize> arr) const
+    {
+        // Extract the value from the variant
+        std::string valueStr = std::visit([](auto const& v) { return fmt::format("{}", v); }, value);
+
+        // Create the error message
+        auto errorMsg =
+            fmt::format("You provided value \"{}\". Key \"{}\"'s value must be one of the following: ", valueStr, key);
+        errorMsg += fmt::format("{}", fmt::join(arr, ", "));
```
godexsoft: I think this could all be simplified to 
```cpp
auto const valueStr = ...;
return fmt::format(R"(Something "{}" key "{}": {})", valueStr, key, fmt::join(arr, ", "));
```

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,455 @@
...
+    checkValueImpl(Value const& val) const = 0;
+};
+
+/**
+ * @brief A constraint to ensure the port number is within a valid range.
+ */
+class PortConstraint final : public Constraint {
+public:
+    constexpr ~PortConstraint()
+    {
```
godexsoft, line 159: `= default;` here does not work because of the gcc bug?

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,455 @@
...
+    /**
+     * @brief Check if the number is positive.
+     *
+     * @param num The number to check
+     * @return An Error object if the constraint is not met, nullopt otherwise
+     */
+    [[nodiscard]] std::optional<Error>
+    checkValueImpl(Value const& num) const override
+    {
+        auto const numValue = std::holds_alternative<double>(num) ? std::get<double>(num) : std::get<int64_t>(num);
```
godexsoft: This is a subtle bug. Take a look at this example i made for you: https://godbolt.org/z/3zYxqjj4b

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -0,0 +1,455 @@
...
+     * @brief Check if the number is positive.
+     *
+     * @param num The number to check
+     * @return An Error object if the constraint is not met, nullopt otherwise
+     */
+    [[nodiscard]] std::optional<Error>
+    checkValueImpl(Value const& num) const override;
+};
+
+static constexpr PortConstraint const validatePort{};
```
godexsoft: These `const`s are redundant - `constexpr` is implicitly `const`.

src/util/newconfig/ConfigFileJson.cpp
godexsoft: I'm surprised that it would be needed for json (which is not a binary format) but i'd keep it because we had the `ios::binary` flag set before too.

godexsoft: You could use `if(auto const in = ifstream(...); in) {...}` syntax to make it a little nicer though.

src/util/newconfig/ConfigFileJson.cpp
godexsoft: This decision usually depends on whether you want to unit-test the function (would have to be declared in the header though). If it's a small helper that is tested through other code it's totally fine to use the anonymous namespace and i'd prefer that myself too.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -0,0 +1,162 @@
...
+ * @param jsonValue The JSON value to extract.
+ * @return A variant containing the same type corresponding to the extracted value.
+ */
+[[nodiscard]] static Value
+extractJsonValue(boost::json::value const& jsonValue)
+{
+    Value variantValue;
+
+    if (jsonValue.is_int64()) {
+        variantValue = jsonValue.as_int64();
```
godexsoft: I'd also not use a temporary here but just `return jsonValue.as_int64();` etc. on each line with an ASSERT at the bottom if you failed to handle all possible types.

src/util/newconfig/ConfigValue.hpp
```diff
@@ -83,47 +61,101 @@ class ConfigValue {
...
+        if (value_.has_value()) {
+            auto const& temp = cons_.value().get();
+            auto const& result = temp.checkConstraint(value_.value());
+            if (result.has_value()) {
+                // useful for specifying clear Error message
+                std::string type;
+                std::visit(
+                    util::OverloadSet{
+                        [&type](bool tmp) { type = fmt::format("bool {}", tmp); },
+                        [&type](std::string tmp) { type = fmt::format("string {}", tmp); },
```
godexsoft: For more complex types like `std::string` it's probably better to avoid a copy here by using `std::string const& tmp`.

godexsoft (changes requested):
Leaving a few nits and findings. Good progress 👍

src/util/newconfig/ConfigConstraints.cpp
```diff
@@ -0,0 +1,197 @@
...
+#include <regex>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <variant>
+
+namespace util::config {
+
+std::optional<Error>
+PortConstraint::checkTypeImpl(Value const& port) const
```
cindyyan317, line 37: Question about the constraints **PortConstraint**, why don't we use general constraint like _Between_ (0,65535)?
Same Question about **ChannelNameConstraint** and **CassandraName**, more general constraint will be like _OneOf_("info", "debug","warning"...) _OneOf_("cassandra"). WDYT? @godexsoft @kuznetsss @PeterChen13579

PeterChen13579: So for the port constraint, I don't think it can part of `NumberValueConstraint`, as port can be represented by strings in the config, and there are conversions needed from string to int. 

For the OneOf though, I think it can be done! I initially misinterpreted your comment to add "OneOf" constraint as part of ConfigValue, and I thought then that the ConfigValue definition would be wayyy to long, because it will look something like this: 
`{"log_level", ConfigValue{ConfigType::String}.defaultValue("info").withConstraint(OneOf("info", "debug", "warning" ......  ))}.` with 50 more lines. 
but now I recognize that I can hide all the constraint stuff and it can be the same as before. Good catch! 
Thanks @cindyyan317  🙇

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: Yes, that's correct. I added an extra godbolt demonstration in the comments

PeterChen13579:
@godexsoft  Just a heads-up: I removed the uint64 constraint since, [in the places where it's used](https://github.com/XRPLF/clio/blob/0ede0ed3515ade1ddba05e92f9f69fc05e0cb5c3/src/util/log/Logger.cpp#L132), the user only inputs a number in the thousands, and we multiply that number by a constant. Therefore, I really don't think the uint64 constraint is necessary

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -72,6 +74,11 @@ static constexpr std::array<char const*, 3> loadCacheMode = {
     "none",
 };
 
+/**
+ * @brief specific values that are accepted for database type in config.
+ */
+static constexpr std::array<char const*, 1> databaseType = {"cassandra"};
```
godexsoft: nit: We usually `CAPITALIZE_SNAKE` `static constexpr` stuff

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -391,8 +313,8 @@ class NumberValueConstraint final : public Constraint {
     [[nodiscard]] std::optional<Error>
     checkValueImpl(Value const& num) const override
     {
-        auto const numValue = std::holds_alternative<double>(num) ? std::get<double>(num) : std::get<int64_t>(num);
-        if (numValue >= min_ && numValue <= max_)
+        auto const numValue = std::get<int64_t>(num);
```
godexsoft, line 316: Is it never a double anymore?

godexsoft (changes requested):
Looks good! just a few more tiny things to verify 👍

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: yes; initially I wanted to combine constraints for both a double and int, but there were problems correctly retrieving the value types, so I split it up and have a separate constraint for double. I forgot to remove this part from before :/

godexsoft (approved):

PR #1600, opened 2024-08-12
feat: Add stop to WorkQueue

kuznetsss:
For #442.


src/rpc/WorkQueue.hpp
```diff
@@ -116,6 +126,11 @@ class WorkQueue {
 
                 func(yield);
                 --curSize_.get();
+                if (curSize_.get().value() == 0 && stopping_) {
+                    auto onTasksComplete = onQueueEmpty_.lock();
```
cindyyan317, line 144: Any chance the callback onQueueEmpty will be call twice?

kuznetsss: There is no such chance because `curSize_` is atomic and `onQueueEmpty_` is protected by `Mutex`.
The reason to use mutex is because `onQueueEmpty_` is used from multiple threads: from a thread pool processing io_context's tasks and from a thread calling `stop()`.

kuznetsss: I see, it is possible for `stop()` to be called after line 128 but before 130, so `onQueueEmpty_` will be called twice.

src/rpc/WorkQueue.cpp
```diff
@@ -19,13 +19,40 @@
...
+WorkQueue::OneTimeCallable::setCallable(std::function<void()> func)
+{
+    func_ = func;
+}
+
+void
+WorkQueue::OneTimeCallable::operator()()
+{
+    if (not called_) {
+        func_();
```
cindyyan317, line 47: I feel there is still race condition here. Can worker thread and stop thread both be running func_ ?

kuznetsss: It is inside `Mutex`, so this can't be called from multiple threads.

cindyyan317 (approved):
LGTM

PR #1601, opened 2024-08-13
refactor: Subscription Manager uses async framework

cindyyan317:
Instead of spawn threads on our own, this PR will use async util.
For testing, we want to use sync executor. So this PR also change the context to a template class.


src/feed/SubscriptionManager.hpp
```diff
@@ -55,53 +52,67 @@ namespace feed {
...
-        , validationsFeed_(ioContext, "validations")
-        , ledgerFeed_(ioContext)
-        , bookChangesFeed_(ioContext)
-        , transactionFeed_(ioContext)
-        , proposedTransactionFeed_(ioContext)
+    ~SubscriptionManager() override
     {
+        ctx_.stop();
+        ctx_.join();
     }
```
kuznetsss, line 95: This is probably not needed because `ExecutionContext` will do the same in its destructor.

src/feed/SubscriptionManager.hpp
```diff
@@ -55,53 +52,67 @@ namespace feed {
...
-    impl::BookChangesFeed bookChangesFeed_;
-    impl::TransactionFeed transactionFeed_;
-    impl::ProposedTransactionFeed proposedTransactionFeed_;
+    ExecutionContext ctx_;
+    impl::ForwardFeed<ExecutionContext> manifestFeed_;
+    impl::ForwardFeed<ExecutionContext> validationsFeed_;
+    impl::LedgerFeed<ExecutionContext> ledgerFeed_;
+    impl::BookChangesFeed<ExecutionContext> bookChangesFeed_;
+    impl::TransactionFeed<ExecutionContext> transactionFeed_;
+    impl::ProposedTransactionFeed<ExecutionContext> proposedTransactionFeed_;
```
kuznetsss, line 67: Maybe use `AnyExecutionContext` and avoid templates?
https://github.com/XRPLF/clio/tree/develop/src/util/async#type-erasure

src/feed/SubscriptionManager.hpp
```diff
@@ -178,124 +222,176 @@ class SubscriptionManager : public SubscriptionManagerInterface {
...
+        };
+    }
 };
 
 /**
- * @brief The help class to run the subscription manager. The container of io_context which is used to publish the
- * feeds.
+ * @brief The help class to run the subscription manager with a pool executor.
  */
 class SubscriptionManagerRunner {
```
kuznetsss, line 390: This class looks very small now, maybe move this functionality into `SubscriptionManager`?

src/rpc/common/impl/HandlerProvider.hpp
```diff
@@ -40,7 +39,7 @@ namespace rpc {
 class Counters;
 }  // namespace rpc
 namespace feed {
-class SubscriptionManager;
+class SubscriptionManagerInterface;
```
kuznetsss, line 42: I think it would be ok here to include `feed/SubscriptionManagerInterface.hpp` and don't make forward declaration.

tests/common/util/SyncExecutionContextFixture.hpp
```diff
@@ -0,0 +1,32 @@
...
+#pragma once
+
+#include "util/LoggerFixtures.hpp"
+#include "util/async/context/SyncExecutionContext.hpp"
+
+/**
+ * @brief Fixture with an embedded sync execution context
+ *
+ */
+struct SyncExecutionCtxFixture : virtual public NoLoggerFixture {
```
kuznetsss, line 29: Maybe we don't need a fixture for such a simple thing? I would say using `util::async::SyncExecutionContext ctx;` is ok.

tests/unit/rpc/handlers/SubscribeTests.cpp
```diff
@@ -604,16 +602,16 @@ TEST_F(RPCSubscribeHandlerTest, StreamsWithoutLedger)
         })"
     );
     runSpawn([&, this](auto yield) {
-        auto const handler = AnyHandler{SubscribeHandler{backend, subManager_}};
+        auto const handler = AnyHandler{SubscribeHandler{backend, mockSubscriptionManagerPtr}};
+        EXPECT_CALL(*mockSubscriptionManagerPtr, subTransactions).Times(1);
```
kuznetsss, line 606: Minor thing, feel free to ignore: `Times(1)` is by default when `EXPECT_CALL()` used, so you don't need to specify it. But it's ok to leave it as it is.

src/feed/SubscriptionManager.hpp
cindyyan317: ExecutionContext's destructor will only call stop, not join

src/feed/SubscriptionManager.hpp
cindyyan317: I will start another PR to address this since it is better to have all cpp back .

PR #1605, opened 2024-08-14
refactor: Subscription Manager uses async framework 

cindyyan317:


src/feed/SubscriptionManager.hpp
```diff
@@ -71,24 +70,31 @@ class SubscriptionManager : public SubscriptionManagerInterface {
...
     {
     }
 
+    /**
+     * @brief Destructor of the SubscriptionManager object. It will block until all running jobs finished.
+     */
+    ~SubscriptionManager() override
+    {
+        ctx_.stop();
+        ctx_.join();
```
kuznetsss, line 95: Why not to add this to the destructor of `AnyExecutionContext`?

src/feed/SubscriptionManager.hpp
```diff
@@ -286,16 +292,15 @@ class SubscriptionManager : public SubscriptionManagerInterface {
 };
 
 /**
- * @brief The help class to run the subscription manager. The container of io_context which is used to publish the
- * feeds.
+ * @brief The help class to run the subscription manager. The container of PoolExecutionContext which is used to publish
+ * the feeds.
  */
 class SubscriptionManagerRunner {
```
kuznetsss, line 298: I think this class now could be merged with `SubscriptionsManager`.

src/feed/impl/ProposedTransactionFeed.cpp
```diff
@@ -101,17 +101,18 @@ ProposedTransactionFeed::pub(boost::json::object const& receivedTxJson)
...
-        signal_.emit(pubMsg);
-        // Prevent the same connection from receiving the same message twice if it is subscribed to multiple accounts
-        // However, if the same connection subscribe both stream and account, it will still receive the message twice.
-        // notified_ can be cleared before signal_ emit to improve this, but let's keep it as is for now, since rippled
-        // acts like this.
-        notified_.clear();
-        for (auto const& account : affectedAccounts)
-            accountSignal_.emit(account, pubMsg);
-    });
+    [[maybe_unused]] auto task =
```
kuznetsss, line 104: Maybe I'm wrong but I thought @godexsoft said task is cancelled when it is deleted.

src/feed/SubscriptionManager.hpp
cindyyan317: I think the join() which might block the thread might not need for every case of destructions.

src/feed/SubscriptionManager.hpp
cindyyan317: Yea. I saw your comment in the previous one. I want to test SubscriptionsManager in sync way, so I have this runner to inject the pool executor to it.

src/feed/impl/ProposedTransactionFeed.cpp
cindyyan317: That's bad. Because in this case, I don't care about the result.

cindyyan317: I have confirmed the tasks would not be cancelled when token is deleted.

src/feed/SubscriptionManager.hpp
cindyyan317: We will fix this after we can move context.

kuznetsss (approved):
👍

PR #1609, opened 2024-08-15
feat: Move/copy support in async framework

### LINKED ISSUE #1608: Fix AsyncFramework's immovability
## Summary
AsyncFramework (util/async) execution contexts are non-copyable and immovable. This is because the underlying `boost::asio::thread_pool` is too.
This is proving to be inconvenient. Sometimes we want to
1) Move contexts around
2) AnyExecutionContext to actually own the context, not just refer to it
3) Copy/move AnyExecutionContext instances around much like we do copy `boost::asio::yield`

## Motivation

It will make usage of AnyAsyncContext as well as the async framework as a whole simpler.

## Solution
Make our contexts movable and in some cases copyable. 
While it does not make sense to copy the actual execution context, it does make sense to allow copying the type-erased wrapper.
The suggested way is to change `pimpl_` inside of AnyExecutionContext to be a `shared_ptr` instead of a `unique_ptr`. The overhead is minimal and the effect is worth it.

godexsoft:
Fixes #1608

### DIFF
--- a/src/util/async/AnyExecutionContext.hpp
+++ b/src/util/async/AnyExecutionContext.hpp
@@ -43,16 +43,41 @@ public:
     /**
      * @brief Construct a new type-erased Execution Context object
      *
+     * @note Stores the Execution Context by reference.
+     *
      * @tparam CtxType The type of the execution context to wrap
      * @param ctx The execution context to wrap
      */
     template <typename CtxType>
-        requires(not std::is_same_v<std::decay_t<CtxType>, AnyExecutionContext>)
     /* implicit */
-    AnyExecutionContext(CtxType&& ctx) : pimpl_{std::make_unique<Model<CtxType>>(std::forward<CtxType>(ctx))}
+    AnyExecutionContext(CtxType&& ctx)
+        requires(not std::is_same_v<std::decay_t<CtxType>, AnyExecutionContext> and std::is_lvalue_reference_v<decltype(ctx)>)
+        : pimpl_{std::make_unique<NonOwningModel<CtxType>>(ctx)}
     {
     }
 
+    /**
+     * @brief Construct a new type-erased Execution Context object
+     *
+     * @note Stores the Execution Context by moving it into the AnyExecutionContext.
+     *
+     * @tparam CtxType The type of the execution context to wrap
+     * @param ctx The execution context to wrap
+     */
+    template <typename CtxType>
+    /* implicit */
+    AnyExecutionContext(CtxType&& ctx)
+        requires(not std::is_same_v<std::decay_t<CtxType>, AnyExecutionContext> and std::is_rvalue_reference_v<decltype(ctx)>)
+        : pimpl_{std::make_unique<OwningModel<CtxType>>(std::forward<CtxType>(ctx))}
+    {
+    }
+
+    AnyExecutionContext(AnyExecutionContext const&) = default;
+    AnyExecutionContext(AnyExecutionContext&&) = default;
+    AnyExecutionContext&
+    operator=(AnyExecutionContext const&) = default;
+    AnyExecutionContext&
+    operator=(AnyExecutionContext&&) = default;
     ~AnyExecutionContext() = default;
 
     /**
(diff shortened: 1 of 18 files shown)


src/util/async/AnyExecutionContext.hpp
```diff
@@ -43,16 +43,41 @@ class AnyExecutionContext {
...
+     * @note Stores the Execution Context by reference.
+     *
+     * @tparam CtxType The type of the execution context to wrap
+     * @param ctx The execution context to wrap
+     */
+    template <typename CtxType>
+    /* implicit */
+    AnyExecutionContext(CtxType&& ctx)
+        requires(not std::is_same_v<std::decay_t<CtxType>, AnyExecutionContext> and std::is_lvalue_reference_v<decltype(ctx)>)
+        : pimpl_{std::make_unique<NonOwningModel<CtxType>>(ctx)}
```
cindyyan317: Should it be make_shared?

src/util/async/AnyExecutionContext.hpp
```diff
@@ -221,10 +246,49 @@ class AnyExecutionContext {
     };
 
     template <typename CtxType>
-    struct Model : Concept {
+    struct OwningModel : Concept {
```
kuznetsss: Why do you need `OwningModel` and `NonOwningModel`? `CtxType` is a template parameter anyway, so you could pass it here as a reference or move.

src/util/async/AnyExecutionContext.hpp
```diff
@@ -43,16 +43,41 @@ class AnyExecutionContext {
...
      * @brief Construct a new type-erased Execution Context object
      *
+     * @note Stores the Execution Context by reference.
+     *
+     * @tparam CtxType The type of the execution context to wrap
+     * @param ctx The execution context to wrap
+     */
+    template <typename CtxType>
+    /* implicit */
+    AnyExecutionContext(CtxType&& ctx)
```
kuznetsss: Could you declare these constructors with `CtxType&` and `CtxType&&`? Or it will not work?

kuznetsss: Hm, why couldn't you forward to `Model` whatever you got here? Then it will be possible to pass `ctx` as rvalue reference or lvalue.

src/util/async/AnyStrand.hpp
```diff
@@ -165,7 +167,7 @@ class AnyStrand {
     };
 
 private:
-    std::unique_ptr<Concept> pimpl_;
+    std::shared_ptr<Concept> pimpl_;
```
kuznetsss, line 168: Why is this shared now?

godexsoft: The overhead is not that much but it will allow us to copy these type-erased wrappers without a care in the world, just like we do with yield_context 👍 
Consider that this is only for AnyExecutionContext and AnyStrand, not for AnyOperation. This is because conceptually there will not be much copying of the ExecutionContext/Strand around (outside of DI). Operation on the other hand might be stored somewhere and removed later which can lead to shared state manipulation and that can lead to contention on the atomic inside if multiple threads are involved.

src/util/async/context/SyncExecutionContext.hpp
```diff
@@ -56,6 +56,24 @@ struct SameThreadContext {
...
+    void
+    stop() const noexcept
+    {
+        executor.stop();
+    }
+
+    void
+    join() noexcept
+    {
+        executor.join();
```
cindyyan317, line 68: I think Executor's join can be updated to const. Then we can change join interface to const too.

src/util/async/Concepts.hpp
```diff
@@ -150,4 +150,19 @@ concept SomeStdDuration = requires {
 template <typename T>
 concept SomeOptStdDuration = requires(T v) { SomeStdDuration<decltype(v.value())>; };
 
+/**
+ * @brief Checks that decayed T s not of the same type as Erased
+ */
+template <typename T, typename Erased>
+concept NotSameAs = not std::is_same_v<std::decay_t<T>, Erased>;
```
cindyyan317, line 169: Old issue: we don't have unittests for this file.

src/util/async/context/SyncExecutionContext.hpp
godexsoft: In the very deep end, `boost::asio::thread_pool::join` is not const, so I don't think we can. However I do see that some of our joins are const, which is strange :rage1:

godexsoft: I tried and turns out it still compiles fine. Strange but let's roll with it 👍

src/util/async/Concepts.hpp
godexsoft: We can add some static_asserts

kuznetsss (approved):
👍

cindyyan317 (approved):
👍

PR #1623, opened 2024-09-04
refactor: Remove SubscriptionManagerRunner

cindyyan317:


src/app/ClioApplication.cpp
```diff
@@ -101,9 +102,10 @@ ClioApplication::run()
     auto backend = data::make_Backend(config_);
 
     // Manages clients subscribed to streams
-    auto subscriptionsRunner = feed::SubscriptionManagerRunner(config_, backend);
-
-    auto const subscriptions = subscriptionsRunner.getManager();
+    auto const workersNum = config_.valueOr<std::uint64_t>("subscription_workers", 1);
+    LOG(util::LogService::info()) << "Starting subscription manager with " << workersNum << " workers";
+    auto subscriptions =
+        std::make_shared<feed::SubscriptionManager>(util::async::PoolExecutionContext(workersNum), backend);
```
kuznetsss: Maybe better to create a function `make_SubscriptionManager(config)` in `SubscriptionManager.hpp`?
Or a `SubscriptionManager`'s constructor taking config and backend?

src/feed/SubscriptionManager.hpp
```diff
@@ -74,9 +72,9 @@ class SubscriptionManager : public SubscriptionManagerInterface {
      * @param backend The backend to use
      */
     template <class ExecutorCtx>
```
kuznetsss: Do we still need a template here? I think there is `AnyExecutionContext`.

src/app/ClioApplication.cpp
cindyyan317: The execution ctx is one of the constructor parameter. The number of threads is needed for only pool ctx. So logically, the cfg should not be sent to the constructor of subscription manager. Maybe better create a function. But creating this function will include the config in SubcriptionManager.hpp.

kuznetsss: I was thinking that knowledge of which config option to take is related to `SubscriptionManagaer` so probably better to have all the information related to it in its own files. Including config shouldn't do any harm.

kuznetsss (approved):
LGTM 👍

Issue #1620, opened 2024-09-03
SubscriptionSource bugs


kuznetsss:
## Issue Description

- No detection of a situation when websocket connection to rippled is still open but not receiving anything.
- Potential race condition in `LoadBalancer::choseForwardingSource()`

PR #1626, opened 2024-09-04
fix: Subscription source bugs fix

kuznetsss:
For #1620.

- Add timeouts for websocket operations for connections to rippled. Without these timeouts if connection hangs for some reason, clio wouldn't know the connection is hanging.
- Fix potential data race in choosing new subscription source which will forward messages to users.
- Optimise switching between subscription sources.

### DIFF
--- a/src/etl/impl/SubscriptionSource.cpp
+++ b/src/etl/impl/SubscriptionSource.cpp
@@ -265,8 +265,6 @@ SubscriptionSource::handleMessage(std::string const& message)
                 } else if (object.contains(JS(type)) && object.at(JS(type)) == JS_ManifestReceived) {
                     LOG(log_.debug()) << "Forwarding manifest: " << object;
                     subscriptions_->forwardManifest(object);
-                } else {
-                    LOG(log_.error()) << "Unknown message: " << object;
                 }
             }
         }
--- a/src/etl/impl/SubscriptionSource.cpp
+++ b/src/etl/impl/SubscriptionSource.cpp
@@ -287,9 +287,9 @@ void
 SubscriptionSource::handleError(util::requests::RequestError const& error, boost::asio::yield_context yield)
 {
     isConnected_ = false;
-    isForwarding_ = false;
+    bool const wasForwarding = isForwarding_.exchange(false);
     if (not stop_) {
-        onDisconnect_();
+        onDisconnect_(wasForwarding);
         LOG(log_.info()) << "Disconnected";
     }
 
--- a/src/etl/impl/SubscriptionSource.cpp
+++ b/src/etl/impl/SubscriptionSource.cpp
@@ -24,6 +24,8 @@
 #include "rpc/JS.hpp"
 #include "util/Retry.hpp"
 #include "util/log/Logger.hpp"
+#include "util/prometheus/Label.hpp"
+#include "util/prometheus/Prometheus.hpp"
 #include "util/requests/Types.hpp"
 
 #include <boost/algorithm/string/classification.hpp>
@@ -66,7 +68,7 @@ SubscriptionSource::SubscriptionSource(
     OnConnectHook onConnect,
     OnDisconnectHook onDisconnect,
     OnLedgerClosedHook onLedgerClosed,
-    std::chrono::steady_clock::duration const connectionTimeout,
+    std::chrono::steady_clock::duration const wsTimeout,
     std::chrono::steady_clock::duration const retryDelay
 )
     : log_(fmt::format("SubscriptionSource[{}:{}]", ip, wsPort))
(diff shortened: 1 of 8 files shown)


tests/unit/etl/LoadBalancerTests.cpp
```diff
@@ -288,29 +285,25 @@ TEST_F(LoadBalancerOnDisconnectHookTests, source0DisconnectsAndConnectsBack)
     EXPECT_CALL(sourceFactory_.sourceAt(0), setForwarding(false));
     EXPECT_CALL(sourceFactory_.sourceAt(1), isConnected()).WillOnce(Return(true));
     EXPECT_CALL(sourceFactory_.sourceAt(1), setForwarding(true));
-    sourceFactory_.callbacksAt(0).onDisconnect();
+    sourceFactory_.callbacksAt(0).onDisconnect(true);
```
godexsoft, line 288: Let's group expectations together, blank line and then group the test calls together 👍

src/etl/impl/SubscriptionSource.hpp
```diff
@@ -103,7 +109,7 @@ class SubscriptionSource {
      * @param onNewLedger The onNewLedger hook. Called when a new ledger is received
      * @param onLedgerClosed The onLedgerClosed hook. Called when the ledger is closed but only if the source is
      * forwarding
-     * @param connectionTimeout The connection timeout. Defaults to 30 seconds
+     * @param wsTimeout A timeout for websocket operations. Defaults to 30 seconds
```
cindyyan317, line 112: We are sure the ledger stream will send us data every ledger. The ledger close time may exceed 30s for some ledger like [89898224](https://livenet.xrpl.org/ledgers/89898224) to [89898225](https://livenet.xrpl.org/ledgers/89898225 ) .  20s gap seems happen from time to time. [89898219](https://livenet.xrpl.org/ledgers/89898219) to [89898220](https://livenet.xrpl.org/ledgers/89898220). We can relax the timeout to 1min maybe.

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -312,7 +319,11 @@ SubscriptionSource::logError(util::requests::RequestError const& error) const
 void
 SubscriptionSource::setLastMessageTime()
 {
-    lastMessageTime_.lock().get() = std::chrono::steady_clock::now();
+    lastMessageTimeSecondsSinceEpoch_.get().set(
+        std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count()
+    );
+    auto lock = lastMessageTime_.lock();
```
cindyyan317, line 325: I think previously it is locked already. But does it really need to be locked? The call path seems be in a strand.

tests/unit/etl/SubscriptionSourceTests.cpp
```diff
@@ -525,11 +545,35 @@ TEST_F(SubscriptionSourceReadTests, LastMessageTime)
...
+        connection.close(yield);
+    });
+
+    EXPECT_CALL(onConnectHook_, Call());
+    EXPECT_CALL(onDisconnectHook_, Call(false)).WillOnce([this]() { subscriptionSource_.stop(); });
+    EXPECT_CALL(lastMessageTimeMock, set).WillOnce([](int64_t value) {
+        auto const now =
+            std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch())
+                .count();
+        EXPECT_LE(std::abs(value - now), 1);
```
cindyyan317: value-now must be negative, right?

cindyyan317 (commented):
👍  I think the data race problem is solved. 
Only want to double check the timeout value.

src/etl/impl/SubscriptionSource.cpp
kuznetsss: Getter for `lastMessageTime_` is called from another thread.

src/etl/impl/SubscriptionSource.hpp
kuznetsss: We receive messages from all subscriptions by one websocket connection. I think in validation stream there is no long gaps. But in the worst case we will just reconnect which is quite quick. WDYT?

cindyyan317: It looks like the validations stream update quick. I don't know if it is guaranteed.  Yes, in the worst case, we reconnect , it seems not much cost. I am okay keep it 30s.

godexsoft (commented):
Just a little style nit. And of course the crashing test needs to be fixed but frankly it's not possible to tell why it would fail just by looking at the code.. good luck 🤞 :rage4:

tests/unit/etl/LoadBalancerTests.cpp
kuznetsss: The empty line here is correct. Expectations are only for `onDisconnect()`. And there is no expectations for `onConnect()`.

godexsoft: UTs are usually structured in the following blocks:
- Given - setup the SUT and other state
- When - action happens
- Then - expectations checked

Sometimes these groups can repeat if the test sets up and checks more than one thing.

In our case we set up expectations first due to the nature of the framework.
In general, i'd setup all expectations in one block and then trigger all actions in another block. 

Having that said, the broken whitespace is not introduced by you as you only added the argument. 
But then there is the boy scout rule - leave the codebase in a cleaner/better state than you found it 👍

godexsoft: I expanded the file and the issue seems to be on every other line.. up to you if you want to change it or not.

tests/unit/etl/SubscriptionSourceTests.cpp
kuznetsss: I removed `abs()` and made it `now - value`.

godexsoft (approved):
👍

PR #1627, opened 2024-09-04
refactor: Replace all old instances of Config with New Config

### LINKED ISSUE #1184: Refactor Config
It would be nice to refactor `Config` class to parse file once into structure applying all the default values and then use the structure everywhere in the code.

It will also allow us to easily add `--check-config` run option to verify provided config instead of starting Clio.

PeterChen13579:
Fixes #1184 
Previous PR's found [here](https://github.com/XRPLF/clio/pull/1593) and [here](https://github.com/XRPLF/clio/pull/1544)


src/app/ClioApplication.cpp
```diff
@@ -81,7 +82,7 @@ ClioApplication::ClioApplication(util::Config const& config) : config_(config),
 int
 ClioApplication::run()
 {
-    auto const threads = config_.valueOr("io_threads", 2);
+    auto const threads = config_.getValue("io_threads").asIntType<uint16_t>();
     if (threads <= 0) {
         LOG(util::LogService::fatal()) << "io_threads is less than 1";
         return EXIT_FAILURE;
```
kuznetsss: As I understand at this point all the values in config are validated, so this check could be removed.

src/data/cassandra/SettingsProvider.cpp
```diff
@@ -36,43 +36,26 @@
...
-
-    out.contactPoints = obj.valueOrThrow<std::string>("contact_points", "`contact_points` must be a string");
-    out.port = obj.maybeValue<uint16_t>("port");
-
-    return out;
-}
 
 inline Settings::SecureConnectionBundle
-tag_invoke(boost::json::value_to_tag<Settings::SecureConnectionBundle>, boost::json::value const& value)
+invoke_tag_SecureConnections(std::string_view value)
```
kuznetsss: I think you could remove this function now and just create `SecureConnectionBundle` instead.

src/etl/ETLService.cpp
```diff
@@ -278,11 +278,15 @@ ETLService::ETLService(
...
     , amendmentBlockHandler_(ioc, state_)
 {
-    startSequence_ = config.maybeValue<uint32_t>("start_sequence");
-    finishSequence_ = config.maybeValue<uint32_t>("finish_sequence");
-    state_.isReadOnly = config.valueOr("read_only", static_cast<bool>(state_.isReadOnly));
-    extractorThreads_ = config.valueOr<uint32_t>("extractor_threads", extractorThreads_);
-    txnThreshold_ = config.valueOr<size_t>("txn_threshold", txnThreshold_);
+    startSequence_ = config.getValue("start_sequence").hasValue()
+        ? std::make_optional(config.getValue("start_sequence").asIntType<uint32_t>())
+        : std::nullopt;
```
kuznetsss: This is repeating multiple times. Maybe add a method in config? So it could look like:
```cpp
startSequence_ = config.getValue("start_sequence").asOptional<uint32_t>();
```
`asOptional()` should be allowed only for optional values and check that provided type is the same as the stored one.

src/etl/LoadBalancer.cpp
```diff
@@ -71,31 +75,33 @@ LoadBalancer::make_LoadBalancer(
...
     }
 
     static constexpr std::uint32_t MAX_DOWNLOAD = 256;
-    if (auto value = config.maybeValue<uint32_t>("num_markers"); value) {
-        ASSERT(*value > 0 and *value <= MAX_DOWNLOAD, "'num_markers' value in config must be in range 1-256");
-        downloadRanges_ = *value;
+    auto const numMarkers = config.getValue("num_markers");
+    if (numMarkers.hasValue()) {
+        auto const value = numMarkers.asIntType<uint32_t>();
+        ASSERT(value > 0 and value <= MAX_DOWNLOAD, "'num_markers' value in config must be in range 1-256");
```
kuznetsss: This check should be in ConfigDefinition. So the code should be:
```cpp
if (auto const numMarkers = config.getValue("num_markers"); numMarkers.hasValue())
    downloadRanges_ = numMarkers;
```

src/main/Main.cpp
```diff
@@ -37,13 +40,18 @@ try {
     return action.apply(
         [](app::CliArgs::Action::Exit const& exit) { return exit.exitCode; },
         [](app::CliArgs::Action::Run const& run) {
-            auto const config = util::ConfigReader::open(run.configPath);
-            if (!config) {
-                std::cerr << "Couldnt parse config '" << run.configPath << "'." << std::endl;
+            auto const json = ConfigFileJson::make_ConfigFileJson(run.configPath);
+            if (!json.has_value()) {
+                std::cerr << json.error().error << std::endl;
```
kuznetsss, line 45: There should be exit after this.

src/util/SignalsHandler.cpp
```diff
@@ -99,9 +99,7 @@ SignalsHandler::SignalsHandler(Config const& config, std::function<void()> force
 {
     impl::SignalsHandlerStatic::registerHandler(*this);
 
-    gracefulPeriod_ = Config::toMilliseconds(config.valueOr("graceful_period", 10.f));
-    ASSERT(gracefulPeriod_.count() >= 0, "Graceful period must be non-negative");
-
+    gracefulPeriod_ = util::config::ClioConfigDefinition::toMilliseconds(config.getValue("graceful_period").asFloat());
```
kuznetsss: Maybe as a separate PR you could add a method `asMillisecondsFromFloat()`?

src/util/Taggable.hpp
```diff
@@ -178,6 +181,22 @@ class TagDecoratorFactory final {
     Type type_;                        /*< The type of TagDecorator this factory produces */
     ParentType parent_ = std::nullopt; /*< The parent tag decorator to bind */
 
+    static Type
+    invoke_type(std::string_view style)
```
kuznetsss: Probably better to name it `make_Type()`.

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -34,100 +35,18 @@
...
 #include <vector>
 
 namespace util::config {
-/**
- * @brief Full Clio Configuration definition.
- *
- * Specifies which keys are valid in Clio Config and provides default values if user's do not specify one. Those
- * without default values must be present in the user's config file.
- */
-static ClioConfigDefinition ClioConfig = ClioConfigDefinition{
```
kuznetsss, line 55: Why did you move it to header?

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -75,27 +80,42 @@ struct ClioConfigDescription {
...
-        KV{"cache.peers.[].ip", "IP address of peer nodes to cache."},
-        KV{"cache.peers.[].port", "Port number of peer nodes to cache."},
+        KV{"workers", "Number of threads to process RPC requests."},
         KV{"server.ip", "IP address of the Clio HTTP server."},
         KV{"server.port", "Port number of the Clio HTTP server."},
         KV{"server.max_queue_size", "Maximum size of the server's request queue."},
-        KV{"server.workers", "Maximum number of threads for server to run with."},
         KV{"server.local_admin", "Indicates if the server should run with admin privileges."},
-        KV{"server.admin_password", "Password for Clio admin-only APIs."},
+        KV{"server.admin_password", "The password to login to the server"},
```
kuznetsss: The previous description looks better for me.

src/web/Server.hpp
```diff
@@ -364,7 +362,7 @@ make_HttpServer(
         util::TagDecoratorFactory(config),
         dosGuard,
         handler,
-        std::move(adminPassword)
+        std::move(adminPassword.hasValue() ? std::make_optional(adminPassword.asString()) : std::nullopt)
```
kuznetsss: Looks like `std::move()` is not needed here anymore.

src/web/dosguard/DOSGuard.cpp
```diff
@@ -134,15 +138,17 @@ DOSGuard::clear() noexcept
...
-    auto whitelist = config.arrayOr("dos_guard.whitelist", {});
-    auto const transform = [](auto const& elem) { return elem.template value<std::string>(); };
-    return T{
-        boost::transform_iterator(std::begin(whitelist), transform),
-        boost::transform_iterator(std::end(whitelist), transform)
-    };
+    std::unordered_set<std::string> ips;
+    auto const whitelist = config.getArray("dos_guard.whitelist");
+
+    if (whitelist.size() > 0) {
```
kuznetsss: You don't need this `if`

src/web/dosguard/DOSGuard.cpp
```diff
@@ -134,15 +138,17 @@ DOSGuard::clear() noexcept
...
+        for (auto it = whitelist.begin<ValueView>(); it != whitelist.end<ValueView>(); ++it) {
```
kuznetsss: Is it possible to do 
```cpp
for (auto ip : whitelist) {
   ips.insert(ip.asString());
}
```

src/web/dosguard/WhitelistHandler.hpp
```diff
@@ -121,24 +123,25 @@ class WhitelistHandler : public WhitelistHandlerInterface {
...
         // resolve hostnames to ips
         std::unordered_set<std::string> ips;
-        for (auto const& hostname : hostnames) {
-            auto resolvedIps = resolver.resolve(hostname, "");
-            for (auto& ip : resolvedIps) {
-                ips.insert(std::move(ip));
-            }
-        };
+
+        if (whitelist.size() > 0) {
```
kuznetsss: Also don't need this `if`.

tests/common/util/newconfig/ClioConfigFactories.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+#include <boost/json/value.hpp>
+
+#include <thread>
+
+using namespace util::config;
+
+// config used for load balancer test
+inline ClioConfigDefinition
+getParseLoadBalancerConfig(boost::json::value val)
```
kuznetsss: I think it is better to not put all of them into one file because I assume each of these functions is used only in one test file.

tests/unit/util/newconfig/JsonFileTests.cpp
```diff
@@ -0,0 +1,98 @@
...
+    {
+    }
+
+    ConfigFileJson const jsonFileObj;
+};
+
+TEST_F(ParseJson, validateValues)
+{
+    EXPECT_TRUE(jsonFileObj.containsKey("header.text1"));
+    EXPECT_EQ(std::get<std::string>(jsonFileObj.getValue("header.text1")), "value");
```
kuznetsss: Tests should be separated by the feature you are testing. I would prefer to split this test for containsKey and getValue test cases.

tests/unit/util/newconfig/JsonFileTests.cpp
```diff
@@ -0,0 +1,98 @@
...
+    EXPECT_EQ(std::get<int64_t>(jsonFileObj.getValue("dosguard.port")), 44444);
+
+    EXPECT_FALSE(jsonFileObj.containsKey("idk"));
+    EXPECT_FALSE(jsonFileObj.containsKey("optional.withNoDefault"));
+}
+
+TEST_F(ParseJson, validateArrayValue)
+{
+    // validate array.[].sub matches expected values
+    EXPECT_TRUE(jsonFileObj.containsKey("array.[].sub"));
```
kuznetsss, line 60:
```suggestion
    ASSERT_TRUE(jsonFileObj.containsKey("array.[].sub"));
```

src/util/newconfig/ConfigDefinition.cpp
PeterChen13579: I don't think static variables in cpp files can be used in other .hpp files. So i moved it to header

src/etl/LoadBalancer.cpp
PeterChen13579: Ah that's right. I think I need to add a new constraint because the value here must be between 1 and 256.

src/data/cassandra/SettingsProvider.cpp
godexsoft: In any case the function name should change if it had to stay. The reason we had `tag_invoke` is because it's one of the "standard" approaches to provide customization points. We should avoid using snake_case other than for `tag_invoke` and maybe `make_*` (should probably call them `make*` in the first place).

src/data/cassandra/SettingsProvider.cpp
```diff
@@ -108,34 +91,44 @@ Settings
...
-        config_.valueOr<uint32_t>("max_read_requests_outstanding", settings.maxReadRequestsOutstanding);
-    settings.coreConnectionsPerHost =
-        config_.valueOr<uint32_t>("core_connections_per_host", settings.coreConnectionsPerHost);
-    settings.queueSizeIO = config_.maybeValue<uint32_t>("queue_size_io");
-    settings.writeBatchSize = config_.valueOr<std::size_t>("write_batch_size", settings.writeBatchSize);
-
-    auto const connectTimeoutSecond = config_.maybeValue<uint32_t>("connect_timeout");
-    if (connectTimeoutSecond)
-        settings.connectionTimeout = std::chrono::milliseconds{*connectTimeoutSecond * util::MILLISECONDS_PER_SECOND};
+    settings.threads = config_.getValue("threads").asIntType<uint32_t>();
```
godexsoft: nit: Wouldn't this API look more natural if `config_.get<uint32_t>("threads")` (or similar) would do the same job?
You could have a list of supported types inside of the `get` and map them to the correct `asType<T>` internally.

src/etl/ETLService.cpp
godexsoft: `config.maybeValue<uint32_t>("...")` would be even better imo :)

godexsoft: It seems to me that you could have kept the interface almost as it was yet implement all the fancy stuff inside of the framework. It would take much less changes to integrate the new config this way and the API would be more concise. Not saying we should change everything at this point but it's sad that we missed the opportunity there

PeterChen13579: yea, looking back I think it would have better to kept the interface and changed the things that needed to be changed instead of implementing it from scratch 😕. But it was good practice on the bright side 👍

src/web/dosguard/DOSGuard.cpp
PeterChen13579: I checked and it cannot, because Array can hold ValueView and objectView, so unless it holds only one type, it isn't possible

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -122,7 +121,25 @@ class ClioConfigDefinition {
...
+     * @tparam T The type T to return
+     * @param fullKey The config key to search for
+     * @return Value of key of type T
+     */
+    template <typename T>
+    T
+    getValue(std::string_view fullKey) const
+    {
+        ASSERT(map_.contains(fullKey), "key {} does not exist in config", fullKey);
+        if (std::holds_alternative<ConfigValue>(map_.at(fullKey))) {
```
kuznetsss: Save `map_.at(fullKey)` into a variable to not search twice. Maybe compiler will do it for you, but it is better to be sure.

src/data/cassandra/SettingsProvider.cpp
godexsoft: Huh? `tag_invoke` is the only `tag_invoke` we should have. If you want to roll a custom mechanism that's ok but don't call it anything related to `tag_invoke` then :) `tag_invoke` is a special function that's supposed to be called exactly `tag_invoke` and have the first parameter be the tag of what this `tag_invoke` allows to customize.

src/etl/CacheLoaderSettings.cpp
```diff
@@ -47,31 +48,29 @@ CacheLoaderSettings::isDisabled() const
...
+    // Given cursors number fetching from account
+    settings.numCacheCursorsFromAccount = cache.get<std::size_t>("num_cursors_from_account");
 
-        settings.numCacheMarkers = cache.valueOr<size_t>("num_markers", settings.numCacheMarkers);
-        settings.cachePageFetchSize = cache.valueOr<size_t>("page_fetch_size", settings.cachePageFetchSize);
+    settings.numCacheMarkers = cache.get<std::size_t>("num_markers");
+    settings.cachePageFetchSize = cache.get<std::size_t>("page_fetch_size");
+
+    auto const entry = cache.get<std::string>("load");
+    if (boost::iequals(entry, "sync"))
```
godexsoft, line 67: Definitely not for this PR but an idea for future. This would be nice to have as a custom type in the config, like `cache.get<CacheLoaderSettings>("load)`. We would have to setup a small parser (`tag_invoke` likely) per type to support custom types but it would be nicer to use and would allow us to pre-validate the possible options at startup or config validation.

src/util/Taggable.hpp
```diff
@@ -182,7 +182,7 @@ class TagDecoratorFactory final {
     ParentType parent_ = std::nullopt; /*< The parent tag decorator to bind */
 
     static Type
-    invoke_type(std::string_view style)
+    makeType(std::string_view style)
     {
```
godexsoft, line 186: Another instance where it could be useful to have custom types support

src/util/log/Logger.cpp
```diff
@@ -158,7 +156,7 @@ LogService::init(config::ClioConfigDefinition const& config)
     }
 
     // get default severity, can be overridden per channel using the `log_channels` array
-    auto defaultSeverity = invoke_tag(config.getValue("log_level").asString());
+    auto defaultSeverity = invoke_tag(config.get<std::string>("log_level"));
```
godexsoft: This `invoke_tag` is sus. It does not look like our normal `camelCase` style for functions and it re-implements what `tag_invoke` should be but not in the right way (not using ADL).

tests/unit/etl/LoadBalancerTests.cpp
```diff
@@ -90,6 +96,26 @@ constexpr static auto const ThreeSourcesLedgerResponse = R"({
...
+          ConfigValue{ConfigType::Double}.defaultValue(10.0).withConstraint(validatePositiveDouble)},
+         {"allow_no_etl", ConfigValue{ConfigType::Boolean}.defaultValue(false)},
+         {"etl_sources.[].ip", Array{ConfigValue{ConfigType::String}.optional().withConstraint(validateIP)}},
+         {"etl_sources.[].ws_port", Array{ConfigValue{ConfigType::String}.optional().withConstraint(validatePort)}},
+         {"etl_sources.[].grpc_port", Array{ConfigValue{ConfigType::String}.optional()}},
+         {"num_markers", ConfigValue{ConfigType::Integer}.optional().withConstraint(validateNumMarkers)}}
+    };
+
+    auto const errors = config.parse(ConfigFileJson{val.as_object()});
+    ASSERT(!errors.has_value(), "Error parsing Json for clio config for load balancer test");
```
godexsoft: We should probably be using gtest's `ASSERT_FALSE` here (and in similar places in tests), not our `ASSERT` that will cause the whole thing to crash.

tests/unit/util/newconfig/ClioConfigDefinitionTests.cpp
```diff
@@ -113,14 +128,14 @@ struct NewConfigDeathTest : NewConfigTest {};
 
 TEST_F(NewConfigDeathTest, GetNonExistentKeys)
 {
-    EXPECT_DEATH({ [[maybe_unused]] auto a_ = configData.getValue("head."); }, ".*");
```
godexsoft, line 115: These `a_` would get renamed to `a` by clang-tidy once we have it setup. I recommend we use `unused` until we have C++26 with its new placeholder (`_`).

godexsoft: Generally speaking, `_` suffix is for members, not for local variables

godexsoft (changes requested):
Looks like a massive effort, good job 👍 
I left a few questions and suggestions for future. Overall it looks pretty good 🥇

src/app/ClioApplication.cpp
```diff
@@ -32,8 +32,8 @@
 #include "rpc/common/impl/HandlerProvider.hpp"
 #include "util/Assert.hpp"
 #include "util/build/Build.hpp"
-#include "util/config/Config.hpp"
 #include "util/log/Logger.hpp"
+#include "util/newconfig/ConfigDefinition.hpp"
```
cindyyan317, line 35: Can we rename `newconfig` to `config` now?

src/etl/LoadBalancer.cpp
```diff
@@ -52,12 +55,13 @@
 #include <vector>
 
 using namespace util;
```
cindyyan317: looks like you don't need `util` here

src/etl/LoadBalancer.cpp
```diff
@@ -106,10 +110,12 @@ LoadBalancer::LoadBalancer(
         }
     };
 
-    auto const forwardingTimeout = Config::toMilliseconds(config.valueOr<float>("forwarding.request_timeout", 10.));
-    for (auto const& entry : config.array("etl_sources")) {
+    auto const forwardingTimeout =
+        ClioConfigDefinition::toMilliseconds(config.get<float>("forwarding.request_timeout"));
+    auto const etlArray = config.getArray("etl_sources");
+    for (auto it = etlArray.begin<ObjectView>(); it != etlArray.end<ObjectView>(); ++it) {
```
cindyyan317, line 115: can it be const?

src/rpc/WorkQueue.cpp
```diff
@@ -92,12 +90,12 @@ WorkQueue::stop(std::function<void()> onQueueEmpty)
...
-WorkQueue::make_WorkQueue(util::Config const& config)
+WorkQueue::make_WorkQueue(util::config::ClioConfigDefinition const& config)
 {
     static util::Logger const log{"RPC"};
-    auto const serverConfig = config.section("server");
-    auto const numThreads = config.valueOr<uint32_t>("workers", std::thread::hardware_concurrency());
-    auto const maxQueueSize = serverConfig.valueOr<uint32_t>("max_queue_size", 0);  // 0 is no limit
+    auto const serverConfig = config.getObject("server");
+    auto const numThreads = config.get<uint32_t>("workers");
+    auto const maxQueueSize = serverConfig.get<uint32_t>("max_queue_size");  // 0 is no limit
```
cindyyan317: I feel the comment should be moved to ConfigDefinition 🤔

src/rpc/WorkQueue.cpp
```diff
@@ -92,12 +90,12 @@ WorkQueue::stop(std::function<void()> onQueueEmpty)
...
 WorkQueue
-WorkQueue::make_WorkQueue(util::Config const& config)
+WorkQueue::make_WorkQueue(util::config::ClioConfigDefinition const& config)
 {
     static util::Logger const log{"RPC"};
-    auto const serverConfig = config.section("server");
-    auto const numThreads = config.valueOr<uint32_t>("workers", std::thread::hardware_concurrency());
-    auto const maxQueueSize = serverConfig.valueOr<uint32_t>("max_queue_size", 0);  // 0 is no limit
+    auto const serverConfig = config.getObject("server");
+    auto const numThreads = config.get<uint32_t>("workers");
```
cindyyan317, line 97: Where is this "workers" defined? I don't see it in ConfigDefinition.

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -229,4 +279,108 @@ class ClioConfigDefinition {
...
+     {"server.local_admin", ConfigValue{ConfigType::Boolean}.optional()},
+     {"server.admin_password", ConfigValue{ConfigType::String}.optional()},
+     {"server.processing_policy",
+      ConfigValue{ConfigType::String}.defaultValue("parallel").withConstraint(validateProcessingPolicy)},
+     {"server.parallel_requests_limit", ConfigValue{ConfigType::Integer}.optional()},
+     {"server.ws_max_sending_queue_size", ConfigValue{ConfigType::Integer}.defaultValue(1500)},
+     {"server.__ng_web_server", ConfigValue{ConfigType::Boolean}.defaultValue(false)},
+     {"prometheus.enabled", ConfigValue{ConfigType::Boolean}.defaultValue(true)},
+     {"prometheus.compress_reply", ConfigValue{ConfigType::Boolean}.defaultValue(true)},
+     {"io_threads", ConfigValue{ConfigType::Integer}.defaultValue(2).withConstraint(validateUint16)},
```
cindyyan317: If we set io_threads 1, does Clio exit with error as it does previously?

src/util/log/Logger.hpp
```diff
@@ -53,6 +53,9 @@
 namespace util {
 
 class Config;
```
cindyyan317: Can be removed

src/app/ClioApplication.cpp
PeterChen13579: I was thinking after this PR is merged, ill create a new PR that deletes the old config and rename my `newconfig` to `config`.  so the PR isn't messy. WDYT?

src/etl/LoadBalancer.cpp
PeterChen13579: like `auto const it`? It can't be const cuz we are getting the Object of every `etl_source`

src/util/newconfig/ConfigDefinition.hpp
PeterChen13579: @cindyyan317  I set `io_threads` to 1 in the old config and it seems to work perfectly fine. We can maybe have a look?

src/etl/CacheLoaderSettings.cpp
PeterChen13579: 💯

src/util/newconfig/ConfigDefinition.hpp
cindyyan317: Forgive me. https://github.com/XRPLF/clio/blob/a7074dbf0f2b63dd76fdd48b1879b468513a78d0/src/app/ClioApplication.cpp#L107. I think i should ask if we set it 0, does Clio exit with error.

PeterChen13579: Thank you for the catch 🙏. The constraint for `io_threads` needs to be changed

tests/unit/web/dosguard/WhitelistHandlerTests.cpp
```diff
@@ -47,7 +47,7 @@ getParseWhitelistHandlerConfig(boost::json::value val)
     ConfigFileJson const jsonVal{val.as_object()};
     auto config = ClioConfigDefinition{{"dos_guard.whitelist.[]", Array{ConfigValue{ConfigType::String}}}};
     auto const errors = config.parse(jsonVal);
-    ASSERT(!errors.has_value(), "Cannot parse Json Correctly");
+    [&]() { ASSERT_FALSE(errors.has_value()); }();
```
godexsoft, line 50: fyi: if you want to preserve the custom error message you can still do it this way:
```cpp
ASSERT_FALSE(condition) << "Custom error message";
```

tests/unit/web/ng/ResponseTests.cpp
```diff
@@ -150,7 +156,9 @@ TEST_F(ResponseTest, createFromStringAndConnection)
 
 TEST_F(ResponseTest, createFromJsonAndConnection)
 {
-    util::TagDecoratorFactory const tagDecoratorFactory{util::Config{}};
+    util::TagDecoratorFactory tagDecoratorFactory{
```
godexsoft, line 159: Can't be `const` anymore?

godexsoft (approved):
Looks good. I'm just leaving a few more small suggestions but they are not a blocker for merging this if others don't find any issues

tests/unit/web/ng/ResponseTests.cpp
PeterChen13579: should definitely be const 💯 . Im going to let clang-tidy fix it, instead of pushing and needing approval again

PR #1628, opened 2024-09-05
fix: not forward admin API

cindyyan317:
Don't forward admin API to rippled

### DIFF
--- a/src/rpc/RPCEngine.hpp
+++ b/src/rpc/RPCEngine.hpp
@@ -134,9 +134,9 @@ public:
     {
         if (forwardingProxy_.shouldForward(ctx)) {
             // Disallow forwarding of the admin api, only user api is allowed for security reasons.
-            if (isAdminCmd(ctx.method, ctx.params)) {
+            if (isAdminCmd(ctx.method, ctx.params))
                 return Result{Status{RippledError::rpcNO_PERMISSION}};
-            }
+
             return forwardingProxy_.forward(ctx);
         }
 
--- a/src/rpc/RPCEngine.hpp
+++ b/src/rpc/RPCEngine.hpp
@@ -22,6 +22,7 @@
 #include "data/BackendInterface.hpp"
 #include "rpc/Counters.hpp"
 #include "rpc/Errors.hpp"
+#include "rpc/RPCHelpers.hpp"
 #include "rpc/WorkQueue.hpp"
 #include "rpc/common/HandlerProvider.hpp"
 #include "rpc/common/Types.hpp"
@@ -131,8 +132,13 @@ public:
     Result
     buildResponse(web::Context const& ctx)
     {
-        if (forwardingProxy_.shouldForward(ctx))
+        if (forwardingProxy_.shouldForward(ctx)) {
+            // Disallow forwarding of the admin api, only user api is allowed for security reasons.
+            if (isAdminCmd(ctx.method, ctx.params)) {
+                return Result{Status{RippledError::rpcNO_PERMISSION}};
+            }
             return forwardingProxy_.forward(ctx);
+        }
 
         if (backend_->isTooBusy()) {
             LOG(log_.error()) << "Database is too busy. Rejecting request";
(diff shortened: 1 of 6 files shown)


src/rpc/RPCHelpers.cpp
```diff
@@ -1273,6 +1273,23 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+    auto const isFieldSet = [&request](auto const field) {
+        return request.contains(field) and request.at(field).is_bool() and request.at(field).as_bool();
+    };
+    if (method == JS(ledger)) {
+        if (isFieldSet(JS(full)) or isFieldSet(JS(accounts)) or isFieldSet(JS(type))) {
+            return true;
+        }
+    }
+
+    if (method == JS(feature) and request.contains(JS(vetoed)))
```
cindyyan317, line 1288: Should we allow forwarding if vetoed is false?

godexsoft: Absolutely not. Clio should never be connecting to validator nodes in the first place.

src/rpc/RPCHelpers.cpp
```diff
@@ -1273,6 +1273,23 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
     return false;
 }
 
+bool
+isAdminCmd(std::string const& method, boost::json::object const& request)
+{
+    auto const isFieldSet = [&request](auto const field) {
+        return request.contains(field) and request.at(field).is_bool() and request.at(field).as_bool();
+    };
```
godexsoft, line 1281: style nit: Let's separate the lambda from its usage by a whitespace line.

src/rpc/RPCEngine.hpp
```diff
@@ -131,8 +132,13 @@ class RPCEngine {
     Result
     buildResponse(web::Context const& ctx)
     {
-        if (forwardingProxy_.shouldForward(ctx))
+        if (forwardingProxy_.shouldForward(ctx)) {
+            // Disallow forwarding of the admin api, only user api is allowed for security reasons.
+            if (isAdminCmd(ctx.method, ctx.params)) {
```
godexsoft: nit: Would be ok to remove the curlies imo.

src/rpc/RPCHelpers.cpp
```diff
@@ -1273,6 +1273,23 @@ specifiesCurrentOrClosedLedger(boost::json::object const& request)
...
+    if (method == JS(ledger)) {
+        if (isFieldSet(JS(full)) or isFieldSet(JS(accounts)) or isFieldSet(JS(type))) {
```
godexsoft: nit: Curlies not needed imo

src/rpc/RPCHelpers.hpp
```diff
@@ -557,6 +557,16 @@ parseIssue(boost::json::object const& issue);
 bool
 specifiesCurrentOrClosedLedger(boost::json::object const& request);
 
+/**
+ * @brief Check if a request requires ADMIN role to execute in rippled.
```
godexsoft: Maybe something like "Check whether a request is considered to require administrative privileges on `rippled` side" is more accurate? In reality some APIs require IDENTIFIED role, not ADMIN on `rippled` side.

godexsoft (changes requested):

src/rpc/RPCHelpers.cpp
cindyyan317: I see. The `vetoed` is to really vote on something.

godexsoft (dismissed):
LGTM 👍

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -539,3 +539,42 @@ TEST_F(RPCHelpersTest, ParseIssue)
         std::runtime_error
     );
 }
+
+TEST_F(RPCHelpersTest, isAdminCmd)
```
kuznetsss: Maybe refactor this to a parameterized tests?

kuznetsss (approved):
👍

PR #1629, opened 2024-09-05
fix: not forward admin API

cindyyan317:
To merge to 2.2.3 branch.
It is different from the #1628 . I think this branch still forward feature API to rippled.


kuznetsss (approved):
👍

PR #1633, opened 2024-09-06
fix: Subscription source bugs fix (#1626)

### LINKED ISSUE #1620: SubscriptionSource bugs

kuznetsss:
Fixes #1620.
Cherry pick of #1626 into develop.

- Add timeouts for websocket operations for connections to rippled. Without these timeouts if connection hangs for some reason, clio wouldn't know the connection is hanging.
- Fix potential data race in choosing new subscription source which will forward messages to users.
- Optimise switching between subscription sources.

### DIFF
--- a/src/etl/LoadBalancer.cpp
+++ b/src/etl/LoadBalancer.cpp
@@ -111,10 +111,13 @@ LoadBalancer::LoadBalancer(
             validatedLedgers,
             forwardingTimeout,
             [this]() {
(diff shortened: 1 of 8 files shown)


godexsoft (approved):
🚀

PR #1643, opened 2024-09-10
fix: Add more restrictions to admin fields 

cindyyan317:


godexsoft (approved):
Horrible workarounds we have to do because of `rippled` 😒 Hopefully OpenAPI will solve all this nonsense once and for all. Meanwhile, thanks for the fix 👍

PR #1644, opened 2024-09-10
fix: no restriction on type field

cindyyan317:
'type' should not matter if 'full' or 'accounts' is false. Relax the restriction for 'type'


godexsoft (approved):
Not sure we need to do this but it looks like it does what it says.

PR #1647, opened 2024-09-13
feat: add 'force_forward' field to request

cindyyan317:
Clio will forward the request to rippled when 'ledger_index' is 'current' or 'closed'. When 'ledger_index' is 'validated' or other valid number input, there is no way to forward it to rippled. 

For purpose of debugging, we allow admin to send request to its p2p rippled with 'force_forward' being true.

### DIFF
--- a/src/rpc/common/impl/ForwardingProxy.hpp
+++ b/src/rpc/common/impl/ForwardingProxy.hpp
@@ -69,6 +69,9 @@ public:
         if (specifiesCurrentOrClosedLedger(request))
             return true;
 
+        if (isForcedForward(ctx))
+            return true;
+
         auto const checkAccountInfoForward = [&]() {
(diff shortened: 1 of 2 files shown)


src/rpc/common/impl/ForwardingProxy.hpp
```diff
@@ -138,6 +141,14 @@ class ForwardingProxy {
     {
         return handlerProvider_->contains(method) || isProxied(method);
     }
+
+    bool
+    isForcedForward(web::Context const& ctx) const
+    {
+        static constexpr auto forceForwardFieldName = "force_forward";
```
godexsoft: We usually use SNAKE_UPPER_CASE style for these.

godexsoft (changes requested):
The change looks fine to me 👍 just one style remark 🔢

godexsoft (approved):
🚀

PR #1648, opened 2024-09-16
Update conanfile.py

mst-rajatmishra:
Removed the `fPIC` Option:
We took out the `fPIC` option from both the `options` and `default_options` sections because it wasn’t being used in your configuration.
Updated Dependency Options: We added a specific setting for the `benchmark` dependency in the `default_options` section. This sets `benchmark/*:header_only` to `True`, assuming this is the right configuration for your setup.
Cleaned Up Comments:

We updated and cleaned up comments to make sure they accurately reflect the changes and provide clearer explanations where needed.
General Cleanup: We did some tidying up by removing unnecessary mentions and placeholder comments that were left over after removing the `fPIC` option.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,6 @@ class Clio(ConanFile):
     settings = 'os', 'compiler', 'build_type', 'arch'
     options = {
         'static': [True, False],              # static linkage
-        'fPIC': [True, False],                # unused?
         'verbose': [True, False],
         'tests': [True, False],               # build unit tests; create `clio_tests` binary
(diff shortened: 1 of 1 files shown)


conanfile.py
```diff
@@ -21,20 +20,8 @@ class Clio(ConanFile):
         'lint': [True, False],                # run clang-tidy checks during compilation
     }
 
-    requires = [
```
godexsoft, line 24: This block should stay as it was. Other than that the pr looks ok to me.

godexsoft (changes requested):

PR #1649, opened 2024-09-16
feat: Delete-before support in data removal tool

godexsoft:
Adds support for `delete-before` and `delete-after` commands to prune Clio's cassandra/scylladb.


tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -0,0 +1,566 @@
...
+func (c *ClioCass) prepareSimpleDeleteQueries(
+	fromLedgerIdx uint64,
+	toLedgerIdx uint64,
+	deleteQueryTemplate string,
+) deleteInfo {
+	var info = deleteInfo{Query: deleteQueryTemplate}
+
+	// Note: we deliberately add 1 extra ledger to make sure we delete any data Clio might have written
+	// if it crashed or was stopped in the middle of writing just before it wrote ledger_range.
+	for i := fromLedgerIdx; i <= toLedgerIdx+1; i++ {
```
cindyyan317: It seems like `toLedgerId+1` is not okay for **delete_before** case.
if delete_before  parameter is 1000, then toLedgerIdx will be 999. In this for loop, it will delete 1000.

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -0,0 +1,566 @@
...
+								rowsRetrieved++
+
+								if keepLastValid && !slices.Equal(previousKey, key) {
+									previousKey = key
+									foundLastValid = false
+								}
+
+								// only grab the rows that are in the correct range of sequence numbers
+								if fromLedgerIdx.HasValue() && fromLedgerIdx.Value() <= seq {
+									outChannel <- deleteParams{Seq: seq, Blob: key}
```
cindyyan317, line 399: I feel below condition will remove more data.
```
else if toLedgerIdx.HasValue() {
			if seq <= toLedgerIdx.Value() &&( !keepLastValid || foundLastValid) {
				outChannel <- deleteParams{Seq: seq, Blob: key}
			} else if seq<=toLedgerIdx.Value()+1{
				foundLastValid = true
			}
		}
```

if `toLedgerIdx.Value()+1` exists, it means we can delete anything <=toLedgerIdx.Value()

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -0,0 +1,566 @@
...
+		totalDeletes += deleteCount
+	}
+
+	// diff queries
+	if !c.settings.SkipDiffTable {
+		log.Println("Generating delete queries for diff table")
+		info = c.prepareSimpleDeleteQueries(rangeFrom, rangeTo,
+			"DELETE FROM diff WHERE seq = ?")
+		log.Printf("Total delete queries: %d\n\n", len(info.Data))
+		deleteCount, errCount = c.performDeleteQueries(&info, columnSettings{UseBlob: true, UseSeq: true})
```
cindyyan317: useBlob should be false?

godexsoft: Yes, good catch. Luckily UseSeq is checked first so it was working correct already.

cindyyan317 (approved):
👍  offline-deletion 😄

PR #1660, opened 2024-09-20
fix: add no lint to ignore clang-tidy

PeterChen13579:
Fixes build for [#1659](https://github.com/XRPLF/clio/actions/runs/10956058143/job/30421296417)

### DIFF
--- a/src/util/newconfig/ConfigConstraints.hpp
+++ b/src/util/newconfig/ConfigConstraints.hpp
@@ -155,7 +155,7 @@ protected:
  */
 class PortConstraint final : public Constraint {
 public:
-    constexpr ~PortConstraint()
+    constexpr ~PortConstraint() override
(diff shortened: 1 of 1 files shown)


src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -156,7 +156,8 @@ class Constraint {
 class PortConstraint final : public Constraint {
 public:
     constexpr ~PortConstraint() override
-    = default;
+    {
```
godexsoft, line 159: Can we have them all in one line or does clang-format change it to multiline here?

PeterChen13579: clang format changes it to multi-line 😔

godexsoft (approved):

PR #1666, opened 2024-09-23
fix: Workaround for gcc12 bug

godexsoft:
Attempt to workaround the gcc bug via `constinit`. Appears to compile and work on my machine, let's see if CI agrees.


PeterChen13579 (approved):
👍 🚀

PR #1667, opened 2024-09-25
chore: Update libxrpl to 2.3.0-b4

kuznetsss:


src/rpc/common/Validators.cpp
```diff
@@ -200,16 +200,10 @@ CustomValidator CustomValidators::SubscribeStreamValidator =
             "ledger", "transactions", "transactions_proposed", "book_changes", "manifests", "validations"
         };
 
-        static std::unordered_set<std::string> const reportingNotSupportStreams = {
```
cindyyan317, line 203: we can create a new ClioError "NotSupported". Users may feel strange that such stream is valid according to document.

cindyyan317 (approved):
LGTM

PR #1671, opened 2024-09-30
feat: server info cache

cindyyan317:
Convert ForwardCache to ResponseExpirationCache.
Add cache to RPC part. 
The cache invalidation is triggered by timeout cfg.


src/util/ResponseExpirationCache.hpp
```diff
@@ -0,0 +1,140 @@
...
+namespace util {
+
+/**
+ * @brief Cache of requests' responses with TTL support and configurable cachable commands
+ */
+class ResponseExpirationCache {
+    /**
+     * @brief A class to store a cache entry.
+     */
+    class CacheEntry {
```
kuznetsss: Maybe name it just `Entry` since it is in the cache now?

src/util/ResponseExpirationCache.hpp
```diff
@@ -0,0 +1,140 @@
...
+         * @brief Put a response into the cache
+         *
+         * @param response The response to store
+         */
+        void
+        put(boost::json::object response)
+        {
+            response_ = std::move(response);
+            lastUpdated_ = std::chrono::steady_clock::now();
+        }
```
kuznetsss: I think implementation could be moved to cpp file for methods of this class.

tests/unit/rpc/RPCEngineTests.cpp
```diff
@@ -0,0 +1,513 @@
...
+    }
+})JSON";
+}  // namespace
+
+class RPCEngineTest : public util::prometheus::WithPrometheus,
+                      public MockBackendTest,
+                      public MockCountersTest,
+                      public MockLoadBalancerTest,
+                      public SyncAsioContextTest {
+protected:
```
kuznetsss:
```suggestion
struct RPCEngineTest : util::prometheus::WithPrometheus,
                      MockBackendTest,
                      MockCountersTest,
                      MockLoadBalancerTest,
                      SyncAsioContextTest {

```

tests/unit/rpc/RPCEngineTests.cpp
```diff
@@ -0,0 +1,513 @@
...
+        {"ForwardedSubmit",
+         isAdmin,
+         "submit",
+         "{}",
+         forwarded,
+         neverCalledIsTooBusy,
+         neverCalledUnknownCmd,
+         !handlerReturnError,
+         rpc::Status{},
+         boost::json::parse(FORWARD_REPLY).as_object()},
```
kuznetsss: I think to improve readability it is better here to use values but with named fields initialisation:
```cpp
{.testName = "ForwardedSubmit",
  .isAdmin = true,
  .method = "submit",
...
}
```

tests/unit/rpc/RPCEngineTests.cpp
```diff
@@ -0,0 +1,513 @@
...
+INSTANTIATE_TEST_CASE_P(
+    RPCEngineFlow,
+    RPCEngineFlowParameterTest,
+    ValuesIn(generateTestValuesForParametersTest()),
+    tests::util::NameGenerator
+);
+
+TEST_P(RPCEngineFlowParameterTest, Test)
+{
+    auto const testBundle = GetParam();
```
kuznetsss: nit: I think `GetParam()` returns a reference, so there is no reason to do a copy here.

src/rpc/RPCEngine.hpp
```diff
@@ -140,6 +163,11 @@ class RPCEngine {
             return forwardingProxy_.forward(ctx);
         }
 
+        if (not ctx.isAdmin and responseCache_) {
+            if (auto res = responseCache_->get(ctx.method); res.has_value())
+                return Result{std::move(res.value())};
```
kuznetsss:
```suggestion
                return Result{std::move(res).value()};
```

tests/common/util/MockResponseExpirationCache.hpp
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <boost/json/object.hpp>
+#include <gmock/gmock.h>
+
+#include <optional>
+#include <string>
+#include <unordered_set>
+
+struct MockResponseExpirationCache {
```
kuznetsss: Is this struct used anywhere?

cindyyan317: Opps , forgot to delete

tests/unit/rpc/RPCEngineTests.cpp
```diff
@@ -200,7 +193,8 @@ TEST_P(RPCEngineFlowParameterTest, Test)
 
     if (testBundle.isTooBusy.has_value()) {
         EXPECT_CALL(*backend, isTooBusy).WillOnce(Return(*testBundle.isTooBusy));
-        EXPECT_CALL(*mockCountersPtr, onTooBusy).Times(1);
+        if (testBundle.isTooBusy.value())
+            EXPECT_CALL(*mockCountersPtr, onTooBusy).Times(1);
```
kuznetsss: nit: `.Times(1)` is not needed because 1 time is by default.

kuznetsss (dismissed):
LGTM

kuznetsss (approved):
👍

Issue #1678, opened 2024-10-02
Support Deletion of Account_tx


PeterChen13579:
## Summary

Currently deletion script does not delete account_tx tables from our database. We should support it.

PR #1679, opened 2024-10-02
fix: deletion script will not OOM

### LINKED ISSUE #1676: Fix Deletion Script OOM
## Summary
Deletion script currently stores all queries in memory before executing them, resulting in OOM if we delete millions of ledgers at a time. We should prevent this. Also good idea to add a file to support resume deletion from where it had left off in case of the program crashing

PeterChen13579:
fixes #1676 and #1678

### DIFF
--- a/tools/cassandra_delete_range/cassandra_delete_range.go
+++ b/tools/cassandra_delete_range/cassandra_delete_range.go
@@ -79,17 +79,18 @@ func main() {
 
 	if *resume {
 		// format of file continue.txt is
-		// previous user command (must match the same command to resume deletion)
-		// table name (ie. objects, ledger_hashes etc)
-		// deletion method (ie. token_range or ledger_range)
-		// the values of token_ranges (each pair of values seperated line by line) or just a single int for ledger_range
+		/*
+		 Previous user command (must match the same command to resume deletion)
+		 Table name (ie. objects, ledger_hashes etc)
+		 Deletion method (ie. token_range). We don't store ledger_range because these tables are extremely small
+		 Values of token_ranges (each pair of values seperated line by line) or just a single int for ledger_range
+		*/
 
 		file, err := os.Open("continue.txt")
 		if err != nil {
 			log.Fatal("continue.txt does not exist. Aborted")
 		}
 		defer file.Close()
-		log.Print("RESUMED2")
 
 		if err != nil {
 			log.Fatalf("Failed to open file: %v", err)
@@ -136,7 +137,6 @@ func main() {
 		case "objects":
 			*skipSuccessorTable = true
 		}
-		log.Print("RESUMED3")
 
 		scanner.Scan()
 		if scanner.Text() == "token_range" {
@@ -160,24 +160,9 @@ func main() {
 					log.Fatalf("Error converting integer: %s, %s", err1, err2)
 				}
 				rangeRead[start] = end
-				log.Printf("%d, %d", start, end)
 			}
 			ledgerOrTokenRange = &util.StoredRange{}
 			ledgerOrTokenRange.TokenRange = maybe.Set(rangeRead)
-		} else if scanner.Text() == "ledger_range" {
-			scanner.Scan()
-			line := scanner.Text()
-			startStr := strings.TrimSpace(line)
-
-			// convert ledger_index to uint64
-			toContinue, err1 := strconv.ParseUint(startStr, 10, 64)
-			if err1 != nil {
-				log.Fatalf("Error converting integer: %s", err1)
-			}
-			ledgerOrTokenRange = &util.StoredRange{}
-			ledgerOrTokenRange.LedgerRange = maybe.Set(toContinue)
-		} else {
-			log.Fatalf("%s is not token_range or ledger_range. Aborted..", scanner.Text())
 		}
 	}
 
--- a/tools/cassandra_delete_range/cassandra_delete_range.go
+++ b/tools/cassandra_delete_range/cassandra_delete_range.go
@@ -5,9 +5,11 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
 	"log"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 	"xrplf/clio/cassandra_delete_range/internal/cass"
@@ -15,6 +17,7 @@ import (
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/gocql/gocql"
+	"github.com/pmorelli92/maybe"
 )
 
 const (
@@ -44,6 +47,7 @@ var (
 	clusterCQLVersion     = app.Flag("cql-version", "The CQL version to use").Short('l').Default("3.0.0").String()
 	clusterPageSize       = app.Flag("cluster-page-size", "Page size of results").Short('p').Default("5000").Int()
 	keyspace              = app.Flag("keyspace", "Keyspace to use").Short('k').Default("clio_fh").String()
+	resume                = app.Flag("resume", "Whether to resume deletion from the previous command due to something crashing").Default("false").Bool()
 
 	userName = app.Flag("username", "Username to use when connecting to the cluster").String()
 	password = app.Flag("password", "Password to use when connecting to the cluster").String()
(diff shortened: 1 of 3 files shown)


tools/cassandra_delete_range/cassandra_delete_range.go
```diff
@@ -208,3 +225,92 @@ func prepareDb(dbHosts *string) (*gocql.ClusterConfig, error) {
...
+			start, err1 := strconv.ParseInt(startStr, 10, 64)
+			end, err2 := strconv.ParseInt(endStr, 10, 64)
+
+			if err1 != nil || err2 != nil {
+				log.Fatalf("Error converting integer: %s, %s", err1, err2)
+			}
+			rangeRead[start] = end
+		}
+		ledgerOrTokenRange = &util.StoredRange{}
+		ledgerOrTokenRange.TokenRange = maybe.Set(rangeRead)
```
cindyyan317: Does TokenRange need to be Maybe?

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -308,54 +410,167 @@ func (c *ClioCass) prepareSimpleDeleteQueries(
...
+				atomic.AddUint64(&totalErrors, err)
+				info = deleteInfo{Query: deleteQueryTemplate}
+			}
+		}
+		// delete the rest of queries if exists
+		if len(info.Data) > 0 {
+			_, err := c.performDeleteQueries(&info, session, colSettings)
+			atomic.AddUint64(&totalDeletes, uint64(len(info.Data)))
+			atomic.AddUint64(&totalErrors, err)
+			info = deleteInfo{Query: deleteQueryTemplate}
```
cindyyan317: No need to reset here

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -180,100 +189,193 @@ func (c *ClioCass) pruneData(
...
+		}
+		fmt.Fprintf(file, "%s\n", c.settings.Command)
+		file.WriteString("diff\n")
+
 		log.Println("Generating delete queries for diff table")
-		info = c.prepareSimpleDeleteQueries(rangeFrom, rangeTo,
-			"DELETE FROM diff WHERE seq = ?")
-		log.Printf("Total delete queries: %d\n\n", len(info.Data))
-		deleteCount, errCount = c.performDeleteQueries(&info, columnSettings{UseBlob: false, UseSeq: true})
+		deleteCount, errCount = c.prepareSimpleDeleteQueries(rangeFrom, rangeTo,
```
cindyyan317: Maybe consider rename this function. It is not only "prepare" anymore.

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -368,70 +583,78 @@ func (c *ClioCass) prepareDeleteQueries(
...
-								atomic.AddUint64(&totalErrors, 1)
-							}
+						if !prepareDeleteResult {
+							log.Printf("ERROR: page iteration failed: %s\n", err)
+							fmt.Fprintf(os.Stderr, "FAILED QUERY: %s\n", fmt.Sprintf("%s [from=%d][to=%d][pagestate=%x]", queryTemplate, r.StartRange, r.EndRange, pageState))
+							atomic.AddUint64(&totalErrors, 1)
 						}
 
 						if len(nextPageState) == 0 {
+							// Checks for delete queries every page
```
cindyyan317: When nextPageState is 0, we have iterated all pages for this query. So this comment looks off here.

tools/cassandra_delete_range/cassandra_delete_range.go
```diff
@@ -208,3 +225,92 @@ func prepareDb(dbHosts *string) (*gocql.ClusterConfig, error) {
...
+		fallthrough
+	case "transactions":
+		*skipLedgerHashesTable = true
+		fallthrough
+	case "ledger_hashes":
+		*skipObjectsTable = true
+		fallthrough
+	case "objects":
+		*skipSuccessorTable = true
+	}
```
kuznetsss: What if none of these matched?

tools/cassandra_delete_range/cassandra_delete_range.go
```diff
@@ -208,3 +225,92 @@ func prepareDb(dbHosts *string) (*gocql.ClusterConfig, error) {
...
+
+	scanner.Scan()
+	if scanner.Text() == "token_range" {
+		rangeRead := make(map[int64]int64)
+
+		// now go through all the ledger range and load it to a set
+		for scanner.Scan() {
+			line := scanner.Text()
+			Range := strings.Split(line, ",")
```
kuznetsss:
```suggestion
			range := strings.Split(line, ",")
```

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -368,70 +583,78 @@ func (c *ClioCass) prepareDeleteQueries(
...
+					if c.settings.RangesRead != nil {
+						if c.settings.RangesRead.TokenRange.HasValue() {
+							if value, exists := c.settings.RangesRead.TokenRange.Value()[r.StartRange]; exists {
+								// Check for end range
+								if value == r.EndRange {
+									continue
+								}
+							}
+						} else {
+							log.Fatal("token_range should exist in continue.txt")
```
cindyyan317: The last run could be terminated when the first slice is processing. So we can not error out here. If the map is empty, we  just do nothing.

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -308,54 +410,167 @@ func (c *ClioCass) prepareSimpleDeleteQueries(
...
+			if keepLastValid && !slices.Equal(previousKey, key) {
+				previousKey = key
+				foundLastValid = false
+			}
+
+			// only grab the rows that are in the correct range of sequence numbers
+			if fromLedgerIdx.HasValue() && fromLedgerIdx.Value() <= seq {
+				info.Data = append(info.Data, deleteParams{Seq: seq, Blob: key})
+			} else if toLedgerIdx.HasValue() {
+				if seq <= toLedgerIdx.Value() && (!keepLastValid || foundLastValid) {
```
cindyyan317, line 503: I saw you changed the "<" to "<=". Is it for bug fixing?

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -180,100 +189,193 @@ func (c *ClioCass) pruneData(
 
 	// successor queries
 	if !c.settings.SkipSuccessorTable {
+		file, err := os.Create("continue.txt")
+		if err != nil {
+			fmt.Println("Error creating file in Successor table:", err)
```
cindyyan317: Maybe the error should be "Error creating file for successor"?

tools/cassandra_delete_range/internal/cass/cass.go
```diff
@@ -180,100 +189,193 @@ func (c *ClioCass) pruneData(
 
 	// successor queries
 	if !c.settings.SkipSuccessorTable {
+		file, err := os.Create("continue.txt")
```
cindyyan317: We might be able to tidy up all logic about "continue.txt" to a class. Then you can get rid of hardcoded file name and repeated logic. You can also leverage the object scope to manage the file creation and deletion.

cindyyan317 (changes requested):
The logic looks good to me. 
But the bookmark can be put into a class. 
The workerCount seems incorrec. If I understand correctly, the number of goroutine running concurrently is workerCount*workerCount now.

tools/cassandra_delete_range/internal/cass/cass.go
PeterChen13579: Yes, in the initial script, I tested and there was a 1 off bug. (so if i delete before 100, it doesn't delete 99). This is because we already did -1 [here](https://github.com/XRPLF/clio/blob/2c18fd5465cfc3139a390a7ac834efdd803f9bcc/tools/cassandra_delete_range/internal/cass/cass.go#L80) so It has to be "<="

tools/cassandra_delete_range/cassandra_delete_range.go
PeterChen13579: I think range is a special key word in Go, so it can't be used to declare variables. I'll rename it to tokenRange instead to obey the naming convention 👍

tools/cassandra_delete_range/cassandra_delete_range.go
PeterChen13579: Nothing will happen. Now thinking about it, it'll be a good idea to handle such case because at least one of the cases should match

cindyyan317:
Maybe we can pack the continue.txt logic all in one class like:
```
type Marker struct {
	cmd  string
	file *os.File
}

func NewMarker(cmd string) *Marker {
	return &Marker{cmd: cmd}
}

func CloseMarker(m *Marker) {
	if m.file != nil {
		m.file.Close()
	}
	os.Remove("continue.txt")
}

func (m *Marker) EnterTable(table string) error {
	// Create the file
	file, err := os.OpenFile("continue.txt", os.O_WRONLY|os.O_TRUNC, 0644)
	m.file = file
	if err != nil {
		return fmt.Errorf("failed to create file: %w", err)
	}
	fmt.Fprintf(m.file, "%s\n", m.cmd)
	m.file.WriteString(fmt.Sprintf("%s\n", table))
	return nil
}

func (m *Marker) MarkProgress(x int64, y int64) {
	fmt.Fprintf(m.file, "%d, %d \n", x, y)
}

func (m *Marker) ExitTable() {
	m.file.Close()
	m.file = nil
}

```

PeterChen13579:
oh yea, this way is definitely better

> Maybe we can pack the continue.txt logic all in one class like:
> 
> ```
> type Marker struct {
> 	cmd  string
> 	file *os.File
> }
> 
> func NewMarker(cmd string) *Marker {
> 	return &Marker{cmd: cmd}
> }
> 
> func CloseMarker(m *Marker) {
> 	if m.file != nil {
> 		m.file.Close()
> 	}
> 	os.Remove("continue.txt")
> }
> 
> func (m *Marker) EnterTable(table string) error {
> 	// Create the file
> 	file, err := os.OpenFile("continue.txt", os.O_WRONLY|os.O_TRUNC, 0644)
> 	m.file = file
> 	if err != nil {
> 		return fmt.Errorf("failed to create file: %w", err)
> 	}
> 	fmt.Fprintf(m.file, "%s\n", m.cmd)
> 	m.file.WriteString(fmt.Sprintf("%s\n", table))
> 	return nil
> }
> 
> func (m *Marker) MarkProgress(x int64, y int64) {
> 	fmt.Fprintf(m.file, "%d, %d \n", x, y)
> }
> 
> func (m *Marker) ExitTable() {
> 	m.file.Close()
> 	m.file = nil
> }
> ```

cindyyan317 (approved):
Thanks ~ LGTM

PR #1691, opened 2024-10-16
chore: Add counter for total messages waiting to be sent

cindyyan317:


kuznetsss (approved):
👍

src/web/impl/WsBase.hpp
```diff
@@ -134,7 +150,9 @@ class WsBase : public ConnectionBase, public std::enable_shared_from_this<WsBase
     void
     onWrite(boost::system::error_code ec, std::size_t)
     {
+        LOG(perfLog_.info()) << tag() << "xinmeng sent message";
```
kennyzlei, line 153: @cindyyan317 was this intended to be included in the PR here?  seems a bit user specific 🤔

cindyyan317: Hi That was a mistake. Thanks for pointing it out. Will remove it.

Issue #919, opened 2023-10-11
Refactor web server


kuznetsss:
It would be nice to hide all the implementation of the web server into a library and make usage simple. Usage should be like:
```cpp
Server server(ip, port);
server.onGet("/", getCallback);
serveer.onPost("/", postCallback);
server.run();
```

kuznetsss:
This is needed for #442 so moving it to the 2.3 milestone.

Issue #1680, opened 2024-10-04
Refactor constants and enumerators to CamelCase


godexsoft:
## Summary

Let's follow the core guidelines and avoid ALL_CAPS style.

## Motivation

May unintentionally clash with Macros.

## Solution

Update the names and add a clang-tidy rule if one exists.

PR #1699, opened 2024-10-21
refactor: Coroutine based webserver

kuznetsss:
Code of new coroutine-based web server. The new server is not connected to Clio and not ready to use yet.
For #919.


src/web/impl/AdminVerificationStrategy.cpp
```diff
@@ -79,4 +80,20 @@ make_AdminVerificationStrategy(std::optional<std::string> password)
...
+    auto const localAdmin = serverConfig.maybeValue<bool>("local_admin");
+    bool const localAdminEnabled = localAdmin && localAdmin.value();
+
+    if (localAdminEnabled == adminPassword.has_value()) {
+        if (adminPassword.has_value())
+            return std::unexpected{"Admin config error, local_admin and admin_password can not be set together."};
+        return std::unexpected{"Admin config error, either local_admin and admin_password must be specified."};
+    }
+
+    return make_AdminVerificationStrategy(std::move(adminPassword));
```
godexsoft, line 96: Should this function not be able to make localAdmin strategy too? 
From the tests I believe this happens inside of the string version of the function but it's not obvious. Consider having two make functions, one for each strategy and then call the correct one depending on config from this main make function.

src/web/ng/Connection.hpp
```diff
@@ -0,0 +1,148 @@
...
+class Connection : public util::Taggable {
+protected:
+    std::string ip_;  // client ip
+    boost::beast::flat_buffer buffer_;
+
+public:
+    /**
+     * @brief The default timeout for send, receive, and close operations.
+     */
+    static constexpr std::chrono::steady_clock::duration DEFAULT_TIMEOUT = std::chrono::seconds{30};
```
godexsoft, line 57: All caps is correct old default but let's use Camel case now (#1680)

src/web/ng/Request.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+namespace web::ng {
+
+namespace {
+
+template <typename HeadersType, typename HeaderNameType>
+std::optional<std::string_view>
+getHeaderValue(HeadersType const& headers, HeaderNameType const& headerName)
+{
+    auto it = headers.find(headerName);
```
godexsoft: nit: const could be added here

src/web/ng/Request.hpp
```diff
@@ -0,0 +1,145 @@
...
+     * @param request The WebSocket request.
+     * @param headers The headers of the HTTP request initiated the WebSocket connection
+     */
+    Request(std::string request, HttpHeaders const& headers);
+
+    /**
+     * @brief Method of the request.
+     * @note WEBSOCKET is not a real method, it is used to distinguish WebSocket requests from HTTP requests.
+     */
+    enum class Method { GET, POST, WEBSOCKET, UNSUPPORTED };
```
godexsoft: Let's use Camel case for enumerators too

src/web/ng/Request.hpp
```diff
@@ -0,0 +1,145 @@
...
+private:
+    /**
+     * @brief Get the HTTP request.
+     * @note This function assumes that the request is an HTTP request. So if data_ is not an HTTP request,
+     * the behavior is undefined.
+     *
+     * @return The HTTP request.
+     */
+    HttpRequest const&
+    httpRequest() const;
```
godexsoft, line 142: We could make it defined by an assert but since it's used internally only, perhaps it is ok as is.

src/web/ng/Response.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+template <typename MessageType>
+std::optional<Response::HttpData>
+makeHttpData(boost::beast::http::status status, Request const& request)
+{
+    if (request.isHttp()) {
+        auto const& httpRequest = request.asHttpRequest()->get();
+        return Response::HttpData{
+            .status = status,
+            .contentType = std::is_same_v<MessageType, std::string> ? Response::HttpData::ContentType::TEXT_HTML
```
godexsoft: Idk if the compiler optimizes this (chances are it does) but you can move this above the `return` and make it a `constexpr` so that it's guaranteed to only compute once at compile time

godexsoft: I'd also drop cv qualifiers just to be sure that the `string` type is always detected correctly if some day we add another function that is not taking `message` by value

src/web/ng/Response.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+    return message_;
+}
+
+boost::beast::http::response<boost::beast::http::string_body>
+Response::intoHttpResponse() &&
+{
+    ASSERT(httpData_.has_value(), "Response must have http data to be converted into http response");
+
+    boost::beast::http::response<boost::beast::http::string_body> result{httpData_->status, httpData_->version};
```
godexsoft: Consider `using` for this long type

src/web/ng/Response.cpp
```diff
@@ -0,0 +1,112 @@
...
+    result.keep_alive(httpData_->keepAlive);
+    result.body() = std::move(message_);
+    result.prepare_payload();
+    return result;
+}
+
+boost::asio::const_buffer
+Response::asConstBuffer() const&
+{
+    ASSERT(not httpData_.has_value(), "Loosing existing http data");
```
godexsoft: Losing?

src/web/ng/Response.hpp
```diff
@@ -0,0 +1,106 @@
...
+class Response {
+public:
+    /**
+     * @brief The data for an HTTP response.
+     */
+    struct HttpData {
+        /**
+         * @brief The content type of the response.
+         */
+        enum class ContentType { APPLICATION_JSON, TEXT_HTML };
```
godexsoft: More of these Camel case things 👍

src/web/ng/Server.cpp
```diff
@@ -0,0 +1,322 @@
...
+    boost::asio::ip::tcp::endpoint endpoint,
+    std::optional<boost::asio::ssl::context> sslContext,
+    impl::ConnectionHandler connectionHandler,
+    util::TagDecoratorFactory tagDecoratorFactory
+)
+    : ctx_{ctx}
+    , sslContext_{std::move(sslContext)}
+    , connectionHandler_{std::move(connectionHandler)}
+    , endpoint_{std::move(endpoint)}
+    , tagDecoratorFactory_{tagDecoratorFactory}
```
godexsoft, line 189: Worth checking if we should move this instead. I believe we were relying on copy elision but it may not work as expected

src/web/ng/impl/Concepts.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include <boost/beast/core/basic_stream.hpp>
+#include <boost/beast/core/tcp_stream.hpp>
+
+#include <type_traits>
+
+namespace web::ng::impl {
+
+template <typename T>
+concept IsTcpStream = std::is_same_v<T, boost::beast::tcp_stream>;
```
godexsoft: Consider `decay_t` here to make sure any cv-qualified `T` can match

tests/common/util/TestHttpClient.hpp
```diff
@@ -0,0 +1,99 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft, line 4: 2024 🔢

tests/common/util/TestHttpClient.cpp
```diff
@@ -0,0 +1,250 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: 2024 🔢

tests/common/util/TestHttpClient.hpp
godexsoft: Or this is a moved file 🤔

tests/common/util/TestWebSocketClient.cpp
```diff
@@ -0,0 +1,225 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2023, the clio developers.
```
godexsoft: Same thing

godexsoft (changes requested):
The code looks great. It's a huge improvement over the current webserver implementation. 
I left a few questions and suggestions. Also, feel free to sprinkle `[[nodiscard]]` where it makes sense and maybe `noexcept` too.

src/web/ng/Connection.hpp
kuznetsss: Maybe it would be better to change all of them at once to not have two different styles in the code. WDYT?

src/web/ng/Server.cpp
kuznetsss: Clang-tidy says that `util::TagDecoratorFactory` is trivally copiable type, so it shouldn't be moved.

tests/common/util/TestHttpClient.hpp
kuznetsss: I think this file is old.

src/web/impl/AdminVerificationStrategy.cpp
kuznetsss: The implementation of this function is very simple:
```cpp
std::shared_ptr<AdminVerificationStrategy>
make_AdminVerificationStrategy(std::optional<std::string> password)
{
    if (password.has_value()) {
        return std::make_shared<PasswordAdminVerificationStrategy>(std::move(*password));
    }
    return std::make_shared<IPAdminVerificationStrategy>();
}
```
It makes both of them based on whether password is presented.

src/web/ng/Connection.hpp
godexsoft: Yeah, both are ok for me with slight preference towards stopping adding ALL_CAPS earlier.

godexsoft (approved):

src/web/ng/impl/WsConnection.hpp
```diff
@@ -0,0 +1,178 @@
...
+        buffer_.consume(buffer_.size());
+
+        return Request{std::move(request), initialRequest_};
+    }
+
+    void
+    close(boost::asio::yield_context yield, std::chrono::steady_clock::duration timeout = DEFAULT_TIMEOUT) override
+    {
+        boost::beast::websocket::stream_base::timeout wsTimeout{};
+        stream_.get_option(wsTimeout);
```
cindyyan317, line 146: do we need timeout here ? seems we don't care about the result of async_close.

src/web/ng/impl/ConnectionHandler.cpp
```diff
@@ -0,0 +1,285 @@
...
+    }
+}
+
+bool
+ConnectionHandler::parallelRequestResponseLoop(Connection& connection, boost::asio::yield_context yield)
+{
+    // atomic_bool is not needed here because everything happening on coroutine's strand
+    bool stop = false;
+    bool closeConnectionGracefully = true;
+    util::CoroutineGroup tasksGroup{yield, maxParallelRequests_};
```
cindyyan317, line 215: Maybe the limit of group size should be got from DosGuard

src/web/ng/Response.cpp
```diff
@@ -0,0 +1,116 @@
...
+    result.keep_alive(httpData_->keepAlive);
+    result.body() = std::move(message_);
+    result.prepare_payload();
+    return result;
+}
+
+boost::asio::const_buffer
+Response::asConstBuffer() const&
+{
+    ASSERT(not httpData_.has_value(), "Losing existing http data");
```
cindyyan317, line 112: Do we must check here ? Why message() does not need this check?

cindyyan317 (approved):
💯  Nice ! I just leave some questions.

src/web/ng/impl/WsConnection.hpp
kuznetsss: We don't want this coroutine to hang because client is not responding.

src/web/ng/Response.cpp
kuznetsss: `message()` just returns underlying message. `asConstBuffer()` is only used to send it by a websocket connection, so this method is only called inside websocket connection processing loop.

src/web/ng/impl/ConnectionHandler.cpp
kuznetsss: Webserver should be completely independent from any other functionality. Also dosguard will do that just on a higher level. This is a protection to not spawn too many coroutines.

PR #1700, opened 2024-10-22
fix: Fix timer spurious calls

### LINKED ISSUE #1634: Timer callback spurious calls
## Issue Description

When using a timer from asio there is a possibility for a spurious call with `success` error code. Such behaviour could lead to UB or a crash by accessing deleted objects if code in timer callback check for error code and if it is a `success` it will do something with local objects.
https://stackoverflow.com/questions/55200310/can-asio-timer-cancel-call-a-spurious-success

We should check timer usages in our code and fix any possible issue. An example of fixing such issue is [in our PR](https://github.com/XRPLF/clio/pull/1633/files#diff-274ff8e1374a8e173db8301965ff3626fff52faf17769d83c19676ec06ae2d27R128).
Maybe if there are many places to be fixed we should write a wrapper for asio timer.

kuznetsss:
Fixes #1634.
I also checked other timers and they don't have the issue.

### DIFF
--- a/src/util/Retry.hpp
+++ b/src/util/Retry.hpp
@@ -24,6 +24,7 @@
 #include <boost/asio/steady_timer.hpp>
 #include <boost/asio/strand.hpp>
 
+#include <atomic>
 #include <chrono>
 #include <cstddef>
 #include <memory>
@@ -80,6 +81,7 @@ class Retry {
     RetryStrategyPtr strategy_;
     boost::asio::steady_timer timer_;
     size_t attemptNumber_ = 0;
+    std::shared_ptr<std::atomic_bool> canceled_{std::make_shared<std::atomic_bool>(false)};
 
 public:
     /**
@@ -91,6 +93,11 @@ public:
     Retry(RetryStrategyPtr strategy, boost::asio::strand<boost::asio::io_context::executor_type> strand);
 
     /**
+     * @brief Destroy the Retry object
+     */
+    ~Retry();
+
+    /**
      * @brief Schedule a retry
      *
      * @tparam Fn The type of the callable to execute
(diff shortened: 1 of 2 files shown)


godexsoft (commented):
Looks good and correct but would be great to add a test or two for this new behaviour 🙏

src/util/Retry.hpp
```diff
@@ -100,15 +107,18 @@ class Retry {
...
     {
+        *canceled_ = false;
         timer_.expires_after(strategy_->getDelay());
         strategy_->increaseDelay();
-        timer_.async_wait([this, func = std::forward<Fn>(func)](boost::system::error_code const& ec) {
-            if (ec == boost::asio::error::operation_aborted) {
-                return;
+        timer_.async_wait(
+            [this, canceled = canceled_, func = std::forward<Fn>(func)](boost::system::error_code const& ec) {
+                if (ec == boost::asio::error::operation_aborted or *canceled) {
```
cindyyan317, line 115: The error check is different here https://github.com/XRPLF/clio/blob/develop/src/util/requests/impl/WsConnectionImpl.hpp#L137. 
Not sure which one is correct.

cindyyan317 (approved):
Just a small question about error code.

src/util/Retry.hpp
kuznetsss: It is `if (...) do_something();` there and here it is `if (...) return;` that's why they are different.

kuznetsss:
> Maybe we should rename test to something else

Changed class is already covered with tests and the spurious call is not so easy to catch, because it should be very specific timing.

godexsoft (approved):

PR #1701, opened 2024-10-23
fix: Add queue size limit for websocket

cindyyan317:
For slow clients, we will disconnect with it if the message queue is too long.

### DIFF
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -77,7 +77,7 @@
         // send a reply for each request whenever it is ready.
         "parallel_requests_limit": 10, // Optional parameter, used only if "processing_strategy" is "parallel". It limits the number of requests for one client connection processed in parallel. Infinite if not specified.
         // Max number of responses to queue up before sent successfully. If a client's waiting queue is too long, the server will close the connection.
-        "ws_max_sending_queue_size": 80000
+        "ws_max_sending_queue_size": 10000
     },
     // Time in seconds for graceful shutdown. Defaults to 10 seconds. Not fully implemented yet.
     "graceful_period": 10.0,
--- a/docs/examples/config/example-config.json
+++ b/docs/examples/config/example-config.json
@@ -77,7 +77,7 @@
         // send a reply for each request whenever it is ready.
         "parallel_requests_limit": 10, // Optional parameter, used only if "processing_strategy" is "parallel". It limits the number of requests for one client connection processed in parallel. Infinite if not specified.
         // Max number of responses to queue up before sent successfully. If a client's waiting queue is too long, the server will close the connection.
-        "ws_max_sending_queue_size": 10000
+        "ws_max_sending_queue_size": 1500
     },
     // Time in seconds for graceful shutdown. Defaults to 10 seconds. Not fully implemented yet.
     "graceful_period": 10.0,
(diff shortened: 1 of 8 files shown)


src/web/HttpSession.hpp
```diff
@@ -51,13 +52,15 @@ template <SomeServerHandler HandlerType>
 class HttpSession : public impl::HttpBase<HttpSession, HandlerType>,
                     public std::enable_shared_from_this<HttpSession<HandlerType>> {
     boost::beast::tcp_stream stream_;
+    util::Config config_;
```
cindyyan317: Let me know if reference is preferred. 👉

docs/examples/config/example-config.json
```diff
@@ -70,7 +70,9 @@
         "admin_password": "xrp",
         // If local_admin is true, Clio will consider requests come from 127.0.0.1 as admin requests
         // It's true by default unless admin_password is set,'local_admin' : true and 'admin_password' can not be set at the same time
-        "local_admin": false
+        "local_admin": false,
+        // Max number of responses to queue up before sent successfully. If a client's waiting queue is too long, the server will close the connection.
+        "ws_max_sending_queue_size": 80000
```
kuznetsss: Maybe 800 is enough?

src/web/HttpSession.hpp
kuznetsss: Why not just the limit for the queue size?

src/web/impl/WsBase.hpp
```diff
@@ -102,22 +101,19 @@ class WsBase : public ConnectionBase, public std::enable_shared_from_this<WsBase
...
-              "The total length of messages in the queue"
-          ))
-        , buffer_(std::move(buffer))
-        , dosGuard_(dosGuard)
-        , handler_(handler)
+        : ConnectionBase(tagFactory, ip), buffer_(std::move(buffer)), dosGuard_(dosGuard), handler_(handler)
     {
+        // If the transactions number is 200 per ledger, A client which subscribes everything will send 400 feeds for
+        // each ledger. Assume 2s per ledger, we allow user delay 80000/400 * 2 = 400s
+        maxSendingQueueSize_ = config.valueOr("ws_max_sending_queue_size", 80000);
```
kuznetsss: This should be initialised before `{`, otherwise there is an uninitialised field for a bit.

src/web/impl/WsBase.hpp
```diff
@@ -181,8 +174,12 @@ class WsBase : public ConnectionBase, public std::enable_shared_from_this<WsBase
         boost::asio::dispatch(
             derived().ws().get_executor(),
             [this, self = derived().shared_from_this(), msg = std::move(msg)]() {
+                if (messages_.size() > maxSendingQueueSize_) {
+                    wsFail(boost::asio::error::timed_out, "Client is too slow");
```
kuznetsss, line 177: Should we change `wsFail()` to log such errors in webserver's log?

src/web/HttpSession.hpp
cindyyan317: I think maybe in future there will be more config?

src/web/impl/WsBase.hpp
cindyyan317: Good idea

src/web/HttpSession.hpp
kuznetsss: This code will be deleted soon, so there will be not many changes here.

src/web/impl/WsBase.hpp
cindyyan317: We 'd better not to remove the perf log. Maybe we can leave it as it is, since it is error level.

kuznetsss (approved):
👍

PR #1712, opened 2024-11-01
feat: Add Support Credentials for Clio

### LINKED ISSUE #1727: Support Credentials on Clio Side
Rippled released new [Credential feature](https://github.com/XRPLF/rippled/pull/5103)

Clio should support it

PeterChen13579:
fixes #1727


src/rpc/CredentialHelpers.hpp
```diff
@@ -0,0 +1,48 @@
...
+/** @brief Check if credential is expired
+ *
+ * @param sleCred The credential to check
+ * @param closed The time that the ledger was closed
+ * @return true if credential not expired, false otherwise
+ */
+bool
+checkExpired(ripple::SLE const& sleCred, ripple::NetClock::time_point const& closed);
+
+/** @brief Check if if there are duplicate Account and CredentialID pairs
```
cindyyan317: Judging by this comment, I would expect the function returns a boolean. But this function seems be a converter.

src/rpc/common/Validators.hpp
```diff
@@ -530,4 +527,38 @@ struct CustomValidators final {
...
+     */
+    [[nodiscard]] static MaybeError
+    verify(boost::json::value const& value, std::string_view key)
+    {
+        if (not value.is_object() or not value.as_object().contains(key.data()))
+            return {};  // ignore. If field is supposed to exist, let 'required' fail instead
+
+        auto const& res = value.as_object().at(key.data());
+        if (!res.is_array()) {
+            return Error{Status{RippledError::rpcINVALID_PARAMS, "Credentials an array of CredentialID(hash256)"}};
```
cindyyan317: We'd better not to mention `Credentials` here. It is supposed to be a generic validator for array of uint256.

cindyyan317: Please add unittest.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +103,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
         auto const owner = util::parseBase58Wrapper<ripple::AccountID>(
             boost::json::value_to<std::string>(input.depositPreauth->at(JS(owner)))
         );
-        auto const authorized = util::parseBase58Wrapper<ripple::AccountID>(
-            boost::json::value_to<std::string>(input.depositPreauth->at(JS(authorized)))
-        );
+        // Only one of authorize or authorize_credentials MUST exist;
+        if (!input.depositPreauth->contains(JS(authorized)) &&
```
cindyyan317: Can we merge these two checks into one? Can below check work?
```
input.depositPreauth->contains(JS(authorized) == input.depositPreauth->contains(JS(authorize_credentials))
```

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +103,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
+            auto const authorized = util::parseBase58Wrapper<ripple::AccountID>(
+                boost::json::value_to<std::string>(input.depositPreauth->at(JS(authorized)))
+            );
+            key = ripple::keylet::depositPreauth(*owner, *authorized).key;
+        } else {
+            auto const authorizedCredentials =
+                util::parseAuthorizeCredentials(input.depositPreauth->at(JS(authorize_credentials)).as_array());
+            // TODO: Move this check into validation.hpp
+            if (authorizedCredentials.size() > ripple::maxCredentialsArraySize)
+                return Error{Status{RippledError::rpcINVALID_PARAMS, "an array of CredentialID(hash256)"}};
```
cindyyan317: More descriptive error message will be useful.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+
+            depositAuthorized =
+                sharedPtrBackend_
+                    ->fetchLedgerObject(
+                        ripple::keylet::depositPreauth(*destinationAccountID, sorted).key, lgrInfo.seq, ctx.yield
+                    )
+                    .has_value();
+        } else {
+            depositAuthorized = sharedPtrBackend_
+                                    ->fetchLedgerObject(
```
cindyyan317: I think we can put the `fetchLedgerObject` outside the if branch. In If branch, we only decide which key we should fetch.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+            auto const credKeylet = ripple::keylet::credential(credHash).key;
+            auto const credLedgerObject = sharedPtrBackend_->fetchLedgerObject(credKeylet, lgrInfo.seq, ctx.yield);
+            auto credIt = ripple::SerialIter{credLedgerObject->data(), credLedgerObject->size()};
+            auto sleCred = ripple::SLE{credIt, credKeylet};
+
+            if (!credLedgerObject || (sleCred.getType() != ripple::ltCREDENTIAL) ||
+                ((sleCred.getFieldU32(ripple::sfFlags) & ripple::lsfAccepted) == 0u))
+                return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+            if (checkExpired(sleCred, lgrInfo.closeTime))
```
cindyyan317: In rippled's PR, it is `parentCloseTime`. Please double check, make sure we are not comparing with wrong time.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+                return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+            if (reqAuth) {
```
cindyyan317: if reqAuth is false, do we need to loop the `credentials`?

src/rpc/CredentialHelpers.hpp
cindyyan317: I understood this function was from rippled's code. 
I think we can do better than `Return empty if there is duplicate` .
Please add a separate function to do duplication check. By adhering to `Single Responsibility Principle `, we ensure that a function is responsible for just one task , making the code easier to understand, test, and modify without unintended side effects.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +103,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
-        key = ripple::keylet::depositPreauth(*owner, *authorized).key;
+            auto const authorized = util::parseBase58Wrapper<ripple::AccountID>(
+                boost::json::value_to<std::string>(input.depositPreauth->at(JS(authorized)))
+            );
+            key = ripple::keylet::depositPreauth(*owner, *authorized).key;
+        } else {
+            auto const authorizedCredentials =
+                util::parseAuthorizeCredentials(input.depositPreauth->at(JS(authorize_credentials)).as_array());
+            // TODO: Move this check into validation.hpp
+            if (authorizedCredentials.size() > ripple::maxCredentialsArraySize)
```
cindyyan317: Do we need to check the emptiness of `authorizedCredentials`?

src/rpc/handlers/LedgerEntry.hpp
```diff
@@ -194,7 +196,8 @@ class LedgerEntryHandler {
                       meta::WithCustomError{
                           validation::CustomValidators::AccountBase58Validator, Status(ClioError::rpcMALFORMED_OWNER)
                       }},
-                     {JS(authorized), validation::Required{}, validation::CustomValidators::AccountBase58Validator},
+                     {JS(authorized), validation::CustomValidators::AccountBase58Validator},
+                     {JS(authorize_credentials), validation::Type<boost::json::array>{}, validation::Hex256ItemType()}
```
cindyyan317: What is the expected `authorize_credentials` ? array of Hex256?

cindyyan317 (changes requested):
👍  Most of work are done.  Let's polish it a bit.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
 
+    auto it = ripple::SerialIter{dstAccountLedgerObject->data(), dstAccountLedgerObject->size()};
+    auto sleDest = ripple::SLE{it, dstKeylet};
+    bool const reqAuth = ((sleDest.getFieldU32(ripple::sfFlags) & ripple::lsfDepositAuth) != 0u) &&
+        (sourceAccountID != destinationAccountID);
+    bool const credentialsPresent = input.credentials.has_value();
+
+    ripple::STArray authCreds;
+    if (credentialsPresent) {
+        // TODO: move this check into validation.hpp
```
kuznetsss: Is it not outdated TODO?

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+        }
+    }
+
+    // If the two accounts are the same OR if that flag is
+    // not set, then the deposit should be fine.
+    bool depositAuthorized = true;
+
+    if (reqAuth) {
+        if (credentialsPresent) {
+            auto const sorted = makeSorted(authCreds);
```
kuznetsss:
```suggestion
            auto const sorted = makeSorted(std::move(authCreds));
```
Probably you will need to update `makeSorted()` as well.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,81 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+            if (checkExpired(sleCred, lgrInfo.closeTime))
+                return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+            if (reqAuth) {
+                auto credential = ripple::STObject::makeInnerObject(ripple::sfCredential);
+                credential.setAccountID(ripple::sfIssuer, sleCred.getAccountID(ripple::sfIssuer));
+                credential.setFieldVL(ripple::sfCredentialType, sleCred.getFieldVL(ripple::sfCredentialType));
+                authCreds.push_back(std::move(credential));
+            }
+        }
```
kuznetsss: Maybe move this loop into a separate function returning `std::expected<ripple::STArray, Error>`?

src/rpc/handlers/LedgerEntry.cpp
kuznetsss: I think it doesn't work if both are false.

tests/unit/rpc/handlers/DepositAuthorizedTests.cpp
```diff
@@ -531,3 +553,285 @@ TEST_F(RPCDepositAuthorizedTest, DifferentAccountsWithDepositAuthFlagReturnsTrue
...
+        }})",
+        CREDENTIALHASH  // CREDENTIALHASH should match credentialIndex
+    );
+
+    backend->setRange(10, 30);
+
+    auto ledgerHeader = CreateLedgerHeader(LEDGERHASH, 30);
+
+    ON_CALL(*backend, fetchLedgerByHash(ripple::uint256{LEDGERHASH}, _)).WillByDefault(Return(ledgerHeader));
+    EXPECT_CALL(*backend, fetchLedgerByHash).Times(1);
```
kuznetsss: `Times(1)` is by default. You don't need to explicitly specify it.

Use `EXPECT_CALL(...).WillOnce(...)` instead of two lines (here and below).

src/rpc/handlers/DepositAuthorized.cpp
PeterChen13579: Doesn't seem to work with std::move(). Clang-tidy gives `Rvalue reference parameter 'in' is never moved from inside the function body`

src/rpc/handlers/LedgerEntry.cpp
PeterChen13579: it works if both are false 😄

src/rpc/handlers/LedgerEntry.cpp
PeterChen13579: checked

src/rpc/handlers/DepositAuthorized.cpp
PeterChen13579: I added this myself

src/rpc/CredentialHelpers.cpp
```diff
@@ -54,4 +58,13 @@ makeSorted(ripple::STArray const& in)
     return out;
 }
 
-}  // namespace rpc
+std::expected<std::set<std::pair<ripple::AccountID, ripple::Slice>>, Status>
+createAuthCredentials(ripple::STArray const& in)
```
kuznetsss: You should move the body of `makeSorted()` into this function and return unexpected when insert failed.

src/rpc/RPCHelpers.cpp
```diff
@@ -1327,4 +1329,56 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
     return obj;
 }
 
+std::expected<ripple::STArray, Status>
+fetchCredentials(
```
kuznetsss: Could this be useful for any other handler? I would prefer to have it as a local function in LedgerEntry.cpp instead.

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -86,65 +86,30 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
-                credential.setAccountID(ripple::sfIssuer, sleCred.getAccountID(ripple::sfIssuer));
-                credential.setFieldVL(ripple::sfCredentialType, sleCred.getFieldVL(ripple::sfCredentialType));
-                authCreds.push_back(std::move(credential));
-            }
-        }
+    if (credentialsPresent && reqAuth) {
+        auto const creds = fetchCredentials(input.credentials, sharedPtrBackend_, lgrInfo, ctx);
+        if (!creds.has_value())
+            return Error{creds.error()};
+        authCreds = creds.value();
```
kuznetsss:
```suggestion
            return Error{std::move(creds).error()};
        authCreds = std::move(creds).value();
```

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -86,65 +86,30 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
-
-            depositAuthorized =
-                sharedPtrBackend_
-                    ->fetchLedgerObject(
-                        ripple::keylet::depositPreauth(*destinationAccountID, sorted).key, lgrInfo.seq, ctx.yield
-                    )
-                    .has_value();
+            auto const sorted = credentials::createAuthCredentials(authCreds);
+            if (!sorted)
+                return Error{sorted.error()};
```
kuznetsss:
```suggestion
                return Error{std::move(sorted).error()};
```

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -104,27 +104,23 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
             boost::json::value_to<std::string>(input.depositPreauth->at(JS(owner)))
         );
         // Only one of authorize or authorize_credentials MUST exist;
-        if (!input.depositPreauth->contains(JS(authorized)) &&
-            !input.depositPreauth->contains(JS(authorize_credentials)))
+        if (input.depositPreauth->contains(JS(authorized)) == input.depositPreauth->contains(JS(authorize_credentials)))
```
kuznetsss: Should it be an error if they are both false?

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -104,27 +104,23 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
                 util::parseAuthorizeCredentials(input.depositPreauth->at(JS(authorize_credentials)).as_array());
-            // TODO: Move this check into validation.hpp
-            if (authorizedCredentials.size() > ripple::maxCredentialsArraySize)
-                return Error{Status{RippledError::rpcINVALID_PARAMS, "an array of CredentialID(hash256)"}};
 
-            auto const sorted = makeSorted(authorizedCredentials);
-            key = ripple::keylet::depositPreauth(owner.value(), sorted).key;
+            auto const sorted = credentials::createAuthCredentials(authorizedCredentials);
+            if (!sorted.has_value())
+                return Error{sorted.error()};
```
kuznetsss:
```suggestion
                return Error{std::move(sorted).error()};
```

src/rpc/handlers/LedgerEntry.cpp
PeterChen13579: yes, it should error out

src/rpc/CredentialHelpers.cpp
PeterChen13579: I think the reason I kept `makeSorted()` is because rippled has the exact same function with the same logic and we wanted to copy it. Although I definitely think it would be better to just rename `createAuthCredentials ` to `makeSorted()` and move all logic inside. WDYT @cindyyan317 ?

cindyyan317: I think my point is we should not return empty set but want to express there is duplicate in array. If array is empty, `makeSorted` will return empty set but it does not have duplicate. 
I'd rather merge makeSorted logic into `createAuthCredentials` .

src/rpc/RPCHelpers.cpp
PeterChen13579: Just noticed this function is used in [Deposit_authorized](https://github.com/XRPLF/clio/pull/1712/files#diff-55fbb4e01cb85ae75c92a13d5d06dc8be977bf9f132302fe95f8c1339af6870bR90) RPC call not in LedgerEntry, so i don't think it can go there. It may be used by other RPC methods in the future if they require to retrieve credentials

src/rpc/RPCHelpers.cpp
```diff
@@ -1347,18 +1348,13 @@ fetchCredentials(
...
 
         ripple::uint256 credHash;
-        if (!credHash.parseHex(boost::json::value_to<std::string>(elem))) {
-            return Error{Status{
-                RippledError::rpcINVALID_PARAMS, "Invalid field 'credentials', not an array of CredentialID(hash256)."
-            }};
-        }
+        ASSERT(
+            credHash.parseHex(boost::json::value_to<std::string>(elem)),
+            "should already be checked in validators.hpp that elem is a uin256 hex"
```
shawnxie999:
```suggestion
            "should already be checked in validators.hpp that elem is a uint256 hex"
```

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,46 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
 
     Output response;
 
+    auto it = ripple::SerialIter{dstAccountLedgerObject->data(), dstAccountLedgerObject->size()};
+    auto sleDest = ripple::SLE{it, dstKeylet};
```
shawnxie999:
```suggestion
    auto const sleDest = ripple::SLE{it, dstKeylet};
```

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,46 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+    bool const reqAuth = ((sleDest.getFieldU32(ripple::sfFlags) & ripple::lsfDepositAuth) != 0u) &&
```
shawnxie999:
```suggestion
    bool const reqAuth = sleDest.isFlag(ripple::lsfDepositAuth) &&
```

src/rpc/CredentialHelpers.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <set>
+#include <utility>
+
+namespace rpc::credentials {
+
+bool
+checkExpired(ripple::SLE const& sleCred, ripple::LedgerHeader const& ledger)
+{
+    if (sleCred.isFieldPresent(ripple::sfExpiration)) {
+        std::uint32_t exp = sleCred.getFieldU32(ripple::sfExpiration);
```
cindyyan317: why not make it also a const?

src/util/AccountUtils.hpp
```diff
@@ -64,4 +77,39 @@ parseBase58Wrapper(ripple::TokenType type, std::string const& str)
...
+ * @param jv The boost json array of credentials to parse
+ * @return Array of credentials after parsing
+ */
+inline ripple::STArray
+parseAuthorizeCredentials(boost::json::array const& jv)
+{
+    ripple::STArray arr;
+    for (auto const& jo : jv) {
+        auto const issuer = ripple::parseBase58<ripple::AccountID>(
+            static_cast<std::string>(jo.at(ripple::jss::issuer.c_str()).as_string())
```
cindyyan317: Can we use JS(issuer) ?

src/util/AccountUtils.hpp
```diff
@@ -64,4 +77,39 @@ parseBase58Wrapper(ripple::TokenType type, std::string const& str)
...
+        );
+        ASSERT(
+            issuer.has_value(), "issuer must be present, should already be checked in AuthorizeCredentialValidator."
+        );
+
+        auto const credentialType =
+            ripple::strUnHex(static_cast<std::string>(jo.at(ripple::jss::credential_type.c_str()).as_string()));
```
cindyyan317: JS(credential_type)

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +101,26 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
         auto const owner = util::parseBase58Wrapper<ripple::AccountID>(
             boost::json::value_to<std::string>(input.depositPreauth->at(JS(owner)))
         );
-        auto const authorized = util::parseBase58Wrapper<ripple::AccountID>(
-            boost::json::value_to<std::string>(input.depositPreauth->at(JS(authorized)))
-        );
+        // Only one of authorize or authorized_credentials MUST exist;
+        if (input.depositPreauth->contains(JS(authorized)) ==
```
cindyyan317, line 105: Do we have unittest for this error? 
Both "authorized" "authorized_credentials" exist / not exist

src/util/AccountUtils.hpp
```diff
@@ -64,4 +77,39 @@ parseBase58Wrapper(ripple::TokenType type, std::string const& str)
...
 }
 
+/**
+ * @brief Parses each credential object and makes sure the credential type and value are correct
+ *
+ * @param jv The boost json array of credentials to parse
+ * @return Array of credentials after parsing
+ */
+inline ripple::STArray
+parseAuthorizeCredentials(boost::json::array const& jv)
```
cindyyan317: It is not really related with Account. Can we move it to CredentialsHelper?

src/rpc/RPCHelpers.cpp
```diff
@@ -1329,4 +1332,51 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
...
+std::expected<ripple::STArray, Status>
+fetchCredentials(
+    std::optional<boost::json::array> const& credID,
+    std::shared_ptr<BackendInterface> const& backend,
+    ripple::LedgerHeader const& info,
+    Context const& ctx
+)
+{
+    ripple::STArray authCreds;
+    if (credID.value().size() > ripple::maxCredentialsArraySize) {
```
cindyyan317: Do you still need this? The error message looks off too. I think we shall leave the field validation job in handler not helper function. Consider it maybe used by other place.

src/rpc/RPCHelpers.cpp
```diff
@@ -1329,4 +1332,51 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
...
+
+        if (!credLedgerObject || (sleCred.getType() != ripple::ltCREDENTIAL) ||
+            ((sleCred.getFieldU32(ripple::sfFlags) & ripple::lsfAccepted) == 0u))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+        if (credentials::checkExpired(sleCred, info))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+        auto credential = ripple::STObject::makeInnerObject(ripple::sfCredential);
+        credential.setAccountID(ripple::sfIssuer, sleCred.getAccountID(ripple::sfIssuer));
```
cindyyan317: I am thinking do we need to convert it to STObject?  in `createAuthCredentials` , you will depack it anyway?

src/rpc/handlers/LedgerEntry.cpp
PeterChen13579: will add 👍

src/rpc/RPCHelpers.cpp
PeterChen13579: Yea, the error is required for [this](url) test. The error message needs to be fixed though

src/rpc/RPCHelpers.cpp
PeterChen13579: Mainly because rippled's side converted to a credential object with `issuer` and `credential_type` field; but yes you're right that it will get depacked anyways. Do u think we should just not convert then or leave it the same?

src/rpc/RPCHelpers.cpp
```diff
@@ -1391,4 +1395,51 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
...
+
+        if (!credLedgerObject || (sleCred.getType() != ripple::ltCREDENTIAL) ||
+            ((sleCred.getFieldU32(ripple::sfFlags) & ripple::lsfAccepted) == 0u))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+        if (credentials::checkExpired(sleCred, info))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS}};
+
+        auto const [it, ins] = authCreds.insert(
+            {sleCred.getAccountID(ripple::sfIssuer), ripple::makeSlice(sleCred.getFieldVL(ripple::sfCredentialType))}
```
cindyyan317: slice does not hold the memory. Is it okay? `sleCred.getFieldVL(ripple::sfCredentialType)` returns a temporary  Blob, slice will point to invalid memory.

src/rpc/CredentialHelpers.hpp
```diff
@@ -0,0 +1,48 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
godexsoft, line 19: Looks like missing `#pragma once`

src/rpc/RPCHelpers.hpp
```diff
@@ -747,4 +749,21 @@ getDeliveredAmount(
...
+ * @brief Get Array of Credential objects
+ *
+ * @param credID Array of CredentialID's to parse
+ * @param backend backend interface
+ * @param info The ledger header
+ * @param yield The coroutine context
+ * @return Array of credential objects, error if failed otherwise
+ */
+std::expected<ripple::STArray, Status>
+fetchCredentialArray(
```
godexsoft: Is there any specific reason why this is not in CredentialHelpers? 

I'd love to try and avoid adding more things to RPCHelpers (or any *Helpers). Imo adding a new file like you did is a better choice than adding to `RPCHelpers` that are already out of control in size and contain multiple functions that are only used once (or never) :rage4: 

Also, if the functions are to only be used in one place, maybe better keep them in the `impl` of the RPC handler (or other unit) that uses the function? Is there a real reason to create more shared helpers here?
I'm not opposed the current design (and we can always change it later) but in an ideal world i'd like to evict as many functions from `*Helpers` as possible and give them a new home in `impl` of their relevant handlers/feeds 👍

src/rpc/handlers/DepositAuthorized.hpp
```diff
@@ -59,6 +61,7 @@ class DepositAuthorizedHandler {
         std::string destinationAccount;
         std::string ledgerHash;
         uint32_t ledgerIndex{};
+        boost::json::array credentials;
         // validated should be sent via framework
```
godexsoft, line 66: Looked better with a space, no? 😅

godexsoft (changes requested):
Noticed a few things while browsing 🔢

src/rpc/RPCHelpers.hpp
PeterChen13579: I think I left it in RPC helpers because it uses all the backend interface/ledger/context yield stuff, but yea, ill move it to credentialHelpers 😄 .
Also was thinking in the future this `fetchCredentialArray` func would be useful for any RPC that uses credentials, so perhaps will be used more than once in the future. 
I def think everything in `impl` with the relevant handlers is the best scenario yea. 👍

src/rpc/RPCHelpers.cpp
```diff
@@ -21,10 +21,12 @@
 
 #include "data/BackendInterface.hpp"
 #include "data/Types.hpp"
+#include "rpc/CredentialHelpers.hpp"
```
cindyyan317: RPCHelpers.hpp and RPCHelpers.cpp only have headers change. Do we still need to update these two files in this PR?

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -71,26 +79,44 @@ DepositAuthorizedHandler::process(DepositAuthorizedHandler::Input input, Context
...
+    auto const sleDest = ripple::SLE{it, dstKeylet};
+    bool const reqAuth = sleDest.isFlag(ripple::lsfDepositAuth) && (sourceAccountID != destinationAccountID);
+    bool const credentialsPresent = input.credentials.has_value();
+
+    ripple::STArray authCreds;
+    // If the two accounts are the same OR if that flag is
+    // not set, then the deposit should be fine.
+    bool depositAuthorized = true;
+    if (credentialsPresent && reqAuth) {
+        auto const creds = credentials::fetchCredentialArray(input.credentials, *sharedPtrBackend_, lgrInfo, ctx.yield);
```
cindyyan317: Can input.credentials be empty?

src/rpc/common/Validators.cpp
```diff
@@ -253,4 +255,58 @@ CustomValidator CustomValidators::CurrencyIssueValidator =
...
+        return MaybeError{};
+    }};
+
+CustomValidator CustomValidators::AuthorizeCredentialValidator =
+    CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (not value.is_array())
+            return Error{Status{RippledError::rpcINVALID_PARAMS, std::string(key) + " not array"}};
+
+        auto const& authCred = value.as_array();
+        if (authCred.size() == 0) {
```
cindyyan317, line 284: While rippled will return 
```
      
{
  "result": {
    "error": "malformedAuthorizedCredentials",
    "error_message": null,
    "ledger_current_index": 5782360,
    "request": {
      "command": "ledger_entry",
      "deposit_preauth": {
        "authorized_credentials": [],
        "owner": "rNYYfp6ivxYdzMz1deFLAg3RoeBpCc7j3u"
      }
    },
    "status": "error",
    "validated": false
  }
}
    
```
Clio returns
```
{
  "result": {
    "error": "invalidParams",
    "error_code": 31,
    "error_message": "Requires at least one element in authorized_credentials array",
    "status": "error",
    "type": "response",
    "request": {
      "method": "ledger_entry",
      "params": [
        {
          "deposit_preauth": {
            "owner": "rNYYfp6ivxYdzMz1deFLAg3RoeBpCc7j3u",
            "authorized_credentials": []
          }
        }
      ]
    }
  },
  "warnings": [
    {
      "id": 2001,
    },
    {
      "id": 2002,
    }
  ]
}
```
for request
```
{
    "method": "ledger_entry",
    "params": [
        {
               "deposit_preauth": {
                "owner" : "rNYYfp6ivxYdzMz1deFLAg3RoeBpCc7j3u",
                "authorized_credentials":[]
                }
        }
    ]
}
```

src/rpc/CredentialHelpers.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+    ripple::STArray arr;
+    for (auto const& jo : jv) {
+        auto const issuer =
+            ripple::parseBase58<ripple::AccountID>(static_cast<std::string>(jo.at(JS(issuer)).as_string()));
+        ASSERT(
+            issuer.has_value(), "issuer must be present, should already be checked in AuthorizeCredentialValidator."
+        );
+
+        auto const credentialType = ripple::strUnHex(static_cast<std::string>(jo.at(JS(credential_type)).as_string()));
```
cindyyan317, line 96: There is a check for credential_type's length against `maxCredentialTypeLength`. We need to also add it.

src/rpc/handlers/DepositAuthorized.hpp
```diff
@@ -99,6 +104,7 @@ class DepositAuthorizedHandler {
             {JS(destination_account), validation::Required{}, validation::CustomValidators::AccountValidator},
             {JS(ledger_hash), validation::CustomValidators::Uint256HexStringValidator},
             {JS(ledger_index), validation::CustomValidators::LedgerIndexValidator},
+            {JS(credentials), validation::Type<boost::json::array>{}, validation::Hex256ItemType()}
```
cindyyan317, line 107: rippled returns 
```
      
{
  "result": {
    "error": "invalidParams",
    "error_code": 31,
    "error_message": "Invalid field 'credentials', not is non-empty array of CredentialID(hash256).",
    "request": {
      "command": "deposit_authorized",
      "credentials": [],
      "destination_account": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF",
      "source_account": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
    },
    "status": "error"
  }
}
```
while clio returns 
```
      
{
  "result": {
    "deposit_authorized": true,
    "source_account": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
    "destination_account": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF",
    "ledger_hash": "11878C8BD7DC0EF08E186F8D537D5650A205087CA68530AB65978F376FC4113E",
    "ledger_index": 5622773,
    "validated": true,
    "credentials": [],
    "status": "success"
  },
  "warnings": [
    {
      "id": 2001,
    },
    {
      "id": 2002,
    }
  ]
}
```
when the request is
```
{
    "method": "deposit_authorized",
    "params": [
        {
            "source_account" : "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
            "destination_account": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF",
            "credentials":[
            ]
            
        }
    ]
}
```

src/rpc/common/Validators.hpp
```diff
@@ -559,6 +557,53 @@ struct CustomValidators final {
...
+    {
+        if (not value.is_object() or not value.as_object().contains(key.data()))
+            return {};  // ignore. If field is supposed to exist, let 'required' fail instead
+
+        auto const& res = value.as_object().at(key.data());
+
+        // loop through each item in the array and make sure it is uint256 hex string
+        for (auto const& elem : res.as_array()) {
+            ripple::uint256 num;
+            if (!num.parseHex(elem.as_string()))
```
cindyyan317: What happen if elem is not a string?
```
{
    "method": "deposit_authorized",
    "params": [
        {
            "source_account" : "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
            "destination_account": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF",
            "credentials":[123]
        }
    ]
}
```
Clio returns `internal error`. We need to fix it.

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +101,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
```
cindyyan317, line 105: rippled returns
```
{
  "result": {
    "error": "malformedRequest",
    "error_message": null,
    "ledger_current_index": 5784418,
    "request": {
      "command": "ledger_entry",
      "deposit_preauth": {
        "authorized": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF",
        "authorized_credentials": [
          {
            "credential_type": "123",
            "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
          }
        ],
        "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
      }
    },
    "status": "error",
    "validated": false
  }
}
```
while Clio returns
```
{
  "result": {
    "error": "badCredentials",
    "error_code": 95,
    "error_message": "Must have one of authorized or authorized_credentials.",
    "status": "error",
    "type": "response",
    "request": {
      "method": "ledger_entry",
      "params": [
        {
          "deposit_preauth": {
            "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
            "authorized_credentials": [
              {
                "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                "credential_type": "123"
              }
            ],
            "authorized": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF"
          }
        }
      ]
    }
  },
  "warnings": [
    {
      "id": 2001,
    },
    {
      "id": 2002,
    }
  ]
}
```
when request is
```
{
    "method": "ledger_entry",
    "params": [
        {
            "deposit_preauth" : {
                "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                "authorized_credentials":[
                    {
                        "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                        "credential_type": "123"
                    }
                ],
                "authorized": "r4gQWHn4tZk9ziKjA69XyPqX3jGYztbxqF"
            }
        }
    ]
}
```

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -97,11 +101,29 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input input, Context const& ctx)
...
+                boost::json::value_to<std::string>(input.depositPreauth->at(JS(authorized)))
+            );
+            key = ripple::keylet::depositPreauth(*owner, *authorized).key;
+        } else {
+            auto const authorizedCredentials = rpc::credentials::parseAuthorizeCredentials(
+                input.depositPreauth->at(JS(authorized_credentials)).as_array()
+            );
 
-        key = ripple::keylet::depositPreauth(*owner, *authorized).key;
+            auto const sorted = credentials::createAuthCredentials(authorizedCredentials);
```
cindyyan317: When there is duplicate 
request like
```
{
    "method": "ledger_entry",
    "params": [
        {
            "deposit_preauth" : {
                "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                "authorized_credentials":[
                    {
                        "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                        "credential_type": "123"
                    },
                    {
                        "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                        "credential_type": "123"
                    }
                ]
            }
        }
    ]
}
```
rippled returns:
```
      
{
  "result": {
    "error": "malformedAuthorizedCredentials",
    "error_message": null,
    "ledger_current_index": 5784511,
    "request": {
      "command": "ledger_entry",
      "deposit_preauth": {
        "authorized_credentials": [
          {
            "credential_type": "123",
            "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
          },
          {
            "credential_type": "123",
            "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
          }
        ],
        "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d"
      }
    },
    "status": "error",
    "validated": false
  }
}
    
```
Clio returns:
```
      
{
  "result": {
    "error": "badCredentials",
    "error_code": 95,
    "error_message": "duplicates in credentials.",
    "status": "error",
    "type": "response",
    "request": {
      "method": "ledger_entry",
      "params": [
        {
          "deposit_preauth": {
            "owner": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
            "authorized_credentials": [
              {
                "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                "credential_type": "123"
              },
              {
                "issuer": "rJ7ywemHxbCkGNskVXs8oWciSL77Nnn7d",
                "credential_type": "123"
              }
            ]
          }
        }
      ]
    }
  },
  "warnings": [
    {
      "id": 2001,
    },
    {
      "id": 2002,
    }
  ]
}
    
```

src/rpc/RPCHelpers.cpp
```diff
@@ -81,6 +81,7 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
```
PeterChen13579, line 84: fyi, this is used line [375](https://github.com/XRPLF/clio/pull/1712/files#diff-63a13cef09d0b504ba9246cab2dfade8752f6cca5640b218f6885f3881d385cfL375)

src/rpc/CredentialHelpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+        ripple::uint256 credHash;
+        ASSERT(
+            credHash.parseHex(boost::json::value_to<std::string>(elem)),
+            "should already be checked in validators.hpp that elem is a uint256 hex"
+        );
+
+        auto const credKeylet = ripple::keylet::credential(credHash).key;
+        auto const credLedgerObject = backend.fetchLedgerObject(credKeylet, info.seq, yield);
+        if (!credLedgerObject)
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS, "credentials aren't accepted."}};
```
cindyyan317: message needs to be changed

src/rpc/CredentialHelpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+        auto credIt = ripple::SerialIter{credLedgerObject->data(), credLedgerObject->size()};
+        auto const sleCred = ripple::SLE{credIt, credKeylet};
+
+        if (!credLedgerObject || (sleCred.getType() != ripple::ltCREDENTIAL) ||
```
cindyyan317: no need to check `!credLedgerObject` at this point.

src/rpc/common/Validators.cpp
```diff
@@ -253,4 +255,67 @@ CustomValidator CustomValidators::CurrencyIssueValidator =
...
+            return Error{Status{
+                ClioError::rpcMALFORMED_AUTHORIZED_CREDENTIALS,
+                fmt::format("Requires at least one element in authorized_credentials array")
+            }};
+        }
+
+        if (authCred.size() > ripple::maxCredentialsArraySize) {
+            return Error{Status{
+                RippledError::rpcINVALID_PARAMS,
```
cindyyan317: Error code should be malformedAuthorizedCredentials?

src/rpc/common/Validators.hpp
```diff
@@ -559,6 +557,54 @@ struct CustomValidators final {
...
+
+        auto const& res = value.as_object().at(key.data());
+
+        // loop through each item in the array and make sure it is uint256 hex string
+        for (auto const& elem : res.as_array()) {
+            ripple::uint256 num;
+            if (!elem.is_string() || !num.parseHex(elem.as_string())) {
+                return Error{Status{
+                    RippledError::rpcINVALID_PARAMS,
+                    "Invalid field 'credentials', not an array of CredentialID(hash256)."
```
cindyyan317: You can use a more generic message, like "item is not a valid uint256 type" here. This validator can be used in other place.

src/rpc/CredentialHelpers.cpp
```diff
@@ -0,0 +1,162 @@
...
+        auto const sleCred = ripple::SLE{credIt, credKeylet};
+
+        if ((sleCred.getType() != ripple::ltCREDENTIAL) ||
+            ((sleCred.getFieldU32(ripple::sfFlags) & ripple::lsfAccepted) == 0u))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS, "credentials aren't accepted"}};
+
+        if (credentials::checkExpired(sleCred, info))
+            return Error{Status{RippledError::rpcBAD_CREDENTIALS, "credentials are expired"}};
+
+        if (sleCred.getAccountID(ripple::sfIssuer) != srcAcc)
```
cindyyan317: We should check `sfSubject` not `sfIssuer`

cindyyan317 (changes requested):
Looks good to me . Let's fix the "sfSubject"

cindyyan317 (dismissed):
Thanks @PeterChen13579 for addressing all the comments. Let's ship it

tests/unit/rpc/handlers/CredentialHelpersTests.cpp
```diff
@@ -0,0 +1,164 @@
...
+        ripple::strUnHex(static_cast<std::string>(credential1[JS(credential_type)].as_string())).value();
+
+    EXPECT_EQ(cred.getAccountID(ripple::sfIssuer), expectedIssuer);
+    EXPECT_EQ(cred.getFieldVL(ripple::sfCredentialType), expectedCredentialType);
+}
+
+class CredentialHelperTest : public util::prometheus::WithPrometheus,
+                             public MockBackendTest,
+                             public SyncAsioContextTest {
+    void
```
godexsoft: `SetUp` and `TearDown` are not needed afaik.

tests/unit/rpc/handlers/DepositAuthorizedTests.cpp
```diff
@@ -792,6 +799,48 @@ TEST_F(RPCDepositAuthorizedTest, DuplicateCredentialsReturnsFalse)
...
+    ON_CALL(*backend, doFetchLedgerObject(ripple::keylet::account(GetAccountIDWithString(ACCOUNT)).key, _, _))
+        .WillByDefault(Return(account1Root.getSerializer().peekData()));
+    ON_CALL(*backend, doFetchLedgerObject(ripple::keylet::account(GetAccountIDWithString(ACCOUNT2)).key, _, _))
+        .WillByDefault(Return(account2Root.getSerializer().peekData()));
+
+    EXPECT_CALL(*backend, doFetchLedgerObject).Times(2);
+
+    auto const input = json::parse(fmt::format(
+        R"({{
+        "source_account": "{}",
```
godexsoft: Please format with one tab to the right, like the next test 👍

godexsoft (changes requested):
Spotted a few more small things 👍 
You can fix the `ALL_CAPS` style of statics to `NotAllCaps` or leave it as is and we will get to it in #1680

godexsoft (approved):
Nicely done 👍

PR #1713, opened 2024-11-01
feat: ETLng Registry

godexsoft:
This PR introduces the `Registry`. A central component of the updated ETL. 
`RegistryInterface`, alongside declaring the interface of a Registry, documents what `hooks` valid extensions could define to get data during various stages of the ETL process.

The idea is to use the registry to dispatch the following events to extensions:
- New `ledger data` extracted
- Initial `ledger data` extracted
- Initial ledger loader sent more `objects` for the ledger

Extensions can define one or more valid hook to participate in the load process of ETL. 

The included tests check all the added safety nets (concepts) as well as enforce the correct order of hook execution.


tests/unit/etlng/RegistryTests.cpp
```diff
@@ -0,0 +1,546 @@
...
+#include <xrpl/protocol/TxMeta.h>
+
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+using namespace etlng::impl;
+
+namespace staticTests {
```
kuznetsss: We usually use snake_case for namespaces.

tests/unit/etlng/RegistryTests.cpp
```diff
@@ -0,0 +1,546 @@
...
+    using spec = etlng::model::Spec<
+        ripple::TxType::ttNFTOKEN_CREATE_OFFER,
+        ripple::TxType::ttNFTOKEN_CANCEL_OFFER,
+        ripple::TxType::ttNFTOKEN_ACCEPT_OFFER>;
+    MOCK_METHOD(void, onInitialTransaction, (uint32_t, etlng::model::Transaction const&), (const));
+};
+
+}  // namespace
+
+namespace util {
```
kuznetsss: It is better to create a test structure instead.

tests/unit/etlng/RegistryTests.cpp
```diff
@@ -0,0 +1,546 @@
...
+{
+    auto transactions = std::vector{
+        util::CreateTransaction(ripple::TxType::ttNFTOKEN_BURN),
+        util::CreateTransaction(ripple::TxType::ttNFTOKEN_BURN),
+        util::CreateTransaction(ripple::TxType::ttNFTOKEN_CREATE_OFFER),
+    };
+
+    auto ext = MockExtLedgerData{};
+
+    EXPECT_CALL(ext, onLedgerData(testing::_)).Times(1);  // 1 batch (dispatch call)
```
kuznetsss: `Times(1)` is by default.

tests/unit/etlng/RegistryTests.cpp
```diff
@@ -0,0 +1,546 @@
...
+    });
+}
+
+TEST(RegistryTest, InitialObjectsCorrectOrderOfHookCalls)
+{
+    uint32_t order = 0;
+    auto extObjs = MockExtInitialObjects{};
+    auto extObj = MockExtInitialObject{};
+
+    EXPECT_CALL(extObjs, onInitialObjects).Times(1).WillRepeatedly([&](auto, auto const&, auto) {
```
kuznetsss:
```suggestion
    EXPECT_CALL(extObjs, onInitialObjects).WillOnce([&](auto, auto const&, auto) {
```

tests/unit/etlng/RegistryTests.cpp
```diff
@@ -0,0 +1,546 @@
...
+        .successors = {},
+        .header = header,
+        .rawHeader = {},
+        .seq = SEQ
+    });
+}
+
+TEST(RegistryTest, InitialObjectsCorrectOrderOfHookCalls)
+{
+    uint32_t order = 0;
```
kuznetsss: There is `testing::Sequence` to check mock calls in sequence.

tests/unit/etlng/RegistryTests.cpp
godexsoft: We don't actually have any snake_case namespaces. i'll rename to 'compiletime::checks' i guess.

kuznetsss (dismissed):
👍

src/etlng/Models.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace etlng::model {
+namespace impl {
+
+static consteval auto
+checkNoDuplicates(auto&&... types)
```
cindyyan317: This method seems can be used other place. Can we move it to util. Concepts.hpp?

godexsoft: I'll move it if there is going to be another place to use it in the future 👍

cindyyan317 (dismissed):

cindyyan317 (approved):
Thanks adding it to util 👍

PR #1718, opened 2024-11-07
feat: Upgrade to `libxrpl 2.3.0-rc1`

### LINKED ISSUE #1717: Upgrade to `libxrpl 2.3.0-rc1`
## Summary
New version of `libxrpl` is available. We can start using it.

## Motivation
Keep up with upcoming features of xrpl.

## Solution
Use new package and add new amendments as "known"

godexsoft:
Fixes #1717 

Note that this just silences the unit test for amendments check by listing the new 4 amendments as supported by Clio.
In reality there are separate PRs for some of them and others don't need any special treatment in Clio:
• AMMClawback - no special treatment needed
• fixAMMv1_2 - no special treatment needed
• Credentials - #1712 
• MPTokensV1 - #1147

### DIFF
--- a/src/data/AmendmentCenter.hpp
+++ b/src/data/AmendmentCenter.hpp
@@ -127,6 +127,10 @@ struct Amendments {
     REGISTER(fixInnerObjTemplate2);
     REGISTER(fixNFTokenPageLinks);
     REGISTER(InvariantsV1_1);
+    REGISTER(Credentials);
+    REGISTER(AMMClawback);
+    REGISTER(fixAMMv1_2);
+    REGISTER(MPTokensV1);
 
     // Obsolete but supported by libxrpl
     REGISTER(CryptoConditionsSuite);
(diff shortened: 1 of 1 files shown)


src/data/AmendmentCenter.hpp
```diff
@@ -127,6 +127,10 @@ struct Amendments {
     REGISTER(fixInnerObjTemplate2);
     REGISTER(fixNFTokenPageLinks);
     REGISTER(InvariantsV1_1);
+    REGISTER(Credentials);
+    REGISTER(AMMClawback);
+    REGISTER(fixAMMv1_2);
+    REGISTER(MPTokensV1);
```
shawnxie999: nit: we should reserve the order. MPT amendment should be the first

godexsoft: Is there any specific order? In Clio the amendments end up in an `unordered_set` so it should not matter how we list them here. But i'm happy to change the order if you have some specific one in mind :)

shawnxie999: Not important, but the following order is how rippled orders them and the sequence they were merged:
```
    REGISTER(MPTokensV1);
    REGISTER(fixAMMv1_2);
    REGISTER(AMMClawback);
    REGISTER(Credentials);
```

(also i believe this is the way it's ordered in the MPT and credential clio PRs)

shawnxie999 (approved):

cindyyan317 (approved):

PR #1722, opened 2024-11-08
feat: Integrate new webserver

kuznetsss:
For #919.
The new web server is not using dosguard yet. It will be fixed by a separate PR.


src/web/ng/RPCServerHandler.hpp
kuznetsss, line 1: Need to cover this file with tests.

src/app/ClioApplication.hpp
```diff
@@ -42,10 +42,12 @@ class ClioApplication {
     /**
      * @brief Run the application
      *
+     * @param useNgWebServer Whether to use the new web server
+     *
      * @return exit code
      */
     int
-    run();
+    run(bool useNgWebServer);
```
godexsoft, line 50: Glad doxygen did not agro because this is `const` in the impl.

src/util/CoroutineGroup.cpp
```diff
@@ -25,13 +25,14 @@
...
 #include <cstddef>
+#include <cstdint>
 #include <functional>
 #include <optional>
 #include <utility>
 
 namespace util {
 
-CoroutineGroup::CoroutineGroup(boost::asio::yield_context yield, std::optional<int> maxChildren)
+CoroutineGroup::CoroutineGroup(boost::asio::yield_context yield, std::optional<int64_t> maxChildren)
```
godexsoft: Why not `std::size_t`? or `std::uint64_t`?

src/web/AdminVerificationStrategy.cpp
```diff
@@ -81,19 +81,21 @@ make_AdminVerificationStrategy(std::optional<std::string> password)
...
-        if (adminPassword.has_value())
-            return std::unexpected{"Admin config error, local_admin and admin_password can not be set together."};
-        return std::unexpected{"Admin config error, either local_admin and admin_password must be specified."};
+    auto adminPassword = config.maybeValue<std::string>("server.admin_password");
+    auto const localAdmin = config.maybeValue<bool>("server.local_admin");
+
+    if (adminPassword.has_value() and localAdmin.has_value() and *localAdmin)
+        return std::unexpected{"Admin config error: 'local_admin' and admin_password can not be set together."};
+
+    if (localAdmin.has_value() and not*localAdmin and not adminPassword.has_value()) {
```
godexsoft: `not*localAdmin` :rage4: 😆

src/web/interface/ConnectionBase.hpp
```diff
@@ -94,6 +82,15 @@ struct ConnectionBase : public util::Taggable {
...
     }
 
+    /**
+     * @brief Get the subscription context for this connection.
+     *
+     * @param factory Tag TagDecoratorFactory to use to create the context.
+     * @return The subscription context for this connection.
+     */
+    virtual SubscriptionContextPtr
+    subscriptionContext(util::TagDecoratorFactory const& factory) = 0;
```
godexsoft: nit: The name seems a little off.. maybe `makeSubscriptionContext` or something like that is more self-explanatory?

src/web/ng/Connection.hpp
```diff
@@ -28,26 +28,77 @@
...
+     * @note This function is lazy, it will update isAdmin only if it is not set yet.
+     *
+     * @tparam T The invocable type of the function to call to set the isAdmin.
+     * @param setter The function to call to set the isAdmin.
+     */
+    template <std::invocable T>
+    void
+    setIsAdmin(T&& setter)
+    {
+        if (not isAdmin_.has_value())
```
godexsoft, line 92: Maybe we can use ASSERT here? Or is it not a logic error if this is called more than once?

src/web/ng/RPCServerHandler.hpp
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        coroutineGroup.asyncWait(yield);
+        ASSERT(response.has_value(), "Woke up coroutine without setting response");
+        return std::move(response).value();
+    }
+
+private:
+    Response
+    handleRequest(
```
godexsoft, line 180: nit: It may be time to split these big functions into some smaller ones 💭

src/web/ng/Response.cpp
```diff
@@ -34,83 +35,98 @@
...
+
 namespace web::ng {
 
 namespace {
 
-std::string_view
-asString(Response::HttpData::ContentType type)
+template <typename T>
+consteval bool
+isString()
```
godexsoft, line 50: This is neat. I think you can even do something like this these days:
```cpp
template <typename T>
constinit bool const isString = std::is_same_v<T, std::string>;
```

src/web/ng/Response.hpp
```diff
@@ -100,7 +91,7 @@ class Response {
      * @return The message of the response as a const buffer.
      */
     boost::asio::const_buffer
-    asConstBuffer() const&;
+    intoWsResponse() const&;
```
godexsoft: nit: Out of curiosity, why `into` and not `as`?

godexsoft (changes requested):
Looks very good. Few comments and questions 👍

src/app/ClioApplication.cpp
```diff
@@ -126,9 +140,91 @@ ClioApplication::run()
...
 
+    if (useNgWebServer or config_.valueOr("server.__ng_web_server", false)) {
+        web::ng::RPCServerHandler<RPCEngineType, etl::ETLService> handler{config_, backend, rpcEngine, etl};
+
+        auto expectedAdminVerifier = web::make_AdminVerificationStrategy(config_);
+        if (not expectedAdminVerifier.has_value()) {
+            LOG(util::LogService::error()) << "Error creating admin verifier: " << expectedAdminVerifier.error();
+            return EXIT_FAILURE;
+        }
+        auto adminVerifier = std::move(expectedAdminVerifier).value();
```
cindyyan317: const

src/web/ng/SubscriptionContext.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+void
+SubscriptionContext::send(std::shared_ptr<std::string> message)
+{
+    if (disconnected_)
+        return;
+
+    if (maxSendQueueSize_.has_value() and tasksGroup_.size() >= *maxSendQueueSize_) {
+        tasksGroup_.spawn(yield_, [this](boost::asio::yield_context innerYield) {
```
cindyyan317, line 61: Maybe add a warning log here

src/web/ng/SubscriptionContext.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+uint32_t
+SubscriptionContext::apiSubversion() const
+{
+    return apiSubversion_;
+}
+
+void
+SubscriptionContext::disconnect(boost::asio::yield_context yield)
```
cindyyan317, line 94: is it only called by test?

src/web/ng/SubscriptionContext.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+}
+
+void
+SubscriptionContext::send(std::shared_ptr<std::string> message)
+{
+    if (disconnected_)
+        return;
+
+    if (maxSendQueueSize_.has_value() and tasksGroup_.size() >= *maxSendQueueSize_) {
```
cindyyan317, line 60: It might have problem if we only restrict the size of subscription message size.
If a user sends requests without reading their response, will it cause the responses piling up?

src/feed/Types.hpp
```diff
@@ -19,12 +19,12 @@
...
-#include "web/interface/ConnectionBase.hpp"
+#include "web/SubscriptionContextInterface.hpp"
 
 #include <memory>
 
 namespace feed {
-using Subscriber = web::ConnectionBase;
+using Subscriber = web::SubscriptionContextInterface;
 using SubscriberPtr = Subscriber*;
 using SubscriberSharedPtr = std::shared_ptr<Subscriber>;
```
cindyyan317, line 30: is it web::SubscriptionContextPtr ? Can we unify the names here? In subscribe.hpp, we are using `SubscriptionContextPtr`, in unsubscribe.hpp, we are using `SubscriberSharedPtr`.

src/web/Context.hpp
```diff
@@ -43,7 +42,7 @@ struct Context : util::Taggable {
     std::string method;
     std::uint32_t apiVersion;
```
cindyyan317, line 43: Can remove this apiVersion now?

src/web/ng/impl/ConnectionHandler.cpp
```diff
@@ -47,7 +56,8 @@ namespace {
 
 Response
 handleHttpRequest(
-    ConnectionContext const& connectionContext,
+    ConnectionMetadata& connectionMetadata,
+    SubscriptionContextPtr& subscriptionContext,
```
cindyyan317, line 60: Do we need "subscriptionContext" for http request?

src/web/ng/impl/WsConnection.hpp
```diff
@@ -112,20 +125,29 @@ class WsConnection : public Connection {
...
+        boost::asio::const_buffer buffer,
         boost::asio::yield_context yield,
-        std::chrono::steady_clock::duration timeout = DEFAULT_TIMEOUT
+        std::chrono::steady_clock::duration timeout = Connection::DEFAULT_TIMEOUT
     ) override
     {
-        auto error = util::withTimeout(
-            [this, &response](auto&& yield) { stream_.async_write(response.asConstBuffer(), yield); }, yield, timeout
-        );
+        auto error =
```
cindyyan317, line 134: I saw there is no longer a queue here. Do the multiple async_write overlap?
From the [document](https://boost.org/doc/libs/1_80_0/doc/html/boost_asio/reference/async_write/overload1.html)
```
The program must ensure that the stream performs no other write operations (such as async_write, the stream's async_write_some function, or any other composed operations that perform writes) until this operation completes.
``` 

Is it a problem?

src/util/CoroutineGroup.cpp
kuznetsss: Because `childrenCounter_` is signed. And it is signed because there is a code to reduce it, so to not accidentally catch overflow.

src/web/ng/Connection.hpp
kuznetsss: It may be called more than once on one connection, so it is on purpose to make this function lazy to verify admin only once for a connection.

src/web/ng/RPCServerHandler.hpp
kuznetsss: Maybe as a separate task? Here I copied the old RPCServerHandler and adapted it to the new interface.

src/web/ng/Response.hpp
kuznetsss: Good point, should be `as` since it is lvalue method.

src/web/ng/SubscriptionContext.cpp
kuznetsss: What do you mean? Here we are restricting the amount of messages waiting in the queue to be sent. Just instead of an actual queue we use scheduled coroutines.

src/web/ng/SubscriptionContext.cpp
kuznetsss: You mean if spawn failed?

src/web/ng/SubscriptionContext.cpp
kuznetsss: Good catch, it should be called in `ConnectionHandler::processConnection()` when processing is done.

src/web/ng/impl/ConnectionHandler.cpp
kuznetsss: It is `nullptr` for http requests.

src/web/ng/impl/WsConnection.hpp
kuznetsss: Connection is processed in a coroutine and each coroutine has its own strand, so it is fine here.

src/web/ng/SubscriptionContext.cpp
cindyyan317: Log when the size is over max

src/web/ng/SubscriptionContext.cpp
cindyyan317: But only subscription messages count here?

src/web/ng/Response.cpp
kuznetsss: `IsString` is a function but you are suggesting to define it as a variable or I don't understand something.

src/web/ng/SubscriptionContext.cpp
kuznetsss: But `taskGroup_` is initialised without a limit here, so it should always be true.

src/web/ng/SubscriptionContext.cpp
kuznetsss: Yeah, it is only for subscription messages. All other messages are working by request-response model.

src/web/Context.hpp
kuznetsss: No, It is used to build `rpc::Context`.

src/feed/Types.hpp
kuznetsss: I changed the naming in `subscribe.cpp`.

src/web/ng/Response.cpp
godexsoft: It's just one more way you can do the same thing. Yes, it's a constant per T. Safe to assume that in both cases all of this is gone after compile time anyway.

src/web/ng/SubscriptionContext.cpp
cindyyan317: i see. the request-response does not have a queue, so the responses would not pile up.

cindyyan317 (approved):
Nice refactor !

godexsoft (approved):
Looks good 👍

Issue #1596, opened 2024-08-12
ETL ng extractor


godexsoft:
## Summary
For #1594 

Rewrite the extractor using ranges and AsyncFramework.
Note that there are going to be multiple extractors operating at the same time. 
Make it testable and cover with unit tests.

PR #1733, opened 2024-11-14
feat: Extraction basics

godexsoft:
For #1596 

This PR implements the transformation from PB to our internal models to be used in ETL ng.


src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+    LOG(log_.debug()) << "Extracting DIFF " << seq;
+
+    auto [batch, time] = ::util::timed<std::chrono::duration<double>>([this, seq] {
+        return fetcher_->fetchDataAndDiff(seq).and_then(makeExtractor(seq));
+    });
+
+    LOG(log_.debug()) << "Extracted and Transformed diff for " << seq << " in " << time << "ms";
+
+    // can be nullopt. this means that either the server is stopping or another node took over ETL writing.
```
cindyyan317, line 205: I feel it still can fetch valid data from grpc when another node took over.

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    rg::copy(
+        data.book_successors() | vs::transform([](auto&& obj) { return extractSuccessor(obj); }),
+        std::back_inserter(output)
+    );
+    return output;
+}
+
+auto
+Extractor::makeExtractor(uint32_t seq)
```
cindyyan317: No need this seq parameter. We can use header.seq.

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+    };
+}
+
+std::optional<model::LedgerData>
+Extractor::extractLedgerWithDiff(uint32_t seq)
+{
+    LOG(log_.debug()) << "Extracting DIFF " << seq;
+
+    auto [batch, time] = ::util::timed<std::chrono::duration<double>>([this, seq] {
+        return fetcher_->fetchDataAndDiff(seq).and_then(makeExtractor(seq));
```
cindyyan317: When I read this line, I feel `depack` is better. It will look like 
`fetchDataAndDiff(seq).and_then(depack())` .

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+    // can be nullopt. this means that either the server is stopping or another node took over ETL writing.
+    return batch;
+}
+
+std::optional<model::LedgerData>
+Extractor::extractLedgerOnly(uint32_t seq)
+{
+    LOG(log_.debug()) << "Extracting FULL " << seq;
+
+    auto [batch, time] = ::util::timed<std::chrono::duration<double>>([this, seq] {
```
cindyyan317: const

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+    // TODO: should be simplified with ranges::to<> when available
+    std::vector<model::Object> output;
+    output.reserve(objects.size());
+
+    rg::move(objects | vs::transform([](auto&& obj) { return extractObj(obj); }), std::back_inserter(output));
+    return output;
+}
+
+model::BookSuccessor
+extractSuccessor(PBBookSuccessorType successor)
```
cindyyan317, line 147: Do we want to move from successor?

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+    output.reserve(transactions.size());
+
+    rg::move(transactions | vs::transform([seq](auto&& tx) { return extractTx(tx, seq); }), std::back_inserter(output));
+    return output;
+}
+
+model::Object
+extractObj(PBObjType obj)
+{
+    auto key = ripple::uint256::fromVoidChecked(obj.key());
```
cindyyan317: const

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+model::Transaction
+extractTx(PBTxType tx, uint32_t seq)
+{
+    auto raw = std::move(*tx.mutable_transaction_blob());
+    ripple::SerialIter it{raw.data(), raw.size()};
+    ripple::STTx const sttx{it};
+    ripple::TxMeta meta{sttx.getTransactionID(), seq, tx.metadata_blob()};
+
+    static constexpr std::size_t KeySize = 32;
+    std::string keyStr{reinterpret_cast<char const*>(sttx.getTransactionID().data()), KeySize};
```
cindyyan317: Can uint256ToString do the same job?

tests/unit/etlng/ExtractionTests.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+    EXPECT_CALL(*fetcher, fetchDataAndDiff(Seq)).WillOnce(testing::Return(original));
+    auto res = extractor.extractLedgerWithDiff(Seq);
+
+    EXPECT_TRUE(res.has_value());
+    EXPECT_EQ(res->objects.size(), 10);
+    EXPECT_EQ(res->transactions.size(), 10);
+    EXPECT_TRUE(res->successors.has_value());
+    EXPECT_EQ(res->successors->size(), 10);
+    EXPECT_FALSE(res->edgeKeys.has_value());  // this is set separately in ETL
```
cindyyan317, line 278: 🤔 Why the comment is "this is set"? but it is not set

src/etlng/impl/Extraction.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+model::Object
+extractObj(PBObjType obj)
+{
+    auto key = ripple::uint256::fromVoidChecked(obj.key());
+    ASSERT(key.has_value(), "Failed to deserialize key from void");
+
+    auto value_or = [](std::string const& maybe, std::string fallback) -> std::string {
+        if (maybe.empty())
+            return fallback;
```
cindyyan317, line 117: do we need to move here?

cindyyan317 (commented):
It looks like more structured 💯 
I left some comments about makeExtractor and uint256ToString

src/etlng/impl/Extraction.cpp
godexsoft: This depends on the fetcher.. out of scope of this PR i'd say. Note that this stuff will still change more and fetcher itself will be updated later too. This comment basically migrated from previous logic

src/etlng/impl/Extraction.cpp
godexsoft: Makes sense, let's try that

src/etlng/impl/Extraction.cpp
godexsoft: `depack` is not a word but maybe `unpack` would work. My concern is that the function name `unpack` does not actually explain what it does on itself.. but it does look good in this context 🤔

src/etlng/impl/Extraction.cpp
godexsoft: I think we can. Notice how it's a copy in the arg, not a ref of some sort

src/etlng/impl/Extraction.cpp
godexsoft: I'm not sure it would be any different in the final assembly but it probably does not hurt to move

src/etlng/impl/Extraction.cpp
godexsoft: Good question. This code was moved as is from the old version. Let me try

src/etlng/impl/Extraction.cpp
godexsoft: Ah, so it does not hurt C++ but it does trigger a warning which becomes an error. So let's not move here

src/etlng/impl/Extraction.cpp
godexsoft: Seems to pass unit tests at least. Let's use it now and if something breaks we fix in future PR

cindyyan317 (approved):
Nice !!!!

Issue #1597, opened 2024-08-12
ETL ng loader


godexsoft:
## Summary
For #1594 

The new loader will support plugins, backfill and likely online_delete-like features.
In this task the existing loader needs to be refactored and covered by tests.
There are separate tasks for in-depth work on backfill and plugins.

PR #1745, opened 2024-11-20
feat: GrpcSource for ETL ng

godexsoft:
For #1596 and #1597 

This is a slightly modified version of the existing `GrpcSource` and a renamed `AsyncCallData`.
Instead of writing to the database directly from the async downloader we introduce an interface for observing the initial ledger load which allows us to separate the downloading from the writing.

The included unit tests are inspired by the existing `etl/GrpcSourceTests` and use the same approach with the mocked gRPC server.

### DIFF
--- a/src/etlng/impl/GrpcSource.cpp
+++ b/src/etlng/impl/GrpcSource.cpp
@@ -1,0 +1,168 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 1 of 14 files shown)


src/etlng/impl/GrpcSource.cpp
```diff
@@ -0,0 +1,168 @@
...
+std::string
+toString(auto const& endpoint)
+{
+    std::stringstream ss;
+    ss << endpoint;
+    return ss.str();
+}
+
+std::string
+resolve(boost::asio::io_context& ctx, std::string const& ip, std::string const& port)
```
kuznetsss: We already have `web::Resolver` doing the same thing.

src/etlng/impl/GrpcSource.cpp
```diff
@@ -0,0 +1,168 @@
...
+{
+    boost::asio::ip::tcp::resolver resolver{ctx};
+
+    if (auto const result = resolver.resolve(ip, port); not result.empty())
+        return toString(result.begin()->endpoint());
+
+    throw std::runtime_error("Failed to resolve " + ip + ":" + port);
```
kuznetsss, line 59: Better to use `std::expected`.

tests/unit/etlng/GrpcSourceTests.cpp
```diff
@@ -0,0 +1,290 @@
...
+    GrpcSourceNgTests()
+        : WithMockXrpLedgerAPIService("localhost:0"), grpcSource_("localhost", std::to_string(getXRPLMockPort()))
+    {
+    }
+
+    testing::StrictMock<MockLoadObserver> observer_;
+    etlng::impl::GrpcSource grpcSource_;
+};
+
+class KeyStore {
```
kuznetsss: It is better to create all classes and structs inside the related fixture to avoid naming collision. Anonymous namespace should save us from naming collision here, but just to be sure.

src/etlng/impl/GrpcSource.cpp
godexsoft: Yes but this is called in the constructor. Let's address this later with some sort of static make function.

tests/unit/etlng/GrpcSourceTests.cpp
godexsoft: We can move it into the fixture, no problem.

kuznetsss (approved):
👍

PR #1751, opened 2024-11-22
feat: Healthcheck endpoint

godexsoft:
This implement a simple endpoint for system healthcheck.

### DIFF
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -65,6 +65,14 @@ namespace app {
 
 namespace {
 
+auto constexpr HealthCheckHTML = R"html(
+    <!DOCTYPE html>
+    <html>
+        <head><title>Test page for Clio</title></head>
+        <body><h1>Clio Test</h1><p>This page shows Clio http(s) connectivity is working.</p></body>
+    </html>
+)html";
+
 /**
  * @brief Start context threads
  *
@@ -175,6 +183,16 @@ ClioApplication::run(bool const useNgWebServer)
                 );
                 ASSERT(maybeResponse.has_value(), "Got unexpected request for Prometheus");
                 return web::ng::Response{std::move(maybeResponse).value(), request};
+            }
+        );
+
+        httpServer->onGet(
+            "/health",
+            [](web::ng::Request const& request,
+               web::ng::ConnectionMetadata&,
+               web::SubscriptionContextPtr,
+               boost::asio::yield_context) -> web::ng::Response {
+                return web::ng::Response{boost::beast::http::status::ok, HealthCheckHTML, request};
             }
         );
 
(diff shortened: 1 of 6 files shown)


src/etlng/impl/GrpcSource.cpp
```diff
@@ -49,10 +48,8 @@ resolve(std::string const& ip, std::string const& port)
 {
     web::Resolver resolver;
 
-    if (auto const results = resolver.resolve(ip, port); not results.empty()) {
-        std::cout << "resolved ip: '" << results.at(0) << '\n';
```
godexsoft, line 53: Oops :)

src/app/ClioApplication.cpp
```diff
@@ -178,6 +186,16 @@ ClioApplication::run(bool const useNgWebServer)
             }
         );
 
+        httpServer->onGet(
```
godexsoft, line 189: This will be a nuisance when merging back into `release/2.3.0`

src/app/ClioApplication.cpp
```diff
@@ -178,6 +186,16 @@ ClioApplication::run(bool const useNgWebServer)
...
+            "/health",
```
kuznetsss, line 190: Maybe `/ping`? Or `/healthcheck`?

tests/common/util/TestHttpClient.cpp
```diff
@@ -58,7 +59,7 @@ using tcp = boost::asio::ip::tcp;
 
 namespace {
 
-std::string
+std::pair<boost::beast::http::status, std::string>
```
kuznetsss, line 62: Maybe a structure would be better?

src/app/ClioApplication.cpp
cindyyan317: Maybe /ping will be confusing. Because we have ping API

src/etlng/impl/GrpcSource.cpp
cindyyan317: 😨

cindyyan317 (approved):

src/app/ClioApplication.cpp
godexsoft: `/health` is fine

tests/common/util/TestHttpClient.cpp
godexsoft: That's what `pair` is for. If it had 3 or more members then yeah.

kuznetsss: There is `tuple` for 3 and more 🙂 . I would say `pair` is bad for readability in general. But I'm ok if you don't want to change this part.

kuznetsss (approved):

PR #1754, opened 2024-11-25
feat: Add v3 support

godexsoft:
This is enabling API v3 in order to support forwarding for the fix introduced in https://github.com/XRPLF/rippled/pull/4924

### DIFF
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -950,7 +950,7 @@ public:
         std::vector<Statement> statements;
         statements.reserve(data.size());
         for (auto [mptId, holder] : data)
-            statements.push_back(schema_->insertMPTHolder.bind(std::move(mptId), std::move(holder)));
+            statements.push_back(schema_->insertMPTHolder.bind(mptId, holder));
 
         executor_.write(std::move(statements));
     }
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -950,7 +950,7 @@ public:
         std::vector<Statement> statements;
         statements.reserve(data.size());
         for (auto [mptId, holder] : data)
-            statements.push_back(schema_->insertMPTHolder.bind(mptId, holder));
+            statements.push_back(schema_->insertMPTHolder.bind(std::move(mptId), std::move(holder)));
 
         executor_.write(std::move(statements));
     }
--- a/src/rpc/common/APIVersion.hpp
+++ b/src/rpc/common/APIVersion.hpp
@@ -34,16 +34,13 @@ static constexpr uint32_t API_VERSION_DEFAULT = 1u;
 
 /**
  * @brief Minimum API version supported by this build
- *
- * Note: Clio does not natively support v1 and only supports v2 or newer.
- * However, Clio will forward all v1 requests to rippled for backward compatibility.
  */
 static constexpr uint32_t API_VERSION_MIN = 1u;
 
 /**
  * @brief Maximum API version supported by this build
  */
-static constexpr uint32_t API_VERSION_MAX = 2u;
+static constexpr uint32_t API_VERSION_MAX = 3u;
 
 /**
  * @brief A baseclass for API version helper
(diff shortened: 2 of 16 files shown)


src/rpc/common/APIVersion.hpp
```diff
@@ -34,16 +34,13 @@ static constexpr uint32_t API_VERSION_DEFAULT = 1u;
 
 /**
  * @brief Minimum API version supported by this build
- *
- * Note: Clio does not natively support v1 and only supports v2 or newer.
```
godexsoft, line 38: This is outdated and not true.

src/data/CassandraBackend.hpp
```diff
@@ -950,7 +950,7 @@ class BasicCassandraBackend : public BackendInterface {
         std::vector<Statement> statements;
         statements.reserve(data.size());
         for (auto [mptId, holder] : data)
-            statements.push_back(schema_->insertMPTHolder.bind(std::move(mptId), std::move(holder)));
+            statements.push_back(schema_->insertMPTHolder.bind(mptId, holder));
```
kuznetsss: I don't think it is a related change.

src/rpc/common/APIVersion.hpp
```diff
@@ -34,16 +34,13 @@ static constexpr uint32_t API_VERSION_DEFAULT = 1u;
...
- * However, Clio will forward all v1 requests to rippled for backward compatibility.
  */
 static constexpr uint32_t API_VERSION_MIN = 1u;
 
 /**
  * @brief Maximum API version supported by this build
  */
-static constexpr uint32_t API_VERSION_MAX = 2u;
+static constexpr uint32_t API_VERSION_MAX = 3u;
```
kuznetsss, line 43: Maybe we should have a test about this?

src/data/CassandraBackend.hpp
godexsoft: It's not. it's a drive-by fix as stated in the description

src/rpc/common/APIVersion.hpp
godexsoft: What would the test do? in `rippled` they have some ASSERTS for this but i'm not sure what the value is

src/data/CassandraBackend.hpp
godexsoft: Fine, i'll move it to the other branch i have which is `clang-tidy` related

src/rpc/common/APIVersion.hpp
kuznetsss: Some handler fails when api version is not 1 or 2 and now 3.

cindyyan317:
Shall we use the rippled's apiMaximumSupportedVersion in xrpl directly for max

tests/unit/web/RPCServerHandlerTests.cpp
```diff
@@ -870,3 +828,84 @@ TEST_F(WebRPCServerHandlerTest, WsRequestNotJson)
...
+    (*handler)(request, session);
+
+    auto response = boost::json::parse(session->message);
+    EXPECT_TRUE(response.is_object());
+    EXPECT_TRUE(response.as_object().contains("error"));
+    EXPECT_EQ(response.at("error").as_string(), "invalid_API_version");
+    EXPECT_TRUE(response.as_object().contains("error_message"));
+    EXPECT_EQ(response.at("error_message").as_string(), GetParam().wsMessage);
+    EXPECT_TRUE(response.as_object().contains("error_code"));
+    EXPECT_EQ(response.at("error_code").as_int64(), 6000);
```
kuznetsss: You could use the enum value of the error here.

kuznetsss (dismissed):
LGTM

godexsoft:
> apiMaximumSupportedVersion

@cindyyan317 I checked and I don't think we can use it. `rippled` sets it to 2 while we want 3. They have a concept of beta version but we don't.

cindyyan317 (dismissed):

kuznetsss (approved):

PR #1755, opened 2024-11-25
fix: Check ledger range in every handler

### LINKED ISSUE #1565: Check ledger range in every handler
## Issue Description

Ledger range is not checked for having a value inside optional which could lead to undefined behaviour.
For example #1564 fixing UB in a test. It is better to check range and throw an exception (or add assert) in each handler.

PeterChen13579:
fixes #1565

### DIFF
--- a/tests/unit/rpc/handlers/AllHandlerTests.cpp
+++ b/tests/unit/rpc/handlers/AllHandlerTests.cpp
@@ -1,0 +1,214 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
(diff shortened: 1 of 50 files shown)


tests/unit/rpc/handlers/AllHandlerTests.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    HandlerType handler_;
+
+private:
+    HandlerType
+    initHandler()
+    {
+        if constexpr (std::is_same_v<HandlerType, AccountInfoHandler> || std::is_same_v<HandlerType, FeatureHandler>) {
+            return HandlerType{this->backend, this->mockAmendmentCenterPtr};
```
kuznetsss, line 139: Don't you need to initialise `mockAmendmentCenterPtr` before providing it into the handler here? Looks like it is `nullptr` now, but maybe I'm wrong.

tests/unit/rpc/handlers/AllHandlerTests.cpp
```diff
@@ -0,0 +1,213 @@
...
+TYPED_TEST(AllHandlersDeathTest, NoRangeAvailable)
+{
+    // doesn't work without 'this'
+    this->runSpawn(
+        [&](boost::asio::yield_context yield) {
+            TypeParam handler = this->handler_;
+
+            auto const input = createInput<TypeParam>();
+            auto const context = Context{yield};
+            EXPECT_DEATH({ [[maybe_unused]] auto _unused = handler.process(input, context); }, ".*");
```
kuznetsss: Maybe you could add a pattern `Assertion .* failed.*`.

tests/unit/rpc/handlers/AllHandlerTests.cpp
```diff
@@ -105,6 +105,7 @@ struct AllHandlersDeathTest : HandlerBaseTest, testing::WithParamInterface<std::
     {
     }
 
+    StrictMockAmendmentCenterSharedPtr mockAmendmentCenterPtr;
```
kuznetsss, line 131: Maybe you could add an ASSERT in `initHandler()` to prevent the error in the future.

cindyyan317 (changes requested):
It's not **every handler.** I did not see `subscribe`  `unsubscribe`, they are immune?

cindyyan317 (approved):

PR #1760, opened 2024-11-26
chore: No ALL_CAPS

godexsoft:
This PR renames identifiers according to the updated `clang-tidy` configuration.
It also fixes the placement of `constexpr` when used with `static` and small other cosmetic fixes.

Let me know what you think.
I don't really like how non-const statics became `kALL_CAPS` formatted but `clang-tidy` does not seem to be able to make the distinction very well.


cindyyan317 (dismissed):
I think kALL_CAPS is fine. I don't down what k means thou.

I am happy to see `make_` gone...

PR #1763, opened 2024-11-28
refactor: setRange in tests

PeterChen13579:
There are a few files that cannot move the setRange into constructor of the test because either the place that calls setRange matters or tests checks range doesn't exist


cindyyan317 (approved):
LGTM

PR #1768, opened 2024-12-03
feat: Migration framework

cindyyan317:
This PR implemented the migration framework, which contains the command line interface to execute migration and helps to migrate data easily.
Please read README.md for more information about this framework.


src/app/CliArgs.cpp
```diff
@@ -45,6 +46,8 @@ CliArgs::parse(int argc, char const* argv[])
         ("version,v", "print version and exit")
         ("conf,c", po::value<std::string>()->default_value(defaultConfigPath), "configuration file")
         ("ng-web-server,w", "Use ng-web-server")
+        ("migrate", po::value<std::string>(),"start migration helper")
+        ("migrate_rollback", po::value<std::string>(),"rollback the given migration")
```
godexsoft: `migration-rollback` or simply `rollback` would be more suitable as a cli option 👍

src/data/README.md
```diff
@@ -262,3 +262,15 @@ CREATE TABLE clio.nf_token_transactions (
...
 The `nf_token_transactions` table serves as the NFT counterpart to `account_tx`, inspired by the same motivations and fulfilling a similar role within this context. It drives the `nft_history` API.
+
+### migrated_features
+
+```
+CREATE TABLE clio.migrated_features (
+	status TEXT, 	  				# The migration status of the feature
+	feature_name TEXT,				# The name of the feature
+	PRIMARY KEY (status, feature_name)
+	)
```
godexsoft: Let's put the `)` at the beginning of the line to align like others?

src/data/cassandra/Schema.hpp
```diff
@@ -270,6 +270,18 @@ class Schema {
             qualifiedTableName(settingsProvider_.get(), "mp_token_holders")
         ));
 
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                    status TEXT,
```
godexsoft: Look carefully at how other CQL is formatted in the file and adjust accordingly please 🔢

src/data/cassandra/Schema.hpp
```diff
@@ -768,6 +780,17 @@ class Schema {
                 qualifiedTableName(settingsProvider_.get(), "ledger_range")
             ));
         }();
+
+        PreparedStatement selectMigratedFeatures = [this]() {
+            return handle_.get().prepare(fmt::format(
+                R"(
+                SELECT feature_name
+                  FROM {}
+                WHERE status = 'migrated'
```
godexsoft: More formatting issues.

src/main/Main.cpp
```diff
@@ -46,6 +47,16 @@ try {
             app::ClioApplication clio{config};
 
             return clio.run(run.useNgWebServer);
+        },
+        [](app::CliArgs::Action::Migrate const& migrate) {
+            auto const config = util::ConfigReader::open(migrate.configPath);
```
godexsoft: What is the idea of having a separate config? is it so that we can setup different log levels etc. for the duration of migration?

src/migration/MigrationApplication.cpp
```diff
@@ -0,0 +1,119 @@
...
+            [this](Cmd::Rollback const& cmdBundle) { return rollback(cmdBundle.migratorName); }
+        },
+        cmd_.state
+    );
+}
+
+int
+MigratorApplication::printStatus()
+{
+    std::cout << "Current Migration Status:" << std::endl;
```
godexsoft, line 68: `cout` is probably not what we want? is the concern that using the logger could have no output if misconfigured?

src/migration/MigrationApplication.cpp
```diff
@@ -0,0 +1,119 @@
...
+    std::cout << "Migration for " << migratorName << " has finished" << std::endl;
+    return EXIT_SUCCESS;
+}
+
+int
+MigratorApplication::rollback(std::string const& migratorName)
+{
+    auto const status = migrationManager_->getMigratorStatusByName(migratorName);
+    if (status == migration::MigratorStatus::NotMigrated) {
+        std::cout << "Migrator " << migratorName << " yet not migrated" << std::endl;
```
godexsoft: "yet not migrated" sounds strange. "Migrator xyz did not yet migrate" or "Migrator xyz not yet migrated" sounds more correct 🔢

src/migration/MigrationApplication.hpp
```diff
@@ -0,0 +1,125 @@
...
+//==============================================================================
+
+#pragma once
+
+#include "migration/MigrationManagerInterface.hpp"
+#include "util/config/Config.hpp"
+
+#include <memory>
+#include <string>
+#include <variant>
```
godexsoft, line 27: Let's keep the spacing the same with other files in the project 🔢

src/migration/MigrationApplication.hpp
```diff
@@ -0,0 +1,125 @@
...
+    int
+    printStatus();
+
+    int
+    migrate(std::string const& name);
+
+    int
+    rollback(std::string const& name);
+
+    Cmd cmd_;
```
godexsoft: This belongs with other data at the top

src/migration/MigrationManagerFactory.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <memory>
+
+namespace migration {
+
+/**
+ * @brief The factory to create a MigrationManagerInferface
+ *
+ * @param config The configuration of the migration application, it contains the database connection configuration and
+ * other migration specific configurations
```
godexsoft: This will be slightly problematic with the new config system. You will need to write a new specification for it. I guess no other choice anyway.

src/migration/MigratorsRegister.hpp
```diff
@@ -0,0 +1,184 @@
...
+    std::shared_ptr<Backend> backend_;
+
+    template <typename Migrator>
+    void
+    callMigration(std::string const& name, util::Config const& config)
+    {
+        if (name == Migrator::name) {
+            LOG(log_.info()) << "Running migration: " << name;
+            Migrator::runMigration(backend_, config);
+            backend_->writeMigratedMigrator(name);
```
godexsoft: nit: Maybe something like `markMigrationDone` to avoid "MigratedMigrator"? 😃

src/migration/MigratorsRegister.hpp
```diff
@@ -0,0 +1,184 @@
...
+     * @brief The backend type which is used by the migrators
+     */
+    using BackendType = Backend;
+
+    /**
+     * @brief Construct a new Migrators Register object
+     *
+     * @param backend The backend shared pointer
+     */
+    MigratorsRegister(std::shared_ptr<Backend> backend) : backend_{std::move(backend)}
```
godexsoft: Don't you want to use `BackendType` here?

src/migration/MigratorsRegister.hpp
```diff
@@ -0,0 +1,184 @@
...
+     */
+    std::vector<std::tuple<std::string, MigratorStatus>>
+    getMigratorsStatus() const
+    {
+        auto const migratedList = data::synchronous([&](auto yield) { return backend_->fetchMigratedFeatures(yield); });
+        auto const fullList = getMigratorNames();
+
+        std::vector<std::tuple<std::string, MigratorStatus>> status;
+
+        for (auto const i : fullList) {
```
godexsoft: nit: `i` is not a very descriptive name for this

src/migration/MigratorsRegister.hpp
```diff
@@ -0,0 +1,184 @@
...
+            if (migratedList != std::nullopt &&
+                std::find(migratedList->begin(), migratedList->end(), i) != migratedList->end()) {
```
godexsoft: Could use ranges instead

src/migration/README.md
```diff
@@ -0,0 +1,101 @@
+
+# Clio Migration
```
godexsoft, line 2: This entire readme needs to be updated for spelling and English grammar. 
@mDuo13 can you assist with this please? 🙇

src/migration/MigrationManagerFactory.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <stdexcept>
+#include <string>
+#include <utility>
+
+namespace migration {
+
+std::shared_ptr<MigrationManagerInterface>
+makeMigrationManager(util::Config const& config)
+{
+    static util::Logger const log{"Migration"};
```
godexsoft: "Migration" is not a channel that is registered in the log system. It should still work but the downside i believe is that in reality it uses "General" log's settings.

src/migration/cassandra/CassandraMigrationBackend.hpp
```diff
@@ -0,0 +1,127 @@
...
+    void
+    migrateInTokenRange(
+        std::int64_t const& start,
+        std::int64_t const& end,
+        auto const& callback,
+        boost::asio::yield_context yield
+    )
+    {
+        LOG(log_.debug()) << "Travsering token range: " << start << " - " << end
+                          << "- table: " << TableDesc::TABLE_NAME;
```
godexsoft: nit: This looks like it will end up squashing `end` and `-`

godexsoft: Perhaps replace the last `-` with`;`?

src/migration/cassandra/CassandraMigrationBackend.hpp
```diff
@@ -0,0 +1,127 @@
...
+     *@brief Record the migration status of a migrator in the database
+     *
+     *@param migratorName The name of the migrator
+     */
+    void
+    writeMigratedMigrator(std::string const& migratorName)
+    {
+        auto const& preparedStatement = migrationSchema_.getPreparedInsertMigratedMigrator(handle_);
+        this->executor_.writeSync(
+            preparedStatement, data::cassandra::Text("migrated"), data::cassandra::Text{migratorName}
```
godexsoft: Does it need to be a string for the second arg? why not a bool?

src/migration/cassandra/CassandraMigrationSchema.hpp
```diff
@@ -0,0 +1,116 @@
...
+    data::cassandra::PreparedStatement
+    getPreparedFullScanStatement(
+        data::cassandra::Handle const& handler,
+        std::string const& tableName,
+        std::string const& key
+    )
+    {
+        return handler.prepare(fmt::format(
+            R"(
+                SELECT * FROM {}
```
godexsoft: Formatting here too.

src/migration/cassandra/CassandraMigrationSchema.hpp
```diff
@@ -0,0 +1,116 @@
...
+     * @return The prepared statement to insert into migrated_features table
+     */
+    data::cassandra::PreparedStatement const&
+    getPreparedInsertMigratedMigrator(data::cassandra::Handle const& handler)
+    {
+        static auto prepared = handler.prepare(fmt::format(
+            R"(
+                INSERT INTO {} (status, feature_name)
+                VALUES (?, ?)
+                )",
```
godexsoft: And some more formatting :)

src/migration/cassandra/CassandraMigrationSchema.hpp
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @param handler The database handler
+     * @return The prepared statement to remove from migrated_features table
+     */
+    data::cassandra::PreparedStatement const&
+    getPreparedRemoveMigratedMigrator(data::cassandra::Handle const& handler)
+    {
+        static auto prepared = handler.prepare(fmt::format(
+            R"(
+                DELETE FROM {} WHERE status = 'migrated'
```
godexsoft: And even more formatting 🔢

src/migration/cassandra/FullTableScanner.hpp
```diff
@@ -0,0 +1,172 @@
...
+        , reader_{std::move(reader)}
+    {
+        ASSERT(workersNum > 0, "workersNum for full table scanner must be greater than 0");
+        auto const cursors = TokenRangesProvider{cursorsNum_}.getRanges();
+        std::ranges::for_each(cursors, [this](auto const& cursor) { queue_.push(cursor); });
+        load(workersNum);
+    }
+
+    /**
+     *
```
godexsoft: Stray empty line here

godexsoft (changes requested):
It's a nice system but needs some polish 🔢 👍 

Among other things it would be nice to add some `static_assert` tests for the introduced concepts, come up with better names for several functions and move some/majority of the code to `impl`.

src/main/Main.cpp
cindyyan317: It can be the same one. But it will need a config. Here migrate.configPath is the config parsed from cmd.

src/migration/MigrationApplication.cpp
cindyyan317: yes. It kinda conflict with clio_server config. This is the output of the command line, We'd like it to print it to console definitely , like -help

src/migration/MigrationManagerFactory.cpp
cindyyan317: I will add it to log channel.

src/migration/cassandra/CassandraMigrationBackend.hpp
cindyyan317: Has changed the schema.

src/CMakeLists.txt
```diff
@@ -5,5 +5,6 @@ add_subdirectory(etlng)
 add_subdirectory(feed)
 add_subdirectory(rpc)
 add_subdirectory(web)
+add_subdirectory(migration)
```
kuznetsss, line 8: Shouldn't it be in data?

src/data/CassandraBackend.hpp
```diff
@@ -835,6 +838,28 @@ class BasicCassandraBackend : public BackendInterface {
         return results;
     }
 
+    std::optional<std::unordered_set<std::string>>
+    fetchMigratedFeatures(boost::asio::yield_context yield) const override
+    {
+        auto const res = executor_.read(yield, schema_->selectMigratedFeatures);
+        if (not res) {
+            LOG(log_.error()) << "Could not fetch migrated features: " << res.error();
```
kuznetsss: This is probably not an error. Current database doesn't have it.

src/data/cassandra/Types.hpp
```diff
@@ -55,6 +57,35 @@ struct Limit {
...
+    explicit Text(std::string text) : text{std::move(text)}
+    {
+    }
+
+    /**
+     * @brief Construct a new Text object from char const* type
+     *
+     * @param text The text to wrap
+     */
+    explicit Text(char const* text) : text{text}
```
kuznetsss: Shouldn't here be size as well or is it always null-terminated strings?

src/main/Main.cpp
kuznetsss: I think it would be better to make migration process as simple as possible from a user perspective. So I think it would be better to not have an extra config required for that.

src/migration/MigrationApplication.cpp
```diff
@@ -0,0 +1,119 @@
...
+}
+
+int
+MigratorApplication::run()
+{
+    return std::visit(
+        util::OverloadSet{
+            [this](Cmd::Status const&) { return printStatus(); },
+            [this](Cmd::Migration const& cmdBundle) { return migrate(cmdBundle.migratorName); },
+            [this](Cmd::Rollback const& cmdBundle) { return rollback(cmdBundle.migratorName); }
```
kuznetsss: Do we need to support a rollback?

src/migration/MigrationApplication.cpp
```diff
@@ -0,0 +1,119 @@
...
+    std::cout << "Current Migration Status:" << std::endl;
+    auto const allMigratorsStatus = migrationManager_->allMigratorsStatus();
+
+    if (allMigratorsStatus.empty()) {
+        std::cout << "No migrator found" << std::endl;
+    }
+
+    for (auto const& [migrator, status] : allMigratorsStatus) {
+        std::cout << "Migrator: " << migrator << " - "
+                  << (status == migration::MigratorStatus::Migrated ? "migrated" : "not migrated") << std::endl;
```
kuznetsss: Maybe it should be `Feature:` instead of `Migrator:` here. Migrator is tool which changes versions of features.
And also maybe it should be version instead of status, so something like:
```
Feature: NFT  current version: 1.0     latest version:1.2 - Requires migration
Feature: Successors  current version: 1.0      latest version: 1.0 - OK
```

src/migration/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+    
+    Current Migration Status:
+    Migrator: ExampleMigrator - not migrated
+    
+
+### To start a migration:
+    
+    
+    ./clio_server --migrate ExampleMigrator  ~/config/migrator.json
```
kuznetsss, line 30: As I understand `ExampleMigrator` is a feature name. Why does a user have to specify each feature to migrate separately? I think it would be convenient to have a list of features in the main Clio config and run `./clio_server --migrate config.json` to migrate all the features.

src/migration/cassandra/FullTableScanner.hpp
```diff
@@ -0,0 +1,172 @@
...
+        {
+            auto const minValue = std::numeric_limits<std::int64_t>::min();
+            auto const maxValue = std::numeric_limits<std::int64_t>::max();
+
+            // Safely calculate the range size using uint64_t to avoid overflow
+            uint64_t rangeSize = (static_cast<uint64_t>(maxValue) * 2) / numRanges_;
+
+            std::vector<TokenRange> ranges;
+            for (std::int64_t i = 0; i < numRanges_; ++i) {
+                int64_t start = minValue + i * static_cast<uint64_t>(rangeSize);
```
kuznetsss: `rangeSize` is already `uint64_t`.

src/migration/cassandra/FullTableScanner.hpp
```diff
@@ -0,0 +1,172 @@
...
+        [[nodiscard]] std::vector<TokenRange>
+        getRanges() const
+        {
+            auto const minValue = std::numeric_limits<std::int64_t>::min();
+            auto const maxValue = std::numeric_limits<std::int64_t>::max();
+
+            // Safely calculate the range size using uint64_t to avoid overflow
+            uint64_t rangeSize = (static_cast<uint64_t>(maxValue) * 2) / numRanges_;
+
+            std::vector<TokenRange> ranges;
```
kuznetsss:
```cpp
ranges.reserve(numRanges_);
```

src/util/Concepts.hpp
```diff
@@ -46,4 +48,25 @@ hasNoDuplicates(auto&&... values)
...
+ * @returns true if no duplicates exist; false otherwise
+ */
+template <typename... Types>
+constexpr bool
+hasNoDuplicateNames()
+{
+    constexpr std::array<std::string_view, sizeof...(Types)> names = {Types::name...};
+    for (std::size_t i = 0; i < names.size(); ++i) {
+        for (std::size_t j = i + 1; j < names.size(); ++j) {
+            if (names[i] == names[j]) {
```
kuznetsss: Is there a guarantee that `Types` are sorted by name?

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
```diff
@@ -0,0 +1,334 @@
...
+        testMigrationBackend = testBundle.second;
+
+        setupDatabase();
+    }
+
+    void
+    TearDown() override
+    {
+        testMigrationManager.reset();
+        testMigrationBackend.reset();
```
kuznetsss: Why do you need to explicitly call it here?

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+    std::shared_ptr<MigrationManagerInterface> testMigrationManager;
+    std::shared_ptr<CassandraMigrationTestBackend> testMigrationBackend;
+
+    void
+    SetUp() override
+    {
+        auto const testBundle = make_MigrationTestManagerAndBackend(cfg);
+        testMigrationManager = testBundle.first;
+        testMigrationBackend = testBundle.second;
```
kuznetsss, line 149:
```suggestion
        auto const [testMigrationManager, testMigrationBackend] = make_MigrationTestManagerAndBackend(cfg);
```

kuznetsss: Also why not to make it in constructor?

src/migration/MigrationApplication.cpp
cindyyan317: Yes. It is a good idea to have version for feature. The version info should given by a feature management in future.

src/migration/README.md
cindyyan317: I think the migration is a one time job. If we put it into a config.json, this config can only be used for one time, right?

src/util/Concepts.hpp
cindyyan317: I think the Types don't need to be sorted to find duplicate, right?

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
cindyyan317: Copied from otherwhere 😄

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
cindyyan317: I think i can. will modify it later.

src/migration/README.md
kuznetsss: Migration is one time job for one Clio version or one set of features. But once one of these changed migration is required again. I think there should be a list of desired features in `config.json` and the list of actual features is in database.

src/data/CassandraBackend.hpp
cindyyan317: I think "not res" means there is error while reading. Current database will have this table for sure , just empty maybe.

src/CMakeLists.txt
cindyyan317: I think migration is not only the data part. To achieve the 0 down time, it needs to have ETL feature as well. The ETL part will be integrated after ETL refactor.

src/data/cassandra/Types.hpp
cindyyan317: I think it will only be null-terminated string.

src/migration/README.md
cindyyan317: I don't think it is the time to put the name in config now. 
1 What happen if we remove the item from the config?
2 For some migrators , it is really one time job. For example , if we decide the remove diff table, we don't want to carry this migrator name in config after the migration has done.
3 The feature or plugin can be enabled in config in future, after we have more solid design for feature system.

src/main/Main.cpp
cindyyan317: No extra config, it will use the same config with clio_server. Its main purpose is to provide db setting.

src/util/Concepts.hpp
kuznetsss: As I understand you go through an array of string_views and check each pair for equality. It will work if the array is sorted `{ Type_A, Type_A, Type_B }` but if it is not sorted it will not work, for example for `{ Type_A, Type_B, Type_A }` this function will not find duplicate.

kuznetsss: Sorry, my bad, the code is correct.

src/data/CassandraBackend.hpp
```diff
@@ -835,6 +838,26 @@ class BasicCassandraBackend : public BackendInterface {
...
+            LOG(log_.error()) << "Could not fetch migrator status: " << res.error();
+            return {};
+        }
+
+        auto const& results = res.value();
+        if (not results) {
+            return {};
+        }
+
+        for (auto [statusString] : extract<std::string>(results))
```
kuznetsss, line 854: Do you need `[]` here? And should it be a loop because there is return on the first iteration?

src/migration/cassandra/CassandraMigrationBackend.hpp
```diff
@@ -0,0 +1,102 @@
...
+     *@param start The start token
+     *@param end The end token
+     *@param callback The callback to call for each row
+     *@param yield The boost asio yield context
+     */
+    template <impl::TableSpec TableDesc>
+    void
+    migrateInTokenRange(
+        std::int64_t const& start,
+        std::int64_t const& end,
```
kuznetsss:
```suggestion
        std::int64_t const start,
        std::int64_t const end,
```

src/migration/cassandra/CassandraMigrationBackend.hpp
```diff
@@ -0,0 +1,102 @@
...
+        LOG(log_.debug()) << "Travsering token range: " << start << " - " << end
+                          << " ; table: " << TableDesc::TABLE_NAME;
+        // for each table we only have one prepared statement
+        static auto statementPrepared =
+            migrationSchema_.getPreparedFullScanStatement(handle_, TableDesc::TABLE_NAME, TableDesc::PARTITION_KEY);
+
+        auto const statement = statementPrepared.bind(start, end);
+
+        auto const res = this->executor_.read(yield, statement);
+        auto const& results = res.value();
```
kuznetsss, line 93: There should be `ASSERT()` here that `res` has value.

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -0,0 +1,173 @@
...
+
+            // Safely calculate the range size using uint64_t to avoid overflow
+            uint64_t rangeSize = (static_cast<uint64_t>(maxValue) * 2) / numRanges_;
+
+            std::vector<TokenRange> ranges;
+            ranges.reserve(numRanges_);
+
+            for (std::int64_t i = 0; i < numRanges_; ++i) {
+                int64_t start = minValue + i * rangeSize;
+                int64_t end = (i == numRanges_ - 1) ? maxValue : start + static_cast<int64_t>(rangeSize) - 1;
```
kuznetsss, line 99: There could be an overflow here if `numRanges_` is `1`.

src/migration/impl/MigrationManagerFactory.cpp
```diff
@@ -0,0 +1,59 @@
...
+makeMigrationManager(util::Config const& config)
+{
+    static util::Logger const log{"Migration"};
+    LOG(log.info()) << "Constructing MigrationManager";
+
+    auto const type = config.value<std::string>("database.type");
+
+    if (not boost::iequals(type, "cassandra")) {
+        LOG(log.error()) << "Unknown database type to migrate: " << type;
+        throw std::runtime_error("Invalid database type");
```
kuznetsss: Maybe it would be better to return `std::expected<std::shared_ptr<impl::MigrationManagerInterface>, std::string>`?

tests/common/util/MockMigrationBackend.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <boost/json/object.hpp>
+#include <gmock/gmock.h>
+#include <xrpl/basics/base_uint.h>
+#include <xrpl/protocol/AccountID.h>
+#include <xrpl/protocol/LedgerHeader.h>
+
+struct MockMigrationBackend : public MockBackend {
+    MockMigrationBackend(util::Config) : MockBackend(util::Config{})
+    {
+    }
```
kuznetsss:
```suggestion
      using MockBackend::MockBackend;
```

tests/common/util/MockMigrationBackend.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "util/MockBackend.hpp"
+#include "util/config/Config.hpp"
+
+#include <boost/asio/spawn.hpp>
+#include <boost/json/object.hpp>
+#include <gmock/gmock.h>
+#include <xrpl/basics/base_uint.h>
+#include <xrpl/protocol/AccountID.h>
+#include <xrpl/protocol/LedgerHeader.h>
```
kuznetsss: Looks like you don't need these includes here.

tests/integration/migration/cassandra/CassandraMigrationTestBackend.hpp
```diff
@@ -0,0 +1,344 @@
...
+    CassandraMigrationTestBackend(data::cassandra::SettingsProvider settingsProvider)
+        : migration::cassandra::CassandraMigrationBackend(settingsProvider)
+        , settingsProvider_(std::move(settingsProvider))
+
+    {
+        if (auto const res = handle_.executeEach(createTablesSchema()); not res)
+            throw std::runtime_error("Could not create schema: " + res.error());
+    }
+
+    ~CassandraMigrationTestBackend() override = default;
```
kuznetsss: I think this is not needed.

tests/unit/migration/MigrationManagerBaseTests.cpp
```diff
@@ -0,0 +1,96 @@
...
+TEST_F(MigrationManagerBaseTest, AllStatus)
+{
+    EXPECT_CALL(*backend, fetchMigratorStatus("SimpleTestMigrator", testing::_)).WillOnce(testing::Return("Migrated"));
+    EXPECT_CALL(*backend, fetchMigratorStatus("SimpleTestMigrator2", testing::_))
+        .WillOnce(testing::Return("NotMigrated"));
+    auto const status = migrationManager->allMigratorsStatus();
+    EXPECT_EQ(status.size(), 2);
+    EXPECT_TRUE(
+        std::find(
+            status.begin(), status.end(), std::make_tuple("SimpleTestMigrator", migration::MigratorStatus::Migrated)
```
kuznetsss, line 71: Why `make_tuple`? I thought `Status` has a constructor.

cindyyan317: It is tuple of the name of the migrator and its status.

src/data/CassandraBackend.hpp
kuznetsss: It seems there is no other way to do it.

tests/unit/migration/MigrationManagerBaseTests.cpp
kuznetsss: So `status` here is a vector of tuples containing name and status. Maybe then it would be better to call it `migratorNameStatusPairs`.

kuznetsss (dismissed):
👍

src/app/CliArgs.cpp
```diff
@@ -45,6 +46,7 @@ CliArgs::parse(int argc, char const* argv[])
         ("version,v", "print version and exit")
         ("conf,c", po::value<std::string>()->default_value(defaultConfigPath), "configuration file")
         ("ng-web-server,w", "Use ng-web-server")
+        ("migrate", po::value<std::string>(),"start migration helper")
```
godexsoft: Space missing 🔢

src/data/BackendInterface.hpp
```diff
@@ -673,6 +683,15 @@ class BackendInterface {
     bool
     finishWrites(std::uint32_t ledgerSequence);
 
+    /**
+     *@brief Mark the migration status of a migrator as Migrated in the database
```
godexsoft: Here and below broken comments alignment

src/data/cassandra/impl/Statement.hpp
```diff
@@ -119,6 +119,9 @@ class Statement : public ManagedObject<CassStatement> {
             // reinterpret_cast is needed here :'(
             auto const rc = bindBytes(reinterpret_cast<unsigned char const*>(value.data()), value.size());
             throwErrorIfNeeded(rc, "Bind string (as bytes)");
+        } else if constexpr (std::is_convertible_v<DecayedType, Text>) {
+            auto const rc = cass_statement_bind_string(*this, idx, value.text.c_str());
```
godexsoft: This is not an issue now but will be later for sure - there probably is `cass_statement_bind_string_n` where you can pass the size of the string as last arg

src/migration/MigratiorStatus.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * @brief The status of a migrator, it provides the helper functions to convert the status to string and vice versa
+ */
+class MigratorStatus {
+public:
+    /**
+     * @brief The status of a migrator
+     */
+    enum Status { Migrated, NotMigrated, NotKnown, Count };
```
godexsoft: Probably should be `enum class`

src/migration/MigratiorStatus.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+private:
+    // Private static array for mapping
+    static constexpr std::array<char const*, static_cast<size_t>(Count)> statusStrMap = {
+        "Migrated",
+        "NotMigrated",
+        "NotKnown"
+    };
+
+    Status status_;
```
godexsoft, line 88: Data is normally at the top in our codebase

src/migration/cassandra/CassandraMigrationManager.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "migration/impl/MigratorsRegister.hpp"
+
+namespace migration::cassandra {
+
+// Register migrators here
+// MigratorsRegister<BackendType, ExampleMigrator>
+template <typename BackendType>
+using CassandraSupportedMigrators = migration::impl::MigratorsRegister<BackendType>;
+
+/// Register with MigrationBackend which proceeds the migration
```
godexsoft: We don't use `///` syntax unless it's `///<` on the same line for member variable docs (and even then, prefer `/**<`)

src/migration/cassandra/impl/CassandraMigrationSchema.hpp
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param handler The database handler
+     * @return The prepared statement to insert into migrator_status table
+     */
+    data::cassandra::PreparedStatement const&
+    getPreparedInsertMigratedMigrator(data::cassandra::Handle const& handler)
+    {
+        static auto prepared = handler.prepare(fmt::format(
+            R"(
+            INSERT INTO {}
```
godexsoft, line 89: Still broken formatting

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -0,0 +1,175 @@
...
+    {
+        namespace vs = std::views;
+
+        tasks_.reserve(workerNum);
+
+        for ([[maybe_unused]] auto taskId : vs::iota(0u, workerNum))
+            tasks_.push_back(spawnWorker());
+    }
+
+    static auto constexpr CursorNumMultiplier = 100;  // will generate 100*workersNum cursors
```
godexsoft: Is this an arbitrary number? how was it chosen? Should not be configurable in any way?

src/migration/cassandra/impl/ObjectsAdapter.hpp
```diff
@@ -0,0 +1,80 @@
...
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <optional>
+#include <tuple>
+#include <utility>
+
+namespace migration::cassandra::impl {
+
+/**
```
godexsoft, line 40: Nice to have more docs but in `impl` we agreed we did not need to add them. Let's keep them though since you already added them

src/migration/impl/MigrationManagerInterface.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <tuple>
+#include <vector>
+
+namespace migration::impl {
+
+/**
+ * @brief The interface for the migration manager. This interface is tend to be implemented for specific database. The
+ * application layer will use this interface to run the migrations.
+ */
+struct MigrationManagerInterface {
```
godexsoft: Interfaces should be in the main namespace of the feature. Implementation of interfaces should be in `impl`

src/migration/impl/MigratorsRegister.hpp
```diff
@@ -0,0 +1,184 @@
...
+     * migrator
+     */
+    std::vector<std::tuple<std::string, MigratorStatus>>
+    getMigratorsStatus() const
+    {
+        auto const fullList = getMigratorNames();
+
+        std::vector<std::tuple<std::string, MigratorStatus>> status;
+
+        std::transform(fullList.begin(), fullList.end(), std::back_inserter(status), [&](auto const& migratorName) {
```
godexsoft: Ranges could be used here and anywhere you have `.begin()` and `.end()` used in algorithms

src/util/Concepts.hpp
```diff
@@ -46,4 +48,25 @@ hasNoDuplicates(auto&&... values)
...
+ *
+ * @tparam Types The types to check
+ * @returns true if no duplicates exist; false otherwise
+ */
+template <typename... Types>
+constexpr bool
+hasNoDuplicateNames()
+{
+    constexpr std::array<std::string_view, sizeof...(Types)> names = {Types::name...};
+    for (std::size_t i = 0; i < names.size(); ++i) {
```
godexsoft: `std::ranges::any_of`?

src/util/log/Logger.hpp
```diff
@@ -177,14 +177,15 @@ class Logger final {
     };
 
 public:
-    static constexpr std::array<char const*, 7> CHANNELS = {
+    static constexpr std::array<char const*, 8> CHANNELS = {
```
godexsoft, line 172: out of scope of this PR: We should probably rewrite this with `std::to_array` to avoid the need to modify the length each time we add stuff here

tests/common/util/MockMigrationBackendFixture.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+        MockType<MockMigrationBackend>&
+        operator*()
+        {
+            return *backend;
+        }
+    };
+
+protected:
+    BackendProxy backend;
```
godexsoft: The point of making the data `protected` here is so we can legitimately call it `backend_` to make it more obvious in the test code that this is a member of the test fixture

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    void
+    SetUp() override
+    {
+        setupDatabase();
```
godexsoft, line 155: Don't we need to call superclass `SetUp` and `TearDown`? Perhaps we can avoid needing these and just use constructor and destructor?

tests/integration/migration/cassandra/CassandraMigrationTestBackend.hpp
```diff
@@ -0,0 +1,342 @@
...
+private:
+    std::vector<data::cassandra::Statement>
+    createTablesSchema()
+    {
+        std::vector<data::cassandra::Statement> statements;
+
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  (
```
godexsoft: Formatting needed

tests/integration/migration/cassandra/CassandraMigrationTestBackend.hpp
```diff
@@ -0,0 +1,342 @@
...
+                        PRIMARY KEY (hash) 
+                  ) 
+            )",
+            data::cassandra::qualifiedTableName(settingsProvider_, "tx_index_example")
+        ));
+
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  (
```
godexsoft: Formatting

tests/integration/migration/cassandra/ExampleObjectsMigrator.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <ostream>
+#include <unordered_set>
+
+std::atomic_int64_t ExampleObjectsMigrator::count;
+std::atomic_int64_t ExampleObjectsMigrator::accountCount;
+
+void
+ExampleObjectsMigrator::runMigration(std::shared_ptr<Backend> const& backend, util::Config const& config)
+{
+    // Do something
```
godexsoft: ? 😃

tests/integration/util/CassandraDBHelper.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <fmt/format.h>
+
+#include <cstdint>
+#include <string>
+
+data::cassandra::ResultOrError
+writeTxFromCSVString(std::string const& space, std::string const& record, data::cassandra::Handle const& handler)
+{
+    std::string statement = fmt::format(
+        "insert into {}.transactions (hash, date, ledger_sequence, metadata, transaction) VALUES ({})", space, record
```
godexsoft: Please use `INSERT` and `INTO` the same way we use `VALUES` 🔢

tests/unit/migration/MigrationManagerBaseTests.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    auto const names = migrationManager->allMigratorsNames();
+    EXPECT_EQ(names.size(), 2);
+    EXPECT_EQ(names[0], "SimpleTestMigrator");
+    EXPECT_EQ(names[1], "SimpleTestMigrator2");
+}
+
+TEST_F(MigrationManagerBaseTest, RunMigration)
+{
+    EXPECT_CALL(*backend, writeMigratorStatus("SimpleTestMigrator", "Migrated")).Times(1);
```
godexsoft: `Times(1)` is default

tests/unit/migration/MigratorRegisterTests.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+using EmptyMigratorRegister = migration::impl::MigratorsRegister<MockMigrationBackend>;
+
+struct MigratorRegisterTests : public util::prometheus::WithMockPrometheus, public MockMigrationBackendTest {
+    util::Config cfg;
+};
+
+TEST_F(MigratorRegisterTests, EmptyMigratorRegister)
+{
+    auto mockBackendPtr = backend.operator std::shared_ptr<MockMigrationBackend>();
```
godexsoft: You should not need this? `backend` should implicitly convert to the `shared_ptr` so you should be able to just pass it to `EmptyMigratorRegister` and the like?

tests/unit/migration/cassandra/FullTableScannerTests.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <functional>
+
+namespace {
+
+// Help to verify that the function is called
+struct MockObject {
+    MOCK_METHOD(void, call, (migration::cassandra::impl::TokenRange const&, boost::asio::yield_context));
```
godexsoft: This looks like a `testing::MockFunction`, maybe we can use it instead?

godexsoft (changes requested):
Found more small things 🔢

src/data/cassandra/impl/Statement.hpp
cindyyan317: I will fix it now. I think string data() makes sure the data() ends with '\0' after C++11. But I can not find it in reference.

src/migration/MigratiorStatus.hpp
cindyyan317: Status is defined inside MigratorStatus, So I can not put in at the top.

tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp
cindyyan317: The base classes don't have Setup and TearDown. But in case they add one in future, I will add the callee here.
I can not move setupDatabase to constructor, it is virtual function. I want to it behave polymorphically for different migrator test.

src/migration/MigratiorStatus.hpp
cindyyan317: I compared the scoped enum and old enum. I want to use refer the status like `MigratorStatus::Migrated`, Instead of `MigratorStatus::Status::Migrated`. Is it too verbose ?

src/migration/cassandra/impl/FullTableScanner.hpp
cindyyan317: Because the total cursor numbers is `workersNum* CursorNumMultiplier`, workersNum is configurable coroutine numbers. Each coroutine can handle `CursorNumMultiplier` cursor by average. But the content are not distribute evenly, so `CursorNumMultiplier` split the cursors range to smaller piece. It is hard to decide which number is the best without the insight of data distribution in particular tables. But I think with CursorNumMultiplier 100, user can tweet total cursors number by adjusting the number of coroutines.

src/migration/MigratiorStatus.hpp
cindyyan317: I compared the scoped enum and old enum. I want to use refer the status like `MigratorStatus::Migrated`, Instead of `MigratorStatus::Status::Migrated`. Is it too verbose ?

godexsoft: I also just now noticed the name of the file is messed up too 🔢

src/migration/MigratiorStatus.hpp
godexsoft: You definitely can if you re-arrange everything properly 🥇 Remember that you can use `using` and put structs/enums outside.

src/migration/MigratiorStatus.hpp
```diff
@@ -33,7 +33,7 @@ class MigratorStatus {
     /**
      * @brief The status of a migrator
      */
-    enum Status { Migrated, NotMigrated, NotKnown, Count };
+    enum Status { Migrated, NotMigrated, NotKnown, COUNT };
```
godexsoft: Why the name had to change? Also `Count` is a little too generic for this list end marker. I have seen people call them `NumStatuses`, `TotalNumStatuses` or similar in the wild.

src/migration/cassandra/impl/CassandraMigrationSchema.hpp
```diff
@@ -67,7 +67,7 @@ class CassandraMigrationSchema {
             R"(
             SELECT * 
               FROM {} 
-             WHERE TOKEN({}) >= ? AND TOKEN({}) <= ?
+            WHERE TOKEN({}) >= ? AND TOKEN({}) <= ?
```
godexsoft: Why broken now? 😃

tests/unit/migration/cassandra/FullTableScannerTests.cpp
```diff
@@ -53,25 +55,50 @@ struct FullTableScannerTests : public NoLoggerFixture {};
 
 TEST_F(FullTableScannerTests, workerNumZero)
 {
-    MockObject obj;
+    testing::MockFunction<void(migration::cassandra::impl::TokenRange const&, boost::asio::yield_context)> mockCallback;
     EXPECT_DEATH(
-        migration::cassandra::impl::FullTableScanner<TestScannerAdaper>(1, 0, TestScannerAdaper(obj)),
+        migration::cassandra::impl::FullTableScanner<TestScannerAdaper>(1, 0, 100, TestScannerAdaper(mockCallback)),
```
godexsoft: Everytime i see an API that takes 3+ ints i feel like one struct with settings for this API call instead of 3 ints would be much easier to use and harder to get wrong. Like 
```cpp
FullTableScanner({.one = 1, .two = 2}, TestScannerAdapter(callback));
```

godexsoft (changes requested):
Just a few more small things.
The API change for `FullTableScanner` is optional but encouraged 👍

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -136,33 +136,36 @@ class FullTableScanner {
     TableAdapter reader_;
 
 public:
+    /**
+     * @brief The full table scanner settings.
+     */
+    struct FullTableScannerSettings {
+        std::uint32_t ctxThreadsNum;  // number of threads used in the execution context
```
godexsoft: These comments could be `/**< text */` to make them into the doc 👍

godexsoft (changes requested):
Let's make the field comments into doxygen comments and we should be good to go 👍

godexsoft (approved):
Let's get this merged 🥇

PR #1772, opened 2024-12-05
feat: Add dosguard to new webserver

### LINKED ISSUE #919: Refactor web server

kuznetsss:
For #919. After this the only thing left which is to switch to the new web server by default and delete the old code.


src/app/WebHandlers.hpp
```diff
@@ -0,0 +1,235 @@
...
+            if (not request.isHttp()) {
+                try {
+                    auto requestJson = boost::json::parse(request.message());
+                    if (requestJson.is_object() && requestJson.as_object().contains("id"))
+                        error["id"] = requestJson.as_object().at("id");
+                    error["request"] = request.message();
+                } catch (std::exception const&) {
+                    error["request"] = request.message();
+                }
+            }
```
cindyyan317, line 198: Just want to mention the current Clio would count these error into the Dosguard too. Should it cause problem? like if the request is a long string?

src/app/ClioApplication.cpp
```diff
@@ -158,72 +152,18 @@ ClioApplication::run(bool const useNgWebServer)
         }
         auto const adminVerifier = std::move(expectedAdminVerifier).value();
 
-        auto httpServer = web::ng::make_Server(config_, ioc);
+        auto httpServer = web::ng::make_Server(config_, OnConnectCheck{dosGuard}, DisconnectHook{dosGuard}, ioc);
```
cindyyan317, line 155: is this DisconnectHook the OnDisconnectHook type in server ?
Ha i see. OnDisconnectHook is a function object.

src/web/ng/Server.hpp
```diff
@@ -42,6 +44,19 @@ namespace web::ng {
  * @brief Web server class.
  */
 class Server {
+public:
+    /**
+     * @brief Check to perform for each new client connection. The check takes client ip as input and returns a Response
```
cindyyan317, line 49: It can be : Check to perform for each new client connection. The check takes client ip as input and returns a Response _if check fails_ .  I feel we are expecting void here, response is an error. Can we  use std::expect here?

tests/unit/app/WebHandlersTests.cpp
```diff
@@ -0,0 +1,319 @@
...
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <utility>
+
+using namespace app;
+namespace http = boost::beast::http;
+
+struct WebHandlersTest : virtual NoLoggerFixture {
+    DOSGuardStrictMock dosGuardMock_;
```
cindyyan317, line 55: `_` seems for private member names. But sometimes I do the same thing in test too :P

kuznetsss: Yeah for tests it is more convenient to do this to make it highlight fixture's fields.

cindyyan317 (approved):

godexsoft (approved):
👍

PR #1773, opened 2024-12-07
fix: Fix and re-enable flaky test

### LINKED ISSUE #1752: Flaky test: DataTransferredAndObserverCalledCorrectly
## Issue Description
The test seems flaky. 

## Steps to Reproduce
Run it enough times and observe it fail. `--gtest_repeat=-1` helps.

## Expected Result
Test should pass everytime.

## Actual Result
Test fails sometimes.

godexsoft:
Fixes #1752 

Since the calls happen asynchronously, we could not guarantee the order. The solution is to calculate things in the same `EXPECT_CALL` instead of trying to separate into two expectations with `InSequence`.


kuznetsss (approved):
👍

PR #1774, opened 2024-12-08
chore: Add clang-tidy 19 checks

### LINKED ISSUE #1664: Update tools to LLVM 19.1
## Summary

LLVM 19.1 released. Once it appears in brew (https://github.com/Homebrew/homebrew-core/pull/181351) we should switch to it in CI.
Also [new checks](https://releases.llvm.org/19.1.0/tools/clang/tools/extra/docs/ReleaseNotes.html#improvements-to-clang-tidy) were added to clang-tidy. We should look at it and add some to clang tidy config.

godexsoft:
Fix #1664 

Adds a few useful checks and fixes the code to accommodate them.
[`modernize-use-ranges`](https://releases.llvm.org/19.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize/use-ranges.html) seems to work with a few modifications to the code and exceptions for the linter.


src/web/ng/Response.cpp
```diff
@@ -107,7 +107,9 @@ Response::message() const
     return std::visit(
         util::OverloadSet{
             [](http::response<http::string_body> const& response) -> std::string const& { return response.body(); },
-            [](std::string const& message) -> std::string const& { return message; },
+            [](std::string const& message) -> std::string const& {
+                return message;  // NOLINT(bugprone-return-const-ref-from-parameter)
```
godexsoft, line 111: This one i leave to you @kuznetsss

src/util/prometheus/Label.cpp
```diff
@@ -73,7 +60,7 @@ Label::serialize() const
 
 Labels::Labels(std::vector<Label> labels) : labels_(std::move(labels))
 {
-    std::sort(labels_.begin(), labels_.end());  // NOLINT
+    std::ranges::sort(labels_);
```
godexsoft, line 63: @kuznetsss please check that the changes to support this update don't break anything (tests pass fine). the code is equivalent as far as i can tell.

kuznetsss: Looks good.

src/web/ng/Response.cpp
kuznetsss: This is fine.

src/rpc/CredentialHelpers.cpp
```diff
@@ -53,6 +53,7 @@
 namespace rpc::credentials {
 
 bool
+// NOLINTNEXTLINE(misc-use-internal-linkage)
```
kuznetsss, line 56: Probably we are doing something wrong with this check or it generates too many false positives.

kuznetsss (approved):
👍

cindyyan317 (approved):
💯

src/rpc/CredentialHelpers.cpp
godexsoft: I think it's just the way we have global functions that become unused (within the internal library it compiles for) if we mark them `static`. So the solution is to not have these global freestanding functions. This is not something I wanted to change within this PR. I think this can be addressed separately later. And this check enabled now means that we can no longer add those global functions and shall find a suitable workaround next time we need them. I am sure a simple wrapper class with static functions would suffice (e.g. `NFTUtils`).

PR #1776, opened 2024-12-11
feat: Repeating operations for util::async

godexsoft:
Async framework needed a way to do repeating operations (think simplest cases like AmendmentBlockHandler).
This PR implements the **absolute minimum**, barebones repeating operations that
- Can't return any values (void)
- Do not take any arguments in the user-provided function
- Can't be scheduled (i.e. a delay before starting repeating)
- Can't be stopped from inside the user block of code (i.e. does not have stop token or anything like that) 
- Can be stopped through the RepeatedOperation's `abort()` function (but not from the user-provided repeating function)


src/util/Repeat.hpp
```diff
@@ -43,9 +43,11 @@ class Repeat {
     /**
      * @brief Construct a new Repeat object
      *
-     * @param ioc The io_context to use
+     * @param ctx The io_context-like object to use
      */
-    Repeat(boost::asio::io_context& ioc);
+    Repeat(auto& ctx) : timer_(ctx)
```
godexsoft: This makes it usable with other types of contexts such as `thread_pool` or `strand`.

src/util/async/Operation.hpp
```diff
@@ -181,4 +186,58 @@ using Operation = impl::BasicOperation<Outcome<RetType>>;
...
+    void
+    abort() noexcept
+    {
+        op_.requestStop();
+        repeat_->stop();
+    }
+
+    /**
+     * @brief Wait for the operation to get cancelled
+     * @warning Calling this from inside of the repeating operation yields a deadlock
```
godexsoft: This can be worked around with custom code instead of `util::Repeat` but i decided not to make it more complicated than is necessary now. We can always improve it in the future as long as the interface remains the same or does not change too much.

src/util/Repeat.hpp
kuznetsss: Maybe worth to leave this comment in the code.

src/util/async/Operation.hpp
```diff
@@ -181,4 +186,58 @@ using Operation = impl::BasicOperation<Outcome<RetType>>;
...
+ * is also no mechanism to request the repeating task to stop from inside of the user provided block of code.
+ *
+ * @tparam CtxType The type of the execution context
+ */
+template <typename CtxType>
+class RepeatingOperation {
+    using DataType = std::expected<void, ExecutionError>;
+    using StopSourceType = typename CtxType::StopSource;
+
+    std::unique_ptr<util::Repeat> repeat_;
```
kuznetsss: Why unique_ptr?

src/util/async/Operation.hpp
```diff
@@ -181,4 +186,58 @@ using Operation = impl::BasicOperation<Outcome<RetType>>;
...
+        op_.requestStop();
+        repeat_->stop();
+    }
+
+    /**
+     * @brief Wait for the operation to get cancelled
+     * @warning Calling this from inside of the repeating operation yields a deadlock
+     */
+    void
+    wait() noexcept
```
kuznetsss: Probably this is convenience for the whole framework but from method `wait()` I would expect waiting for an operation to finish.

tests/unit/util/async/AnyExecutionContextTests.cpp
```diff
@@ -270,6 +273,20 @@ TEST_F(AnyExecutionContextTests, TimerWithBoolHandlerExecuted)
     EXPECT_EQ(timer.get().value(), 42);
 }
 
+TEST_F(AnyExecutionContextTests, RepeatingOperation)
+{
+    auto mockRepeatingOp = RepeatingOperationType<std::any>{};
+    EXPECT_CALL(mockRepeatingOp, wait());
+    EXPECT_CALL(
+        mockExecutionContext, executeRepeatedly(An<std::chrono::milliseconds>(), An<std::function<std::any()>>())
```
kuznetsss:
```suggestion
        mockExecutionContext, executeRepeatedly(A<std::chrono::milliseconds>(), A<std::function<std::any()>>())
```
Also why not 1 millisecond?

tests/unit/util/async/AnyExecutionContextTests.cpp
```diff
@@ -270,6 +273,20 @@ TEST_F(AnyExecutionContextTests, TimerWithBoolHandlerExecuted)
...
+    )
+        .WillOnce([&mockRepeatingOp] -> RepeatingOperationType<std::any> const& { return mockRepeatingOp; });
+
+    auto res = ctx.executeRepeatedly(std::chrono::milliseconds(1), [] -> void { throw 0; });
```
kuznetsss: Why throwing? If you don't want this to be called probably strict function mock would be better.

tests/unit/util/async/AsyncExecutionContextTests.cpp
```diff
@@ -167,6 +169,17 @@ TYPED_TEST(ExecutionContextTests, timerUnknownException)
     EXPECT_TRUE(std::string{err}.ends_with("unknown"));
 }
 
+TYPED_TEST(ExecutionContextTests, repeatingOperation)
+{
+    testing::MockFunction<void()> call;
+
+    auto res = this->ctx.executeRepeatedly(std::chrono::milliseconds(1), [&] { call.Call(); });
```
kuznetsss:
```suggestion
    auto res = this->ctx.executeRepeatedly(std::chrono::milliseconds(1), call.AsStdFunction());
```

src/util/async/Operation.hpp
```diff
@@ -181,4 +186,58 @@ using Operation = impl::BasicOperation<Outcome<RetType>>;
...
+     * @param executor The executor to operate on
+     * @param interval Time to wait before repeating the user-provided block of code
+     * @param fn The function to execute repeatedly
+     */
+    RepeatingOperation(auto& executor, std::chrono::steady_clock::duration interval, std::invocable auto&& fn)
+        : repeat_(std::make_unique<util::Repeat>(executor)), op_(&outcome_)
+    {
+        repeat_->start(interval, std::forward<decltype(fn)>(fn));
+    }
+
```
kuznetsss, line 218: Don't you need a custom destructor here? It looks like `op_` will die before `repeat_` now.

src/util/async/Operation.hpp
godexsoft: Yea, that's what it sort of does, waits until "finished" which in case of a repeating block of code is same as "aborted".. the naming may be not the best but it matches others in the framework

src/util/async/Operation.hpp
godexsoft: Repeat is not movable. But let me try with your PR #1775

tests/unit/util/async/AnyExecutionContextTests.cpp
godexsoft: It's just a way to make sure this function does not actually gets called or used - that's not the intention of this test. it's the same with all other checks in this file so i just went with the same style

godexsoft:
@kuznetsss @cindyyan317 I think the code is ok and ready to go but on CI/mac it did fail before with much higher than the amount of calls we expect. It happened again now: https://github.com/XRPLF/clio/actions/runs/12402585393/job/34625374542?pr=1776#step:4:2704
Any ideas welcome 👍

src/util/Repeat.cpp
```diff
@@ -19,24 +19,15 @@
...
-
 void
 Repeat::stop()
 {
-    *stopped_ = true;
-    timer_.cancel();
+    control_->stopping = true;
+    control_->timer.cancel();
+    control_->semaphore.acquire();
+    control_->stopping = false;
```
kuznetsss: Maybe move it to `start()` and add an assert to check that `stopping` is `false`?

src/util/Repeat.cpp
```diff
@@ -19,24 +19,15 @@
...
 namespace util {
 
-Repeat::Repeat(boost::asio::io_context& ioc) : timer_(ioc)
-{
-}
-
-Repeat::~Repeat()
-{
-    *stopped_ = true;
-}
```
kuznetsss, line 33: Currently without a destructor it is not safe to delete `Repeat` when it is running.

src/util/Repeat.hpp
```diff
@@ -83,11 +102,10 @@ class Repeat {
...
     startImpl(std::chrono::steady_clock::duration interval, Action&& action)
     {
-        timer_.expires_after(interval);
-        timer_.async_wait([this, interval, stopping = stopped_, action = std::forward<Action>(action)](
-                              auto const& errorCode
-                          ) mutable {
-            if (errorCode or *stopping) {
+        control_->timer.expires_after(interval);
+        control_->timer.async_wait([this, interval, action = std::forward<Action>(action)](auto const&) mutable {
+            if (control_->stopping) {
```
kuznetsss:
```suggestion
        control_->timer.async_wait([this, interval, action = std::forward<Action>(action)](auto const& ec) mutable {
            if (ec or control_->stopping) {
```

kuznetsss (approved):

PR #1780, opened 2024-12-13
fix: Fix bugs in new webserver

### LINKED ISSUE #919: Refactor web server

kuznetsss:
Fixes #919.

Fixes bugs for new webserver:
- Unhandled exception when closing already closed websocket
- No pings for plain websocket connection
- Server drops websocket connection when client responds to pings but doesn't send anything

Also changing API of ng connections. Now timeout is set by a separate method instead of providing it for each call.


src/web/ng/Connection.hpp
```diff
@@ -105,7 +105,7 @@ class Connection : public ConnectionMetadata {
     /**
      * @brief The default timeout for send, receive, and close operations.
      */
-    static constexpr std::chrono::steady_clock::duration DEFAULT_TIMEOUT = std::chrono::seconds{30};
+    static constexpr std::chrono::steady_clock::duration DEFAULT_TIMEOUT = std::chrono::seconds{10};
```
cindyyan317: I don't see this change in PR summary. Is it typo? or intended?

cindyyan317 (dismissed):
LGTM .
Left a question about the default value.

src/web/ng/Connection.hpp
kuznetsss: This is intended. I thought it is better to disconnect inactive clients earlier.

godexsoft (approved):
LGTM 👍

PR #1797, opened 2024-12-18
fix: Optimize `ledger_range` query

godexsoft:
There is a primary key on this table, `is_latest`.  So there can be 0, 1, or 2 rows in this table.
Adding the `inlist` on the primary key, scylla will hash the partition key (which is the primary key - `is_latest`), and only go to token ranges that match. This should be much faster than the current version.


godexsoft:
At the very least this does not break our DB tests and i have been told that this should be a lot faster than what we had.

cindyyan317 (dismissed):
LGTM

cindyyan317 (approved):
LGTM

PR #1800, opened 2024-12-20
fix: Another try to improve Retry

kuznetsss:


kuznetsss:
It looks like it should work but sometimes test `ExecutionContextTests/*.repeatingOperation` hangs.

godexsoft:
> It looks like it should work but sometimes test `ExecutionContextTests/*.repeatingOperation` hangs.
I'm not sure it's worth trying to improve this further if we already have a fully working (blocking) version and this updated one sometimes hangs. It's not like we are using the control block in a busy loop so i'd prefer to leave it as is 👍

godexsoft:
@kuznetsss you want to try this again at some point or should we just drop it?

kuznetsss:
@godexsoft, yeah, I want to investigate this implementation, so lets leave it for now.

PR #1801, opened 2024-12-20
feat: Graceful shutdown

### LINKED ISSUE #442: Handle SIGHUP and friends gracefully

kuznetsss:
Fixes #442.


src/app/Stopper.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <utility>
+
+namespace app {
+
+/**
+ * @brief Application stopper class. On stop it will create a new thread to run all the shutdown tasks.
+ */
+class Stopper {
+    boost::asio::io_context ctx_;
+    std::thread worker_;
```
godexsoft, line 44: Should we use async framework here or is this deliberately not using it?

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -157,59 +152,55 @@ SubscriptionSource::validatedRange() const
...
-            if (not connection) {
-                handleError(connection.error(), yield);
-                return;
-            }
+    boost::asio::spawn(strand_, [this, _ = boost::asio::make_work_guard(strand_)](boost::asio::yield_context yield) {
+        auto connection = wsConnectionBuilder_.connect(yield);
+        if (not connection) {
+            handleError(connection.error(), yield);
+            return;
+        }
```
godexsoft:
```suggestion
        if (auto connection = wsConnectionBuilder_.connect(yield); connection) {
            wsConnection_ = std::move(connection).value();
        } else {
            handleError(connection.error(), yield);
            return;
        }
```

This scopes `connection` so that there is no way anyone can use it after we moved from it, even if they wanted to

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -157,59 +152,55 @@ SubscriptionSource::validatedRange() const
...
+        wsConnection_ = std::move(connection).value();
 
-            auto const& subscribeCommand = getSubscribeCommandJson();
-            auto const writeErrorOpt = wsConnection_->write(subscribeCommand, yield, wsTimeout_);
-            if (writeErrorOpt) {
-                handleError(writeErrorOpt.value(), yield);
-                return;
-            }
+        auto const& subscribeCommand = getSubscribeCommandJson();
+        auto const writeErrorOpt = wsConnection_->write(subscribeCommand, yield, wsTimeout_);
```
godexsoft: Same as above

src/etl/impl/SubscriptionSource.cpp
```diff
@@ -157,59 +152,55 @@ SubscriptionSource::validatedRange() const
...
-                auto const handleErrorOpt = handleMessage(message.value());
-                if (handleErrorOpt) {
-                    handleError(handleErrorOpt.value(), yield);
-                    return;
-                }
+            auto const handleErrorOpt = handleMessage(message.value());
+            if (handleErrorOpt) {
+                handleError(handleErrorOpt.value(), yield);
+                return;
             }
```
godexsoft: And more of the same 👍

src/util/StopHelper.cpp
```diff
@@ -0,0 +1,46 @@
...
+    *stopped_ = true;
+}
+
+void
+StopHelper::asyncWaitForStop(boost::asio::yield_context yield)
+{
+    boost::asio::steady_timer timer{yield.get_executor(), std::chrono::steady_clock::duration::max()};
+    onStopReady_.connect([&timer]() { timer.cancel(); });
+    boost::system::error_code error;
+    if (not*stopped_)
```
godexsoft: Maybe in this case `!*stopped_` is better? 😃

src/util/StopHelper.cpp
```diff
@@ -0,0 +1,46 @@
...
+    onStopReady_();
+    *stopped_ = true;
+}
+
+void
+StopHelper::asyncWaitForStop(boost::asio::yield_context yield)
+{
+    boost::asio::steady_timer timer{yield.get_executor(), std::chrono::steady_clock::duration::max()};
+    onStopReady_.connect([&timer]() { timer.cancel(); });
+    boost::system::error_code error;
```
godexsoft, line 41: Should we do anything at all about this error if it's set?

src/web/ng/Server.cpp
```diff
@@ -288,15 +301,32 @@ Server::handleConnection(boost::asio::ip::tcp::socket socket, boost::asio::yield
...
+            LOG(log_.info()) << *connectionExpected.error();
         }
         return;
     }
     LOG(log_.trace()) << connectionExpected.value()->tag() << "Connection created";
 
+    if (connectionHandler_.isStopping()) {
+        boost::asio::spawn(
+            ctx_.get(),
+            [connection = std::move(connectionExpected).value()](boost::asio::yield_context yield) {
```
godexsoft, line 313: nit: Here and other places - we usually pass `yield` as `auto` in lambdas

src/app/ClioApplication.cpp
```diff
@@ -158,6 +161,22 @@ ClioApplication::run(bool const useNgWebServer)
...
+        appStopper_.setOnStop([&](boost::asio::yield_context yield) {
+            auto serverStopped = boost::asio::spawn(
+                yield, [&httpServer](auto innerYield) { httpServer->stop(innerYield); }, boost::asio::use_future
+            );
+            auto balancerStopped = boost::asio::spawn(
+                yield, [&balancer](auto innerYield) { balancer->stop(innerYield); }, boost::asio::use_future
+            );
+            serverStopped.get();
+            balancerStopped.get();
+            // etl->stop();
```
godexsoft: Are these left intentionally for future or just forgotten to be removed?

godexsoft (changes requested):
Adding some small suggestions. Looks pretty good 👍

src/app/Stopper.hpp
kuznetsss: Hm, I don't need context switching or stop token here, but probably still could switch.

src/etl/impl/SubscriptionSource.cpp
kuznetsss: It is an old code, but I will update 👍

src/util/StopHelper.cpp
kuznetsss: The error will be something like `operation_aborted` when timer is cancelled. If we don't provide an error, exception will be thrown when timer is cancelled.

src/app/ClioApplication.cpp
kuznetsss: It is still draft, so it will be fixed later.

src/web/ng/Server.cpp
kuznetsss: Why? Clangd couldn't guess the type here when `yield` is `auto`, but I understand that more typing is also not very convenient.

godexsoft: No particular reason other than the usual preference towards using `auto` when possible. 
Without `auto` there is indeed more typing as well as the need to include the header that provides the type of `yield` (because we enforce `iwyu`).

src/app/Stopper.hpp
kuznetsss: I tried but it seems async framework doesn't fit here because it doesn't allow to create a background task without saving a task token somewhere.

src/web/ng/Server.cpp
```diff
@@ -256,8 +268,9 @@ Server::run()
 }
 
 void
-Server::stop()
+Server::stop(boost::asio::yield_context yield)
 {
+    connectionHandler_.stop(yield);
```
kuznetsss, line 273: This should be covered by test.

src/app/Stopper.hpp
```diff
@@ -0,0 +1,118 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
```
cindyyan317, line 4: 2025

cindyyan317 (commented):
Do we need to consider that shutdown happens during cache loading?

src/app/Stopper.hpp
kuznetsss: I created it in 2024 😄

cindyyan317 (approved):
LGTM

PR #1807, opened 2025-01-02
fix: Incorrect log values

PeterChen13579:
Fixes the incorrect value from [here](https://github.com/XRPLF/clio/pull/1627/files#diff-c32eb3bef779b5a92f8b23a305167e04a4f84d240cf6aa392959e070203ca0bbL132)


godexsoft (changes requested):
Shall we separate the automated changes into a separate PR?

src/util/BytesConverter.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace util {
+
+/**
+ * @brief Convert megabytes to bytes
+ * @param mb Number of megabytes to convert
+ * @return The equivalent number of bytes
+ */
+constexpr std::uint64_t
+mbToBytes(int mb)
```
godexsoft: I'm not sure if we want to take 'int' as input though. Why not use the same type we get from the config values we use it with?

src/util/BytesConverter.hpp
```diff
@@ -0,0 +1,37 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
godexsoft, line 19: .hpp file missing a `#pragma once`

tests/unit/util/BytesConverterTests.cpp
```diff
@@ -0,0 +1,40 @@
...
+using namespace util;
+
+TEST(MBToBytesTest, SimpleValues)
+{
+    EXPECT_EQ(mbToBytes(0), 0);
+    EXPECT_EQ(mbToBytes(1), 1024 * 1024);
+    EXPECT_EQ(mbToBytes(2), 2 * 1024 * 1024);
+}
+
+TEST(MBToBytesTest, limitValues)
```
godexsoft: Let's try to keep the naming of the tests consistent (at least in the same file)

godexsoft (changes requested):

godexsoft (approved):

PR #1814, opened 2025-01-08
feat: add config verify flag

### LINKED ISSUE #1806: Config Validation command
Check whether config file that is provided is valid. Something like `./clio-server PathToConfig --verify`

PeterChen13579:
fixes #1806

### DIFF
--- a/src/app/CliArgs.cpp
+++ b/src/app/CliArgs.cpp
@@ -47,6 +47,7 @@ CliArgs::parse(int argc, char const* argv[])
         ("conf,c", po::value<std::string>()->default_value(kDEFAULT_CONFIG_PATH), "configuration file")
         ("ng-web-server,w", "Use ng-web-server")
         ("migrate", po::value<std::string>(), "start migration helper")
+        ("verify", "Checks the validity of config values")
     ;
     // clang-format on
     po::positional_options_description positional;
@@ -74,6 +75,9 @@ CliArgs::parse(int argc, char const* argv[])
             return Action{Action::Migrate{.configPath = std::move(configPath), .subCmd = MigrateSubCmd::status()}};
         return Action{Action::Migrate{.configPath = std::move(configPath), .subCmd = MigrateSubCmd::migration(opt)}};
     }
+
+    if (parsed.count("verify") != 0u)
+        return Action{Action::VerifyConfig{.configPath = std::move(configPath)}};
 
     return Action{Action::Run{.configPath = std::move(configPath), .useNgWebServer = parsed.count("ng-web-server") != 0}
     };
--- a/src/app/VerifyConfig.hpp
+++ b/src/app/VerifyConfig.hpp
@@ -1,0 +1,55 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2025, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include "util/newconfig/ConfigDefinition.hpp"
+#include "util/newconfig/ConfigFileJson.hpp"
+
+#include <cstdlib>
+#include <iostream>
+#include <string_view>
+
+namespace app {
+
+/**
+ * @brief Verifies user's config values are correct
+ *
+ * @param configPath The path to config
+ * @return true if config values are all correct, false otherwise
+ */
+inline bool
+verifyConfig(std::string_view configPath)
+{
+    using namespace util::config;
+
+    auto const json = ConfigFileJson::makeConfigFileJson(configPath);
+    if (!json.has_value()) {
+        std::cerr << json.error().error << std::endl;
+        return false;
+    }
+    auto const errors = gClioConfig.parse(json.value());
+    if (errors.has_value()) {
+        for (auto const& err : errors.value())
+            std::cerr << err.error << std::endl;
+        return false;
+    }
+    return true;
+}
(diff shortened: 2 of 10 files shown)


src/app/CliArgs.cpp
```diff
@@ -47,6 +47,7 @@ CliArgs::parse(int argc, char const* argv[])
         ("conf,c", po::value<std::string>()->default_value(kDEFAULT_CONFIG_PATH), "configuration file")
         ("ng-web-server,w", "Use ng-web-server")
         ("migrate", po::value<std::string>(), "start migration helper")
+        ("verify", "Checks the validity of config values")
```
kuznetsss, line 50: Maybe this flag should take path to config as well?

src/main/Main.cpp
```diff
@@ -29,9 +29,28 @@
 #include <cstdlib>
 #include <exception>
 #include <iostream>
+#include <string_view>
 
 using namespace util::config;
 
+static int
+handleVerifyConfig(std::string_view configPath)
```
kuznetsss: Better to put it into a separate file and then you will be able to test it. Also I would call it `verifyConfig`. And better for it to return bool instead of exit code.

src/util/newconfig/ConfigDefinition.hpp
```diff
@@ -350,8 +350,9 @@ static ClioConfigDefinition gClioConfig = ClioConfigDefinition{
      {"server.admin_password", ConfigValue{ConfigType::String}.optional()},
      {"server.processing_policy",
       ConfigValue{ConfigType::String}.defaultValue("parallel").withConstraint(gValidateProcessingPolicy)},
-     {"server.parallel_requests_limit", ConfigValue{ConfigType::Integer}.optional()},
-     {"server.ws_max_sending_queue_size", ConfigValue{ConfigType::Integer}.defaultValue(1500)},
+     {"server.parallel_requests_limit", ConfigValue{ConfigType::Integer}.optional().withConstraint(gValidateUint16)},
+     {"server.ws_max_sending_queue_size",
+      ConfigValue{ConfigType::Integer}.defaultValue(1500).withConstraint(gValidateUint32)},
```
kuznetsss: Maybe these changes should be submitted as a separate PR?

src/main/Main.cpp
```diff
@@ -29,9 +29,28 @@
...
+        std::cerr << json.error().error << std::endl;
+        return EXIT_FAILURE;
+    }
+    auto const errors = gClioConfig.parse(json.value());
+    if (errors.has_value()) {
+        for (auto const& err : errors.value())
+            std::cerr << err.error << std::endl;
+        return EXIT_FAILURE;
+    }
+    std::cout << "Config is correct" << std::endl;
```
cindyyan317: When I just run clio server ,will this line be printed out too ?

PeterChen13579: Yes, it'll print "Config is correct" at the very beginning. Do u think this is okay or prefer not to print out this line?

src/app/CliArgs.cpp
PeterChen13579: I think it already does here: https://github.com/XRPLF/clio/pull/1814/files#diff-bad42112163fd471148e3dbe18fbb299652a1d323e8d0a0c1a526cba10928a2dR80

tests/unit/app/VerifyConfigTests.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "app/VerifyConfig.hpp"
+#include "util/TmpFile.hpp"
+#include "util/newconfig/FakeConfigData.hpp"
+
+#include <gtest/gtest.h>
+
+using namespace app;
+using namespace util::config;
+
+TEST(VerifyConfigTest, InvalidConfig)
```
cindyyan317, line 29: Shall we add 2 more testcases:
1 config file not exist
2 config file is not a valid json file

tests/unit/app/VerifyConfigTests.cpp
```diff
@@ -0,0 +1,63 @@
...
+    // invalid json because extra "," after 51233
+    static constexpr auto kINVALID_JSON = R"({
+                                             "server": {
+                                                "ip": "0.0.0.0",
+                                                "port": 51233, 
+                                            }
+                                        })";
+    auto const tmpConfigFile = TmpFile(kINVALID_JSON);
+
+    // current example config should always be compatible with configDefinition
```
cindyyan317: Not sure I understand the comment.

cindyyan317 (dismissed):
Thanks for adding the unittests. LGTM

tests/unit/app/VerifyConfigTests.cpp
PeterChen13579: This comment should be removed. I wanted to test `example config` in our repo, but our CI doesn't support that 😄

cindyyan317 (approved):

src/app/CliArgs.cpp
kuznetsss: I meant to verify config currently user should run command `clio_server -c ./config.json --verify`, right?
I was thinking that maybe `clio_server --verify ./config.json` may be better, but maybe it doesn't matter.
@godexsoft, WDYT?

src/main/Main.cpp
```diff
@@ -40,34 +40,27 @@ try {
...
     return action.apply(
         [](app::CliArgs::Action::Exit const& exit) { return exit.exitCode; },
-        [](app::CliArgs::Action::Run const& run) {
-            auto const json = ConfigFileJson::makeConfigFileJson(run.configPath);
-            if (!json.has_value()) {
-                std::cerr << json.error().error << std::endl;
-                return EXIT_FAILURE;
+        [](app::CliArgs::Action::VerifyConfig const& verify) {
+            if (app::verifyConfig(verify.configPath)) {
+                std::cout << "Config is correct" << "\n";
```
kuznetsss, line 45: Probably better to print config path here, something like
```
Config './config.json` is correct
```

src/app/VerifyConfig.hpp
```diff
@@ -0,0 +1,55 @@
...
+ * @return true if config values are all correct, false otherwise
+ */
+inline bool
+verifyConfig(std::string_view configPath)
+{
+    using namespace util::config;
+
+    auto const json = ConfigFileJson::makeConfigFileJson(configPath);
+    if (!json.has_value()) {
+        std::cerr << json.error().error << std::endl;
```
kuznetsss, line 44: Maybe a bit more verbose:
```cpp
std::cerr << "Error parsing json from config: " << configPath << "\n" << json.error().error << std::endl;
```

src/app/VerifyConfig.hpp
```diff
@@ -0,0 +1,55 @@
...
+        return false;
+    }
+    auto const errors = gClioConfig.parse(json.value());
+    if (errors.has_value()) {
+        for (auto const& err : errors.value())
```
kuznetsss, line 49: And the same here:
```cpp
std::cerr << "Issues found in provided config '" << configPath << "':\n" ;
```

tests/common/util/newconfig/FakeConfigData.hpp
```diff
@@ -208,3 +208,11 @@ static constexpr auto kINVALID_JSON_DATA = R"JSON({
...
     }
 })JSON";
+
+// used to Verify Config test
+static constexpr auto kVALID_JSON_DATA = R"JSON({
+     "server": {
+        "ip": "0.0.0.0",
+        "port": 51233
+    }
+})JSON";
```
kuznetsss, line 218: This should be in the file where it is used.

PeterChen13579:
@kuznetsss The verify flag can work both with and without the `-c`. ie `clio_server -c ./config.json --verify` and `clio_server --verify ./config.json` or `clio_server  ./config.json --verify` all works.

PR #1815, opened 2025-01-09
fix: AsyncFramework RAII

### LINKED ISSUE #1812: AsyncFramework operation needs proper RAII
## Issue Description
At the moment, if any of the operations that the user is holding onto is deleted, the underlying async operation is not automatically cancelled/aborted. This does not feel right.

## Expected Result
Any operation in async framework should automatically abort itself if the user no longer holds onto its handle.

godexsoft:
Fixes #1812 

AsyncFramework operations were not handling their destruction properly. With this fix operations will automatically cancel/abort if they go out of scope (i.e. the user is no longer holding onto them).

### DIFF
--- a/src/util/WasMoved.hpp
+++ b/src/util/WasMoved.hpp
@@ -32,8 +32,8 @@ protected:
      * @brief The function to be used by clients in order to check whether the instance was moved from
      * @return true if moved from; false otherwise
      */
-    bool
-    wasMoved() const
+    [[nodiscard]] bool
+    wasMoved() const noexcept
     {
         return wasMoved_;
     }
--- a/src/util/WasMoved.hpp
+++ b/src/util/WasMoved.hpp
@@ -19,6 +19,8 @@
 
 #pragma once
 
+#include <utility>
+
 namespace util {
 
 /**
@@ -38,11 +40,10 @@ protected:
         return wasMoved_;
     }
 
-protected:
     WasMoved() = default;  // should only be used via inheritance
 
 public:
-    ~WasMoved() = default;
+    virtual ~WasMoved() = default;
 
     /**
      * @brief Move constructor sets the internal flag to true on the moved-from object
@@ -50,7 +51,7 @@ public:
      */
     WasMoved(WasMoved&& other)
     {
-        other.wasMoved_ = true;
+        *this = std::move(other);
     }
 
     /**
@@ -62,6 +63,7 @@ public:
     operator=(WasMoved&& other)
     {
         other.wasMoved_ = true;
+        wasMoved_ = false;
         return *this;
     }
 
--- a/src/util/WasMoved.hpp
+++ b/src/util/WasMoved.hpp
@@ -52,6 +52,7 @@ public:
     {
         other.wasMoved_ = true;
     }
+
     /**
      * @brief Move operator sets the internal flag to true on the moved-from object
      * @param other the moved-from object
--- a/src/util/WasMoved.hpp
+++ b/src/util/WasMoved.hpp
@@ -1,0 +1,72 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2025, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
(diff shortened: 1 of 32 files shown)


src/util/WasMoved.hpp
```diff
@@ -0,0 +1,73 @@
...
+     * @brief The function to be used by clients in order to check whether the instance was moved from
+     * @return true if moved from; false otherwise
+     */
+    [[nodiscard]] bool
+    wasMoved() const noexcept
+    {
+        return wasMoved_;
+    }
+
+protected:
```
kuznetsss:
```suggestion

```
`protected` 2 times.

src/util/WasMoved.hpp
```diff
@@ -0,0 +1,73 @@
...
+    WasMoved() = default;  // should only be used via inheritance
+
+public:
+    ~WasMoved() = default;
```
kuznetsss:
```suggestion
    virtual ~WasMoved() = default;
```

src/util/WasMoved.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @brief Move constructor sets the internal flag to true on the moved-from object
+     * @param other the moved-from object
+     */
+    WasMoved(WasMoved&& other)
+    {
+        other.wasMoved_ = true;
```
kuznetsss: Hm, shouldn't we set here `wasMoved_ = false;`?
To fix such case:
```cpp
WasMoved a;
WasMoved b = std::move(a);
a = std::move(b);
EXPECT_FALSE(a.wasMoved());
```

kuznetsss: Also probably better to reuse code:
```cpp
*this = std::move(other);
```

src/util/async/Operation.hpp
```diff
@@ -149,7 +161,8 @@ struct BasicScheduledOperation {
  * @tparam StopSourceType The type of the stop source
  */
 template <typename RetType, typename StopSourceType>
-class StoppableOperation : public impl::BasicOperation<StoppableOutcome<RetType, StopSourceType>> {
+class StoppableOperation : public impl::BasicOperation<StoppableOutcome<RetType, StopSourceType>>,
+                           public util::WasMoved {
```
kuznetsss: Shouldn't the `BasicOperation` class inherit `WasMoved` instead?

src/util/WasMoved.hpp
godexsoft: This one was on purpose actually. It's a common pattern, usually used to separate data from logic but in this case perhaps it's not really helping much so i'll apply your suggestion.

src/util/WasMoved.hpp
godexsoft: Oh yeah, i actually remembered about this in bed last night 😃

src/util/async/Operation.hpp
godexsoft: I deliberately inherited on the same level where we actually need the `wasMoved()` call.

src/util/WasMoved.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @brief Move operator sets the internal flag to true on the moved-from object
+     * @param other the moved-from object
+     * @return reference to self
+     */
+    WasMoved&
+    operator=(WasMoved&& other)
+    {
+        other.wasMoved_ = true;
```
cindyyan317: We may need to check if other is the current object to handle self assignment .
```
WasMoved a; 
a = std::move(a); 
EXPECT_FALSE(a.wasMoved());
```

src/util/WasMoved.hpp
godexsoft: Good suggestions 👍

tests/unit/util/MoveTrackerTests.cpp
```diff
@@ -38,3 +38,13 @@ TEST(WasMovedTests, SimpleChecks)
...
+
+TEST(MoveTrackerTests, SupportReuse)
+{
+    auto original = MoveMe();
+    auto other = std::move(original);
+
+    original = std::move(other);
+    EXPECT_FALSE(original.wasMoved());
+    EXPECT_TRUE(other.wasMoved());
+}
```
kuznetsss, line 50: Lets also add a test about moving into itself: `moveMe = std::move(moveMe);`

src/util/WasMoved.hpp
godexsoft: That's caught by the compiler already.. explicit self-move is not allowed. 
We can however trick the compiler to do it anyway and the current behaviour ends up with `wasMoved()` equal to `false` which feels correct because you don't actually move anything (if all client move code is doing the correct things). We could also ASSERT instead in such case. WDYT?

godexsoft: Ok I think we figured out the correct logic. Take a look at the current tests 👍

kuznetsss (approved):
👍

cindyyan317 (approved):
👍

PR #1820, opened 2025-01-10
feat: ETLng scheduling

godexsoft:
For #1596 

This implements forward and backfill schedulers and some tests for them. These will be used by the extractors in order to obtain the next piece of work to extract and consequently load.

### DIFF
--- a/src/etlng/impl/Scheduling.hpp
+++ b/src/etlng/impl/Scheduling.hpp
@@ -1,0 +1,150 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2025, the clio developers.
+
+    Permission to use, copy, modify, and distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#pragma once
+
+#include "etl/NetworkValidatedLedgersInterface.hpp"
+#include "etlng/Models.hpp"
+#include "etlng/SchedulerInterface.hpp"
+
+#include <sys/types.h>
+
+#include <atomic>
+#include <cstdint>
+#include <limits>
+#include <memory>
+#include <optional>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+
+namespace etlng::impl {
+
+template <typename T>
+concept SomeScheduler = std::is_base_of_v<SchedulerInterface, std::decay_t<T>>;
+
+class ForwardScheduler : public SchedulerInterface {
+    std::shared_ptr<etl::NetworkValidatedLedgersInterface> ledgers_;
+
+    uint32_t startSeq_;
+    std::optional<uint32_t> maxSeq_;
+    std::atomic_uint32_t seq_;
+
+public:
+    ForwardScheduler(ForwardScheduler const& other)
+        : ledgers_(other.ledgers_), startSeq_(other.startSeq_), maxSeq_(other.maxSeq_), seq_(other.seq_.load())
+    {
+    }
+
+    ForwardScheduler(
+        std::shared_ptr<etl::NetworkValidatedLedgersInterface> ledgers,
+        uint32_t startSeq,
+        std::optional<uint32_t> maxSeq = std::nullopt
+    )
+        : ledgers_(std::move(ledgers)), startSeq_(startSeq), maxSeq_(maxSeq), seq_(startSeq)
+    {
+    }
+
+    [[nodiscard]] std::optional<model::Task>
+    next() override
+    {
+        static constexpr auto kMAX = std::numeric_limits<uint32_t>::max();
+        uint32_t seq = seq_.load(std::memory_order_relaxed);
+
+        if (ledgers_->getMostRecent() >= seq) {
+            while (seq < maxSeq_.value_or(kMAX)) {
+                if (seq_.compare_exchange_weak(seq, seq + 1, std::memory_order_acq_rel)) {
+                    return {{.priority = 1u, .seq = seq}};
+                }
+            }
+        }
+
+        return std::nullopt;
+    }
+};
+
+class BackfillScheduler : public SchedulerInterface {
+    uint32_t startSeq_;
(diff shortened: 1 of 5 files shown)


src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+    [[nodiscard]] std::optional<model::Task>
+    next() override
+    {
+        static constexpr auto kMAX = std::numeric_limits<uint32_t>::max();
+        uint32_t seq = seq_.load(std::memory_order_relaxed);
+
+        if (ledgers_->getMostRecent() >= seq) {
+            while (seq < maxSeq_.value_or(kMAX)) {
+                if (seq_.compare_exchange_weak(seq, seq + 1, std::memory_order_acq_rel)) {
```
cindyyan317: Is `seq+1` will larger than the available ledger sequence if `seq == ledgers_->getMostRecent()`?

cindyyan317: I see. The when next next() called, it will compare it with current getMostRecent. ✅

src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+        std::optional<model::Task> task;
+        auto const expand = [&](auto& s) {
+            if (task.has_value())
+                return false;
+
+            task = s.next();
+            return task.has_value();
+        };
+
+        std::apply([&expand](auto&&... xs) { (... || expand(xs)); }, schedulers_);
```
cindyyan317, line 137: It seems like the priority is decided by the order of the template parameter? I wonder what is the `priority` in `Task` used for then?

tests/unit/etlng/SchedulingTests.cpp
```diff
@@ -0,0 +1,187 @@
...
+    auto scheduler = impl::BackfillScheduler(10u, 5u);
+
+    for (auto i = 10u; i > 5u; --i) {
+        auto maybeTask = scheduler.next();
+
+        EXPECT_TRUE(maybeTask.has_value());
+        EXPECT_EQ(maybeTask->seq, i);
+    }
+
+    auto empty = scheduler.next();
```
cindyyan317: small thing: const empty

tests/unit/etlng/SchedulingTests.cpp
```diff
@@ -0,0 +1,187 @@
...
+TEST_F(ForwardSchedulerTests, ExhaustsSchedulerIfMostRecentLedgerIsNewerThanRequestedSequence)
+{
+    auto scheduler = impl::ForwardScheduler(networkValidatedLedgers_, 0u, 10u);
+    EXPECT_CALL(*networkValidatedLedgers_, getMostRecent()).Times(11).WillRepeatedly(testing::Return(11u));
+
+    for (auto i = 0u; i < 10u; ++i) {
+        auto maybeTask = scheduler.next();
+
+        EXPECT_TRUE(maybeTask.has_value());
+        EXPECT_EQ(maybeTask->seq, i);
```
cindyyan317, line 69: There is no place to check the other parameter of `Task` - `priority`.

src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+        : ledgers_(other.ledgers_), startSeq_(other.startSeq_), maxSeq_(other.maxSeq_), seq_(other.seq_.load())
+    {
+    }
+
+    ForwardScheduler(
+        std::shared_ptr<etl::NetworkValidatedLedgersInterface> ledgers,
+        uint32_t startSeq,
+        std::optional<uint32_t> maxSeq = std::nullopt
+    )
+        : ledgers_(std::move(ledgers)), startSeq_(startSeq), maxSeq_(maxSeq), seq_(startSeq)
```
cindyyan317: Can `maxSeq_` just be an uint32_t ? 🤔  like `minSeq_` in `BackfillScheduler`

src/etlng/impl/Scheduling.hpp
godexsoft: The order of template parameters only dictates which schedulers in the chain will be queried first. So in a way this is the first layer of prioritization. 
All tasks acquired by `next()` end up in a `priority_queue` which will prioritize according to the `priority` field. That part will come in another PR 🔢

tests/unit/etlng/SchedulingTests.cpp
godexsoft: Correct. That's for another PR - these smaller PRs are extracted out of a larger, more complete branch. Just so you don't have to review 20k lines at once 😃

src/etlng/impl/Scheduling.hpp
godexsoft: In forward fill we may not want to ever stop, so it's optional. In backward we always know to stop either on the value specified or on seq `0` (in practice it is not even really 0 but whatever the first ledger is).

src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+#include <type_traits>
+#include <utility>
+
+namespace etlng::impl {
+
+template <typename T>
+concept SomeScheduler = std::is_base_of_v<SchedulerInterface, std::decay_t<T>>;
+
+class ForwardScheduler : public SchedulerInterface {
+    std::shared_ptr<etl::NetworkValidatedLedgersInterface> ledgers_;
```
kuznetsss: Shouldn't this be `std::ref<>` instead?

src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+    )
+        : ledgers_(std::move(ledgers)), startSeq_(startSeq), maxSeq_(maxSeq), seq_(startSeq)
+    {
+    }
+
+    [[nodiscard]] std::optional<model::Task>
+    next() override
+    {
+        static constexpr auto kMAX = std::numeric_limits<uint32_t>::max();
+        uint32_t seq = seq_.load(std::memory_order_relaxed);
```
kuznetsss: Why do you need to be so specific with atomic? Why not simply
```cpp
uint32_t seq = seq_;
```

Also `seq` and `seq_` are very similar names, maybe `currentSeq`?

src/etlng/impl/Scheduling.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+        if (ledgers_->getMostRecent() >= seq) {
+            while (seq < maxSeq_.value_or(kMAX)) {
+                if (seq_.compare_exchange_weak(seq, seq + 1, std::memory_order_acq_rel)) {
+                    return {{.priority = 1u, .seq = seq}};
```
kuznetsss: Maybe make `priority` an enum to not use magic numbers when creating a task?

godexsoft: Good catch. I should've done that from the start instead of leaving it for when i split main feature branch into PRs.

src/etlng/impl/Scheduling.hpp
godexsoft: For the atomic question: it's potentially a free optimization which does not change the correctness of this code. By default it would use `memory_order_seq_cst` which seems stricter than needed in this case. I'm no expert on atomics but my understanding is that both ways are ok, this one is just more explicit about the intention and leaves optimization potential (allow compiler to rearrange operations if needed). If you have a strong opinion on this, we can switch to the default. It's not like this code is in a very tight loop so it should not impact performance much/at all.

src/etlng/impl/Scheduling.hpp
godexsoft: In best spirits of not storing `shared_ptr`s that we don't "own"? We have so much of this already across the codebase that it may be worth creating an issue to fix it everywhere in one go, wdyt? I can change this one in this PR.

kuznetsss: Yeah, we could create an issue instead.

godexsoft: https://github.com/XRPLF/clio/issues/1824

cindyyan317 (approved):

kuznetsss (approved):

PR #1828, opened 2025-01-14
fix: Check config is valid

kuznetsss:


kuznetsss:
Let me make a better fix.

PR #1831, opened 2025-01-14
fix: Add more constraints to config

PeterChen13579:
Log file size and rotation should also not allowed to be 0.

### DIFF
--- a/src/util/newconfig/ConfigConstraints.hpp
+++ b/src/util/newconfig/ConfigConstraints.hpp
@@ -363,6 +363,7 @@ static constinit NumberValueConstraint<uint16_t> gValidateUint16{
     std::numeric_limits<uint16_t>::max()
 };
 
+// log file size minimum is 1mb, log rotation time minimum is 1hr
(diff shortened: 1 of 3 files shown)


kuznetsss (dismissed):
👍

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -362,6 +362,9 @@ static constinit NumberValueConstraint<uint16_t> gValidateUint16{
     std::numeric_limits<uint16_t>::min(),
     std::numeric_limits<uint16_t>::max()
 };
+
+// log file size minimum is 1mb, log rotation time minimum is 1hr
+static constinit NumberValueConstraint<uint32_t> gValidateLogs{1, std::numeric_limits<uint32_t>::max()};
```
cindyyan317: I think the limit for time and size happen to be the same. But I will use 2 constraints.

cindyyan317 (approved):

PR #1834, opened 2025-01-15
feat: Block clio if migration is blocking

cindyyan317:
Add:
- Block server if migration is blocking
- Initialise the migration related table when server starts against empty DB

Add MigrationInspectorInterface. server uses inspector to check the migrators status.

### DIFF
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -106,10 +106,10 @@ ClioApplication::run(bool const useNgWebServer)
 
     {
         auto const migrationInspector = migration::makeMigrationInspector(config_, backend);
-        // Check if any migration is blocking server starting.
-        if (migrationInspector->isBlockingServer() and backend->hardFetchLedgerRangeNoThrow()) {
+        // Check if any migration is blocking Clio server starting.
+        if (migrationInspector->isBlockingClio() and backend->hardFetchLedgerRangeNoThrow()) {
             LOG(util::LogService::error())
-                << "Existing Migration is blocking server, Please complete the database migration first.";
+                << "Existing Migration is blocking Clio, Please complete the database migration first.";
             return EXIT_FAILURE;
         }
     }
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -26,6 +26,7 @@
 #include "etl/LoadBalancer.hpp"
 #include "etl/NetworkValidatedLedgers.hpp"
 #include "feed/SubscriptionManager.hpp"
+#include "migration/MigrationInspectorFactory.hpp"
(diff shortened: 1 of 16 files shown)


tests/unit/migration/MigrationInspectorBaseTests.cpp
```diff
@@ -0,0 +1,130 @@
...
+TEST_F(MigrationInspectorBaseTest, getMigratorStatusByName)
+{
+    EXPECT_CALL(*backend_, fetchMigratorStatus("SimpleTestMigrator", testing::_)).WillOnce(testing::Return("Migrated"));
+    EXPECT_CALL(*backend_, fetchMigratorStatus("SimpleTestMigrator2", testing::_))
+        .WillOnce(testing::Return("NotMigrated"));
+
+    EXPECT_EQ(migrationInspector_->getMigratorStatusByName("SimpleTestMigrator"), migration::MigratorStatus::Migrated);
+    EXPECT_EQ(
+        migrationInspector_->getMigratorStatusByName("SimpleTestMigrator2"), migration::MigratorStatus::NotMigrated
+    );
```
kuznetsss: Nit: probably better to change order here a bit to make it: expectation then action.

tests/unit/migration/MigrationInspectorFactoryTests.cpp
```diff
@@ -0,0 +1,73 @@
...
+struct MigrationInspectorFactoryTests : public util::prometheus::WithPrometheus, public MockBackendTest {
+protected:
+    util::config::ClioConfigDefinition const readerConfig_ = util::config::ClioConfigDefinition{
+        {"read_only", util::config::ConfigValue{util::config::ConfigType::Boolean}.defaultValue(true)}
+    };
+};
+
+TEST_F(MigrationInspectorFactoryTests, NullBackend)
+{
+    EXPECT_DEATH(migration::makeMigrationInspector(readerConfig_, nullptr), ".*");
```
kuznetsss, line 46: Tests with `EXPECT_DEATH` should have `DeathTest` suffix in fixture name.

src/app/ClioApplication.cpp
```diff
@@ -103,6 +104,16 @@ ClioApplication::run(bool const useNgWebServer)
     // Interface to the database
     auto backend = data::makeBackend(config_);
 
+    {
+        auto const migrationInspector = migration::makeMigrationInspector(config_, backend);
+        // Check if any migration is blocking server starting.
+        if (migrationInspector->isBlockingServer() and backend->hardFetchLedgerRangeNoThrow()) {
```
kuznetsss: I think the word `server` makes a reader think about web server. I would change it to `Clio`.

kuznetsss (approved):
👍

PR #1835, opened 2025-01-15
fix: Copyright format

cindyyan317:
Aligned the copyright notice to match the format used in other files. Updated Copyright (c) 2022-2024 to Copyright (c) 2024 to ensure consistency.

### DIFF
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2022, the clio developers.
+    Copyright (c) 2024, the clio developers.
 
     Permission to use, copy, modify, and distribute this software for any
     purpose with or without fee is hereby granted, provided that the above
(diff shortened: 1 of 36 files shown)


src/main/Main.cpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2022-2024, the clio developers.
+    Copyright (c) 2024, the clio developers.
```
kuznetsss, line 4: This should be 2022. As I understand year here is the year copyright started to apply.

src/app/ClioApplication.cpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2024, the clio developers.
+    Copyright (c) 2022, the clio developers.
```
godexsoft: Not sure i agree with this change

godexsoft (changes requested):
I think the application file was created in 2024. The rest is fine

godexsoft (approved):

PR #1840, opened 2025-01-20
feat: LPT freeze

shawnxie999:
https://github.com/XRPLF/rippled/pull/5227

Introduces amendment gating in `accountHolds` that so that the assets in the AMM are not frozen


shawnxie999:
@godexsoft @cindyyan317 This PR is ready for review. There is still a test CI issue around `SubscribeManager` that needs to be resolved, but I don’t believe it should block the review process. However, I will have limited capacity to investigate this issue further over the next two weeks.

src/feed/SubscriptionManager.hpp
shawnxie999, line 1: @cindyyan317 Would you mind giving this (and SubscriptionManagerTests.cpp) a quick review when you get a chance? If you spot anything obvious that could be causing the segfault in SubscriptionManagerAsyncTest, that’d be super helpful! I thought my changes were pretty minimal but it may have broke it somewhere that I could not reproduce on my mac

cindyyan317: Will have a look today

cindyyan317: I would suggest you double check the declare order of SubscriptionManagerBaseTest. You use mockAmendmentCenterPtr_ before u declare it.

src/rpc/RPCHelpers.cpp
```diff
@@ -1036,6 +1053,93 @@ accountHolds(
...
+            if (issuerSle.isFieldPresent(ripple::sfAMMID)) {
+                auto const ammKeylet = ripple::keylet::amm(issuerSle[ripple::sfAMMID]);
+                auto const ammBlob = backend.fetchLedgerObject(ammKeylet.key, sequence, yield);
+
+                if (!ammBlob)
+                    return false;
+
+                ripple::SLE const ammSle{ripple::SerialIter{ammBlob->data(), ammBlob->size()}, ammKeylet.key};
+
+                if (isLPTokenFrozen(
```
cindyyan317: Maybe `return !isLPTokenFrozen(...)` is easier

src/rpc/RPCHelpers.cpp
```diff
@@ -19,6 +19,8 @@
 
 #include "rpc/RPCHelpers.hpp"
 
+#include "data/AmendmentCenter.hpp"
```
cindyyan317, line 22: No need this header?

src/rpc/RPCHelpers.cpp
```diff
@@ -991,11 +1008,11 @@ accountFunds(
         return amount;
     }
 
-    return accountHolds(backend, sequence, id, amount.getCurrency(), amount.getIssuer(), true, yield);
+    return accountHolds(backend, amendmentCenter, sequence, id, amount.getCurrency(), amount.getIssuer(), true, yield);
 }
 
 ripple::STAmount
-accountHolds(
+ammAccountHolds(
```
cindyyan317, line 1016: ammAccountHolds and accountHolds seem share some common code , can we refactor them a bit?

src/rpc/AMMHelpers.cpp
```diff
@@ -65,7 +68,9 @@ getAmmLpHolds(
 )
 {
     auto const lptCurrency = ammLPTCurrency(cur1, cur2);
-    return accountHolds(backend, sequence, lpAccount, lptCurrency, ammAccount, true, yield);
+
```
cindyyan317, line 71: This number will be shown in amm_info response , right? If the accout's lptoken is considered "frozen" to it, should we still show the number?

shawnxie999: AFAIK, `getAmmLpHolds` is only used to find the LPToken balance between the AMM Account and an LPToken holder, and it never considers the freeze status of the trustline (since AMM account is a psuedo account that can't freeze anyone anyways).

This is why i kept the existing behavior -we don't need to check if the LPToken holder is frozen for the underlying assets in the pool.

Eg. 
1. Alice deposited XRP and USD into AMM pool, receiving `100 LPToken`
2. Issuer freezes Alice's USD
3. When `amm_info` is called with Alice, we still want return Alice's true balance of `100 LPToken` using the `ammAccountHolds `. Because otherwise, the new `accountHolds` fetches Alice's USD trustline and return `0 LPToken` for her LPT balance (which defeats the purpose because we'd like to know Alice's LPT balance regardless her USD is frozen or not)

src/feed/SubscriptionManager.hpp
shawnxie999: thanks that fixed it!

src/rpc/AMMHelpers.cpp
cindyyan317: Do we have any place we can check if the holder's LPT is frozen or not?

shawnxie999: i think `book_offers` is the main place in clio that checks if LPT is frozen to decide if the offer is funded or not. But the change should pretty much apply to everywhere that uses `accountHolds`. `amm_info` should not have used `accountHolds` in the first place because it is a more general purpose function.

src/rpc/RPCHelpers.cpp
shawnxie999: i don't think we benefit much from refactoring to share common code. these two functions should be independent from each other anyways

cindyyan317: My main thought was that pulling out the common code could make things easier to maintain and avoid duplication. If we ever need to update that logic, we'd only have to do it in one place.
Screenshot 2025-01-31 at 10 37 36
For example, the green part is the shared code, if there is bug in it, we need to change two places.
Are these two functions really independent? I saw `accountHolds` just does one more check comparing with `ammAcountHolds`. So the former can call latter like this:
```
ripple::STAmount
accountHolds(
    BackendInterface const& backend,
    data::AmendmentCenterInterface const& amendmentCenter,
    std::uint32_t sequence,
    ripple::AccountID const& account,
    ripple::Currency const& currency,
    ripple::AccountID const& issuer,
    bool const zeroIfFrozen,
)
{
    auto amountBeforeCheckingLP = ammAccountHolds(backend, sequence, account, currency, issuer, zeroIfFrozen, yield);

    if (!zeroIfFrozen || amountBeforeCheckingLP == amountBeforeCheckingLP.zeroed())
        return amountBeforeCheckingLP;

    auto const isLptFrozen = [&]() {
        if (amendmentCenter.isEnabled(yield, data::Amendments::fixFrozenLPTokenTransfer, sequence)) {
            auto const issuerBlob = backend.fetchLedgerObject(ripple::keylet::account(issuer).key, sequence, yield);

            if (!issuerBlob)
                return true;

            ripple::SLE const issuerSle{
                ripple::SerialIter{issuerBlob->data(), issuerBlob->size()}, ripple::keylet::account(issuer).key
            };

            // if the issuer is an amm account, then currency is lptoken, so we will need to check if the
            // assets in the pool are frozen as well
            if (issuerSle.isFieldPresent(ripple::sfAMMID)) {
                auto const ammKeylet = ripple::keylet::amm(issuerSle[ripple::sfAMMID]);
                auto const ammBlob = backend.fetchLedgerObject(ammKeylet.key, sequence, yield);

                if (!ammBlob)
                    return true;

                ripple::SLE const ammSle{ripple::SerialIter{ammBlob->data(), ammBlob->size()}, ammKeylet.key};

                return isLPTokenFrozen(
                    backend,
                    sequence,
                    account,
                    ammSle[ripple::sfAsset].get<ripple::Issue>(),
                    ammSle[ripple::sfAsset2].get<ripple::Issue>(),
                    yield
                );
            }
        }
        return false;
    };

    return isLptFrozen() ? amountBeforeCheckingLP.zeroed() : amountBeforeCheckingLP;
}
```
`

cindyyan317 (changes requested):
Please add unittests for RPCHelper in `RPCHelperTests.cpp`. 
For example:
when `fixFrozenLPTokenTransfer` enable/disable
   when `asset` is frozen and `asset2` is not
   when `asset` is not and `asset2` is frozen
   when `sfAMMID` present/not present
...

src/rpc/RPCHelpers.cpp
shawnxie999: truthfully, the code in ammAccountHolds:

```
    if (ripple::isXRP(currency))
        return {xrpLiquid(backend, sequence, account, yield)};
```

should be replaced with an assertion `ASSERT(!ripple::isXRP(currency))`, which is why I think these two functions are independent

cindyyan317: If you genuinely feel certain modifications are necessary, please go ahead and adjust the code accordingly. IMO, refactoring common logic doesn’t mean you can’t make case-specific tweaks; it just helps keep the codebase cleaner and easier to maintain.

shawnxie999: they may share some common piece of code now, but that  may not be the case soon. In fact i believe the implementation `ammAccountHolds` will soon diverge from `accountHolds` due to the DeepFreeze feature.  For the incoming deepfreeze feature,`accountHolds` will need to add `isDeepFreeze` check (that's not gated by an amendment).

cindyyan317: >accountHolds will need to add isDeepFreeze check 

Would it make it impossible for `accountHolds` calls `ammAccountHolds`? 

Since you prefer to keep it as is, I'll respect your decision.

src/rpc/RPCHelpers.cpp
```diff
@@ -1036,6 +1053,91 @@ accountHolds(
...
+    auto const blob = backend.fetchLedgerObject(key, sequence, yield);
+
+    if (!blob) {
+        amount.setIssue(ripple::Issue(currency, issuer));
+        amount.clear();
+        return amount;
+    }
+
+    ripple::SerialIter it{blob->data(), blob->size()};
+    ripple::SLE const sle{it, key};
```
cindyyan317: You don't need it if `allowBalance` is false.

src/rpc/RPCHelpers.cpp
shawnxie999: I'd prefer leaving as it as for now. In the future there some more incoming amendments that will change this function and two functions will look completely different

shawnxie999: but i'd be open to make this change later if you think it is still necessary.

src/rpc/RPCHelpers.cpp
shawnxie999: fixed

cindyyan317 (dismissed):
Thanks for fixing the comments.

tests/common/util/TestObject.cpp
```diff
@@ -260,6 +261,10 @@ createAccountRootObject(
     accountRoot.setFieldH256(ripple::sfPreviousTxnID, ripple::uint256{previousTxnID});
     accountRoot.setFieldU32(ripple::sfPreviousTxnLgrSeq, previousTxnSeq);
     accountRoot.setFieldU32(ripple::sfTransferRate, transferRate);
+
+    if (ammID != ripple::uint256{0})
```
godexsoft: Do we not have a constant for this zero somewhere? Consider adding if not

tests/unit/feed/TransactionFeedTests.cpp
```diff
@@ -865,7 +868,11 @@ TEST_F(FeedTransactionTest, PubTransactionWithOwnerFund)
 
     EXPECT_CALL(*mockSessionPtr, apiSubversion).WillOnce(testing::Return(1));
     EXPECT_CALL(*mockSessionPtr, send(sharedStringJsonEq(kTRANSACTION_FOR_OWNER_FUND))).Times(1);
-    testFeedPtr->pub(trans1, ledgerHeader, backend_);
+    EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(testing::_, Amendments::fixFrozenLPTokenTransfer, testing::_))
+        .Times(1);
```
godexsoft: Times(1) is default and can be omitted

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -547,6 +556,264 @@ TEST_F(RPCHelpersTest, ParseIssue)
...
+    auto account = getAccountIdWithString(kACCOUNT);
+    auto account2 = getAccountIdWithString(kACCOUNT2);
+
+    auto const usdRippleState =
+        createRippleStateLedgerObject("USD", kACCOUNT2, 100, kACCOUNT, 100, kACCOUNT2, 100, kTXN_ID, 3);
+    auto const usdRippleStateKk = ripple::keylet::line(account2, account, ripple::to_currency("USD")).key;
+
+    // trustline fetched twice. once in accountHolds and once in isFrozen
+    EXPECT_CALL(*backend_, doFetchLedgerObject(usdRippleStateKk, testing::_, testing::_)).Times(2);
+    ON_CALL(*backend_, doFetchLedgerObject(usdRippleStateKk, testing::_, testing::_))
```
godexsoft: All these ON_CALL should be instead setup in the EXPECT_CALL with WillOnce or WillRepeatedly

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -547,6 +556,264 @@ TEST_F(RPCHelpersTest, ParseIssue)
...
+    EXPECT_CALL(*backend_, doFetchLedgerObject(lptRippleStateKk, testing::_, testing::_)).Times(2);
+    ON_CALL(*backend_, doFetchLedgerObject(lptRippleStateKk, testing::_, testing::_))
+        .WillByDefault(testing::Return(lptRippleState.getSerializer().peekData()));
+
+    EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(testing::_, Amendments::fixFrozenLPTokenTransfer, testing::_))
+        .Times(1);
+    ON_CALL(*mockAmendmentCenterPtr_, isEnabled(testing::_, Amendments::fixFrozenLPTokenTransfer, testing::_))
+        .WillByDefault(testing::Return(true));
+
+    auto const ammID = ripple::uint256{54321};
```
godexsoft: We usually add constants like these at the top of the file in an anonymous namespace and make them constexpr

godexsoft (changes requested):
Adding a few suggestions. Please note that these suggestions can be applied across all changes even though the comments are only for select lines

tests/unit/rpc/RPCHelpersTests.cpp
shawnxie999: fixed

tests/unit/rpc/RPCHelpersTests.cpp
shawnxie999: fixed

tests/unit/feed/TransactionFeedTests.cpp
shawnxie999: fixed

tests/common/util/TestObject.cpp
shawnxie999: i've actually changed it to optional which i think makes more sense

godexsoft (approved):
LGTM 🚀

PR #1842, opened 2025-01-21
feat: Generate config descriptions

### LINKED ISSUE #1460: Add descriptive info about various fields of Clio config.
## Summary

Currently there's no documentation or description about various fields of Clio config.

Feature request is to create a document or add more info about each key and acceptable values of Clio config in [https://github.com/XRPLF/clio/blob/develop/docs/examples/config/example-config.json](url)

## Motivation

Lack of documentation or description about various fields of Clio config makes it ambiguous for the user to understand the purpose of the keys and customize their own values.

## Solution

Add info about keys and acceptable values in example config.json similar to rippled.
Rippled example config: [https://github.com/XRPLF/rippled/blob/develop/cfg/rippled-example.cfg](url)

PeterChen13579:
Fixes #1460


src/app/CliArgs.cpp
```diff
@@ -44,6 +45,7 @@ CliArgs::parse(int argc, char const* argv[])
     description.add_options()
         ("help,h", "print help message and exit")
         ("version,v", "print version and exit")
+        ("description,d", "generate config description")
```
kuznetsss:
```suggestion
        ("config-description,d", "generate config description")
```

kuznetsss: Also it probably should take a string where to put output file.

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +70,62 @@ struct ClioConfigDescription {
...
+     */
+    [[nodiscard]] static std::expected<void, Error>
+    getMarkdown()
+    {
+        std::ofstream file(kCONFIG_DESCRIPTION_FILE_NAME);
+
+        if (!file.is_open())
+            return std::unexpected<Error>{fmt::format("failed to create file: {}", kCONFIG_DESCRIPTION_FILE_NAME)};
+
+        file << "# Config Description Markdown File\n";
```
kuznetsss:
```suggestion
        file << "# Clio Config Description\n";
```

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +70,62 @@ struct ClioConfigDescription {
...
+        file << "This is a **markdown** file listing all Clio Configurations in detail.\n\n";
```
kuznetsss: Why do you think it is important to specify that this is a markdown file?

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +70,62 @@ struct ClioConfigDescription {
...
+        file << "## Configuration Details\n\n";
+
+        for (auto const& [key, val] : kCONFIG_DESCRIPTION) {
+            file << "### " << key << "\n";
+            file << val << "\n\n";
```
kuznetsss: Would be good to also add type and constraints for each config entry.

tests/unit/app/CliArgsTests.cpp
```diff
@@ -145,3 +147,24 @@ TEST_F(CliArgsTests, Parse_VerifyConfig)
         returnCode
     );
 }
+
+TEST_F(CliArgsTests, Parse_ConfigDescription)
+{
+    using namespace util::config;
+    std::array argv{"clio_server", "--description"};  // NOLINT(bugprone-suspicious-stringview-data-usage)
```
kuznetsss: Shouldn't the comment be on the next line?

src/util/newconfig/ConfigDescription.hpp
PeterChen13579: Yea true, it's not important to say this is a markdown file

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +72,78 @@ struct ClioConfigDescription {
...
+
+        file.close();
+        std::cout << "Markdown file generated successfully: " << fullFilePath << "\n";
+        return {};
+    }
+
 private:
     static constexpr auto kCONFIG_DESCRIPTION = std::array{
-        KV{.key = "database.type", .value = "Type of database to use. Default is Scylladb."},
+        KV{.key = "database.type",
```
PeterChen13579, line 134: @maria-robobug This part of the PR provides detailed explanations for each config. @kennyzlei mentioned it would be good if the description is user-friendly, ie. anyone without the knowledge of CLIO can understand them. Whenever you have the time, I'd appreciate a review for this section. Tysm 🙇 🥇

maria-robobug: Sounds good, I'll set some time to go over the descriptions.

PeterChen13579: Decided for Maria's review when markdown file is generated in CI.

src/app/CliArgs.cpp
```diff
@@ -48,6 +49,7 @@ CliArgs::parse(int argc, char const* argv[])
         ("ng-web-server,w", "Use ng-web-server")
         ("migrate", po::value<std::string>(), "start migration helper")
         ("verify", "Checks the validity of config values")
+        ("config-description,d", po::value<std::string>(),"generate config description")
```
godexsoft: Let's add a space after the comma 🔢

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +72,78 @@ struct ClioConfigDescription {
...
+     * @param path The path location to generate the Config-description file
+     * @return An Error if generating markdown fails, otherwise nothing
+     */
+    [[nodiscard]] static std::expected<void, Error>
+    getMarkdown(std::string_view path)
+    {
+        namespace fs = std::filesystem;
+
+        // Construct the full file path
+        auto const fullFilePath = fmt::format("{}/{}", path, kCONFIG_DESCRIPTION_FILE_NAME);
```
godexsoft: You probably should use `std::filesystem` to compose the paths instead

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +72,78 @@ struct ClioConfigDescription {
...
     }
 
+    /**
+     * @brief Generate markdown file of all the clio config descriptions
+     *
+     * @param path The path location to generate the Config-description file
+     * @return An Error if generating markdown fails, otherwise nothing
+     */
+    [[nodiscard]] static std::expected<void, Error>
+    getMarkdown(std::string_view path)
```
godexsoft: Would it be better to take a `std::filesystem::path`?

godexsoft: By the name of this function I'd expect it to return the MD as a string or an error if generation failed. It does more than the name suggests - creates a file, writes to file, prints to standard output. I think it would be cleaner if the handling of the generated MD was outside of this function. Consider splitting to smaller functions.

src/util/newconfig/ValueView.hpp
```diff
@@ -141,6 +143,17 @@ class ValueView {
...
 
+    /**
+     * @brief Retrieves the constraint associated with the ConfigValue in this ValueView, if any.
+     *
+     * @return An optional reference to the associated Constraint
+     */
+    [[nodiscard]] constexpr std::optional<std::reference_wrapper<Constraint const>>
+    getConstraint() const
+    {
+        return configVal_.get().getConstraint();
```
godexsoft, line 154: Can this ever be nullopt?

tests/unit/util/newconfig/ClioConfigDefinitionTests.cpp
```diff
@@ -165,7 +165,10 @@ TEST(ConfigDescription, GetValues)
 {
     ClioConfigDescription const definition{};
 
-    EXPECT_EQ(definition.get("database.type"), "Type of database to use. Default is Scylladb.");
+    EXPECT_EQ(
+        definition.get("database.type"),
+        "Type of database to use. We currently support Cassandra and Scylladb. We default to Scylladb."
```
godexsoft, line 170: Does this means that `database.type` is "scylladb" by default and it's an error not to provide "database.scylladb" config?

godexsoft (changes requested):
No real blockers but a few things to consider and a few questions to answer 👍

src/util/newconfig/ConfigDescription.hpp
cindyyan317: I feel the name can be better

src/util/newconfig/Array.hpp
```diff
@@ -98,6 +99,20 @@ class Array {
...
     end() const;
 
+    /**
+     * @brief Custom output stream for Array
+     *
+     * @param stream The output stream
+     * @param arr The Array
+     * @return The same ostream we were given
+     */
+    friend std::ostream&
```
cindyyan317: you are calling its public function. `friend` is not needed?

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -206,6 +240,17 @@ class ValidIPConstraint final : public Constraint {
...
+
+    /**
+     * @brief Prints to the output stream for this specific constraint.
+     *
+     * @param stream The output stream
+     */
+    void
+    print(std::ostream& stream) const override
+    {
+        stream << fmt::format("The value must be a valid IP address") << '\n';
```
cindyyan317: fmt::format is not needed

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -260,6 +305,17 @@ class OneOf final : public Constraint {
...
 
+    /**
+     * @brief Prints to the output stream for this specific constraint.
+     *
+     * @param stream The output stream
+     */
+    void
+    print(std::ostream& stream) const override
+    {
+        stream << fmt::format("The value must be one of the following: `{}`", fmt::join(arr_, ", ")) << '\n';
```
cindyyan317: Why not put `\n` inside the format string?

src/util/newconfig/ConfigConstraints.hpp
```diff
@@ -341,6 +408,17 @@ class PositiveDouble final : public Constraint {
...
+
+    /**
+     * @brief Prints to the output stream for this specific constraint.
+     *
+     * @param stream The output stream
+     */
+    void
+    print(std::ostream& stream) const override
+    {
+        stream << fmt::format("The value must be a positive float number") << '\n';
```
cindyyan317: float? it seems must be positive int64 or double.

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -34,6 +42,9 @@ namespace util::config {
  */
 struct ClioConfigDescription {
 public:
+    /** @brief Name of the Markdown file containing detailed descriptions of all configuration values. */
+    static constexpr auto kCONFIG_DESCRIPTION_FILE_NAME = "Config-Descriptions.md";
```
cindyyan317: Does it need to be public?

src/util/newconfig/ConfigDescription.hpp
cindyyan317: If user want to save the file under the current folder, "" empty string is passed. In this case, we will save it at "/kCONFIG_DESCRIPTION_FILE_NAME".

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +72,78 @@ struct ClioConfigDescription {
...
+        auto const fullFilePath = fmt::format("{}/{}", path, kCONFIG_DESCRIPTION_FILE_NAME);
+
+        // Validate the directory exists
+        auto const dir = fs::path(fullFilePath).parent_path();
+        if (!dir.empty() && !fs::exists(dir))
+            return std::unexpected<Error>{fmt::format("Error: File path {} does not exist", fullFilePath)};
+
+        std::ofstream file(fullFilePath);
+        if (!file.is_open())
+            return std::unexpected<Error>{fmt::format("failed to create file: {}", kCONFIG_DESCRIPTION_FILE_NAME)};
```
cindyyan317: better log the `fullFilePath`

src/util/newconfig/ConfigValue.hpp
```diff
@@ -201,6 +202,29 @@ class ConfigValue {
...
+     *
+     * @param stream The output stream
+     * @param val The config value to output to osstream
+     * @return The same ostream we were given
+     */
+    friend std::ostream&
+    operator<<(std::ostream& stream, ConfigValue val)
+    {
+        stream << "      -   **Required**: " << (val.isOptional() ? "False" : "True") << "\n";
+        stream << "      -   **Type**: : " << val.type() << "\n";
```
cindyyan317: Extra `:`?

cindyyan317 (changes requested):
Good work! 
Please check if we need to handle empty path input.

src/util/newconfig/ValueView.hpp
PeterChen13579: Yes, it can be nullopt, which is fine because that means the config value has no constraint 👍. Already handled the nullopt [here](https://github.com/XRPLF/clio/pull/1842/files#diff-364e6348cd0545ea7dacce670413cc9d5747417f9f6c9dabd08855240202454aR220) in output stream

tests/unit/util/newconfig/ClioConfigDefinitionTests.cpp
PeterChen13579: This part only tests the key-value descriptions, not the actual config values. It means that it expects the description of database.type to match it's corresponding value defined in [configdescription.hpp](https://github.com/XRPLF/clio/pull/1842/files#diff-9d6f9624dd401f7f45974d6ab5496da7648148c93e2a9e86c2f625d0b50cb27eR123)

src/util/newconfig/ConfigConstraints.hpp
PeterChen13579: yes, it should be double here

src/util/newconfig/ConfigDescription.hpp
PeterChen13579: For saving under current folder, you need to pass "." 
It'll say `Markdown file generated successfully: "./Config-Descriptions.md"`
Empty string is not allowed. It'll fail to generate

cindyyan317: Sorry , i did not see the emptiness check

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +69,89 @@ struct ClioConfigDescription {
...
+        fs::path const fullFilePath = path.string();
+
+        // Validate the directory exists
+        auto const dir = fs::path(fullFilePath).parent_path();
+        if (!dir.empty() && !fs::exists(dir))
+            return std::unexpected<Error>{fmt::format("Error: File path {} does not exist", fullFilePath.string())};
+
+        std::ofstream file(fullFilePath);
+        if (!file.is_open())
+            return std::unexpected<Error>{fmt::format("failed to create file: {}", fullFilePath.string())};
```
cindyyan317: `failed`->`Failed`

cindyyan317 (dismissed):
👍 LGTM

src/app/CliArgs.cpp
```diff
@@ -67,6 +70,19 @@ CliArgs::parse(int argc, char const* argv[])
         return Action{Action::Exit{EXIT_SUCCESS}};
     }
 
+    if (parsed.count("config-description") != 0u) {
+        std::filesystem::path filePath = parsed["config-description"].as<std::string>();
+        if (!filePath.empty() && !filePath.string().ends_with(".md"))
+            filePath += ".md";
```
kuznetsss: This is quite unexpected behaviour. I think Clio should use exactly what user passed, without any modiffications.

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +69,89 @@ struct ClioConfigDescription {
...
+    {
+        namespace fs = std::filesystem;
+
+        // Construct the full file path
+        fs::path const fullFilePath = path.string();
+
+        // Validate the directory exists
+        auto const dir = fs::path(fullFilePath).parent_path();
+        if (!dir.empty() && !fs::exists(dir))
+            return std::unexpected<Error>{fmt::format("Error: File path {} does not exist", fullFilePath.string())};
```
kuznetsss:
```suggestion
            return std::unexpected<Error>{fmt::format("Error: Directory {} does not exist or provided path is invalid", dir.string())};
```

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -61,26 +69,89 @@ struct ClioConfigDescription {
...
+
+        std::ofstream file(fullFilePath);
+        if (!file.is_open())
+            return std::unexpected<Error>{fmt::format("Failed to create file: {}", fullFilePath.string())};
```
kuznetsss: Is it possible to specify the error why the file was not created here?

src/util/newconfig/Types.hpp
```diff
@@ -22,18 +22,70 @@
...
 
+/**
+ * @brief Prints the specified config type to output stream
+ *
+ * @param stream The output stream
+ * @param type The config type
+ * @return The same ostream we were given
+ */
+inline std::ostream&
+operator<<(std::ostream& stream, ConfigType type)
```
kuznetsss: Would be better to create a cpp file and move it there.

tests/unit/app/CliArgsTests.cpp
```diff
@@ -35,6 +41,8 @@ struct CliArgsTests : testing::Test {
     testing::StrictMock<testing::MockFunction<int(CliArgs::Action::VerifyConfig)>> onVerifyMock;
 };
 
+static constexpr auto kCONFIG_DESCRIPTION_FILE_NAME = "../configDescription.md";
```
kuznetsss: Better to use a tmp file for each test.

tests/unit/app/CliArgsTests.cpp
```diff
@@ -145,3 +153,70 @@ TEST_F(CliArgsTests, Parse_VerifyConfig)
...
+    for (auto const& key : gClioConfig)
+        EXPECT_TRUE(fileContent.find(key.first));
+
+    std::filesystem::remove(kCONFIG_DESCRIPTION_FILE_NAME);
+}
+
+TEST_F(CliArgsTests, Parse_ConfigDescriptionInvalidPath)
+{
+    using namespace util::config;
+    std::array argv{"clio_server", "--config-description", ""};
```
kuznetsss: Also add test to create a file in `/proc/` please.

tests/unit/app/CliArgsTests.cpp
```diff
@@ -154,15 +153,37 @@ TEST_F(CliArgsTests, Parse_VerifyConfig)
...
+{
+    std::array argv{"clio_server", "--config-description", tmpFile.path.c_str()};
     auto const action = CliArgs::parse(argv.size(), argv.data());
     EXPECT_CALL(onExitMock, Call).WillOnce([](CliArgs::Action::Exit const& exit) { return exit.exitCode; });
 
     // user provide config markdown file name as well
-    ASSERT_TRUE(std::filesystem::exists(kCONFIG_DESCRIPTION_FILE_NAME));
-    std::filesystem::remove(kCONFIG_DESCRIPTION_FILE_NAME);
+    ASSERT_TRUE(std::filesystem::exists(tmpFile.path.c_str()));
+    std::filesystem::remove(tmpFile.path.c_str());
```
kuznetsss: Do you need `c_str()` here and below?

PeterChen13579: Yea, you're right. This isn't needed 👍

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -79,22 +78,42 @@ struct ClioConfigDescription {
...
      */
     [[nodiscard]] static std::expected<void, Error>
-    getMarkdown(std::string_view path)
+    generateConfigDescriptionToFile(std::filesystem::path path)
     {
         namespace fs = std::filesystem;
 
         // Construct the full file path
-        auto const fullFilePath = fmt::format("{}/{}", path, kCONFIG_DESCRIPTION_FILE_NAME);
+        fs::path const fullFilePath = path.string();
```
godexsoft: I don't understand this change. You are given a path and then you copy it into another `path` object by first getting a string out of it. Does this differ from the original `path` you were given at all?

godexsoft: fyi: what I meant previously by saying you should use `filesystem::path` to compose paths is that you could do something like 
```
auto const fullPath = path / kCONFIG_DESCRIPTION_FILE_NAME;
```
this composes the path correctly on all platforms while string manipulation using `{}/{}` would produce a broken path on say Windows (not that we care).

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -79,22 +78,42 @@ struct ClioConfigDescription {
...
 
         // Validate the directory exists
         auto const dir = fs::path(fullFilePath).parent_path();
```
godexsoft: `fs::path(fullFilePath)` is now already a path object

src/util/newconfig/ConfigDescription.hpp
```diff
@@ -79,22 +78,42 @@ struct ClioConfigDescription {
...
         if (!dir.empty() && !fs::exists(dir))
-            return std::unexpected<Error>{fmt::format("Error: File path {} does not exist", fullFilePath)};
+            return std::unexpected<Error>{
+                fmt::format("Error: Directory {} does not exist or provided path is invalid", dir.string())
```
godexsoft: Consider wrapping the path string in `'{}'` or something like that

tests/unit/app/CliArgsTests.cpp
```diff
@@ -148,41 +153,73 @@ TEST_F(CliArgsTests, Parse_VerifyConfig)
...
+    auto const fileContent = buffer.str();
+    EXPECT_TRUE(fileContent.find("# Clio Config Description") != std::string::npos);
+    EXPECT_TRUE(fileContent.find("This file lists all Clio Configuration definitions in detail.") != std::string::npos);
+    EXPECT_TRUE(fileContent.find("## Configuration Details") != std::string::npos);
+
+    // all keys that exist in clio config should be listed in config description file
+    for (auto const& key : gClioConfig)
+        EXPECT_TRUE(fileContent.find(key.first));
+
+    std::filesystem::remove(tmpFile.path);
```
godexsoft: No need to do this. This is already in the destructor of `TmpFile` 👍

tests/unit/app/CliArgsTests.cpp
```diff
@@ -148,41 +153,73 @@ TEST_F(CliArgsTests, Parse_VerifyConfig)
...
 {
-    using namespace util::config;
-    std::array argv{"clio_server", "--config-description", "...."};
+    std::array argv{"clio_server", "--config-description", tmpFile.path.c_str()};
     auto const action = CliArgs::parse(argv.size(), argv.data());
     EXPECT_CALL(onExitMock, Call).WillOnce([](CliArgs::Action::Exit const& exit) { return exit.exitCode; });
 
+    // user provide config markdown file name as well
+    ASSERT_TRUE(std::filesystem::exists(tmpFile.path));
+    std::filesystem::remove(tmpFile.path);
```
godexsoft: No need to do this. This is already in the destructor of `TmpFile` 👍

godexsoft (changes requested):
Few small things can be improved 👍

tests/unit/app/CliArgsTests.cpp
PeterChen13579: ah shoot yea. I wasn't using a temp file before, and now I am, so this can be deleted 👍

godexsoft (approved):
👍

PR #1843, opened 2025-01-21
feat: ETLng task manager

godexsoft:
Implementation of a task manager for ETL ng.
There are a couple TODOs left that require some changes in AsyncFramework to implement properly - I'd like to leave it for a separate PR and then fix those TODOs in a later stage of ETL ng implementation.


src/etlng/impl/TaskManager.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+void
+TaskManager::wait()
+{
+    for (auto& extractor : extractors_)
+        extractor.wait();
+    for (auto& loader : loaders_)
+        loader.wait();
+}
```
kuznetsss, line 130: Should it be a separate method?

src/feed/SubscriptionManager.cpp
```diff
@@ -50,7 +50,7 @@ void
 SubscriptionManager::pubBookChanges(
     ripple::LedgerHeader const& lgrInfo,
     std::vector<data::TransactionAndMetadata> const& transactions
-) const
```
kuznetsss, line 53: Why?

src/util/StrandedPriorityQueue.hpp
```diff
@@ -0,0 +1,98 @@
...
+class StrandedPriorityQueue {
+    util::async::AnyStrand strand_;
+    std::priority_queue<T, std::vector<T>, Compare> queue_;
+
+public:
+    /**
+     * @brief Construct a new priority queue on a strand
+     * @param strand The strand to use
+     */
+    StrandedPriorityQueue(util::async::AnyStrand&& strand) : strand_(std::move(strand))
```
kuznetsss: Shouldn't there be a size limit for the queue to prevent overload?

src/etlng/impl/TaskManager.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+TaskManager::~TaskManager()
+{
+    stop();
+}
+
+void
+TaskManager::run()
+{
+    static constexpr auto kEXTRACTION_WORKERS = 5;
```
kuznetsss: Should this come from config?

tests/unit/etlng/TaskManagerTests.cpp
```diff
@@ -0,0 +1,134 @@
...
+    MOCK_METHOD(void, load, (LedgerData const&), (override));
+    MOCK_METHOD(std::optional<ripple::LedgerHeader>, loadInitialLedger, (LedgerData const&), (override));
+};
+
+struct TaskManagerTests : NoLoggerFixture {
+protected:
+    util::async::CoroExecutionContext ctx_{2};
+    std::shared_ptr<MockScheduler> mockSchedulerPtr_ = std::make_shared<MockScheduler>();
+    std::shared_ptr<MockExtractor> mockExtractorPtr_ = std::make_shared<MockExtractor>();
+    std::shared_ptr<MockLoader> mockLoaderPtr_ = std::make_shared<MockLoader>();
```
kuznetsss: Maybe strict or nice mocks?

src/feed/SubscriptionManager.cpp
godexsoft: This is due to removal of const all over the place in Strand. Kind of a chain reaction.
The change in Strand is consistent with the same changes in Contexts themselves though - we just did not notice this before.

src/util/StrandedPriorityQueue.hpp
godexsoft: Yeah, probably. I can add something 👍

src/etlng/impl/TaskManager.cpp
godexsoft: Yeah, ideally

src/etlng/impl/TaskManager.cpp
godexsoft: Not sure i understand the question. It's separate now because it's used in both `stop` and `run`. Please clarify

src/util/async/AnyStrand.hpp
```diff
@@ -61,7 +62,7 @@ class AnyStrand {
      * @return The type-erased operation
      */
     [[nodiscard]] auto
-    execute(SomeHandlerWithoutStopToken auto&& fn) const
+    execute(SomeHandlerWithoutStopToken auto&& fn)
```
cindyyan317, line 65: Why can't it be const any more?

tests/unit/util/StrandedPriorityQueueTests.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+    async::CoroExecutionContext realCtx{6};
+    async::AnyExecutionContext ctx{realCtx};
+    StrandedPriorityQueue<TestData> queue{ctx.makeStrand()};
+
+    EXPECT_TRUE(queue.empty());
+    static constexpr auto kTOTAL_THREADS = 5u;
+    static constexpr auto kTOTAL_ITEMS_PER_THREAD = 100u;
+
+    std::vector<async::AnyOperation<void>> tasks;
```
cindyyan317, line 105: This test will pass if `enqueue` does not work at all. Maybe we can check the size/emptyness of the queue after threads end.

src/etlng/impl/TaskManager.cpp
kuznetsss: Sorry, I mean, what is the use case of it as a separate method? Why not to put this into `stop()`?

godexsoft: Because it's used in both stop and run. It's also private anyways, basically a helper function

tests/unit/util/StrandedPriorityQueueTests.cpp
godexsoft: While writing it i thought so too but i decided it's not really possible if the constants are non-zero. I can add some checks to make sure the queue had data in it.

cindyyan317 (dismissed):
LGTM

tests/unit/util/StrandedPriorityQueueTests.cpp
```diff
@@ -46,8 +50,9 @@ TEST(StrandedPriorityQueueTests, DefaultPriority)
     util::async::CoroExecutionContext ctx;
     StrandedPriorityQueue<TestData> queue{ctx.makeStrand()};
 
-    for (auto i = 0u; i < 100u; ++i)
-        queue.enqueue(TestData{.seq = i});
+    for (auto i = 0u; i < 100u; ++i) {
+        [[maybe_unused]] auto unused = queue.enqueue(TestData{.seq = i});
```
kuznetsss: Sholdn't this be `EXPECT_TRUE()`?

kuznetsss (approved):

cindyyan317 (approved):

PR #1851, opened 2025-01-24
refactor: Use mutex from utils

### LINKED ISSUE #1359: Use mutex from utils
Mutex from utils (https://github.com/XRPLF/clio/blob/develop/src/util/Mutex.hpp) improves readability of code making it obvious which data is protected by mutex. We should refactor code to use it where possible.

kuznetsss:
Fixes #1359.


godexsoft (approved):
Not a fan of how more places need to write `template` now and even less of a fan of `operator[](m)` but it works. You can avoid the extra `template` via a helper function but that makes the API less discoverable so that's a trade off I suppose.

cindyyan317 (approved):

PR #1853, opened 2025-01-28
fix: CacheLoader causes crash when no cache is used

kuznetsss:
If cache is disabled or Clio starts with and empty DB, `loader_` inside cache is not created. So calling `CacheLoader::stop()` or `CacheLoader::wait()` was causing crash.


godexsoft (approved):
Nice! Thanks for fixing this 👍

PR #1859, opened 2025-01-29
refactor: Remove boost filesystem

PeterChen13579:


godexsoft (approved):
LGTM 👍 The TmpFile uses a system function that gives us warnings but that's out of scope for this PR.

PR #1877, opened 2025-02-06
feat: Snapshot exporting tool

cindyyan317:
In this PR:
1 We create a golang grpc client to request data from rippled
2 We store the data into the specific place
3 Add unittests
4 Create build script, the build can be initiated by set conan option `snapshot` being true.  

Please ignore the grpc server part. It will be implemented in importing tool.


tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+		if i != len(markers)-1 {
+			end = markers[i+1][:]
+		}
+
+		fmt.Printf("Got ledger data marker: %x-%x\n", marker, end)
+
+		go func(marker []byte, end []byte) {
+			defer wg.Done()
+			saveLedgerData(client, seq, marker, end, path)
+		}(marker[:], end)
```
kuznetsss: Why not capture them?

tools/snapshot/internal/server/server.go
```diff
@@ -0,0 +1,60 @@
...
+
+// create a server implement the xrpl rpc v1 server interface
+type Server struct {
+	// The server implementation
+	pb.XRPLedgerAPIServiceServer
+}
+
+// GetLedger prints a message and returns a fake response.
+func (s *Server) GetLedger(ctx context.Context, req *pb.GetLedgerRequest) (*pb.GetLedgerResponse, error) {
+	log.Println("GetLedger called with request:", req)
```
kuznetsss: Maybe use some library with logging levels?

tools/snapshot/internal/server/server.go
```diff
@@ -0,0 +1,60 @@
...
+
+// GetLedgerDiff prints a message and returns a fake response.
+func (s *Server) GetLedgerDiff(ctx context.Context, req *pb.GetLedgerDiffRequest) (*pb.GetLedgerDiffResponse, error) {
+	log.Println("GetLedgerDiff called with request:", req)
+	return &pb.GetLedgerDiffResponse{}, nil
+}
+
+func newServer() *Server {
+	s := &Server{}
+	return s
```
kuznetsss: Why not `return &Server{}`?

tools/snapshot/snapshot.go
```diff
@@ -0,0 +1,67 @@
...
+	exportMode := flag.String("export", "", "Set export mode: 'full' (export full ledger data of start_seq) or 'delta' (only export ledger diff data)")
+	seq := flag.Uint("start_seq", 0, "Starting sequence number")
+	endSeq := flag.Uint("end_seq", 0, "Ending sequence number")
+	path := flag.String("path", "", "Path to the data")
+	grpcServer := flag.String("grpc_server", "localhost:50051", "rippled's gRPC server address")
+	serverMode := flag.Bool("server", false, "Start server mode")
+	grpcPort := flag.Int("grpc_port", 0, "Port for gRPC server to listen on")
+	wsPort := flag.Int("ws_port", 0, "Port for WebSocket server to listen on")
+
+	flag.Parse()
```
kuznetsss: Better to create a structure of cli args. Also`flag` doesn't support short versions of cli options, maybe use `pflag` instead.

tools/snapshot/unittests/export_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"xrplf/clio/clio_snapshot/internal/export"
+	"xrplf/clio/clio_snapshot/mocks"
+	pb "xrplf/clio/clio_snapshot/org/xrpl/rpc/v1"
+
+	"github.com/golang/mock/gomock"
+)
+
+func TestExportDeltaLedgerData(t *testing.T) {
+
```
kuznetsss: I saw many places with an empty line after function's first line. Is it for some purpose?

tools/snapshot/internal/server/server.go
kuznetsss, line 1: > Please ignore the grpc server part. It will be implemented in importing tool.

Maybe not add `server` part  if you don't want it to be reviewed? Otherwise we may forget to review already existing code.

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+		log.Fatalf("Error creating gRPC client: %v", err)
+	}
+
+	defer client.Close()
+
+	ExportFromFullLedgerImpl(client.Client, startSeq, endSeq, path)
+
+}
+
+func ExportFromFullLedgerImpl(client pb.XRPLedgerAPIServiceClient, startSeq uint32, endSeq uint32, path string) {
```
kuznetsss: Why is impl function public?

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+
+	getLedgerFullData(client, startSeq, path)
+
+	//We need to fetch the ledger header and txs for startSeq as well
+	for i := startSeq; i <= endSeq; i++ {
```
kuznetsss, line 219: Shouldn't it be `startSeq + 1` because we already go full data for `startSeq`?

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+
+	filePath := filepath.Join(subPath, fmt.Sprintf("%d.dat", seq))
+
+	err = os.WriteFile(filePath, protoData, 0644)
+	if err != nil {
+		log.Fatalf("failed to write file: %v", err)
+	}
+}
+
+func generate256Markers(markerNum uint32) [][32]byte {
```
kuznetsss: Maybe remove `256` from the name?

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	request.Ledger = ledger
+	request.Marker = marker[:]
+	if end != nil {
+		request.EndMarker = end[:]
+	}
+	request.User = grpcUser
+
+	subPath := filepath.Join(path, fmt.Sprintf("ledger_data_%d", seq), fmt.Sprintf("marker_%x", marker))
+	err := os.MkdirAll(subPath, os.ModePerm)
```
kuznetsss, line 147: Will this fail if the directory already exists?

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+
+	for i := 0; i < int(markerNum); i++ {
+		byteArray[0] = byte(i * int(incr)) // Increment the highest byte
+		byteArrayList = append(byteArrayList, byteArray)
+	}
+
+	return byteArrayList
+}
+
+func saveLedgerData(client pb.XRPLedgerAPIServiceClient, seq uint32, marker []byte, end []byte, path string) {
```
kuznetsss: It is a bit confusing that `saveLedgerData()` performs grpc request while `saveLedgerDeltaData()` only saves provided data. Maybe move some parts of `saveLedgerData()` into `getLedgerFullData()` or a new function?

tools/snapshot/unittests/export_test.go
```diff
@@ -0,0 +1,96 @@
...
+			export.ExportFromDeltaLedgerImpl(mockClient, tt.startSeq, tt.endSeq, "test")
+
+			_, err := os.Stat("test")
+
+			assert.Equal(t, os.IsNotExist(err), tt.endSeq < tt.startSeq)
+		})
+	}
+}
+
+func TestExportFullLedgerData(t *testing.T) {
```
kuznetsss: These tests look almost like integration tests. Shouldn't there be unit tests for each function where possible?

tools/snapshot/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
+# Build snapshot tool
+# Need to be installed before building: go protoc-gen-go protoc-gen-go-grpc
+
+set(GO_EXECUTABLE "go")
```
kuznetsss, line 4: Maybe instead of doing this we should create a go package exporting libxrpl's proto file and use the package here. I think it would be more user friendly to be able to build this tool with simply `go build`, not this conan-cmake-go C++ style way.

cindyyan317: I think it is more friendly to use conan-cmake-go way, since it is inside Clio repo. It has the same way of building clio_server.

tools/snapshot/internal/export/export.go
cindyyan317: Because I'd like to test it in unittests.

tools/snapshot/internal/server/server.go
cindyyan317: For now, `log` seems enough for this tool . I will change it if log get complicated .

tools/snapshot/unittests/export_test.go
cindyyan317: No , it is not.  I will remove them. Thanks.

tools/snapshot/internal/server/server.go
cindyyan317: The server part is very simple for now. It just to verify the grpc server can be started. I can remove it if it is noise.

tools/snapshot/internal/export/export.go
cindyyan317: Ledger full data does not contain the header and transactions.

tools/snapshot/internal/export/export.go
cindyyan317: In this case, it does nothing. No error.

tools/snapshot/unittests/export_test.go
cindyyan317: I only test the public interfaces. I keep the tests under unittests module, not the export. I don't want to change functions to private just for testing purpose.

tools/snapshot/CMakeLists.txt
godexsoft: While it's a good idea to have this separated out - let's not spend time doing this now and instead keep it in the repo until we feel we really need to separate this tool. I agree with Cindy that it's more convenient and maintainable to do it this way now.

tools/snapshot/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+# Initialize the options strings
+set(GO_OPTS "")
+set(GRPC_OPTS "")
+
+# Loop through each proto file
+foreach(proto ${PROTO_FILES})
+    get_filename_component(proto_filename ${proto} NAME)
+
+    # Build the --go_opt and --go-grpc_opt mappings
+    set(GO_OPTS ${GO_OPTS} --go_opt=M${GO_IMPORT_PATH}/${proto_filename}=${GO_IMPORT_PATH})
```
godexsoft, line 30: Can any of the paths here have a space? If it does for some reason, will it still work or should we add double quotes everywhere?

tools/snapshot/internal/export/export.go
godexsoft: You should be able to test private functions too: https://stackoverflow.com/questions/24622388/how-to-test-a-unexported-private-function-in-go-golang

tools/snapshot/internal/export/export.go
```diff
@@ -0,0 +1,245 @@
...
+)
+
+const (
+	deltaDataFolderDiv = 10000
+	grpcUser           = "clio-snapshot"
+	markerNum          = 16
+)
+
+type GRPCClient struct {
+	Client pb.XRPLedgerAPIServiceClient
```
godexsoft, line 24: Can client be private?

tools/snapshot/snapshot.go
godexsoft: We use a new library for parsing cli opts every tool we build :rage3:  let's settle on one library and use it? I quite liked the one we used in the DB cleaner.

cindyyan317: Ha after some research,  `pflag` is more common use. I will use `pflag` then.

tools/snapshot/internal/export/export.go
cindyyan317: We can test the private functions if we put the unittests and code in same package. But should we test private functions? In Cpp ut, we only test public functions too. I shall only `ExportFromFullLedger` here, `ExportFromFullLedgerImpl` is to inject the mock grpc client.

tools/snapshot/CMakeLists.txt
cindyyan317: Both GO_IMPORT_PATH and proto_filename are impossible to have space. I think it is okay here. `protoc-gen-go-grpc` fails to send the mapping to golang plugin if we add quotes.

kuznetsss (approved):
👍

PR #1879, opened 2025-02-06
feat: Run tests with sanitizers in CI

### LINKED ISSUE #1049: Build and run tests with sanitizers in CI

### LINKED ISSUE #1075: Create build_clio reusable workflow
It turns out one workflow can call another one.

So to avoid duplications in building clio we could move all the common code into reusable workflow and call it from both regular check and nightly build. And in that case we could move code coverage to a separate workflow and reuse build workflow.

Materials for more info and examples about reusable workflows:
https://docs.github.com/en/actions/using-workflows/reusing-workflows
https://dev.to/n3wt0n/avoid-duplication-github-actions-reusable-workflows-3ae8

godexsoft:
Fixes #1075
Fixes #1049

This PR implements a new workflow that will build and run clio_tests with TSAN and ASAN (UBSAN supported but turned off for now).
If sanitizer reports any errors a report is generated and uploaded to artifacts.
Since at this moment there are lots of errors generated, we disabled the automatic creation of an issue on github on failure, otherwise we would be generating an issue nightly. The plan is to work on fixing the existing problems reported by TSAN and ASAN and enable the automated issue creation once those are resolved in the codebase.

The PR also adjusts the schedule of clang-tidy and nightly build to accommodate the new workflow.


.github/actions/build_clio/action.yml
```diff
@@ -4,12 +4,18 @@ inputs:
   target:
     description: Build target name
     default: all
+  substract_threads:
```
godexsoft, line 7: This is not strictly needed anymore (was a test to see if heavy would survive) but we can keep it in case we need it in the future.

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,107 @@
...
+    name: Build clio tests
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - sanitizer: tsan
+            compiler: gcc
+          - sanitizer: asan
+            compiler: gcc
+          # - sanitizer: ubsan # todo: enable when heavy runners are available
```
godexsoft, line 21: Can't enable right now on gh runners because this job simply takes longer than 6 hours and gets nuked from orbit.

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,107 @@
...
+          #   compiler: gcc
+    runs-on: ubuntu-latest # todo: change to heavy
```
godexsoft, line 51: Can't change to heavy right now as it dies every time with OOM.

cmake/Settings.cmake
```diff
@@ -39,6 +39,29 @@ if (is_appleclang)
   list(APPEND COMPILER_FLAGS -Wreorder-init-list)
 endif ()
 
+if (san)
```
godexsoft, line 42: When building with sanitizers some compilers will actually produce extra warnings/errors. We don't want this yet, at least not until we have fixed all runtime issues reported by the sanitizers. Once that is done we can start removing some of these and trying to fix it in our codebase. We can never remove all of below because most of them are reported from deep inside libraries like boost or libxrpl.

cmake/deps/libbacktrace.cmake
```diff
@@ -1,3 +1,7 @@
-target_compile_definitions(clio_options INTERFACE BOOST_STACKTRACE_LINK)
-target_compile_definitions(clio_options INTERFACE BOOST_STACKTRACE_USE_BACKTRACE)
-find_package(libbacktrace REQUIRED CONFIG)
+if (Boost_STACKTRACE_FOUND)
+  target_compile_definitions(clio_options INTERFACE BOOST_STACKTRACE_LINK)
+  target_compile_definitions(clio_options INTERFACE BOOST_STACKTRACE_USE_BACKTRACE)
+  find_package(libbacktrace REQUIRED CONFIG)
+else ()
+  target_compile_definitions(clio_options INTERFACE CLIO_WITHOUT_STACKTRACE)
```
godexsoft, line 9: Some sanitizers (TSAN and ASAN for sure) can't be used with libbacktrace because they have their own backtracing capabilities and there are conflicts. In any case, this makes sure Clio code knows that backtrace is not available.

.github/actions/build_clio/action.yml
```diff
@@ -4,12 +4,18 @@ inputs:
...
+    description: See get_number_of_threads
```
kuznetsss:
```suggestion
    description: An option for the action get_number_of_threads. See get_number_of_threads 
```

.github/actions/generate/action.yml
```diff
@@ -12,6 +12,14 @@ inputs:
...
     required: true
     default: 'Release'
+  build_integration_tests:
+    description: Whether to build integration tests
+    required: true
+    default: 'true'
+  lint:
+    description: Whether to build with linter enabled
+    required: true
+    default: 'false'
```
kuznetsss: I don't think we need this one. Linting during compilation might be useful for development, but because build fails on the first issue `clang-tidy` found, I think it is not useful in CI.

.github/actions/generate/action.yml
```diff
@@ -33,14 +45,21 @@ runs:
...
+    
     - name: Run cmake
       shell: bash
       env:
         BUILD_TYPE: "${{ inputs.build_type }}"
+        SANITIZER_OPTION: |
+          ${{ inputs.sanitizer == 'tsan' && '-Dsan=thread' ||
+              inputs.sanitizer == 'ubsan' && '-Dsan=undefined' ||
+              inputs.sanitizer == 'asan' && '-Dsan=address' ||
+              '' }}
```
kuznetsss, line 57: Maybe extract it into a separate step and fail the job if `inputs.sanitizer` is incorrect?

.github/workflows/build.yml
```diff
@@ -47,133 +47,44 @@ jobs:
...
-        if: ${{ matrix.code_coverage }}
-        uses: ./.github/actions/code_coverage
-
-  upload_coverage_report:
-    name: Codecov
-    needs: build
-    uses: ./.github/workflows/upload_coverage_report.yml
-    secrets:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+    uses: ./.github/workflows/reusable_build.yml
```
kuznetsss: `reusable_build` sounds more like it creates something reusable. Maybe rename `build` to `on_push_check` and `reusable_build` to `build`?

.github/workflows/reusable_build.yml
```diff
@@ -0,0 +1,198 @@
...
+        description: Whether ccache should be disabled
+        required: false
+        type: boolean
+        default: false
+
+      conan_cache:
+        description: Whether to use conan cache
+        required: false
+        type: boolean
+        default: true
```
kuznetsss: I think we could have only one option `disable_cache`. I couldn't imagine a situation where we are ok with one cache, but want to avoid another one. WDYT?

.github/workflows/nightly.yml
```diff
@@ -21,68 +21,23 @@ jobs:
...
+      runs_on: ${{ matrix.os }}
+      container: ${{ matrix.container }}
+      conan_profile: gcc
+      build_type: ${{ matrix.build_type }}
+      code_coverage: false
+      static: ${{ matrix.static }}
+      unit_tests: true
+      integration_tests: true
+      clio_server: true
+      disable_ccache: true
```
kuznetsss: It seems `conan_cache: false` is missing.

kuznetsss (commented):
Really good improvement, thanks!

kuznetsss:
Let's link #1075 and #1049 please.

.github/workflows/reusable_build.yml
godexsoft: I can remove second cache input.

cmake/Settings.cmake
kuznetsss: Maybe this and the next your comment should be put into related CMake files to not forget about it?

.github/actions/generate/action.yml
godexsoft: Let's do this in the CI improvement ticket if you don't mind

godexsoft: I have addressed that through cmake now. Should be a readable error in the logs 👍

cindyyan317 (dismissed):
💯  Nice !

.github/workflows/build.yml
```diff
@@ -93,37 +90,42 @@ jobs:
...
 
       - name: Run conan and cmake
         uses: ./.github/actions/generate
         with:
           conan_profile: ${{ steps.conan.outputs.conan_profile }}
-          conan_cache_hit: ${{ steps.restore_cache.outputs.conan_cache_hit }}
-          build_type: ${{ matrix.build_type }}
-          code_coverage: ${{ matrix.code_coverage }}
-          static: ${{ matrix.static }}
+          conan_cache_hit: ${{ !inputs.restore_cache && steps.restore_cache.outputs.conan_cache_hit }}
```
kuznetsss: It seems there is no `inputs.restore_cache`.

.github/workflows/sanitizers.yml
kuznetsss: Github runner is only 4 CPU and 16 GB RAM. Probably just add a thread shift to make heavy runner work.

.github/workflows/build.yml
godexsoft: Good catch. I think i had them as separate inputs at first but then decided it could be covered by `conan_cache` but forgot to update this part 🔢  Now that everything is handled by `disable_cache` it should be fairly simple.

.github/workflows/sanitizers.yml
godexsoft: I actually tried with `substract_threads` set to 2 before and it did not help anything.. not sure it's worth spending more time trying for now - we can try again in the next iteration of adding sanitizers.

kuznetsss (approved):

kuznetsss (approved):

.github/workflows/on_push_checks.yml
```diff
@@ -0,0 +1,129 @@
...
+            conan_profile: clang
+            build_type: Debug
+            container: '{ "image": "rippleci/clio_ci:latest" }'
+            code_coverage: false
+            static: true
+          - os: macos15
+            build_type: Release
+            code_coverage: false
+            static: false
+    uses: ./.github/workflows/build.yml
```
mathbunnyru, line 77: Forgot to pass secrets.CODECOV_TOKEN here, commenting for a history

PR #1884, opened 2025-02-07
fix: Array parsing in new config

### LINKED ISSUE #1870: New Config parses arrays incorrectly
## Issue Description

If a config array consists of objects and each object has an optional field, json config might be parsed incorrectly.

This happens because json flattening discards index of each object and pushes back each field into flatten array.

kuznetsss:
Fixes #1870.


src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -165,4 +181,97 @@ ConfigFileJson::flattenJson(boost::json::object const& obj, std::string const& p
...
+    array: [
+        { a: 1},
+        { b: 2}
+    ]
+}
+
+array.[].a
+array.[].b
+
+*/
```
kuznetsss: Forgot to delete 😄

src/util/newconfig/Array.cpp
```diff
@@ -36,14 +36,21 @@ Array::Array(ConfigValue arg) : itemPattern_{std::move(arg)}
 {
 }
 
+std::string_view
+Array::prefix(std::string_view key)
+{
+    ASSERT(key.contains(".[]"), "Provided key is not an array key: {}", key);
+
+    return key.substr(0, key.rfind(".[]") + 3);
```
godexsoft: nit: I think i understand what 3 means here but perhaps it would be nice to extract this into some constant just above to make it more clear and avoid magic numbers.

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -165,4 +181,97 @@ ConfigFileJson::flattenJson(boost::json::object const& obj, std::string const& p
...
+
+void
+ConfigFileJson::flattenJsonNonRecursive(boost::json::object const& jsonRootObject)
```
godexsoft: nit: If we don't need the original implementation anymore, perhaps this can be renamed to `flattenJson` 🔢

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -165,4 +181,97 @@ ConfigFileJson::flattenJson(boost::json::object const& obj, std::string const& p
...
+        std::string prefix;
+        std::optional<size_t> arrayIndex = std::nullopt;
+    };
+
+    std::queue<Task> tasks;
+    tasks.push(Task{.object = jsonRootObject, .prefix = ""});
+
+    std::unordered_map<std::string, size_t> arraysSizes;
+
+    while (tasks.size() > 0) {
```
godexsoft: nit: Queue also provides `.empty()` which probably reads more like english (e.g. `not tasks.empty()`)

src/util/newconfig/Types.hpp
```diff
@@ -52,9 +70,24 @@ getType()
...
+        return ConfigType::Null;
     } else {
         static_assert(util::Unsupported<Type>, "Wrong config type");
     }
 }
 
 }  // namespace util::config
+
+/** @cond */
+// Doxygen could not parse this
```
godexsoft, line 104: Doxygen is so sad :rage4:

tests/unit/util/newconfig/ConfigFileJsonTests.cpp
```diff
@@ -0,0 +1,480 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
```
godexsoft, line 4: 2025 👍

tests/unit/util/newconfig/ConfigValueTests.cpp
```diff
@@ -105,14 +214,14 @@ TEST(ConfigValue, OneOfConstraint)
...
-TEST(ConfigValue, IpConstraint)
+TEST_F(ConstraintTest, IpConstraint)
 {
     auto ip = ConfigValue{ConfigType::String}.defaultValue("127.0.0.1").withConstraint(gValidateIp);
     EXPECT_FALSE(ip.setValue("http://127.0.0.1").has_value());
     EXPECT_FALSE(ip.setValue("http://127.0.0.1.com").has_value());
     auto const err = ip.setValue("123.44");
     EXPECT_TRUE(err.has_value());
-    EXPECT_EQ(err->error, "Ip is not a valid ip address");
+    EXPECT_EQ(err->error, "Unknown_key Ip is not a valid ip address");
```
godexsoft, line 226: nit: The final error message reads really strangely. `[something] Ip is not a valid ip address`. Maybe `The value of [something] is not a valid IP address` or similar is better?

tests/unit/util/newconfig/ConfigFileJsonTests.cpp
godexsoft: Up to you - i see now that this file was moved from another place.

godexsoft (changes requested):
Leaving small suggestions. Overall looks great 👍 🚀

src/util/newconfig/Array.hpp
```diff
@@ -46,6 +46,17 @@ class Array {
      */
     Array(ConfigValue arg);
 
+    /**
+     * @brief Extract array prefix from a key, For example for a key foo.[].bar the method will return foo.[]
+     * @note Provided key must contain '.[]'
+     * @warning Be carriful with string_view! Returned value is valid only while the key is valid
```
PeterChen13579: Be `careful` 😄

tests/unit/util/newconfig/ConfigFileJsonTests.cpp
```diff
@@ -0,0 +1,480 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <string>
+#include <unordered_map>
+#include <variant>
+
+using namespace util::config;
+
+namespace {
+constexpr auto kEPS = 1e-9;
```
PeterChen13579, line 41: what does EPS mean?

PeterChen13579 (commented):
Looks good!!!  🚀

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -69,14 +74,17 @@ extractJsonValue(boost::json::value const& jsonValue)
     if (jsonValue.is_double()) {
         return jsonValue.as_double();
     }
+    if (jsonValue.is_null()) {
+        return NullType{};
+    }
     ASSERT(false, "Json is not of type int, uint, string, bool or double");
     std::unreachable();
```
cindyyan317, line 81: Not for this Pr, do we need `unreachable` here?

src/util/newconfig/ConfigFileJson.cpp
```diff
@@ -69,14 +74,17 @@ extractJsonValue(boost::json::value const& jsonValue)
     if (jsonValue.is_double()) {
         return jsonValue.as_double();
     }
+    if (jsonValue.is_null()) {
+        return NullType{};
+    }
     ASSERT(false, "Json is not of type int, uint, string, bool or double");
```
cindyyan317: maybe add `null` here. "Json is not of type int, uint, string, bool, double or empty")

src/util/newconfig/ConfigFileJson.cpp
kuznetsss: Yeah, because this method returns something. Without `unreachable` compiler is not happy that we don't return anything.

tests/unit/util/newconfig/ConfigFileJsonTests.cpp
kuznetsss: It is short for epsilon - a greek letter which is used in math as a minimal value to see a difference between values.
I don't mind renaming if needed.

tests/unit/util/newconfig/ConfigValueTests.cpp
kuznetsss: I think it is out of the PR's scope, because such style of errors is currently used across 'newconfig` library, so changing it only here would make it worse. I will create a task for refactoring.

kuznetsss: Created #1886.

tests/unit/util/newconfig/ConfigFileJsonTests.cpp
PeterChen13579: Ah I see. EPS is fine.

PeterChen13579 (dismissed):
LGTM

godexsoft (dismissed):
🚀

godexsoft (approved):

PR #1890, opened 2025-02-11
chore: Revert workflow names

godexsoft:
In #1879 we renamed `build.yml` and that was a mistake. Let's revert those names as well as fix a few small issues that slipped through the review last time.


kuznetsss (dismissed):
👍

cindyyan317 (dismissed):

kuznetsss (approved):

PR #1896, opened 2025-02-14
fix: Array parsing in new config

kuznetsss:
Improving array parsing in config:
- Allow null values in arrays for optional fields
- Allow empty array even for required field
- Allow to not put an empty array in config even if array contains required fields


src/util/newconfig/Array.cpp
```diff
@@ -48,13 +48,22 @@ Array::prefix(std::string_view key)
...
-    elements_.emplace_back(std::move(newElem));
+    elements_.emplace_back(std::move(newItem));
+    return std::nullopt;
+}
+
+std::optional<Error>
+Array::addNull(std::optional<std::string_view> key)
+{
+    if (not itemPattern_.isOptional() and not itemPattern_.hasValue()) {
+        return Error{key.value_or("Unknown_key"), "Value for the array (or object field inside array) is required"};
```
PeterChen13579: nit: lower case `v` for `Value`. So error reads as `ConfigKey value for the array...`

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -179,21 +183,60 @@ ClioConfigDefinition::parse(ConfigFileInterface const& config)
...
                               }
             },
             value
         );
     }
+
+    if (!listOfErrors.empty())
+        return listOfErrors;
+
+    for (auto const& item : arrayPrefixesToKeysMap) {
```
PeterChen13579: Perhaps change `item` into `[ _, keys]` and use `keys` instead of `item.second` for readability

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -179,21 +183,60 @@ ClioConfigDefinition::parse(ConfigFileInterface const& config)
...
+        size_t maxSize = 0;
+        std::ranges::for_each(item.second, [&](std::string_view key) {
```
PeterChen13579: I see why this is needed; maybe an extra comment here saying what maxSize checks for and why?

PeterChen13579 (changes requested):
Looks good just a few nits 👍 🚀

PeterChen13579 (dismissed):
LGMT

src/util/newconfig/ConfigDefinition.cpp
```diff
@@ -179,21 +183,61 @@ ClioConfigDefinition::parse(ConfigFileInterface const& config)
...
+            maxSize = std::max(maxSize, arraySize(key));
+        });
+        if (maxSize == 0) {
+            // empty arrays are allowed
+            continue;
+        }
+
+        std::ranges::for_each(keys, [&](std::string_view key) {
+            auto& array = std::get<Array>(map_.at(key));
+            if (array.size() != maxSize) {
```
cindyyan317: looks like this `if` and `while` can merge -> `while(array.size()!=maxSize)`

cindyyan317 (approved):

PR #1898, opened 2025-02-17
feat: ETLng monitor

godexsoft:
This PR
- Adds support for subscribing for and notifying on new ledgers ingested by Clio in `ValidatedNetworkLedgers` class
- Implements a `Monitor` class that uses the `ValidatedNetworkLedgers` as one datasource and Clio's database as another datasource in order to notify subscribers when new ledgers become available
- Modifies existing async framework to support repeated operations over strands as well as support force-invocation on certain types of operations (such as repeated ones)
- Covers above functionality with unit-tests


src/etlng/MonitorInterface.hpp
```diff
@@ -0,0 +1,69 @@
...
+namespace etlng {
+
+/**
+ * @brief An interface for the monitor service
+ * An implementation of this service is responsible for periodically checking various datasources to detect newly
+ * ingested ledgers.
+ */
+class MonitorInterface {
+public:
+    static constexpr auto kDEFAULT_REPEAT_INTERVAL = std::chrono::milliseconds{util::kMILLISECONDS_PER_SECOND};
```
kuznetsss: Maybe simply
```suggestion
    static constexpr auto kDEFAULT_REPEAT_INTERVAL = std::chrono::seconds{1};
```

src/util/async/AnyStrand.hpp
```diff
@@ -131,14 +131,43 @@ class AnyStrand {
...
 
+    /**
+     * @brief Schedule a repeating operation on the execution context
+     *
+     * @param interval The interval at which the operation should be repeated
+     * @param fn The block of code to execute; no args allowed and return type must be void
+     * @return A repeating stoppable operation that can be used to wait for its cancellation
+     */
+    [[nodiscard]] auto
+    executeRepeatedly(SomeStdDuration auto interval, SomeHandlerWithoutStopToken auto&& fn)
```
kuznetsss, line 142: Why does `interval` need to be generic? I thought `std::chrono::steady_clock::duration` is good enough.

src/util/async/Operation.hpp
```diff
@@ -266,6 +269,15 @@ class RepeatingOperation : public util::MoveTracker {
...
         repeat_.stop();
     }
+
+    /**
+     * @brief Force-invoke the operation
+     */
+    void
+    invoke()
+    {
+        action_();
```
kuznetsss, line 282: Maybe it would be good to mention somewhere in docs that action should be safe to run multiple times in parallel unless this is used on a strand.

src/etlng/impl/Monitor.cpp
```diff
@@ -0,0 +1,99 @@
...
+Monitor::subscribe(SignalType::slot_type const& subscriber)
+{
+    return notificationChannel_.connect(subscriber);
+}
+
+void
+Monitor::onNextSequence(uint32_t seq)
+{
+    LOG(log_.debug()) << "rippled published sequence " << seq;
+    repeatedTask_->invoke();  // force-invoke immediately
```
cindyyan317, line 87: If the `doWork()` is running, will this function call `doWork()` or not?

src/etlng/impl/Monitor.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void
+Monitor::doWork()
+{
+    if (auto rng = backend_->hardFetchLedgerRangeNoThrow(); rng) {
+        while (rng->maxSequence >= nextSequence_)
```
cindyyan317, line 94: If a readonly node lags for a while, now it is catching up from ledgerA - 1000:
while it is in this loop notifying ledgerA - 1000 to ledgerA, ledgerA + 1 is validated, so the doWork will be re-enter from onNextSequence path?

cindyyan317 (dismissed):
Looks good to me.
Just need to confirm the repeating work will always be executed in a strand.

src/etlng/impl/Monitor.cpp
godexsoft: It will schedule through the context. If the context is a strand then it can't overlap, if it is not a strand then it can. I'll add this to doxygen as Sergey suggested 👍

src/etlng/impl/Monitor.cpp
godexsoft: Same story with strand vs normal ctx.

src/util/async/AnyStrand.hpp
godexsoft: Initially it was to support any duration type, not just from `steady_clock`, but i agree your suggestion could work too. I don't think we need to change this in the framework though, at least not within this PR

kuznetsss (approved):

PR #1926, opened 2025-02-26
fix: Data race in new webserver

kuznetsss:
There was a data race inside `CoroutineGroup` because internal timer was used from multiple threads in the methods `asyncWait()` and `onCoroutineComplete()`. Changing `registerForeign()` to spawn to the same `yield_context` fixes the problem because now the timer is accessed only from the same coroutine which has an internal strand.

During debugging I also added websocket support for `request_gun` tool.


cindyyan317 (approved):
Fix looks good to me 👍

PR #1938, opened 2025-03-04
fix: Fix dangling reference in new web server

kuznetsss:
Also delete move constructors where moving may be dangerous.

### DIFF
--- a/src/data/CassandraBackend.hpp
+++ b/src/data/CassandraBackend.hpp
@@ -120,7 +120,9 @@ public:
         LOG(log_.info()) << "Created (revamped) CassandraBackend";
     }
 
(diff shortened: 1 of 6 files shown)


src/data/CassandraBackend.hpp
```diff
@@ -120,6 +120,9 @@ class BasicCassandraBackend : public BackendInterface {
         LOG(log_.info()) << "Created (revamped) CassandraBackend";
     }
 
+    // Move constructor is deleted because handle_ is shared by reference with executor_
```
godexsoft: Why not doxygen like in other places? :rage1:

godexsoft (dismissed):

godexsoft (approved):

PR #1941, opened 2025-03-04
fix: change math/rand to crypto/rand

PeterChen13579:


tools/cassandra_delete_range/internal/util/util.go
```diff
@@ -1,10 +1,11 @@
 package util
 
 import (
+	"crypto/rand"
```
kuznetsss, line 4: Don't you also need to do `go mod tidy` to update `go.mod` and `go.sum` files?

PeterChen13579: Oh cool 😄 , I didn't know you have to run `go mod tidy`, but after searching, I see why it is required.
In this case though, `crypto/rand `and `math/rand` are both part of the standard library for Golang, so they aren't stored in `go.mod` and `go.sum` files, so after running, `go mod tidy`, nothing happens

cindyyan317 (approved):

PR #1946, opened 2025-03-06
fix: Improve error message when starting read only mode with empty DB 

### LINKED ISSUE #1721: Unclear error when Clio is in read only mode
## Issue Description

If Clio is configured to work in read only mode but the database is empty (so read only should be false to fill the database), the error message is quite unclear:
```
Exit on exception: Invalid query: Keyspace clio does not exist
```
The error message could be improved to provide a hint to turn off read only mode

kuznetsss:
Fixes #1721


godexsoft (approved):
👍

PR #1947, opened 2025-03-07
test: Add assert mock to avoid death tests

### LINKED ISSUE #1750: Improve ASSERT to speed up tests
## Summary

Improve ASSERT to speed up tests.

## Motivation

Testing ASSERT() cases using gtest's death tests is very slow with code coverage enabled.

## Solution

ASSERT() could be rewritten into something like:
```cpp
class AssertImpl {
  static std::function<void(...)> onAssertFailed_;
public:
   static operator()(bool cond, ... ) {
      if (not cond) {
        if (not onAssertFailed) {
            defaultAssertFailed(...);
        } else {
           onAssertFailed_(...);
        }
      }
   }
   
  static defaultAssertFailed(...) { ... }
  static setAssertCallback(std::function<void(...)>) { ... }
};
```
Having such implementation we could create `WithAssertMock` class for tests which will set mock assert implementation in constructor and reset it in destructor. To mimic runtime break the real assert does, mock could throw an exception and tests could expect the exception instead of expecting death. Probably there are cases where `ASSERT()` is used inside try cache block, but there should be not so many of them and for such cases we could throw some unusual exception which is not inherited from `std::exception`.

kuznetsss:
Fixes #1750


src/util/Assert.cpp
```diff
@@ -0,0 +1,67 @@
...
+void
+OnAssert::resetAction()
+{
+    OnAssert::action = [](std::string_view m) { OnAssert::defaultAction(m); };
+}
+
+void
+OnAssert::defaultAction(std::string_view message)
+{
+    if (boost::log::core::get()->get_logging_enabled()) {
```
godexsoft: Should we add this as `LogService::enabled()`?

godexsoft (dismissed):
Wonderful 🚀 
Just leaving one optional suggestion 👍

godexsoft (approved):

Issue #1200, opened 2024-02-21
Get rid of Cache in BackendInterface


godexsoft:
We have Cache inside of BackendInterface that is accessible via the `cache()` member function. A refactoring effort is needed to get rid of it and inject it where it's needed instead. This way we can have mocked cache in tests as well as better architecture in general.

PR #1955, opened 2025-03-11
feat: Implement and use LedgerCacheInterface

godexsoft:
For #1200 

This PR introduces `LedgerCacheInterface` and modifies the rest of the codebase to use cache (mostly) through this interface.
Removal of `cache()` getters in `BackendInterface` will be done after old ETL code is removed.

### DIFF
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -23,6 +23,7 @@
 #include "app/WebHandlers.hpp"
 #include "data/AmendmentCenter.hpp"
 #include "data/BackendFactory.hpp"
+#include "data/LedgerCache.hpp"
 #include "etl/ETLService.hpp"
 #include "etl/LoadBalancer.hpp"
 #include "etl/NetworkValidatedLedgers.hpp"
@@ -102,9 +103,10 @@ ClioApplication::run(bool const useNgWebServer)
     auto whitelistHandler = web::dosguard::WhitelistHandler{config_};
     auto dosGuard = web::dosguard::DOSGuard{config_, whitelistHandler};
     auto sweepHandler = web::dosguard::IntervalSweepHandler{config_, ioc, dosGuard};
+    auto cache = data::LedgerCache{};
 
     // Interface to the database
-    auto backend = data::makeBackend(config_);
+    auto backend = data::makeBackend(config_, cache);
 
     auto const amendmentCenter = std::make_shared<data::AmendmentCenter const>(backend);
 
--- a/src/data/LedgerCache.cpp
+++ b/src/data/LedgerCache.cpp
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2025, the clio developers.
+    Copyright (c) 2022, the clio developers.
 
     Permission to use, copy, modify, and distribute this software for any
     purpose with or without fee is hereby granted, provided that the above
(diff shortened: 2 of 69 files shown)


src/app/ClioApplication.cpp
```diff
@@ -102,9 +103,10 @@ ClioApplication::run(bool const useNgWebServer)
     auto whitelistHandler = web::dosguard::WhitelistHandler{config_};
     auto dosGuard = web::dosguard::DOSGuard{config_, whitelistHandler};
     auto sweepHandler = web::dosguard::IntervalSweepHandler{config_, ioc, dosGuard};
+    auto cache = data::LedgerCache{};
```
kuznetsss, line 106: It looks like an implementation leak. Maybe create a two overloads of `makeBackend()` with and without cache?

src/data/LedgerCache.cpp
```diff
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of clio: https://github.com/XRPLF/clio
-    Copyright (c) 2022, the clio developers.
+    Copyright (c) 2025, the clio developers.
```
kuznetsss: Why? If this file was created in 2022, it should be 2022.

src/data/LedgerCacheInterface.hpp
```diff
@@ -0,0 +1,146 @@
...
+#include <cstdint>
+#include <optional>
+#include <vector>
+
+namespace data {
+
+/**
+ * @brief Cache for an entire ledger.
+ */
+class LedgerCacheInterface {
```
kuznetsss, line 37: Move constructor should be deleted since we hold it by reference in Backend.

src/data/LedgerCache.hpp
```diff
@@ -83,109 +84,44 @@ class LedgerCache {
     std::unordered_set<ripple::uint256, ripple::hardened_hash<>> deletes_;
 
 public:
-    /**
-     * @brief Update the cache with new ledger objects.
-     *
-     * @param objs The ledger objects to update cache with
-     * @param seq The sequence to update cache for
-     * @param isBackground Should be set to true when writing old data from a background thread
-     */
```
kuznetsss, line 92: Is doxygen ok with the absence of documentation here because of `override`?

godexsoft: Yeah, it takes the docs from the interface instead

src/data/LedgerCache.cpp
godexsoft: It looks weird with the interface created in 2025 😃 I can change back.

src/app/ClioApplication.cpp
godexsoft: What would it be instead? Either an instance inside of backend or a reference from outside? In my understanding backend should have never hosted the cache in the first place, it should have always been a dependency to inject. Can you explain your view?

src/data/LedgerCache.cpp
kuznetsss: I don't think it looks weird. 3 years ago the content of the file could be even completely different. But as I understand, the rule is to put the year of creation and never change it.

src/app/ClioApplication.cpp
kuznetsss: Hm, maybe it is ok for backend to own cache? Here we create cache, pass it to backend and never use it again. So why not place it into backend but with an option to mock?

godexsoft: Initially i wanted the factory function to create the cache but that's not mockable and not really possible without using the heap. Cache is not some implementation detail - it's used across the system and should be injected in multiple components.. at the moment it's not this way simply because backend has the `cache` getters and we use `backend.cache()` in many places we should have injected the cache - this was always a mistake and a dirty hack that we wanted to get rid of. I don't think backend should own cache by any means.

kuznetsss (approved):
👍

Issue #1599, opened 2024-08-12
ETL ng NFT and book dir


godexsoft:
## Summary
For #1594 

Organize code that deals with NFTs and book dir in the existing ETL into corresponding plugins.
This task should be mostly refactoring and covering with tests.

PR #1967, opened 2025-03-18
feat: ETLng extensions

godexsoft:
For #1599 #1597 

This PR implements several extensions for ETLng.


src/data/LedgerCache.cpp
```diff
@@ -139,6 +176,29 @@ LedgerCache::get(ripple::uint256 const& key, uint32_t seq) const
     return {e->second.blob};
 }
 
+std::optional<Blob>
+LedgerCache::getDeleted(ripple::uint256 const& key, uint32_t seq) const
+{
+    if (disabled_)
+        return std::nullopt;
+
+    std::shared_lock const lck{mtx_};
```
kuznetsss, line 185: Could it be that this method is called so often so it blocks updating the cache?

kuznetsss (approved):
👍

src/data/LedgerCache.cpp
godexsoft: Should not be the case. This is only called for deleted objects on each ledger update 👍

PR #1969, opened 2025-03-19
fix: ripple_flag logic in account lines

### LINKED ISSUE #1968: Change ripple_peer and ripple_peer_flag to match rippled in account line
## Issue Description

Match logic in [rippled fix](https://github.com/XRPLF/rippled/pull/5345)

PeterChen13579:
fixes #1968

### DIFF
--- a/src/rpc/handlers/AccountLines.cpp
+++ b/src/rpc/handlers/AccountLines.cpp
@@ -102,6 +102,12 @@ AccountLinesHandler::addLine(
     line.qualityIn = lineQualityIn;
     line.qualityOut = lineQualityOut;
 
+    if (lineNoRipple)
+        line.noRipple = true;
+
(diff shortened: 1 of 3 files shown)


cindyyan317 (approved):
LGTM

src/rpc/handlers/AccountLines.cpp
```diff
@@ -102,6 +102,12 @@ AccountLinesHandler::addLine(
     line.qualityIn = lineQualityIn;
     line.qualityOut = lineQualityOut;
 
+    if (lineNoRipple)
```
godexsoft, line 105: Is the intention to check that the optional is not `nullopt` and unconditionally set `noRipple` to true if it is? This code does not consider nor set it to the actual value inside of the optional - just wanted to check that it is the intention.

PR #1976, opened 2025-03-26
feat: Nodes communication via DB

### LINKED ISSUE #1966: Nodes communication via DB
## Summary

Database could be used for Clio nodes to share useful information or current status.

kuznetsss:
Fixes #1966.


src/cluster/ClioNode.hpp
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @brief Represents a node in the cluster.
+ */
+struct ClioNode {
+    /**
+     * @brief The format of the time to store in the database.
+     */
+    static constexpr char const* kTIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ";
+
+    // enum class WriterRole {
```
PeterChen13579, line 40: Would this be used in the future? Also I think NodeRole would be a better name 😄

src/cluster/ClioNode.hpp
```diff
@@ -0,0 +1,58 @@
...
+    //     ReadOnly,
+    //     NotWriter,
+    //     Writer
+    // };
+
+    std::shared_ptr<boost::uuids::uuid> uuid;          ///< The UUID of the node.
+    std::chrono::system_clock::time_point updateTime;  ///< The time the data about the node was last updated.
+    bool isSelf;                                       ///< Whether this node is the current node.
```
PeterChen13579: I've been looking through the codebase, and I'm not entirely clear on the purpose of this `isSelf` member variable. (I only see it set and checked) Could you clarify its intended use? 🤔

src/data/BackendInterface.hpp
```diff
@@ -682,6 +692,15 @@ class BackendInterface {
...
     writeSuccessor(std::string&& key, std::uint32_t seq, std::string&& successor) = 0;
 
+    /**
+     * @brief Write a node message. Used by ClusterCommunicationService
+     *
+     * @param uuid The UUID of the node
+     * @param message The message to write
+     */
+    virtual void
+    writeNodeMessage(boost::uuids::uuid const& uuid, std::string message) = 0;
```
PeterChen13579, line 702: Should there be limitations on what the message can be written? (I'm thinking the content of the message should be restricted as it's primary case is to tell the other nodes which one is writer or how to select a writer.)

src/data/CassandraBackend.hpp
```diff
@@ -878,6 +880,22 @@ class BasicCassandraBackend : public BackendInterface {
...
+    std::expected<std::vector<std::pair<boost::uuids::uuid, std::string>>, std::string>
+    fetchClioNodesData(boost::asio::yield_context yield) const override
+    {
+        auto const readResult = executor_.read(yield, schema_->selectClioNodesData);
+        if (not readResult)
+            return std::unexpected{readResult.error().message()};
+
+        std::vector<std::pair<boost::uuids::uuid, std::string>> result;
+
+        for (auto [uuid, message] : extract<boost::uuids::uuid, std::string>(*readResult)) {
```
PeterChen13579, line 892: nit: remove `{}` for  `for loop`

PeterChen13579 (commented):
Looks great 🚀  
Just a few questions

src/cluster/ClioNode.hpp
kuznetsss: Yeah, this will be used in the future. NodeRole sounds too general in case there will be other roles.

src/data/CassandraBackend.hpp
kuznetsss: Loop without `{}` may create bugs in the future. We don't have a strong requirement for that, so I would prefer to leave it

src/data/BackendInterface.hpp
kuznetsss: From Backend's perspective, it doesn't matter what to publish. The purpose of this function is only to put  arguments into the related table. Caller is responsible to use the right data

src/cluster/ClioNode.hpp
kuznetsss: Yeah, good idea. Originally, I thought it will be useful. Removed.

PeterChen13579 (dismissed):
LGTM 👍

src/cluster/ClusterCommunicationService.cpp
```diff
@@ -0,0 +1,175 @@
...
+    });
+    return result;
+}
+
+void
+ClusterCommunicationService::doRead(boost::asio::yield_context yield)
+{
+    otherNodesData_.clear();
+
+    auto expectedResult = backend_->fetchClioNodesData(yield);
```
godexsoft: Can be const i believe

src/cluster/ClusterCommunicationServiceInterface.hpp
```diff
@@ -0,0 +1,52 @@
...
+class ClusterCommunicationServiceInterface {
+public:
+    virtual ~ClusterCommunicationServiceInterface() = default;
+
+    /**
+     * @brief Get the data of the current node.
+     *
+     * @return The data of the current node.
+     */
+    virtual ClioNode
```
godexsoft: These interface getters should probably be marked `[[nodiscard]]`

src/data/BackendInterface.hpp
```diff
@@ -568,6 +569,15 @@ class BackendInterface {
     virtual std::optional<std::string>
     fetchMigratorStatus(std::string const& migratorName, boost::asio::yield_context yield) const = 0;
 
+    /**
+     * @brief Fetches the data of all nodes in the cluster.
+     *
+     * @param yield The coroutine context
+     *@return The data of all nodes in the cluster.
+     */
+    virtual std::expected<std::vector<std::pair<boost::uuids::uuid, std::string>>, std::string>
```
godexsoft: Here `[[nodiscard]]` makes sense too

src/data/cassandra/Schema.hpp
```diff
@@ -282,6 +282,19 @@ class Schema {
...
 
+        statements.emplace_back(fmt::format(
+            R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  ( 
+                   node_id UUID,
+                   message TEXT,
+                   PRIMARY KEY (node_id)
+                  )
+            WITH default_time_to_live = 2
```
godexsoft: Fix alignment please

src/data/cassandra/Schema.hpp
```diff
@@ -489,6 +502,17 @@ class Schema {
             ));
         }();
 
+        PreparedStatement updateClioNodeMessage = [this]() {
+            return handle_.get().prepare(fmt::format(
+                R"(
+                UPDATE {}
+                   SET message = ?
+                WHERE node_id = ?
```
godexsoft: Alignment 🔢

src/data/cassandra/impl/Result.hpp
```diff
@@ -99,6 +101,14 @@ extractColumn(CassRow const* row, std::size_t idx)
         auto const rc = cass_value_get_int64(cass_row_get_column(row, idx), &out);
         throwErrorIfNeeded(rc, "Extract int64");
         output = static_cast<DecayedType>(out);
+    } else if constexpr (std::is_convertible_v<DecayedType, boost::uuids::uuid>) {
+        CassUuid uuid;
+        auto const rc = cass_value_get_uuid(cass_row_get_column(row, idx), &uuid);
+        throwErrorIfNeeded(rc, "Extract uuid");
+        std::string uuidStr(CASS_UUID_STRING_LENGTH, '0');
+        cass_uuid_string(uuid, uuidStr.data());
+        uuidStr.pop_back();  // remove the last \0 character
```
godexsoft, line 110: Is there really no cleaner way to do this?

src/util/TimeUtils.cpp
```diff
@@ -38,6 +40,18 @@ systemTpFromUtcStr(std::string const& dateStr, std::string const& format)
...
 }
 
+[[nodiscard]] std::string
+systemTpToUtcStr(std::chrono::system_clock::time_point const& tp, std::string const& format)
+{
+    auto const timeT = std::chrono::system_clock::to_time_t(tp);
+    std::tm timeStruct{};
+    gmtime_r(&timeT, &timeStruct);
+
+    std::ostringstream oss;
```
godexsoft: Please check if there is something in libfmt we could use instead 👍

godexsoft (changes requested):
Nice 👍 Let's add `[[nodiscard]]` where applicable (on interface level) 🔢

src/data/cassandra/impl/Result.hpp
kuznetsss: `CASS_UUID_STRING_LENGTH` is 37 sybmols: 36 for uuid and one for `\0`. I think `cass_uuid_string()` expects a buffer with the length `CASS_UUID_STRING_LENGTH`. So probably there is no better way.

godexsoft (approved):
🚀

PR #1981, opened 2025-03-28
fix: Fix ssl in new webserver

### LINKED ISSUE #1980: SSL is broken in ng web server
## Issue Description

When using new web server (turned off by default so users are not affected) an attempt to use SSL/TLS (`https` and `wss`) returns error while plain connection `http` and `ws` work fine.

kuznetsss:
Fixes #1980.

An SSL handshake was missing and WsConnection should be build from stream not socket because in case SSL connection stream already completed SSL handshake.


src/web/ng/impl/WsConnection.hpp
```diff
@@ -189,25 +172,24 @@ class WsConnection : public WsConnectionBase {
...
-    boost::asio::ssl::context& sslContext,
     util::TagDecoratorFactory const& tagDecoratorFactory,
     boost::asio::yield_context yield
-);
+)
+{
+    auto connection = std::make_unique<WsConnection<StreamType>>(
+        std::forward<StreamType>(stream), std::move(ip), std::move(buffer), std::move(request), tagDecoratorFactory
+    );
+    auto maybeError = connection->performHandshake(yield);
```
godexsoft, line 189: Could be const i suppose 👍

tests/unit/web/ng/impl/HttpConnectionTests.cpp
```diff
@@ -300,8 +300,7 @@ TEST_F(HttpConnectionTests, Upgrade)
         [&]() { ASSERT_TRUE(expectedResult.has_value()) << expectedResult.error().message(); }();
         [&]() { ASSERT_TRUE(expectedResult.value()); }();
 
-        std::optional<boost::asio::ssl::context> sslContext;
-        auto expectedWsConnection = connection.upgrade(sslContext, tagDecoratorFactory_, yield);
+        auto expectedWsConnection = connection.upgrade(tagDecoratorFactory_, yield);
```
godexsoft, line 303: Can be const by the looks of it

godexsoft (approved):
Few suggestions, no blockers though 👍

src/web/ng/impl/WsConnection.hpp
kuznetsss: I think clang-tidy will fix const automatically where possible

PR #1997, opened 2025-04-11
feat: More efficient cache

### LINKED ISSUE #1473: More efficient forwarding cache
## Summary

If we are updating an entry in forwarding cache, all other similar requests could wait for the entry instead of doing more requests to rippled.

## Motivation

Currently when an entry is outdated in forwarded cache all the requests are being forwarded to rippled. So cache is not working for the time we are waiting for response from rippled. Blocking other similar requests from being forwarded and wait for the cache entry to be ready will decrease load on rippled.

Thanks to @sophiax851 for this idea!

kuznetsss:
Fixes #1473.


tests/unit/util/BlockingCacheTests.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+        ASSERT_FALSE(result.has_value());
+        EXPECT_EQ(result.error(), error);
+    });
+}
+
+TEST_F(BlockingCacheTest, asyncGet_EmptyCacheUpdateSuccessButVerifierRejects)
+{
+    runSpawn([&](boost::asio::yield_context yield) {
+        std::expected<int, std::string> result;
```
PeterChen13579: Instead of` int, std::string`, maybe `ValueType, ErrorType`?

src/etl/LoadBalancer.cpp
```diff
@@ -364,4 +362,40 @@ LoadBalancer::chooseForwardingSource()
     }
 }
 
+std::expected<boost::json::object, rpc::CombinedError>
+LoadBalancer::forwardToRippledImpl(
+    boost::json::object const& request,
+    std::optional<std::string> const& clientIp,
+    bool isAdmin,
```
PeterChen13579: const

src/etlng/LoadBalancer.cpp
```diff
@@ -368,4 +366,40 @@ LoadBalancer::chooseForwardingSource()
...
+
+    std::optional<boost::json::object> response;
+    rpc::ClioError error = rpc::ClioError::EtlConnectionError;
+    while (numAttempts < sources_.size()) {
+        auto res = sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield);
+        if (res) {
+            response = std::move(res).value();
+            break;
+        }
+        error = std::max(error, res.error());  // Choose the best result between all sources
```
PeterChen13579, line 391: A bit confused why std::max is used here 🤔 . Do we prioritize errors with higher value?

PeterChen13579 (commented):
LGTM. Found a few small things and probably can add [[nodiscard]] to a few functions.

src/etl/LoadBalancer.cpp
```diff
@@ -364,4 +362,40 @@ LoadBalancer::chooseForwardingSource()
...
+    boost::asio::yield_context yield
+)
+{
+    ASSERT(not sources_.empty(), "ETL sources must be configured to forward requests.");
+    std::size_t sourceIdx = util::Random::uniform(0ul, sources_.size() - 1);
+
+    auto numAttempts = 0u;
+
+    auto xUserValue = isAdmin ? kADMIN_FORWARDING_X_USER_VALUE : kUSER_FORWARDING_X_USER_VALUE;
```
godexsoft, line 377: Could be `const`

src/etl/LoadBalancer.cpp
```diff
@@ -364,4 +362,40 @@ LoadBalancer::chooseForwardingSource()
...
+
+    std::optional<boost::json::object> response;
+    rpc::ClioError error = rpc::ClioError::EtlConnectionError;
+    while (numAttempts < sources_.size()) {
+        auto res = sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield);
+        if (res) {
+            response = std::move(res).value();
+            break;
+        }
+        error = std::max(error, res.error());  // Choose the best result between all sources
```
godexsoft, line 387: What does it mean for an error to be "the best"?

src/etlng/LoadBalancer.cpp
```diff
@@ -368,4 +366,40 @@ LoadBalancer::chooseForwardingSource()
...
+    bool isAdmin,
+    boost::asio::yield_context yield
+)
+{
+    ASSERT(not sources_.empty(), "ETL sources must be configured to forward requests.");
+    std::size_t sourceIdx = util::Random::uniform(0ul, sources_.size() - 1);
+
+    auto numAttempts = 0u;
+
+    auto xUserValue = isAdmin ? kADMIN_FORWARDING_X_USER_VALUE : kUSER_FORWARDING_X_USER_VALUE;
```
godexsoft, line 381: Same `const`

src/rpc/RPCEngine.hpp
```diff
@@ -317,6 +315,53 @@ class RPCEngine {
     {
         return handlerProvider_->contains(method) || forwardingProxy_.isProxied(method);
     }
+
+    Result
+    buildResponseImpl(web::Context const& ctx)
+    {
+        if (backend_->isTooBusy()) {
+            LOG(log_.error()) << "Database is too busy. Rejecting request";
+            notifyTooBusy();  // TODO: should we add ctx.method if we have it?
```
godexsoft, line 324: Let's decide what to do with this before we merge

src/util/BlockingCache.hpp
```diff
@@ -0,0 +1,221 @@
...
+ * @tparam ErrorType The type of error that can occur during updates
+ */
+template <typename ValueType, typename ErrorType>
+    requires(not std::same_as<ValueType, ErrorType>)
+class BlockingCache {
+public:
+    /**
+     * @brief Possible states of the cache
+     */
+    enum class State { Empty, Updating, Full };
```
godexsoft: Perhaps `Empty` and `Full` are not the right names for the actual states they represent?
`Empty` seems to mean `NotCached` and `Full` seems to mean `Cached` or `Available` if i'm reading the code correctly. Wdyt?

src/util/BlockingCache.hpp
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * @param yield The ASIO yield context for coroutine suspension
+     * @param updater Function to generate a new value
+     * @param verifier Function to validate whether a value should be cached
+     * @return std::expected<ValueType, ErrorType> The new value or an error
+     *
+     * Initiates a cache update operation regardless of current state.
+     * If another update is already in progress, waits for it to complete.
+     */
+    std::expected<ValueType, ErrorType>
```
godexsoft: Don't we want `nodiscard` here too or is it implied that the user may just want to update but does not care about the result?

tests/unit/util/BlockingCacheTests.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include <string>
+
+struct BlockingCacheTest : SyncAsioContextTest {
+    using ErrorType = std::string;
+    using ValueType = int;
+    using Cache = util::BlockingCache<ValueType, ErrorType>;
+    using MockUpdater = StrictMock<MockFunction<std::expected<ValueType, ErrorType>(boost::asio::yield_context)>>;
+    using MockVerifier = StrictMock<MockFunction<bool(ValueType const&)>>;
+
+    std::unique_ptr<Cache> cache = std::make_unique<Cache>();
```
godexsoft, line 49: I thought we decided we want members to be protected (and have an underscore at the end)?

godexsoft (commented):
Nice 👍 Leaving a few questions. Overall looks great.

src/etlng/LoadBalancer.cpp
kuznetsss: This logic was moved from another place, I didn't change it in this PR.
ETL related errors defined in `rpc::ClioError` are in the order of progress: connect -> request -> response. So taking the max of the error we are choosing an error with the furtherst progress.

src/etl/LoadBalancer.cpp
kuznetsss: It is a string_view, so const doesn't make any difference here.

src/etl/LoadBalancer.cpp
kuznetsss: This logic was moved from another place, I didn't change it in this PR.
ETL related errors defined in `rpc::ClioError` are in the order of progress: connect -> request -> response. So taking the max of the error we are choosing an error with the furthest progress.

src/rpc/RPCEngine.hpp
kuznetsss: This is a moved old code. I think it is not related to this PR.

src/util/BlockingCache.hpp
kuznetsss: Good point, I will add `nodiscard`.

src/util/BlockingCache.hpp
kuznetsss: I meant to read it as "cache is empty" or "cache is full". I think `Cached` or `NotCached` is rather related to an entry but not to cache.

tests/unit/util/BlockingCacheTests.cpp
kuznetsss: I thought we didn't come to a strict rule about it and both variants are acceptable. But we could choose a style.

godexsoft: Prefer consistency if possible but overall you are not wrong 👍

src/etl/LoadBalancer.cpp
```diff
@@ -365,7 +365,7 @@ std::expected<boost::json::object, rpc::CombinedError>
 LoadBalancer::forwardToRippledImpl(
     boost::json::object const& request,
     std::optional<std::string> const& clientIp,
-    bool isAdmin,
+    bool const isAdmin,
```
godexsoft, line 368: Yikes.

godexsoft (approved):

PR #2001, opened 2025-04-14
fix: CTID issue

### LINKED ISSUE #1998: Ctid was added to all RPC tx transactions and error code change in rippled responses, should be modified on clio responses to maintain the parity.
## Issue Description

- `tx` with unassociated `network_id` used to return `unknown` error, it was changed to `wrongNetwork`
- CTID was addded on all RPC `tx` transactions.

Please refer the rippled [PR] for additional details(https://github.com/XRPLF/rippled/commit/c4308b216f0cc7b1a33a720b4d38bbc850c4a2f1).

PeterChen13579:
fixes #1998

### DIFF
--- a/src/etl/ETLState.hpp
+++ b/src/etl/ETLState.hpp
@@ -38,7 +38,12 @@ namespace etl {
  * @brief This class is responsible for fetching and storing the state of the ETL information, such as the network id
  */
 struct ETLState {
-    std::optional<uint32_t> networkID;
+    /*
+     * NOTE: Rippled NetworkID: Mainnet = 0; Testnet = 1; Devnet = 2
+     * However, if rippled is running on neither of these (ie. standalone mode) rippled will show 0, but
+     * is not included in the stateOpt response. Must manually add it here.
+     */
+    std::optional<uint32_t> networkID{0};
 
     /**
      * @brief Fetch the ETL state from the rippled server
--- a/src/etl/ETLState.hpp
+++ b/src/etl/ETLState.hpp
@@ -40,10 +40,10 @@ namespace etl {
 struct ETLState {
     /*
      * NOTE: Rippled NetworkID: Mainnet = 0; Testnet = 1; Devnet = 2
-     * However, if rippled is running on neither of these (ie. standalone mode) rippled will show 0, but
+     * However, if rippled is running on neither of these (ie. standalone mode) rippled will default to 0, but
      * is not included in the stateOpt response. Must manually add it here.
      */
-    std::optional<uint32_t> networkID{0};
+    uint32_t networkID{0};
 
     /**
      * @brief Fetch the ETL state from the rippled server
--- a/src/etl/LoadBalancer.cpp
+++ b/src/etl/LoadBalancer.cpp
@@ -142,12 +142,11 @@ LoadBalancer::LoadBalancer(
         if (!stateOpt) {
             LOG(log_.warn()) << "Failed to fetch ETL state from source = " << source->toString()
                              << " Please check the configuration and network";
-        } else if (etlState_ && etlState_->networkID && stateOpt->networkID &&
-                   etlState_->networkID != stateOpt->networkID) {
+        } else if (etlState_ && etlState_->networkID != stateOpt->networkID) {
             checkOnETLFailure(fmt::format(
                 "ETL sources must be on the same network. Source network id = {} does not match others network id = {}",
-                *(stateOpt->networkID),
-                *(etlState_->networkID)
+                stateOpt->networkID,
+                etlState_->networkID
             ));
         } else {
             etlState_ = stateOpt;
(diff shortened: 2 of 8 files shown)


tests/unit/rpc/handlers/TxTests.cpp
```diff
@@ -829,6 +833,7 @@ TEST_F(RPCTxTest, CTIDNotMatch)
         ASSERT_FALSE(output);
 
         auto const err = rpc::makeError(output.result.error());
+        // TODO: [PETER13579] change to "WrongNetwork" error string when new libxrpl gets merged
```
godexsoft: It's better to create an issue and link it here in the TODO so that there is less chance to forget about it 👍

src/etl/LoadBalancer.cpp
```diff
@@ -151,6 +151,13 @@ LoadBalancer::LoadBalancer(
             ));
         } else {
             etlState_ = stateOpt;
+            /*
+             * NOTE: Rippled NetworkID: Mainnet = 0; Testnet = 1; Devnet = 2
+             * However, if rippled is running on neither of these (ie. standalone mode) rippled will show 0, but
+             * is not included in the stateOpt response. Must manually add it here.
+             */
+            if (!etlState_->networkID.has_value())
+                etlState_->networkID = 0;
```
godexsoft: You could also default `networkID` to 0 in `ETLState` even though it's an optional. That would move this comment to ETLState (not sure if that's better) and allow you to rollback the multiple one line changes that create `ETLState{.networkID = 0}`.
There also might be some constant in libXRPL that we can use to get this `0` for mainnet.

godexsoft (commented):
Not sure what approach is better but left some ideas 👍

src/etl/ETLState.hpp
```diff
@@ -38,7 +38,12 @@ namespace etl {
  * @brief This class is responsible for fetching and storing the state of the ETL information, such as the network id
  */
 struct ETLState {
-    std::optional<uint32_t> networkID;
+    /*
+     * NOTE: Rippled NetworkID: Mainnet = 0; Testnet = 1; Devnet = 2
+     * However, if rippled is running on neither of these (ie. standalone mode) rippled will show 0, but
+     * is not included in the stateOpt response. Must manually add it here.
+     */
+    std::optional<uint32_t> networkID{0};
```
kuznetsss: If you go with this fix, probably it doesn't need to be optional anymore.

kuznetsss:
Shouldn't this be fixed on rippled's side? It looks incorrect that standalone id and mainnet id are the same.

PeterChen13579:
> Shouldn't this be fixed on rippled's side? It looks incorrect that standalone id and mainnet id are the same.

Just found out in Rippled, the default value of `network_id` is [0 if not set](https://github.com/XRPLF/rippled/blob/405f4613d8d3af5e4443aaa8f952c557a7c7a54b/src/xrpld/core/Config.h#L156). 
So I think this is intentional

tests/unit/etl/ETLStateTests.cpp
```diff
@@ -57,8 +57,8 @@ TEST_F(ETLStateTest, NetworkIdValid)
     EXPECT_CALL(source, forwardToRippled).WillOnce(Return(json.as_object()));
     auto const state = etl::ETLState::fetchETLStateFromSource(source);
     ASSERT_TRUE(state.has_value());
-    ASSERT_TRUE(state->networkID.has_value());
-    EXPECT_EQ(state->networkID.value(), 12);
+    ASSERT_TRUE(state->networkID);
```
mathbunnyru: This doesn't make sense for a non-optional and is wrong for `mainnet`

PeterChen13579: The `ASSERT_TRUE` can be removed 👍 . For this test, we are just checking that `network_id` value that is returned is the same, the value doesn't have to be mainnet/devnet/or testnets. 12 is valid for network_id as it is `uint32_t` in both rippled and Clio.

tests/unit/rpc/handlers/TxTests.cpp
```diff
@@ -71,6 +71,7 @@ constexpr auto kDEFAULT_OUT1 = R"({
     "TakerPays": "300",
     "TransactionType": "OfferCreate",
     "hash": "2E2FBAAFF767227FE4381C4BE9855986A6B9F96C62F6E443731AB36F7BBB8A08",
+    "ctid":"C000006400640000",
```
godexsoft: Please keep the formatting (add space) 👍

tests/unit/rpc/handlers/TxTests.cpp
```diff
@@ -578,6 +582,7 @@ TEST_F(RPCTxTest, MintNFT)
             "SigningPubKey": "74657374",
             "TransactionType": "NFTokenMint",
             "hash": "C74463F49CFDCBEF3E9902672719918CDE5042DC7E7660BEBD1D1105C4B6DFF4",
+            "ctid":"C000006400000000",
```
godexsoft: Same formatting issue

godexsoft (changes requested):
Just some nits 👍

godexsoft (approved):
🚀

PR #2005, opened 2025-04-16
fix: Fix incorrect requests logging

### LINKED ISSUE #2004: Log request instead of params
## Issue Description

Clio logs only part of requests. `requests` and `tag` should be passed here:
https://github.com/XRPLF/clio/blob/develop/src/web/RPCServerHandler.hpp#L200

kuznetsss:
Fixes #2004.

### DIFF
--- a/src/rpc/RPCHelpers.hpp
+++ b/src/rpc/RPCHelpers.hpp
@@ -745,7 +745,8 @@ decodeCTID(T const ctid) noexcept
  * @brief Log the duration of the request processing
  *
  * @tparam T The type of the duration
- * @param ctx The context of the request
+ * @param request The request to log
+ * @param tag The tag of the context of the request
  * @param dur The duration to log
  */
 template <typename T>
--- a/src/rpc/RPCHelpers.hpp
+++ b/src/rpc/RPCHelpers.hpp
@@ -30,6 +30,7 @@
 #include "rpc/Errors.hpp"
 #include "rpc/common/Types.hpp"
 #include "util/JsonUtils.hpp"
+#include "util/Taggable.hpp"
 #include "util/log/Logger.hpp"
 #include "web/Context.hpp"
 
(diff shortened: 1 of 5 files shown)


src/rpc/RPCHelpers.hpp
```diff
@@ -744,12 +745,13 @@ decodeCTID(T const ctid) noexcept
  * @brief Log the duration of the request processing
  *
  * @tparam T The type of the duration
- * @param ctx The context of the request
+ * @param request The request to log
+ * @param tag The tag of the context of the request
  * @param dur The duration to log
  */
 template <typename T>
```
godexsoft: While you are here, can we change `T` to `Duration` or `DurationType`? `T` is such a generic name.. 🙇

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -1166,3 +1174,71 @@ TEST_P(IsAdminCmdParameterTest, Test)
...
+        {
+        }
+    };
+    DummyTaggable taggable{tagFactory};
+};
+
+TEST_P(RPCHelpersLogDurationTest, LogDuration)
+{
+    auto const& tag = taggable.tag();
+    // TOOD: Update after https://github.com/XRPLF/clio/issues/1668
```
godexsoft: This is so going to be lost in time 😆 
What i suggest is let's create a new issue specifically for this change and make it dependant on 1668. Then we eventually will get to it in our list of issues and we will know that it depends on 1668 👍

godexsoft (changes requested):
Cool 🚀 Left a few suggestions 👍

tests/unit/rpc/RPCHelpersTests.cpp
kuznetsss: Created a child issue, updated TODO and left a comment in the issue to not forget.
https://github.com/XRPLF/clio/issues/2008

godexsoft (approved):
🚀

PR #2011, opened 2025-04-18
fix: Add CTID to all RPC's that includes transactions

### LINKED ISSUE #1998: Ctid was added to all RPC tx transactions and error code change in rippled responses, should be modified on clio responses to maintain the parity.

PeterChen13579:
fixes #1998


src/rpc/RPCHelpers.cpp
```diff
@@ -380,6 +380,24 @@ insertMPTIssuanceID(
...
 
+bool
+insertCTID(
+    boost::json::object& jsonObject,
+    uint32_t const ledgerSeq,
+    uint32_t const transactionID,
+    uint32_t const networkID
+)
+{
+    if (transactionID <= 0xFFFFU && ledgerSeq < 0x0FFF'FFFFUL && networkID <= 0xFFFFU) {
```
PeterChen13579: fyi logic taken from rippled: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/Tx.cpp#L179

src/app/ClioApplication.cpp
```diff
@@ -146,6 +146,10 @@ ClioApplication::run(bool const useNgWebServer)
     // ETL is responsible for writing and publishing to streams. In read-only mode, ETL only publishes
     auto etl = etl::ETLService::makeETLService(config_, ioc, backend, subscriptions, balancer, ledgers);
 
+    // inject networkID into subscriptions, as transaction feed require it to inject CTID in response
+    if (auto const state = etl->getETLState(); state)
+        subscriptions->setNetworkID(state->networkID);
+
```
kuznetsss: Maybe this should be done inside `makeETLService()`?

src/feed/SubscriptionManager.hpp
```diff
@@ -332,6 +333,21 @@ class SubscriptionManager : public SubscriptionManagerInterface {
      */
     boost::json::object
     report() const final;
+
+    /**
+     * @brief Set the networkID.
+     * @param networkID The network id to set.
+     */
+    void
+    setNetworkID(uint32_t const networkID) final;
```
kuznetsss: `const` doesn't do anything in declaration.

src/feed/impl/TransactionFeed.cpp
```diff
@@ -205,6 +206,11 @@ TransactionFeed::pub(
         rpc::insertDeliverMaxAlias(pubObj[txKey].as_object(), version);
         rpc::insertMPTIssuanceID(pubObj[JS(meta)].as_object(), tx, meta);
 
+        auto const& metaObj = pubObj[JS(meta)];
+        if (metaObj.is_object() && metaObj.as_object().contains("TransactionIndex") &&
+            metaObj.as_object().at("TransactionIndex").is_int64())
```
kuznetsss: Could this `if` be `false` here? Maybe we should replace it with `ASSERT()`?

src/feed/impl/TransactionFeed.hpp
```diff
@@ -182,12 +182,14 @@ class TransactionFeed {
...
      * @param backend The backend.
+     * @param networkID The network ID.
      */
     void
     pub(data::TransactionAndMetadata const& txMeta,
         ripple::LedgerHeader const& lgrInfo,
         std::shared_ptr<data::BackendInterface const> const& backend,
-        std::shared_ptr<data::AmendmentCenterInterface const> const& amendmentCenter);
+        std::shared_ptr<data::AmendmentCenterInterface const> const& amendmentCenter,
+        uint32_t const networkID = 0);
```
kuznetsss: Probably better to not put a default value here, because having default value we could miss a place where `networkID` should be passed and it will result in sending incorrect data.

src/rpc/RPCHelpers.cpp
kuznetsss: Better to add this comment to the code.

src/rpc/common/impl/HandlerProvider.cpp
```diff
@@ -86,7 +86,7 @@ ProductionHandlerProvider::ProductionHandlerProvider(
           {"account_nfts", {.handler = AccountNFTsHandler{backend}}},
           {"account_objects", {.handler = AccountObjectsHandler{backend}}},
           {"account_offers", {.handler = AccountOffersHandler{backend}}},
-          {"account_tx", {.handler = AccountTxHandler{backend}}},
+          {"account_tx", {.handler = AccountTxHandler{backend, etl}}},
```
kuznetsss, line 89: Maybe we could pass only `networkID` here? I assume it will not change ever in the same network, right?

src/rpc/handlers/Tx.hpp
```diff
@@ -37,6 +37,7 @@
 #include <boost/json/object.hpp>
 #include <boost/json/value.hpp>
 #include <boost/json/value_to.hpp>
+#include <fmt/core.h>
```
kuznetsss, line 40: This seems to be left from debug.

src/feed/impl/TransactionFeed.cpp
PeterChen13579: I think `TransactionIndex` can be false. The `metaObj.is_object` can be `ASSERT() `yea.

src/rpc/common/impl/HandlerProvider.cpp
PeterChen13579: Yea, that's true, unless somehow in the future, rippled allows to swap networks without stopping the server.

src/rpc/handlers/Tx.hpp
PeterChen13579: fmt was already used [in this file](https://github.com/XRPLF/clio/blob/develop/src/rpc/handlers/Tx.hpp#L175) before 😄

tests/unit/feed/TransactionFeedTests.cpp
```diff
@@ -191,10 +191,10 @@ TEST_F(FeedTransactionTest, SubTransactionV1)
 
     EXPECT_CALL(*mockSessionPtr, apiSubversion).WillOnce(testing::Return(1));
     EXPECT_CALL(*mockSessionPtr, send(sharedStringJsonEq(kTRAN_V1)));
-    testFeedPtr->pub(trans1, ledgerHeader, backend_, mockAmendmentCenterPtr_);
+    testFeedPtr->pub(trans1, ledgerHeader, backend_, mockAmendmentCenterPtr_, 0);
```
kuznetsss: Better to created a named constant in the fixture to make it more readable.

src/rpc/RPCHelpers.cpp
```diff
@@ -380,6 +380,25 @@ insertMPTIssuanceID(
     return false;
 }
 
+bool
+insertCTID(
+    boost::json::object& jsonObject,
+    uint32_t const ledgerSeq,
+    uint32_t const transactionID,
+    uint32_t const networkID
+)
```
kuznetsss: Maybe it would be better to change the signature of the function, so the usage will look:
```cpp
jsonObject[JS(ctid)] = makeCTID(ledgerSeq, transactionID, networkID);
```

PeterChen13579:
TODO: We can optimize passing required values into each handler instead of `etl_` https://github.com/XRPLF/clio/issues/2021

src/rpc/RPCHelpers.cpp
PeterChen13579: Now thinking about it more, I think this function isn't really needed as there was already a `makeCTID` and I can change the logic of that function instead as it matches rippled as well

tests/unit/feed/TransactionFeedTests.cpp
PeterChen13579: Made it constexpr.

kuznetsss (approved):
👍

PR #2014, opened 2025-04-21
feat: Cache FetchLedgerBySeq

### LINKED ISSUE #1758: Cache result of fetchLedgerBySequence
This is called almost by every API. We shall cache this response to relieve DB stress. 
The request of current ledger header would cause DB shards access unevenly.

PeterChen13579:
fixes #1758


src/data/CassandraBackend.hpp
```diff
@@ -263,11 +275,19 @@ class BasicCassandraBackend : public BackendInterface {
...
+                return lock->ledger;
+        }
+
         auto const res = executor_.read(yield, schema_->selectLedgerBySeq, sequence);
         if (res) {
             if (auto const& result = res.value(); result) {
                 if (auto const maybeValue = result.template get<std::vector<unsigned char>>(); maybeValue) {
-                    return util::deserializeHeader(ripple::makeSlice(*maybeValue));
+                    auto const header = util::deserializeHeader(ripple::makeSlice(*maybeValue));
+                    ledgerCache_.put(FetchLedgerCache::CacheEntry{header, sequence});
```
kuznetsss, line 282: Probably, you should check that `header` has value before putting it into cache.

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,99 @@
...
+ * fetched via `FetchLedgerBySeq` (often triggered by RPC commands),
+ * the result can be stored here. Subsequent requests for the same ledger
+ * sequence can proceed to retrieve the header from this cache, avoiding unnecessary
+ * database reads and improving performance.
+ */
+class FetchLedgerCache {
+public:
+    FetchLedgerCache()
+    {
+        mutexPtr_ = std::make_unique<util::Mutex<std::optional<CacheEntry>, std::shared_mutex>>();
```
kuznetsss: Better to move it to parameter initialisation (after `:`).

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,99 @@
...
+        uint32_t seq;
+    };
+
+    /**
+     * @brief Put CacheEntry into mutex container.
+     *
+     * @param cacheEntry The Cache to store into mutex container.
+     */
+    void
+    put(CacheEntry const& cacheEntry) const
```
kuznetsss: The implementation could be moved to cpp file.

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -78,6 +82,13 @@ using namespace prometheus;
 
 using namespace data::cassandra;
 
+constexpr auto kRAWHEADER =
```
kuznetsss: Better to make it a static field inside fixture.

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1302,8 +1307,69 @@ TEST_F(BackendCassandraTest, CacheIntegration)
     ASSERT_EQ(done, true);
 }
 
+TEST_F(BackendCassandraTest, CacheFetchLedgerBySeq)
+{
+    std::atomic_bool done = false;
+    std::optional<boost::asio::io_context::work> work;
+    work.emplace(ctx_);
+
+    boost::asio::spawn(ctx_, [this, &done, &work](boost::asio::yield_context yield) {
```
kuznetsss: We have `runSpawn()` for this.

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1302,8 +1307,69 @@ TEST_F(BackendCassandraTest, CacheIntegration)
...
+        auto const testLedgerSeq = lgrInfo.seq;
+        ASSERT_TRUE(backend_->finishWrites(lgrInfo.seq));
+
+        // use mock cache
+        using TestBackendType = data::cassandra::BasicCassandraBackend<
+            SettingsProvider,
+            data::cassandra::impl::DefaultExecutionStrategy<>,
+            MockLedgerHeaderCache&>;
+
+        MockLedgerHeaderCache mockCache{};
```
kuznetsss: If you make it `StrictMock`, I assume the test will fail.

tests/integration/data/cassandra/BackendTests.cpp
PeterChen13579: Didn't know constexpr static field can exist in classes 😄

src/data/CassandraBackend.hpp
PeterChen13579: `util::deserializeHeader` returns a ledgerHeader, so it guarantees to have value

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include <xrpl/protocol/LedgerHeader.h>
+
+#include <cstdint>
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <shared_mutex>
+
+namespace data::cassandra {
```
xinmeng2011: Maybe not Cassandra namespace

src/data/CassandraBackend.hpp
```diff
@@ -71,21 +72,25 @@ namespace data::cassandra {
...
     Schema<SettingsProviderType> schema_;
-
     std::atomic_uint32_t ledgerSequence_ = 0u;
 
 protected:
     Handle handle_;
 
     // have to be mutable because BackendInterface constness :(
     mutable ExecutionStrategyType executor_;
+    FetchLedgerCacheType ledgerCache_;
```
xinmeng2011: I feel this cache should not sit in a specific database impl.

src/data/CassandraBackend.hpp
```diff
@@ -94,13 +100,20 @@ class BasicCassandraBackend : public BackendInterface {
...
+        data::LedgerCacheInterface& cache,
+        bool readOnly,
+        FetchLedgerCacheType ledgerCache = FetchLedgerCache{}
+    )
         : BackendInterface(cache)
         , settingsProvider_{std::move(settingsProvider)}
         , schema_{settingsProvider_}
         , handle_{settingsProvider_.getSettings()}
         , executor_{settingsProvider_.getSettings(), handle_}
+        , ledgerCache_{std::forward<FetchLedgerCacheType>(ledgerCache)}
```
mathbunnyru: Should be std::move, not std::forward here.
FetchLedgerCacheType is not a template for a ctor, and if you take it by value here, then you can move it

src/data/CassandraBackend.hpp
```diff
@@ -263,11 +276,19 @@ class BasicCassandraBackend : public BackendInterface {
     std::optional<ripple::LedgerHeader>
     fetchLedgerBySequence(std::uint32_t const sequence, boost::asio::yield_context yield) const override
     {
+        {
+            auto const lock = ledgerCache_.get();
+            if (lock.has_value() && lock->seq == sequence)
+                return lock->ledger;
+        }
```
mathbunnyru:
```suggestion
        if (auto const lock = ledgerCache_.get(); lock.has_value() && lock->seq == sequence)
            return lock->ledger;
```

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @brief Struct to store ledger header cache entry and the sequence it belongs to
+     */
+    struct CacheEntry {
+        ripple::LedgerHeader ledger;
+        uint32_t seq;
+    };
+
+    /**
+     * @brief Put CacheEntry into mutex container.
```
mathbunnyru: Maybe instead of `mutex` container, we should say `thread-safe container`?
Not just here, but everywhere in this PR.

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * @brief Comparing CacheEntry. Used in testing for EXPECT_CALL
+ *
+ * @param lhs A CacheEntry
+ * @param rhs Another CacheEntry
+ * @return true if two CacheEntry is the same, false otherwise
+ */
+inline bool
+operator==(FetchLedgerCache::CacheEntry const& lhs, FetchLedgerCache::CacheEntry const& rhs)
+{
```
mathbunnyru: Afaik, this should be better implemented inside the class.
The reason is quite interesting.
Let's say you have `if (a == b)` and it doesn't compile (`a` and `b` are some another type)

If it's a free function like here, compiler will give you an error that it tried this function and it didn't work.
If it's part of the class, then it won't try to use this function at all.

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1302,8 +1307,59 @@ TEST_F(BackendCassandraTest, CacheIntegration)
...
+        using TestBackendType = data::cassandra::BasicCassandraBackend<
+            SettingsProvider,
+            data::cassandra::impl::DefaultExecutionStrategy<>,
+            MockLedgerHeaderCache&>;
+
+        testing::StrictMock<MockLedgerHeaderCache> mockCache{};
+
+        backend_ = std::make_unique<TestBackendType>(settingsProvider_, cache_, false, mockCache);
+
+        auto* backendPtr = dynamic_cast<TestBackendType*>(backend_.get());
```
mathbunnyru: backend_ is of type `unique_ptr<TestBackendType>`, you don't need dynamic_cast (or any cast) here at all.

mathbunnyru (changes requested):
Made a few small comments

src/data/LedgerHeaderCache.hpp
PeterChen13579: Interesting. Thanks for the info 😄

src/data/CassandraBackend.hpp
PeterChen13579: The reason why it is forward here is because we need to pass [mockLedgerCache type by reference](https://github.com/XRPLF/clio/pull/2014/files#diff-0e7a9870bf4ec500fae67b0d935d5f805dd8de7e949401f82bd12c7e7bc8907cR1321), as mock disables copy constructor. so using std::move here doesn't work.

tests/integration/data/cassandra/BackendTests.cpp
PeterChen13579: Thanks for the catch!

src/data/CassandraBackend.hpp
```diff
@@ -277,8 +277,7 @@ class BasicCassandraBackend : public BackendInterface {
     fetchLedgerBySequence(std::uint32_t const sequence, boost::asio::yield_context yield) const override
     {
         {
```
mathbunnyru, line 279: Don't need this anymore, I thought my suggestion was without this block

src/data/CassandraBackend.hpp
```diff
@@ -263,11 +276,18 @@ class BasicCassandraBackend : public BackendInterface {
     std::optional<ripple::LedgerHeader>
     fetchLedgerBySequence(std::uint32_t const sequence, boost::asio::yield_context yield) const override
     {
+        {
```
mathbunnyru: Don't need this anymore, I thought my suggestion was without this block

src/data/CassandraBackend.hpp
mathbunnyru: Something dodgy going on, maybe just remove `std::forward<FetchLedgerCacheType>`?

PeterChen13579: That doesn't work because mockLedgerCache is a reference type, and removing `std::forward` breaks the forwarding behavior. I think using `std::forward` is fine here; if the argument is an lvalue reference, it stays an lvalue; if it's an rvalue, it's perfectly forwarded as an rvalue.

src/data/CassandraBackend.hpp
```diff
@@ -71,21 +72,26 @@ namespace data::cassandra {
...
     Schema<SettingsProviderType> schema_;
-
     std::atomic_uint32_t ledgerSequence_ = 0u;
 
 protected:
     Handle handle_;
 
     // have to be mutable because BackendInterface constness :(
     mutable ExecutionStrategyType executor_;
+    // TODO: move to interface level
```
mathbunnyru, line 96: You sure you don't want to do this as part of this PR?

I this is the reason why we have to have both `data::LedgerCacheInterface& cache` and `FetchLedgerCacheType ledgerCache` in the same class? (it might be a bit confusing)

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,86 @@
...
+        }
+    };
+
+    /**
+     * @brief Put CacheEntry into thread-safe container
+     *
+     * @param cacheEntry The Cache to store into thread-safe container.
+     */
+    void
+    put(CacheEntry const& cacheEntry) const;
```
mathbunnyru: This shouldn't be marked `const`; because we use pointers, it is by the language grammar, but semantically it is not.

src/data/CassandraBackend.hpp
PeterChen13579: No, @kuznetsss  made a new cache class, so the cache I wrote here would be replaced anyways. We decided to leave as is.

src/data/LedgerHeaderCache.hpp
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @brief Read CacheEntry from thread-safe container.
+     *
+     * @return Optional CacheEntry, depending on if it exists in thread-safe container or not.
+     */
+    std::optional<CacheEntry>
+    get() const;
+
+private:
+    std::unique_ptr<util::Mutex<std::optional<CacheEntry>, std::shared_mutex>> mutexPtr_;
```
mathbunnyru: Why do we need `unique_ptr` here?

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1302,8 +1307,56 @@ TEST_F(BackendCassandraTest, CacheIntegration)
...
 
+TEST_F(BackendCassandraTest, CacheFetchLedgerBySeq)
+{
+    runSpawn([&](boost::asio::yield_context yield) {
+        auto rawHeaderBlob = hexStringToBinaryString(kRAWHEADER);
+        ripple::LedgerHeader lgrInfo = util::deserializeHeader(ripple::makeSlice(rawHeaderBlob));
+
+        backend_->writeLedger(lgrInfo, std::move(rawHeaderBlob));
+        auto const testLedgerSeq = lgrInfo.seq;
+        ASSERT_TRUE(backend_->finishWrites(lgrInfo.seq));
```
mathbunnyru, line 1331: I'm sorry, I don't get it, should this be here?
You overwrite `backend_` few lines below, maybe you should have a separate test class?

src/data/LedgerHeaderCache.hpp
kuznetsss: Because mutex is not movable.

src/data/CassandraBackend.hpp
mathbunnyru: OK

src/data/LedgerHeaderCache.hpp
PeterChen13579: This is `unique_ptr` here, because of the `std::forward` statement [here](https://github.com/XRPLF/clio/pull/2014/files#diff-5d2896a019d0a49496dd02d914bf5586df50681630e91d5246c3d4e3a9b488e2R116). Without unique_ptr, it won't be movable

tests/integration/data/cassandra/BackendTests.cpp
PeterChen13579: This is here because I first write a ledger to database with sequence `lgrInfo.seq`. Then, I want to test the cache works by fetching for this particular ledger. I guess I can override the `backend_` at the beginning of the test though, instead of later.

src/data/LedgerHeaderCache.hpp
mathbunnyru: I don't think we actually require our LedgerHeaderCache class to be moveable.

We only require it because we combined 2 different constructors into one, and create a temporary object which we move from, if no param was passed, this is what happens now:

```
(FetchLedgerCacheType ledgerCache = FetchLedgerCache{}):
ledgerCache_{std::forward<FetchLedgerCacheType>(ledgerCache)}
```

What I propose. There are 2 cases:
1. Default ctor (prod) for `FetchLedgerCacheType`: then we can simply create `ledgerCache_` in-place with no params.
2. We pass a reference to a mock (test) - then we should just save a reference.

We could probably write 2 constructors, get rid of std::forward and get rid of unique_ptr - this will make access to the header cache faster.
And move the ctor body to some function (and call it from these 2 ctors).

I might be wrong and we might move CassandraBackend somewhere though.

src/data/LedgerHeaderCache.hpp
PeterChen13579: I remember why const is needed here now. [fetchledgerbySeq](https://github.com/XRPLF/clio/pull/2014/files#diff-5d2896a019d0a49496dd02d914bf5586df50681630e91d5246c3d4e3a9b488e2R277) is marked as const, and since the function is used in the method, it has to const here. That's also why pointers are used.

src/data/LedgerHeaderCache.hpp
PeterChen13579: Thanks for the suggestion.I think this would eliminate the pointer overhead, which is great. However, we used pointers here partly because fetchLedgerBySequence is marked as const, and I believe we wanted to preserve that. Removing the const would require changes to several other functions, which we wanted to avoid. @kuznetsss WDYT?

kuznetsss: You could mark the cache as `mutable`, so should be fine.

src/data/LedgerHeaderCache.hpp
mathbunnyru: I think better approach is to:

- not mark this method `const` - semantically cache changes when you put something to it
- still mark mutex here `mutable` - to be able to make `get` const
- mark the `ledgerCache_` mutable in the class where you use it, this fixes your problem

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1294,8 +1299,56 @@ TEST_F(BackendCassandraTest, CacheIntegration)
...
+    runSpawn([&](boost::asio::yield_context yield) {
+        using TestBackendType = data::cassandra::BasicCassandraBackend<
+            SettingsProvider,
+            data::cassandra::impl::DefaultExecutionStrategy<>,
+            MockLedgerHeaderCache&>;
+
+        testing::StrictMock<MockLedgerHeaderCache> mockCache{};
+
+        // use mock cache
+        backend_ = std::make_unique<TestBackendType>(settingsProvider_, cache_, false, mockCache);
```
mathbunnyru: I still think this should be another test class, @kuznetsss wdyt?

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1294,8 +1299,56 @@ TEST_F(BackendCassandraTest, CacheIntegration)
...
+            // Second call: should return from cache
+            auto const ledger = backend_->fetchLedgerBySequence(testLedgerSeq, yield);
+            ASSERT_TRUE(ledger.has_value());
+            EXPECT_EQ(ledger->seq, lgrInfo.seq);
+        }
+    });
+}
+
 struct BackendCassandraNodeMessageTest : BackendCassandraTest {
-    boost::uuids::random_generator generateUuid{};
```
mathbunnyru, line 1298: Why this change is needed?

tests/common/util/MockLedgerHeaderCache.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <gmock/gmock.h>
+#include <xrpl/protocol/LedgerHeader.h>
+
+#include <optional>
+
+struct MockLedgerHeaderCache {
+    MockLedgerHeaderCache() = default;
+    using CacheEntry = data::FetchLedgerCache::CacheEntry;
+
+    MOCK_METHOD(void, put, (CacheEntry), ());
```
mathbunnyru, line 33: Now it's inconsistent - in your implemenation above it says `const, here it's not const.
I explained how to make it non-const and don't change lots of code above.

tests/integration/data/cassandra/BackendTests.cpp
PeterChen13579: clang tidy complains about it for me

mathbunnyru: That's strange, to be honest, but I don't mind.

tests/integration/data/cassandra/BackendTests.cpp
mathbunnyru: Made it another class

tests/integration/data/cassandra/BackendTests.cpp
mathbunnyru: Made a separate test class

src/data/CassandraBackend.hpp
```diff
@@ -94,43 +103,40 @@ class BasicCassandraBackend : public BackendInterface {
...
-                "Failed to prepare the statements: {}; readOnly: {}. ReadOnly should be turned off or another Clio "
-                "node with write access to DB should be started first.",
-                ex.what(),
-                readOnly
-            );
-            LOG(log_.error()) << error;
-            throw std::runtime_error(error);
-        }
+        initializeDatabase(readOnly);
+    }
```
mathbunnyru: You don't need this ctor anymore.

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1294,8 +1299,65 @@ TEST_F(BackendCassandraTest, CacheIntegration)
     ASSERT_EQ(done, true);
 }
 
+class MockCacheClassTest : public BackendCassandraTest {
```
mathbunnyru: I highly suggest making current `BackendCassandraTest` a `BackendCassandraTestBase` class where you don't have `backend_` at all, and then making two simple classes `BackendCassandraTest` and `CacheBackendCassandraTest` that derive from it and define `backend_` field.
This way there will be no overwrite of `backend_` at all.

mathbunnyru (commented):
Overall, looks better, I would remove ctor which is no longer needed and create a hierarchy of test classes where no private fields are overdefined.

tests/integration/data/cassandra/BackendTests.cpp
mathbunnyru: I don't like `MockCacheClassTest` because it's too general, btw, that's why I proposed `CacheBackendCassandraTest` here.

mathbunnyru (approved):
LGTM 🚀

PR #2018, opened 2025-04-23
chore: Run dependabot for all local actions

mathbunnyru:


.github/dependabot.yml
```diff
@@ -1,16 +1,209 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
```
mathbunnyru, line 4: This is not enough to update local actions.
For example, we use `docker/build-push-action@v5` in `.github/actions/build_docker_image/action.yml` even though v6 is available

.github/dependabot.yml
```diff
@@ -1,16 +1,209 @@
...
+      time: 04:00
+      timezone: Etc/GMT
     reviewers:
-      - "godexsoft"
-      - "kuznetsss"
-      - "PeterChen13579"
+      - godexsoft
+      - kuznetsss
+      - PeterChen13579
+      - mathbunnyru
```
mathbunnyru, line 14: Added myself to the list

.github/dependabot.yml
```diff
@@ -1,16 +1,209 @@
...
+      day: monday
+      time: 04:00
+      timezone: Etc/GMT
+    reviewers:
+      - godexsoft
+      - kuznetsss
+      - PeterChen13579
+      - mathbunnyru
+    commit-message:
+      prefix: "[CI] "
```
kuznetsss: Could you please update prefix everywhere to `ci: [DEPENDABOT] `?

mathbunnyru: Sure, done

kuznetsss (approved):

.github/dependabot.yml
bthomee: I would highly recommend to create a follow-up PR and reference a GitHub team. I assume Dependabot supports that. Sergey has the permissions to create a team.

PR #2019, opened 2025-04-23
chore: Pin external actions using hashes

mathbunnyru:
Note: I used exactly the version which are already being used (didn't update anything).
Updates will come when we merge https://github.com/XRPLF/clio/pull/2018


kuznetsss (approved):

mathbunnyru:
The reason for this change - if someone whose lib we use decides to change their action in a malicious way, we will automatically start using it without even knowing something has changed.

I don't pin github provided actions, because if they are compromised, we are probably done already, and it creates too much noise.

PR #2020, opened 2025-04-23
refactor: Add str() method to BaseTagDecorator

### LINKED ISSUE #2008: Make tag easily convertible to string
## Summary

Currently to convert `util::BasicTagDecorator` to string we have to use `std::stringstream`. It would be convenient to have a better way (probably a method) to get a string from the decorator.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2008

### DIFF
--- a/src/util/Taggable.hpp
+++ b/src/util/Taggable.hpp
@@ -96,6 +96,18 @@ public:
         decorator.decorate(os);
         return os;
     }
+
+    /**
+     * @brief Gets the string representation of the tag.
+     *
+     * @return The string representation of the tag
+     */
+    std::string str() const
+    {
+        std::ostringstream oss;
+        decorate(oss);
+        return std::move(oss).str();
+    }
 };
 
 /**
--- a/src/util/Taggable.hpp
+++ b/src/util/Taggable.hpp
@@ -102,7 +102,8 @@ public:
      *
      * @return The string representation of the tag
      */
-    std::string str() const
+    std::string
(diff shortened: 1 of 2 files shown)


tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -1204,19 +1204,13 @@ struct RPCHelpersLogDurationTest : LoggerFixture, testing::WithParamInterface<RP
 TEST_P(RPCHelpersLogDurationTest, LogDuration)
 {
     auto const& tag = taggable.tag();
-    // TOOD: Update in https://github.com/XRPLF/clio/issues/2008
```
mathbunnyru, line 1207: I saw the typo and thought it would be easier to fix the issue rather than fix the typo 😆

src/util/Taggable.hpp
```diff
@@ -96,6 +96,18 @@ class BaseTagDecorator {
         decorator.decorate(os);
         return os;
     }
+
+    /**
+     * @brief Gets the string representation of the tag.
+     *
+     * @return The string representation of the tag
+     */
+    std::string str() const
```
kuznetsss: Maybe lets give it some more verbose name, for example `toString()`?

mathbunnyru: Sure, thanks 👍

kuznetsss (approved):

PR #2023, opened 2025-04-23
refactor: Compare error_code with enum values, not hardcoded ints

### LINKED ISSUE #2002: Update test Error codes
## Summary

There's a couple of places where we do: 
EXPECT_EQ(err.at("error_code").as_uint64(), //number);
it's better to replace the number with the actual `error_code` string

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2002

### DIFF
--- a/tests/unit/app/WebHandlersTests.cpp
+++ b/tests/unit/app/WebHandlersTests.cpp
@@ -181,7 +181,7 @@ TEST_F(RequestHandlerTest, DosguardRateLimited_Http)
 
         auto const body = boost::json::parse(httpResponse.body()).as_object();
         EXPECT_EQ(body.at("error").as_string(), "slowDown");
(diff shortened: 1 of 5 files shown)


mathbunnyru:
I think the best way to go is to `error_code` with enum value (instead of hardcoded int).

I grepped `.at("error_code")` and fixed all the occurrences.

PeterChen13579:
> I think the best way to go is to `error_code` with enum value (instead of hardcoded int).
> 
> I grepped `.at("error_code")` and fixed all the occurrences.

I think the reason why it is not an enum value is because it is not guaranteed the error code int values will increase sequentially. At least, that is what I remember in the case of rippled

PeterChen13579 (approved):
LGTM

PR #2029, opened 2025-04-24
style: Use pre-commit tool and add simple config

mathbunnyru:
I started with really simple pre-commit hooks and will add more on top.

Important files:
- `.pre-commit-config.yaml` - the config for pre-commit
- `.github/workflows/pre-commit.yml` - runs pre-commit hooks in branches and `develop` 
- `.github/workflows/pre-commit-autoupdate.yml` - autoupdates pre-commit hooks once in a month


.github/actions/build_docker_image/action.yml
```diff
@@ -38,6 +38,8 @@ runs:
           password: ${{ env.GITHUB_TOKEN }}
 
       - uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 #v3.6.0
+        with:
+          cache-image: false
```
mathbunnyru, line 42: I got the failure unrelated to my change: https://github.com/XRPLF/clio/actions/runs/14641449795/job/41090823407
I think it happens during running this action and on caching the image, so let's try not to cache it.

kuznetsss (approved):

PR #2037, opened 2025-04-25
chore: Remove git-lfs

mathbunnyru:
`git-lfs` has been introduced more than a year ago: https://github.com/XRPLF/clio/pull/1210

Since then no new files have been added:
```
git-lfs ls-files
16e6f0a44b * docs/img/doxygen-docs-output.png
d1faa73951 * docs/img/xrpl-logo.svg
```

So, I don't see why we should keep it.

It has many downsides though:
- project is more difficult to contribute to (I had to manually install `git-lfs` and learn what it does to start contributing)
- we have to copy-paste code to check if `git-lfs is available`
- and this code is in bash
- it also makes it more difficult to introduce pre-commit as our tool for hooks

So I propose to remove `git-lfs` from the repo


kuznetsss:
As I remember users don't need to have git-lfs installed unless they want to work with the files from git-lfs, so I don't see any problem with it.

docs/img/doxygen-docs-output.png
kuznetsss, line 1: These images are required for documentation: https://xrplf.github.io/clio/

mathbunnyru: These files are added back to the repo as regular files

kuznetsss: Oh, they were modified, not deleted, sorry.

I think it is not a good practice to store images in repository.

mathbunnyru: As far as I understand, git-lfs is to store large files to not bloat the repo.
Could you tell me, what's the downside of having 2 small images in the repo?

Also, nothing prevents from committing images to the repo before this change.

kuznetsss: We decided to add git-lfs to not put images into the repo and we didn't have any problem with it.

I'm against this change, let's see what @godexsoft thinks.

> Could you tell me, what's the downside of having 2 small images in the repo?
<details>
<summary>
I will paste AI answer </summary>

```
Storing large binary files (like images, videos, audio files, compiled executables, large datasets, PDFs, etc.) directly in a standard Git repository is generally considered bad practice for several key reasons:   

Repository Bloat:

Git Stores Full Copies: Git is designed primarily for text files. For text, it can efficiently store changes (deltas) between versions. However, for most binary files, even a small change (like re-saving an image with slightly different compression or modifying a single pixel) can change the entire file's byte sequence.
Inefficient Delta Compression: Git's delta compression often doesn't work well on binaries. As a result, each time you commit a modified binary file, Git often stores a near-complete new copy of that entire file within the repository's history (.git directory).   
Exponential Growth: This means your repository size can grow incredibly quickly, especially if binaries are updated frequently. A 10MB image modified 10 times could easily add close to 100MB to the repository history.   
Slow Performance:

Cloning and Fetching: Every user who clones or fetches from the repository has to download the entire history, including every version of every binary file ever committed. Large repositories take significantly longer to clone and fetch, consuming more bandwidth.
Local Operations: Operations like switching branches (git checkout), checking status (git status), and garbage collection (git gc) can become noticeably slower as Git has to process more data.
Ineffective Diffing:

Understanding Changes: git diff is excellent for showing line-by-line changes in text files, making code reviews straightforward. For binary files, git diff usually just tells you that the files differ, not how they differ in a meaningful way. You can't easily review visual changes to an image or structural changes in a compiled file directly within Git.
Example: git diff output for a binary might just be: Binary files a/logo.png and b/logo.png differ
Merging Difficulties:

No Automatic Merging: Git cannot automatically merge concurrent changes made to the same binary file on different branches. If two people modify the same image, a merge conflict is almost guaranteed.
Manual Conflict Resolution: Resolving such conflicts typically involves manually choosing one version of the file (--ours or --theirs) and completely discarding the changes from the other branch. There's no way for Git to intelligently combine the changes within the binary file itself.
 Storage Costs: Larger repositories consume more disk space locally and on hosting platforms (like GitHub, GitLab, Bitbucket), potentially leading to increased costs.   

The Solution: Git LFS (Large File Storage)

To handle large binary files correctly with Git, the standard solution is Git LFS.

How it works: Instead of storing the large binary file directly in the Git history, Git LFS stores a small "pointer file" (a text file containing a hash and location of the actual file). The actual large binary file is uploaded to a separate LFS storage area (often associated with your Git host).   
Benefits:
Keeps the core Git repository small and fast.
Users only download the versions of the large files they actually need for the commit they check out, not the entire history of every large file.   
Diffing and merging still apply only to the small pointer files within Git, avoiding the main issues (though managing changes to the binaries themselves still requires external tools or careful workflow).
In summary, while Git can technically store binary files, it's highly inefficient and leads to bloated repositories, slow performance, and difficulties with diffing and merging. Using Git LFS is the recommended approach for managing large binary assets within a Git workflow.   
```
</details>

mathbunnyru: I think I can leave `git-lfs` as is, but just remove the hooks and rewrite them to `pre-commit`.
Doesn't seem to be too difficult.

mathbunnyru:
> As I remember users don't need to have git-lfs installed unless they want to work with the files from git-lfs, so I don't see any problem with it.

Unfortunately, it can cause problems during `git clone`.
That's why I had to install it.

```
~/D/oss > git clone git@github.com:XRPLF/clio.git
Cloning into 'clio'...
remote: Enumerating objects: 20675, done.
remote: Counting objects: 100% (315/315), done.
remote: Compressing objects: 100% (218/218), done.
remote: Total 20675 (delta 189), reused 97 (delta 97), pack-reused 20360 (from 3)
Receiving objects: 100% (20675/20675), 8.08 MiB | 11.78 MiB/s, done.
Resolving deltas: 100% (15802/15802), done.
git-lfs filter-process: git-lfs: command not found
fatal: the remote end hung up unexpectedly
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'

~/D/oss > echo $status
128
```

docs/img/doxygen-docs-output.png
godexsoft: Initially we added lfs because we were going to have a revamp of the docs/readmes and we were anticipating multiple new infographics and images. The revamp happened but did not involve multiple such files.
There was another issue with lfs that seems to be fixed now: iPhone github app would not show the icon in the readme because it's on lfs and the app did not support fetching it correctly. 
I can see how it could be nicer for a developer not to have to install git-lfs just to get two small image files but i'm not sure it's worth removing lfs support to simplify one package install. Also i though lfs was not strictly required 🤔

mathbunnyru: Thanks @kuznetsss @godexsoft,

I will try another approach, where git-lfs stays as is, but I remove pre-commit.sh like scripts and/or move it to pre-commit utility.

As it will be another approach, I'm closing this PR

PR #2038, opened 2025-04-25
style: Add markdownlint pre-commit hook

mathbunnyru:


docs/build-clio.md
```diff
@@ -23,9 +23,9 @@ Clio does not require anything other than `compiler.cppstd=20` in your (`~/.cona
 > [!NOTE]
 > Although Clio is built using C++23, it's required to set `compiler.cppstd=20` for the time being as some of Clio's dependencies are not yet capable of building under C++23.
 
-> Mac example:
+Mac example:
```
godexsoft: The `>` was making it into a nice block, not sure how it will look without. There may be a better way than `>` though, not sure.

godexsoft (dismissed):
Looks good to me. Let's consult @maria-robobug for the doc changes 👍

docs/build-clio.md
mathbunnyru: We're still creating a nice block, but after this change text is not part of the block (only the config), which in my opinion looks better:

Before: https://github.com/XRPLF/clio/blob/develop/docs/build-clio.md#conan-configuration

After: https://github.com/mathbunnyru/clio/blob/pre_commit_markdownlint/docs/build-clio.md#conan-configuration

godexsoft: I meant that the example text is in a block of its own 😄 
Let's ask @maria-robobug to review what looks better and go from there 👍

maria-robobug: I see no issue with removing the `>` here, also agree it reads better. If you really want to emphasize the text above the block maybe make it bold.

src/migration/README.md
```diff
@@ -8,7 +8,7 @@ Clio provides a migration command-line tool to migrate data in database.
 
 > Note: We need a **configuration file** to run the migration tool. This configuration file has the same format as the configuration file of the Clio server, ensuring consistency and ease of use. It reads the database configuration from the same session as the server's configuration, eliminating the need for separate setup or additional configuration files. Be aware that migration-specific configuration is under `.migration` session.
```
maria-robobug:
```suggestion
> [!NOTE]
> We need a **configuration file** to run the migration tool. This configuration file has the same format as the configuration file of the Clio server, ensuring consistency and ease of use. It reads the database configuration from the same session as the server's configuration, eliminating the need for separate setup or additional configuration files. Be aware that migration-specific configuration is under `.migration` session.
```

src/migration/README.md
```diff
@@ -51,7 +51,7 @@ Sometimes migrator isn't able to query the historical data by table's partition
 
 Most indexes are based on either ledger states or transactions. We provide the `objects` and `transactions` scanner. Developers only need to implement the callback function to receive the historical data. Please find the examples in `tests/integration/migration/cassandra/ExampleTransactionsMigrator.cpp` and `tests/integration/migration/cassandra/ExampleObjectsMigrator.cpp`.
 
-> **Note** The full table scanner splits the table into multiple ranges by token(https://opensource.docs.scylladb.com/stable/cql/functions.html#token). A few of rows maybe read 2 times if its token happens to be at the edge of ranges. **Deduplication is needed** in the callback function.
+> **Note** The full table scanner splits the table into multiple ranges by token(<https://opensource.docs.scylladb.com/stable/cql/functions.html#token>). A few of rows maybe read 2 times if its token happens to be at the edge of ranges. **Deduplication is needed** in the callback function.
```
maria-robobug:
```suggestion
> [!NOTE]
> The full table scanner splits the table into multiple ranges by token(<https://opensource.docs.scylladb.com/stable/cql/functions.html#token>). A few of rows maybe read 2 times if its token happens to be at the edge of ranges. **Deduplication is needed** in the callback function.
```

maria-robobug (dismissed):
Docs changes LGTM

docs/build-clio.md
mathbunnyru: Thanks, made it bold as well!

mathbunnyru:
> Docs changes LGTM

Thanks! Could you please approve again? Merged latest branch and made the suggested text bold.

maria-robobug (approved):

PR #2039, opened 2025-04-25
style: Add C++ pre-commit hooks

mathbunnyru:


docs/CMakeLists.txt
```diff
@@ -3,6 +3,6 @@ project(docs)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/../cmake/ClioVersion.cmake)
```
mathbunnyru, line 4: This file and file below were wrongfully not formatted by cmake-format, because it only formatted `cmake` folder

tools/snapshot/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
-# Build snapshot tool
-# Need to be installed before building: go protoc-gen-go protoc-gen-go-grpc
+# Build snapshot tool needed to be installed before building: go protoc-gen-go protoc-gen-go-grpc
```
mathbunnyru, line 1: Changed the line slightly

.pre-commit-config.yaml
```diff
@@ -30,3 +30,24 @@ repos:
     hooks:
       - id: prettier
         exclude: ^docs/doxygen-awesome-theme/
+
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v19.1.7
```
mathbunnyru, line 55: Used the same version as being used right now, that's why there is no diff

.pre-commit-config.yaml
```diff
@@ -30,3 +30,24 @@ repos:
...
+    hooks:
+      - id: cmake-format
+        additional_dependencies: [PyYAML]
+
+  - repo: local
+    hooks:
+      - id: check-no-h-files
+        name: No .h files
+        entry: There should be no .h files in this repository
+        language: fail
```
mathbunnyru, line 72: This is quite a nice way of checking that no files with this extension are present in the repo

.githooks/check-format
```diff
@@ -2,13 +2,13 @@
 
 # Note: This script is intended to be run from the root of the repository.
 #
-# This script checks the format of the code and cmake files.
+# This script checks the format of the C++ code.
```
godexsoft, line 5: I suppose check-format is no longer a suitable name for this script. Perhaps we should rename it to something like `fix-local-includes`?

.pre-commit-config.yaml
godexsoft: Cool, yeah. There is a recent discussion in the committee about standardising on `.hpp` or something that is NOT `.h` for C++ headers. Frankly i doubt it will be standardised but definitely good to prevent `.h` in our own repos and code.

docs/CMakeLists.txt
```diff
@@ -3,6 +3,6 @@ project(docs)
...
 
-# Generate `docs` target for doxygen documentation
-# Note: use `cmake --build . --target docs` from your `build` directory to generate the documentation
+# Generate `docs` target for doxygen documentation Note: use `cmake --build . --target docs` from your `build` directory
```
godexsoft: This is a bad change imo.

godexsoft (changes requested):
Overall i like the change. Leaving a few nits.

.githooks/check-format
mathbunnyru: I didn't do it to show diff properly (not delete+add file).
When I switch to pre-commit hooks only, I will move this file to a separate dir, an example: https://github.com/XRPLF/clio/pull/2046/files

.pre-commit-config.yaml
mathbunnyru: Me too 🙂

docs/CMakeLists.txt
mathbunnyru: I agree, was able to disable it using `cmake-format: off` and `cmake-format: on`

tools/snapshot/CMakeLists.txt
mathbunnyru: This look fine to me, so let's keep it this way.

godexsoft (dismissed):

godexsoft (approved):
👍

PR #2041, opened 2025-04-25
style: Add typos pre-commit hook

mathbunnyru:


_typos.toml
```diff
@@ -0,0 +1,23 @@
+[default.extend-identifiers]
```
godexsoft: Oh, this is going to be painful potentially. Is there no way to disable checking long strings that are clearly not real words?

godexsoft (dismissed):
Pretty cool 🚀 I'm only worried about the hashes that we need to maintain in the toml file and also, did you verify all the golang changes work? `TestUnavailableMethods` seems unused entirely.

_typos.toml
mathbunnyru: Great idea, I'll see if it's possible.
I agree it's easier and less error-prone than maintaining a list of exceptions.

mathbunnyru:
> Pretty cool 🚀 I'm only worried about the hashes that we need to maintain in the toml file and also, did you verify all the golang changes work? `TestUnavailableMethods` seems unused entirely.

I think it doesn't have to be used or mentioned anywhere, because it's a name of the test.
So, it runs when `go test` runs

_typos.toml
mathbunnyru: I was able to write few regexes, which allow not to hardcode this list

godexsoft (approved):

PR #2042, opened 2025-04-25
chore: Specify conan profile explicitly and don't override it

mathbunnyru:


.github/workflows/build.yml
```diff
@@ -71,6 +71,7 @@ jobs:
             code_coverage: false
             static: true
           - os: macos15
+            conan_profile: default_apple_clang
```
mathbunnyru, line 74: This change serves several purposes:
1. If default clang version changes, the profile name will still be correct.
2. If we change the runner, it will also stay correct
3. We don't have to patch the name for conan profile in our action

.github/workflows/nightly.yml
```diff
@@ -16,21 +16,24 @@ jobs:
...
           - os: heavy
+            conan_profile: gcc
             build_type: Debug
             static: true
             container: '{ "image": "rippleci/clio_ci:latest" }'
     uses: ./.github/workflows/build_impl.yml
     with:
       runs_on: ${{ matrix.os }}
       container: ${{ matrix.container }}
-      conan_profile: gcc
```
mathbunnyru, line 33: This is the reason why we had to patch the name and this value is not what actually happens on Mac, it's easy to make a mistake and use wrong variable

godexsoft (approved):
Looks reasonable 👍

Issue #1779, opened 2024-12-12
Create releases in CI


kuznetsss:
## Summary

We should create releases automatically on each tag pushed. [git-cliff](https://git-cliff.org/) could be used to generate nice changelog automatically.

PR #2043, opened 2025-04-25
feat: Introduce `release_impl.yml` workflow to automatically create rel…

mathbunnyru:
…eases

Work on: https://github.com/XRPLF/clio/issues/1779

Features:
- works in PRs (everything but actually pushing a release)
- supports custom release notes header
- removes previous release for nightly releases

The idea is that creating releases should be automatic or almost automatic and should work more or less the same for nightly and actual releases.

I haven't yet implemented automatic release process for the actual release (and don't plan to do it here), but this PR will make it much easier to implement it in the future.
I suggest deferring discussion of how actual releases should work till the next PR.


godexsoft (approved):
👍

PR #2044, opened 2025-04-25
fix: Add missing paths to update_docker_ci.yml

mathbunnyru:


.github/workflows/update_docker_ci.yml
```diff
@@ -5,6 +5,7 @@ on:
       - "docker/ci/**"
       - "docker/compilers/**"
       - .github/workflows/update_docker_ci.yml
+      - ".github/actions/build_docker_image/**"
```
mathbunnyru, line 8: This is where all the work happens, so it is definitely worth to trigger the build if something changes there

godexsoft (approved):
👍

PR #2045, opened 2025-04-25
chore: Capitalize Dockerfile to add support of tools

mathbunnyru:
For example, there is not syntax highlighting in GitHub, because file is named `dockerfile` (and not Dockerfile), for example: https://github.com/XRPLF/clio/blob/release/2.4.0/docker/ci/dockerfile

I also grep'ed dockerfile in repo and fixed some other places


godexsoft (approved):
🚀

PR #2046, opened 2025-04-26
style: Apply `go fmt` to go code

mathbunnyru:
~I will add pre-commit hook later if I find a good one~

Found a nice repo, but it is no longer maintained: https://github.com/dnephin/pre-commit-golang
So, I implemented the check as a local hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,3 +30,12 @@ repos:
     hooks:
       - id: prettier
         exclude: ^docs/doxygen-awesome-theme/
+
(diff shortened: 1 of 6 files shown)


godexsoft (approved):
🚀 
Assuming that the pre-commit tool takes care of all deps and the user need not care about having golang installed.

mathbunnyru:
> 🚀 Assuming that the pre-commit tool takes care of all deps and the user need not care about having golang installed.

Yes, I checked it works without having `go` installed 👍

PR #2048, opened 2025-04-27
test: Add `build_and_test` reusable workflow

mathbunnyru:
Reasons:
- workflows become simpler
- no code duplication
- more difficult to build and not test something
- no wrong workflow dependencies: before, clang test was waiting for gcc build, which doesn't make sense

~**I don't want to use `steps.conan_export_output.outputs.conan_profile`, so mac build is expected to fail on artifact download till we merge https://github.com/XRPLF/clio/pull/2042**~


.github/workflows/build.yml
```diff
@@ -37,11 +37,10 @@ jobs:
           ./.githooks/check-docs
         shell: bash
 
-  build:
-    name: Build
-    needs:
-      - check_format
-      - check_docs
```
mathbunnyru: I think there is no reason for build/test to wait for these checks.

When I switch to pre-commit hooks completely, there will be no check_format at all and it will be even in a different workflow file

.github/workflows/build.yml
```diff
@@ -74,63 +73,33 @@ jobs:
...
-      integration_tests: true
-      clio_server: true
-
-  test:
-    name: Run Tests
-    needs: build
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
```
mathbunnyru, line 63: I think I fixed a bug here, because this matrix only has 4 scenarios, but we're building for 5

.github/workflows/build_impl.yml
```diff
@@ -6,13 +6,11 @@ on:
         description: Runner to run the job on
         required: true
         type: string
-        default: heavy
```
mathbunnyru, line 9: These options are already specified everywhere, and default values make it more difficult to reason on the call side what's going on, so let's remove where it's easy

.github/workflows/nightly.yml
```diff
@@ -26,82 +28,35 @@ jobs:
...
-            conan_profile: gcc
-            build_type: Debug
-            container:
-              image: rippleci/clio_ci:latest
-            integration_tests: true
-    runs-on: [self-hosted, "${{ matrix.os }}"]
-    container: ${{ matrix.container }}
-
-    services:
-      scylladb:
```
mathbunnyru, line 71: We don't need to have scylladb for unit tests, that's why in test_impl the service is only run for integration tests

.github/workflows/nightly.yml
```diff
@@ -26,82 +28,35 @@ jobs:
...
+          static: ${{ matrix.static }}
+          unit_tests: true
+          integration_tests: true
+          clio_server: true
+          disable_cache: true
+
+      - name: Test
+        uses: ./.github/workflows/test_impl.yml
         with:
-          name: clio_integration_tests_${{ runner.os }}_${{ matrix.build_type }}_${{ matrix.conan_profile }}
```
mathbunnyru, line 95: Integration tests were waiting for unit tests, but there is no need for that.
In a new design it doesn't happen because these are separate jobs

.github/workflows/test_impl.yml
```diff
@@ -0,0 +1,132 @@
...
+        if: ${{ steps.check_report.outputs.found_report == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.compiler }}_${{ inputs.sanitizer }}_report
+          path: .sanitizer-report/*
+          include-hidden-files: true
+
+      # TODO: enable when we have fixed all currently existing issues from sanitizers
+      - name: Create an issue
+        if: ${{ false && steps.check_report.outputs.found_report == 'true' }}
```
mathbunnyru, line 87: I uncommented this code to make sure it stays correct (code in comments tends to become incorrect over time).
This way the only change we will need (I hope) to enable issue creating in the future is to remove `false` here

.github/workflows/test_impl.yml
```diff
@@ -0,0 +1,132 @@
...
+            Reports are available as artifacts.
+
+  integration_tests:
+    name: Integration test ${{ inputs.container != '' && 'in container' || 'natively' }}
+    runs-on: ${{ inputs.runs_on }}
+    container: ${{ inputs.container != '' && fromJson(inputs.container) || null }}
+
+    services:
+      scylladb:
+        image: ${{ (inputs.integration_tests) && 'scylladb/scylla' || '' }}
```
mathbunnyru: I think we can simplify this and move the if to the job part. Will do.

.github/workflows/build.yml
```diff
@@ -74,63 +73,35 @@ jobs:
...
-
     steps:
-      - name: Clean workdir
-        if: ${{ runner.os == 'macOS' }}
-        uses: kuznetsss/workspace-cleanup@80b9863b45562c148927c3d53621ef354e5ae7ce #v1.0
+      - uses: actions/checkout@v4
 
-      - uses: actions/download-artifact@v4
+      - name: Build
+        uses: ./.github/workflows/build_impl.yml
```
mathbunnyru: Unfortunately, calling a reusable workflow doesn't work as a step of another workflow. I will think how to overcome this, maybe I will make our reusable workflows actions

mathbunnyru: Added a new workflow which simply runs both of build and test

mathbunnyru:
After this change it's much easier to see what has and hasn't been run in GitHub, especially with integration tests.

Before:
Screenshot 2025-04-27 at 13 28 05

After:
Screenshot 2025-04-27 at 13 28 19

mathbunnyru:
Now we don't upload unit_tests bin when code coverage is enabled.
`if: ${{ inputs.unit_tests && !inputs.code_coverage }}`

That's why we build gcc Debug build for code coverage, but don't upload binaries.
This PR won't change this behaviour, but after merging it I'll work on refactoring when and how we:
- upload artifacts
- name them
- run code coverage

.github/workflows/test_impl.yml
```diff
@@ -0,0 +1,128 @@
...
+            echo "found_report=true" >> $GITHUB_OUTPUT
+          else
+            echo "found_report=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Upload sanitizer report
+        if: ${{ steps.check_report.outputs.found_report == 'true' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.conan_profile }}_report
```
mathbunnyru, line 81: `inputs.conan_profile` already has information about sanitizer and compiler, so I slightly changed the artifact name.
The artifact name is now also uniform for builds with sanitizers and without sanitizers (previously, upload happened in different places)

.github/workflows/build.yml
godexsoft: There is no need to clog the CI and build anything if we don't even pass the pre-commit hooks, that's why we are waiting for these to succeed first.

mathbunnyru: Got your point, reverted this change

.github/workflows/nightly.yml
godexsoft: This was already only running for integration. If integration_tests was not set to true the image would resolve to empty string and the whole service would be discarded. It's a hack so i like your approach better anyway.

godexsoft (dismissed):
I like the direction this is going 👍 I saw there is an issue with artifact names in the mac portion (https://github.com/XRPLF/clio/actions/runs/14693922110/job/41233037563). I assume this is fixed or will be fixed in another PR.

.github/workflows/nightly.yml
mathbunnyru: I agree, but when integration_tests was set to true, it was also run for unit tests, which is not great.

It wasn't always running though for unit tests, only when integration_tests was set.

mathbunnyru:
> I like the direction this is going 👍 I saw there is an issue with artifact names in the mac portion (https://github.com/XRPLF/clio/actions/runs/14693922110/job/41233037563). I assume this is fixed or will be fixed in another PR.

Thanks!

Yes, I fixed it in another PR: https://github.com/XRPLF/clio/pull/2042, merged it to develop, and updated this branch.
This PR should be good to go now and if CI passes, I will merge it 👍

.github/workflows/build.yml
kuznetsss: No, tests for gcc debug run in build stage.

mathbunnyru: Ah, got it, thanks, as part of code coverage step.

I think there is no harm to run it with and without coverage.
And I will soon change the place coverage runs after this PR is merged, so I think we can keep it as is for now.

kuznetsss: What do you mean with and without coverage?
Tests compiled by gcc in Debug are always with coverage enabled. Also it takes 40 seconds to run them in debug, so it would be better to avoid running them twice.

.github/workflows/build.yml
```diff
@@ -52,30 +54,36 @@ jobs:
             container: '{ "image": "rippleci/clio_ci:latest" }'
             code_coverage: false
             static: true
+            integration_tests: true
```
kuznetsss: Why?
Integration tests were on only in nightly on purpose to make PR checks as fast as possible.

.github/workflows/nightly.yml
godexsoft: I think the idea was that we only call it with either integration_tests or unit_tests for any given run. I may be misremembering though 👍

.github/workflows/build.yml
godexsoft: Yeah, let's not have integration tests run other than in nightly.

mathbunnyru: Thank you, you're right, it's a bug, fixed it 👍

.github/workflows/build.yml
```diff
@@ -84,54 +87,11 @@ jobs:
       code_coverage: ${{ matrix.code_coverage }}
       static: ${{ matrix.static }}
       unit_tests: true
-      integration_tests: true
```
mathbunnyru, line 87: I will remove this from the build config, I think we can always build them, it's not a huge deal

mathbunnyru, line 54: We can't set this to true now, because it will also test integration tests.
@kuznetsss @godexsoft should I:
1. create separate options to build and to test unit/integration tests (it will be 4 options)
2. just always build unit and integration tests and remove options (and unit tests option) from the `build_impl.yml` config?

I think second option is better - it's simpler, building is not expensive, and we should always try to build under various configurations, but only run when we actually want to run.

.github/workflows/build.yml
mathbunnyru: Sorry, didn't explain myself correctly.

I meant that the tests for gcc debug after my PR will be run twice:
- as part of tests_impl.yml
- inside `build_impl.yml`

If you don't mind, I will change this in a separate PR after we merge this one, because I will also change the place we're running coverage (running it inside build process is not great, as currently mentioned in comments), so both issues will be fixed.

What do you think?

.github/workflows/build.yml
mathbunnyru: Discussed and decided to go with second option

.github/workflows/build.yml
kuznetsss: We shouldn't build and run tests with coverage twice, but it's ok to fix it in a separate PR.

kuznetsss (dismissed):

godexsoft (approved):
🚀

PR #2049, opened 2025-04-27
fix: Use ytanikin/pr-conventional-commits action for conventional com…

mathbunnyru:
…mits check

There are 2 actions with almost the same name from the same author:
- https://github.com/ytanikin/PRConventionalCommits: the one we use, not updated for 7 months
- https://github.com/ytanikin/pr-conventional-commits: seems to be more recent, has more stars and releases

I've merged this change to my develop branch in my fork and created a PR on to test it works: https://github.com/mathbunnyru/clio/pull/3

I will update the upstream issue and suggest author to archive the old action: https://github.com/ytanikin/pr-conventional-commits/issues/19

### DIFF
--- a/.github/workflows/check_pr_title.yml
+++ b/.github/workflows/check_pr_title.yml
@@ -1,4 +1,5 @@
 name: Check PR title
+
 on:
   pull_request:
     types: [opened, edited, reopened, synchronize]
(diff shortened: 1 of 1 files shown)


mathbunnyru:
I think now this action runs in a PR, that's why github token doesn't have any permissions, but when we merge it to develop, it might work fine.

godexsoft (approved):
Let's try the new one and see if it works in develop. Should be easy to fix/change if it does not.

PR #2050, opened 2025-04-27
chore: Don't use wandalen/wretry.action and update codecov/codecov-ac…

mathbunnyru:
…tion

### DIFF
--- a/.github/workflows/upload_coverage_report.yml
+++ b/.github/workflows/upload_coverage_report.yml
@@ -1,4 +1,5 @@
 name: Upload report
+
 on:
   workflow_dispatch:
   workflow_call:
@@ -11,10 +12,6 @@ jobs:
     name: Upload report
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
       - name: Download report artifact
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/upload_coverage_report.yml
```diff
@@ -11,10 +12,6 @@ jobs:
     name: Upload report
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
mathbunnyru: We don't use repo here, so we can omit cloning it

.github/workflows/upload_coverage_report.yml
```diff
@@ -23,13 +20,9 @@ jobs:
 
       - name: Upload coverage report
         if: ${{ hashFiles('build/coverage_report.xml') != '' }}
-        uses: wandalen/wretry.action@ffdd254f4eaf1562b8a2c66aeaa37f1ff2231179 #v3.7.3
```
mathbunnyru, line 26: There was a note that this extension was a bit dodgy: https://github.com/XRPLF/clio/pull/1822#issuecomment-2588152562

So let's remove it!
I'm updating codecov/codecov-action hoping it is more stable now and we don't need action retry at all.

.github/workflows/upload_coverage_report.yml
mathbunnyru: It is probably used used by codecov util, so will try to return it back.

mathbunnyru:
Seems to work the same way, but without extra action 🎉
And will also be automatically updated by dependabot when new versions are released.

godexsoft (approved):
We can try. If it starts failing again we know how to bring this back.

PR #2051, opened 2025-04-27
ci: Try to push to ghcr.io

mathbunnyru:
I don't think I can change where it's pushed and start using it in the same PR, so will try to first update the push part, and then change where the image is used. Only doing it for CI image to not break anything.


.github/workflows/update_docker_ci.yml
```diff
@@ -1,4 +1,5 @@
 name: Update CI docker image
+
```
mathbunnyru, line 2: Had to change this file to actually trigger the build, fixed the issue here: https://github.com/XRPLF/clio/pull/2044

.github/workflows/update_docker_ci.yml
```diff
@@ -27,7 +28,7 @@ jobs:
           DOCKERHUB_PW: ${{ secrets.DOCKERHUB_PW }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          image_name: rippleci/clio_ci
+          image_name: ghcr.io/xrplf/clio-ci
```
kuznetsss: Maybe you should change tags instead?
https://docs.docker.com/build/ci/github-actions/push-multi-registries/

.github/actions/build_docker_image/action.yml
```diff
@@ -48,7 +50,7 @@ runs:
         images: ${{ inputs.image_name }}
         tags: ${{ inputs.tags }}
 
-    - name: Build and push
+    - name: Build and push to DockerHub and GitHub Container Registry
```
kuznetsss: Why was the name changed but implementation is still the same?

mathbunnyru: Will revert this, before I added 2 different steps, but then it was like I could just do it in one, so merged them and names as well

.github/workflows/update_docker_ci.yml
mathbunnyru: We're using `docker/metadata-action` and apply several tags, so I don't want to write down list of all combinations manually.
Bu I think I can provide several `images` values to this action and all tags will be applied to all `images`.
Thanks!

.github/workflows/update_docker_ci.yml
mathbunnyru: Merged it, so it will work better now

godexsoft (approved):
👍

mathbunnyru:
I checked that it produces expected tags, merging this one.
Hope it will be able to push the images to ghcr.

Screenshot 2025-04-28 at 19 03 26

PR #2056, opened 2025-04-28
fix: Don't use steps.conan.outputs.conan_profile as it doesn't exist …

mathbunnyru:
…anymore

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -1,4 +1,5 @@
 name: Clang-tidy check
+
 on:
   schedule:
     - cron: "0 9 * * 1-5"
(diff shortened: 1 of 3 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -42,12 +42,12 @@ jobs:
         with:
           conan_dir: ${{ env.CONAN_USER_HOME }}/.conan
           ccache_dir: ${{ env.CCACHE_DIR }}
-          conan_profile: ${{ steps.conan.outputs.conan_profile }}
+          conan_profile: clang
```
kuznetsss: But that's why it was originally created. Now you have to specify the same profile everywhere and make sure it is the same.
Maybe put it into some variable to fix duplication?

mathbunnyru: Thank you for suggestion, used env variable and removed explicit conan_profile value duplication! 👍

kuznetsss (approved):

PR #2058, opened 2025-04-29
chore: Run sanitizer tests in parallel

mathbunnyru:


mathbunnyru:
It doesn't seem to work faster, at least for tsan.
This PR is not that relevant, so closing for now.

PR #2066, opened 2025-04-30
fix: Upload coverage directly

mathbunnyru:


mathbunnyru:
I'm testing my hypothesis that coverage report value will change if we upload in the same place as we're building, not in a clean environment.

.codecov.yml
```diff
@@ -9,3 +9,9 @@ coverage:
       default:
         target: 20% # Need to bump this number https://docs.codecov.com/docs/commit-status#patch-status
         threshold: 2%
+
+ignore:
+  - "tests"
+  - "src/data/cassandra/"
+  - "src/data/CassandraBackend.hpp"
+  - "src/data/BackendFactory.*"
```
mathbunnyru, line 17: The thing is that if we run `codecov/codecov-action` action where build files exist, it reruns the gcovr, but without exclude list, which makes the coverage value different.

PR #2069, opened 2025-05-01
ci: Don't use concurrency with workflow_call

mathbunnyru:


.github/workflows/build_clio_docker_image.yml
```diff
@@ -38,11 +38,6 @@ on:
         description: Whether to strip clio binary
         default: true
 
-concurrency:
-  # Only cancel in-progress jobs or runs for the current workflow - matches against branch & tags
-  group: ${{ github.workflow }}-${{ github.ref }}
```
mathbunnyru, line 43: Another workflow calls this as workflow_call, in which case `${{ github.workflow }}` value is taken from the caller, which makes workflow want to cancel itself and we get a deadlock issue.

.github/workflows/clang-tidy.yml
```diff
@@ -9,7 +9,6 @@ on:
     paths:
       - .clang_tidy
       - .github/workflows/clang-tidy.yml
-  workflow_call:
```
mathbunnyru, line 12: This is not called from anywhere, so let's remove it.

godexsoft: It's called via `gh` in `clang-tidy_on_fix_merged.yml`.. i'm not 100% sure this `workflow_call` is needed but i suspect it is.

godexsoft (changes requested):
May need some adjustments 🔢

.github/workflows/clang-tidy.yml
mathbunnyru: We use `gh workflow run` which actually creates workflow_dispatch event.
I think someone tried to run `clang-tidy.yml` using `uses:` (workflow_call), but it doesn't work in a step, and this was left here.

godexsoft (approved):
👍

PR #2072, opened 2025-05-01
feat: Implement clang build analyzer

### LINKED ISSUE #936: Print build statistics in CI
We could use [ninjatracing](https://github.com/nico/ninjatracing) to convert ninja's build log into json and collect statistics of how long each file took to build. It would help us to find problematic files which should be refactored.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/936


godexsoft (approved):
👍

PR #2074, opened 2025-05-02
ci: Move clang build analyzer to nightly

mathbunnyru:


.github/workflows/nightly.yml
```diff
@@ -43,13 +43,44 @@ jobs:
       container: ${{ matrix.container }}
       conan_profile: ${{ matrix.conan_profile }}
       build_type: ${{ matrix.build_type }}
-      code_coverage: false
```
mathbunnyru, line 46: This flag doesn't exist anymore in the used workflow, need to remove it to make it work.

.github/workflows/build_impl.yml
```diff
@@ -115,8 +115,7 @@ jobs:
         run: |
           ClangBuildAnalyzer --all build/ build_time_report.bin
           ClangBuildAnalyzer --analyze build_time_report.bin > build_time_report.txt
-          # Running again to have colorized output
-          ClangBuildAnalyzer --analyze build_time_report.bin
```
mathbunnyru, line 119: I checked and this tool doesn't produce colorful output in GitHub Actions, so it's better not to run it twice.

kuznetsss (approved):
👍

PR #2077, opened 2025-05-02
fix: Flaky test in ETLng

### LINKED ISSUE #2075: `ExtractionModelNgTests.LedgerDataCopyableAndEquatable` test is flaky
https://github.com/XRPLF/clio/actions/runs/14794633695/job/41538889602

```
[ RUN      ] ExtractionModelNgTests.LedgerDataCopyableAndEquatable
/__w/clio/clio/tests/unit/etlng/ExtractionTests.cpp:96: Failure
Expected: (first) != (third), actual: 312-byte object ...
```

godexsoft:
Fixes a flaky test

Fix: https://github.com/XRPLF/clio/issues/2075

### DIFF
--- a/tests/common/util/TestObject.cpp
+++ b/tests/common/util/TestObject.cpp
@@ -89,6 +89,8 @@ createLedgerHeader(std::string_view ledgerHash, ripple::LedgerIndex seq, std::op
     ledgerHeader.seq = seq;
 
     if (age) {
+        // Note: be cautios of using small age values as the underlying NetClock precision is seconds
+        // and the small time difference may lead to comparison bugs
         auto const now = duration_cast<seconds>(system_clock::now().time_since_epoch());
         auto const closeTime = (now - seconds{age.value()}).count() - kRIPPLE_EPOCH_START;
         ledgerHeader.closeTime = ripple::NetClock::time_point{seconds{closeTime}};
--- a/tests/common/util/TestObject.cpp
+++ b/tests/common/util/TestObject.cpp
@@ -89,7 +89,7 @@ createLedgerHeader(std::string_view ledgerHash, ripple::LedgerIndex seq, std::op
     ledgerHeader.seq = seq;
 
     if (age) {
-        // Note: be cautios of using small age values as the underlying NetClock precision is seconds
+        // Note: be cautious of using age values close to each other as the underlying NetClock precision is seconds
         // and the small time difference may lead to comparison bugs
         auto const now = duration_cast<seconds>(system_clock::now().time_since_epoch());
         auto const closeTime = (now - seconds{age.value()}).count() - kRIPPLE_EPOCH_START;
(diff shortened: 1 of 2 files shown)


tests/common/util/TestObject.cpp
```diff
@@ -89,6 +89,8 @@ createLedgerHeader(std::string_view ledgerHash, ripple::LedgerIndex seq, std::op
     ledgerHeader.seq = seq;
 
     if (age) {
+        // Note: be cautios of using small age values as the underlying NetClock precision is seconds
+        // and the small time difference may lead to comparison bugs
```
kuznetsss: Maybe we should add `ASSERT()` here?

tests/common/util/TestObject.cpp
```diff
@@ -89,6 +89,8 @@ createLedgerHeader(std::string_view ledgerHash, ripple::LedgerIndex seq, std::op
     ledgerHeader.seq = seq;
 
     if (age) {
+        // Note: be cautios of using small age values as the underlying NetClock precision is seconds
```
mathbunnyru:
```suggestion
        // Note: be cautious of using age values close to each other as the underlying NetClock precision is seconds
```

😆

mathbunnyru: Just using a large value won't help - if you provide 10000 and 10001 as values, the comparison will still be flake.
The idea is to provide values which are not close to each other.

kuznetsss: Why precommit hook didn't catch the typo?

tests/common/util/TestObject.cpp
mathbunnyru: Sorry, I don't get it, what will we assert?

tests/common/util/TestObject.cpp
mathbunnyru: https://github.com/crate-ci/typos?tab=readme-ov-file#why-was--not-corrected

> typos maintains a list of known typo corrections to keep the false positive count low so it can safely run unassisted.

> This is in contrast to most spell checking UIs people use where there is a known list of valid words

I think it might not be in the corrections list.

mathbunnyru: I checked `typos --files` and this test file shows up (as expected).
And looked at the `words.csv` here: https://raw.githubusercontent.com/crate-ci/typos/refs/heads/master/crates/typos-dict/assets/words.csv
Haven't found `cautios` there.

kuznetsss: Submited to typos's list of additions: https://github.com/crate-ci/typos/issues/1264#issuecomment-2847271667

tests/common/util/TestObject.cpp
kuznetsss: Assert that the value is big enough, but if it should be different from other values, we could't add an assertion here.

mathbunnyru: Yep, that's exactly why I was asking. So, the best we could do is to write a good comment.

tests/common/util/TestObject.cpp
mathbunnyru: Nice one, thanks!

mathbunnyru (approved):
👍

PR #2079, opened 2025-05-02
fix: Pass secrets.CODECOV_TOKEN explicitly

mathbunnyru:


mathbunnyru:
This was broken here: https://github.com/XRPLF/clio/pull/1879/files#r2071823493

When using reusable workflows, we need to pass secrets from caller side.
Otherwise, they won't have any secrets and `secrets.SOMETHING will have empty value`.

And because we used `fail_ci_if_error: false`, in the protected branches the codecov report was not uploaded.

godexsoft (approved):
👍

PR #2082, opened 2025-05-06
feat: Dosguard API weights

kuznetsss:
Experimental support for Dosguard API weights.


src/rpc/common/impl/ForwardingProxy.hpp
```diff
@@ -106,22 +107,7 @@ class ForwardingProxy {
...
-            "ledger_current",
-            "ripple_path_find",
-            "manifest",
-            "channel_authorize",
-            "channel_verify",
-            "simulate",
-        };
-
-        return kPROXIED_COMMANDS.contains(method);
+        return RPCCenter::kFORWARDED_RPCS.contains(method);
```
godexsoft: nit: Maybe it could be cleaner if we don't expose the static map and expose a helper like isProxied() instead?

tests/unit/web/dosguard/WeightsTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <boost/json/parse.hpp>
+#include <gtest/gtest.h>
+
+#include <cstddef>
+#include <optional>
+#include <string>
+#include <unordered_map>
+
+using namespace web::dosguard;
+using namespace web::dosguard;
```
godexsoft: Looks like we are double-using dosguard here

tests/unit/web/dosguard/WeightsTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL,  DIRECT,  INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
```
godexsoft, line 18: Please add empty line after the block comment

src/rpc/RPCCenter.hpp
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * The RPCCenter maintains lists of RPC commands that can be handled locally
+ * and those that need to be forwarded to rippled.
+ */
+struct RPCCenter {
+    /** @brief Set of RPC commands that can be handled locally by Clio */
+    static std::unordered_set<std::string_view> const kHANDLED_RPCS;
+
+    /** @brief Set of RPC commands that will be forwarded to rippled */
+    static std::unordered_set<std::string_view> const kFORWARDED_RPCS;
```
godexsoft: nit: Not sure we need to expose the maps, perhaps add few functions like 'isRpcName', 'isForwarded' and is 'isHandled'?

godexsoft (changes requested):
Leaving a few nits and suggestions. Overall very nice implementation 👍

StringHashTest.cpp
mathbunnyru, line 1: Do we need this empty file?

src/web/dosguard/DOSGuard.hpp
```diff
@@ -59,6 +61,7 @@ class DOSGuard : public DOSGuardInterface {
     util::Mutex<State> mtx_;
 
     std::reference_wrapper<WhitelistHandlerInterface const> whitelistHandler_;
+    std::reference_wrapper<WeightsInterface const> weights_;
```
mathbunnyru, line 64: Who is responsible for storage of `weights_`?
I think with this implementation it's easy to have a reference to a local object which goes out of scope.

Maybe we could store `unique_ptr` to the interface instead?

tests/unit/util/newconfig/ConfigConstraintsTests.cpp
```diff
@@ -0,0 +1,50 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of clio: https://github.com/XRPLF/clio
+    Copyright (c) 2024, the clio developers.
```
mathbunnyru: 2025 :)

mathbunnyru (changes requested):

src/rpc/RPCCenter.hpp
mathbunnyru: The maps are still exposed though.
You can remove them from header, and only leave them in cpp, and even better, put them in anonymous namespace.

src/rpc/RPCCenter.cpp
```diff
@@ -0,0 +1,97 @@
...
+*/
+//==============================================================================
+
+#include "rpc/RPCCenter.hpp"
+
+#include <string_view>
+#include <unordered_set>
+
+namespace rpc {
+std::unordered_set<std::string_view> const RPCCenter::kHANDLED_RPCS = {
```
mathbunnyru: nitpicking: need empty line here

src/web/dosguard/DOSGuard.hpp
kuznetsss: It will be not so easy to provide a mock with `unique_ptr`. 

Dosguard has such design for all its components, I created an issue to fix all of them: #2083.

StringHashTest.cpp
kuznetsss: Removed, thanks!

src/rpc/RPCCenter.hpp
kuznetsss: Moved them to cpp into anonymous namespace.

src/web/dosguard/DOSGuard.hpp
mathbunnyru: Thanks!

tests/unit/util/StringHashTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "util/StringHash.hpp"
+
+#include <gtest/gtest.h>
+
+#include <functional>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+
+namespace util {
```
godexsoft: nit: I don't mind but i don't think we normally do this for tests

tests/unit/web/ng/RPCServerHandlerTests.cpp
```diff
@@ -79,15 +80,22 @@ struct NgRpcServerHandlerTest : util::prometheus::WithPrometheus, MockBackendTes
     util::TagDecoratorFactory tagFactory_{config};
     std::string const ip_ = "some ip";
     StrictMockConnectionMetadata connectionMetadata_{ip_, tagFactory_};
+    Request::HttpHeaders const HttpHeaders_;
```
godexsoft: nit: clang-tidy will be upset about the name i think

tools/requests_gun/internal/request_maker/request_maker.go
```diff
@@ -74,6 +74,7 @@ func NewHttp(host string, port uint) *HttpRequestMaker {
 		host = "http://" + host
 	}
 	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DisableKeepAlives = true
```
godexsoft, line 77: Was this change intended or a side-effect of some recent investigation that sneaked in by mistake?

godexsoft (changes requested):
Leaving a few nits 👍

tools/requests_gun/internal/request_maker/request_maker.go
kuznetsss: Yeah, this was intended. As I understand the http library keeps connections alive to reuse them, but we need to avoid this to simulate multiple clients making requests.

godexsoft (dismissed):

src/rpc/RPCCenter.cpp
```diff
@@ -0,0 +1,102 @@
...
+    "get_aggregate_price",
+    "ledger",
+    "ledger_data",
+    "ledger_entry",
+    "ledger_index",
+    "ledger_range",
+    "mpt_holders",
+    "nfts_by_issuer",
+    "nft_history",
+    "nft_buy_offers",
```
mathbunnyru: List is not sorted (maybe other one as well, or this one in another places).

Is there a way not to hardcode this?

src/rpc/RPCCenter.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "rpc/RPCCenter.hpp"
+
+#include <string_view>
+#include <unordered_set>
+
+namespace rpc {
+
+namespace {
+
+std::unordered_set<std::string_view> const kHANDLED_RPCS = {
```
mathbunnyru: Even better is to have functions where you declare these as static and return const reference

src/web/dosguard/DOSGuard.cpp
```diff
@@ -115,14 +123,16 @@ DOSGuard::add(std::string const& ip, uint32_t numObjects) noexcept
...
 {
     if (whitelistHandler_.get().isWhiteListed(ip))
         return true;
 
+    auto const weight = weights_.get().requestWeight(request);
+
     {
         auto lock = mtx_.lock<std::scoped_lock>();
-        lock->ipState[ip].requestsCount++;
+        lock->ipState[ip].requestsCount += weight;
```
mathbunnyru, line 135: It's not `requestsCount` anymore

src/web/dosguard/DOSGuard.hpp
```diff
@@ -133,11 +141,12 @@ class DOSGuard : public DOSGuardInterface {
      * returned otherwise.
      *
      * @param ip
+     * @param request The request as json object
      * @return true
      * @return false
```
mathbunnyru, line 146: This comment is pretty useless, maybe you could fix it?

src/web/dosguard/DOSGuard.cpp
kuznetsss: It is still request count, just some requests cost more than 1 request.

mathbunnyru (approved):
Great job 👍

PR #2086, opened 2025-05-06
test: Remove SetUp and TearDown methods in tests

### LINKED ISSUE #910: Remove SetUp and TearDown from tests
We have this pattern to initialise/destroy everything in SetUp/TearDown methods. It limits code flexibility and makes us to manually call SetUp/TearDown of all parent classes. I think we could try to mostly get rid of such pattern and move all logic to test's constructors/destructors.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/910

### DIFF
--- a/tests/unit/etl/TransformerTests.cpp
+++ b/tests/unit/etl/TransformerTests.cpp
@@ -60,19 +60,12 @@ struct ETLTransformerTest : util::prometheus::WithPrometheus, MockBackendTest {
     using TransformerType = etl::impl::
         Transformer<ExtractionDataPipeType, LedgerLoaderType, LedgerPublisherType, AmendmentBlockHandlerType>;
 
-    void
-    SetUp() override
+    ETLTransformerTest()
     {
         state_.isStopping = false;
         state_.writeConflict = false;
         state_.isReadOnly = false;
         state_.isWriting = false;
-    }
-
-    void
-    TearDown() override
-    {
-        transformer_.reset();
     }
 
 protected:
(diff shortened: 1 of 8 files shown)


tests/unit/etl/TransformerTests.cpp
```diff
@@ -60,21 +60,14 @@ struct ETLTransformerTest : util::prometheus::WithPrometheus, MockBackendTest {
...
         state_.isStopping = false;
         state_.writeConflict = false;
         state_.isReadOnly = false;
         state_.isWriting = false;
     }
 
-    void
-    TearDown() override
-    {
-        transformer_.reset();
```
mathbunnyru, line 75: Don't need this, as destructor will do the same

mathbunnyru:
I checked the repo and there are no more `TearDown`.
One `SetUp` is needed though:
```
tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp:
   90      // This function is used to prepare the database before running the tests
   91:     // It is called in the SetUp function. Different tests can override this function to prepare the database
   92      // differently

  152      void
  153:     SetUp() override
  154      {
  155          setupDatabase();
  156      }
```

And `setupDatabase()` is a virtual method of the class, so we shouldn't call it directly in ctor, so I left this one as is.

tests/unit/etl/TransformerTests.cpp
godexsoft: There was some reason why it was added but other code that depended on it may have been later removed. I recall we had some issues with this before

godexsoft (approved):
Nice! 👍

tests/unit/etl/TransformerTests.cpp
mathbunnyru: Got it, seems to work fine now, so probably the code has been removed.

PR #2093, opened 2025-05-07
fix: Catch exception in ClusterCommunicationService

### LINKED ISSUE #2016: Reading in ClusterCommunicationService could throw
## Issue Description

This call may throw an exception if read times out: 
https://github.com/XRPLF/clio/blob/develop/src/cluster/ClusterCommunicationService.cpp#L130

Currently this exception is not handled and it could cause crash. We should add try-catch block somewhere to handle timeout.

kuznetsss:
Fixes #2016.


godexsoft (approved):
👍

PR #2096, opened 2025-05-08
feat: Add load balancer prometheus metrics

### LINKED ISSUE #2070: Prometheus counters for forwarding
## Summary

Let's introduce prometheus counters for 
- forwarded requests processing time
- forwarding cache usage rate
- how many forwarding sources were attempted before one succeeded
- any other useful counters to do with forwarding of requests

## Motivation

It would be useful to have this on grafana to understand how it impacts overall system stability.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2070


src/etl/LoadBalancer.cpp
```diff
@@ -76,6 +82,8 @@ LoadBalancer::makeLoadBalancer(
     );
 }
 
+using namespace util::prometheus;
```
godexsoft: If you want to do this let's put it at the top of the file next to includes OR (better) inside of the function that uses it, if there is only one/few places.

godexsoft (changes requested):
Just a nit 👍

src/etl/LoadBalancer.cpp
mathbunnyru: It's used inside ctor member init, so can't put it inside function unfortunately.
Put it to the top of the file, thanks!

src/etl/LoadBalancer.cpp
```diff
@@ -84,6 +91,27 @@ LoadBalancer::LoadBalancer(
...
+    : forwardedDurationHistogram_(PrometheusService::histogramInt(
+          "lb_forwarded_duration_milliseconds_histogram",
+          Labels(),
+          kHISTOGRAM_BUCKETS,
+          "The duration of processing forwarded requests"
+      ))
+    , forwardedRetryCounter_(PrometheusService::counterInt(
+          "lb_forwarded_retry_counter",
+          Labels(),
+          "The number of retries before a forwarded request was successful"
```
godexsoft: nit: Maybe worth adding something like "excluding initial attempt"

godexsoft (approved):
👍

src/etl/LoadBalancer.cpp
```diff
@@ -84,6 +91,27 @@ LoadBalancer::LoadBalancer(
...
     SourceFactory sourceFactory
 )
+    : forwardedDurationHistogram_(PrometheusService::histogramInt(
+          "lb_forwarded_duration_milliseconds_histogram",
+          Labels(),
+          kHISTOGRAM_BUCKETS,
+          "The duration of processing forwarded requests"
+      ))
+    , forwardedRetryCounter_(PrometheusService::counterInt(
+          "lb_forwarded_retry_counter",
```
kuznetsss, line 101: I think `lb` prefix is not very discoverable, `load_balancer` or even `forwarding` prefix would be better I think.

kuznetsss:
I think there should be tests that the metrics are updated when the related method is called.

src/etl/LoadBalancer.cpp
```diff
@@ -84,6 +91,27 @@ LoadBalancer::LoadBalancer(
...
+          Labels(),
+          "The number of retries before a forwarded request was successful. Initial attempt excluded"
+      ))
+    , cacheTriedCounter_(PrometheusService::counterInt(
+          "lb_cache_tried_counter",
+          Labels(),
+          "The number of requests that we tried to serve from the cache"
+      ))
```
kuznetsss, line 109: Maybe this could be cache hit counter instead. Tried counter is less intuitive.

src/etl/LoadBalancer.cpp
```diff
@@ -84,6 +91,27 @@ LoadBalancer::LoadBalancer(
     std::shared_ptr<NetworkValidatedLedgersInterface> validatedLedgers,
     SourceFactory sourceFactory
 )
+    : forwardedDurationHistogram_(PrometheusService::histogramInt(
+          "lb_forwarded_duration_milliseconds_histogram",
+          Labels(),
+          kHISTOGRAM_BUCKETS,
+          "The duration of processing forwarded requests"
+      ))
```
kuznetsss, line 99: Do we actually need a histogram here? Maybe simply counter would be enough, because histogram is quite more heavy than a single counter. And probably we don't need to analyse distribution of response time.

src/etl/LoadBalancer.cpp
```diff
@@ -379,11 +406,15 @@ LoadBalancer::forwardToRippledImpl(
     std::optional<boost::json::object> response;
     rpc::ClioError error = rpc::ClioError::EtlConnectionError;
     while (numAttempts < sources_.size()) {
-        auto res = sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield);
+        auto [res, duration] =
+            util::timed([&]() { return sources_[sourceIdx]->forwardToRippled(request, clientIp, xUserValue, yield); });
+        forwardedDurationHistogram_.get().observe(duration);
```
kuznetsss, line 411: Maybe we should separate successful and failed requests here.

src/etl/LoadBalancer.cpp
mathbunnyru: Thanks, renamed variables

src/etl/LoadBalancer.cpp
mathbunnyru: https://github.com/XRPLF/clio/issues/2070 was asking for `forwarded requests processing time`, that's why I implemented it as histograms.

@godexsoft what do you need: processing times or counter?

src/etl/LoadBalancer.cpp
mathbunnyru: Waiting for the response above first

PR #2057, opened 2025-04-29
chore: Use only pre-commit hooks

mathbunnyru:

PR #2099, opened 2025-05-12
fix: Make fix-local-includes.sh work on file-basis

### LINKED ISSUE #2098: Full rebuild after every commit
After every commit, `ninja` rebuilds every file, even if they haven't changed.
This is likely due to `fix-local-includes` pre-commit hook.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2098

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -82,7 +82,6 @@ repos:
         entry: pre-commit-hooks/fix-local-includes.sh
         types: [c++]
         language: system
-        pass_filenames: false
       - id: verify-commits
         name: Verify Commits
         entry: pre-commit-hooks/verify-commits.sh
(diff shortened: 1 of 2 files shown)


mathbunnyru:
The problem was not in adding `pre-commit` tool, but because of `fix-local-includes.sh` implementation: it changed all includes to be in `<>` style and then `""` style for local includes (https://github.com/XRPLF/clio/pull/2057/files#diff-2e447d51e163ef75d5b5fd6685111521da8e5275dbbb6ee800ddbd17b3a340d3)
It was doing it for **all files** in `src|tests` if at least one file there was changed.

Now the `pre-commit` tool only runs this hook for changed files.

mathbunnyru:
Checked that there is no full rebuild after commits:

1. Built the project
2. Changed `src/app/CliArgs.cpp`
3. Ran `ninja` again: only 4 targets were rebuilt

Also checked that the hook still works:
1. Changed some include styles and tried to commit.
2. Got an error and autofix worked fine

godexsoft (approved):
Looks good 👍

PR #2100, opened 2025-05-12
fix: Guarantee async behaviour of WsBase::send

godexsoft:
`dispatch` can (and sometimes will) execute the otherwise async code on the thread of the caller if it's already on the same `io_context`. `wsFail` closes the socket if an error happened which can happen while the Subscription subsystem is publishing something into the same socket and hence holding the lock that is also required to unsubscribe once the connection is no longer needed. This can lead to double locking and things hang indefinitely.


kuznetsss (approved):
All good, just change the PR title prefix to `fix` please.

PR #2102, opened 2025-05-12
fix: Make fix-local-includes.sh work with multiple files

mathbunnyru:


pre-commit-hooks/fix-local-includes.sh
```diff
@@ -2,26 +2,38 @@
...
+        # make all includes to be <...> style
+        sed -i -E 's|#include "(.*)"|#include <\1>|g' "$file_path"
+
+        # make local includes to be "..." style
+        sed -i -E "s|#include <(($main_src_dirs)/.*)>|#include \"\1\"|g" "$file_path"
+    fi
+}
+
+for file in $files; do
+    fix_includes "$file"
```
kuznetsss, line 38: Maybe this could be done in parallel now

mathbunnyru: It's already done in parallel by `pre-commit` tool itself.
On my machine it spins up to 12 threads and runs this script.
So, no need to think about executing in parallel internally.

And, since we're only running on the changed files now because of `pre-commit` tool (when running locally), it is extremely fast.

kuznetsss (approved):
👍

PR #2104, opened 2025-05-12
style: Use `codespell` instead of `typos` pre-commit hook

mathbunnyru:


kuznetsss:
Why do we need both typos and codespell? I thought the idea of typos was that it ignores unknown words, so we don't have to maintain a big list of exceptions. While for codespell we have to maintain a list of exceptions.

mathbunnyru:
> Why do we need both typos and codespell? I thought the idea of typos was that it ignores unknown words, so we don't have to maintain a big list of exceptions. While for codespell we have to maintain a list of exceptions.

I am adding codespell, because apparently typos is not fixing all the problems present in the codespell. I'm pretty sure it's true the other way around as well.

We actually maintain similar list for typos as well: https://github.com/XRPLF/clio/blob/develop/_typos.toml

And, it's not that big :)

pre-commit-hooks/codespell_ignore.txt
```diff
@@ -0,0 +1,9 @@
+ser
+onWs
+datas
+AtLeast
+AtMost
+compiletime
+tring
+trings
```
kuznetsss, line 8: Where did this come from? I found it only in typos config.

.pre-commit-config.yaml
```diff
@@ -42,6 +42,16 @@ repos:
     hooks:
       - id: typos
```
kuznetsss: Maybe remove typos then? I don't think we need two spell checkers.

pre-commit-hooks/codespell_ignore.txt
mathbunnyru: https://github.com/XRPLF/clio/blob/develop/src/rpc/JS.hpp#L24

.pre-commit-config.yaml
mathbunnyru: As I said, I don't think one of them is better than another, and I don't see a problem of having both.

godexsoft: I'm not a fan of having two spell-checkers either.. but it does seem to have fixed obvious mistakes that `typos` should also have discovered. What i really don't like is having two separate places to specify exceptions :rage4: If there is a way to have the same exceptions file then i really don't care how many spell-checkers there are in the pre-commit tool as long as it works 👍

kuznetsss (approved):
👍

PR #2105, opened 2025-05-12
ci: Attempt to remove whitespace to fix sanitizer build

godexsoft:
Let's see if this does the trick. 
Alternatives are
- more complicated cleanup using sed or similar
- don't multiline the expression (e.g. "${{ multiline expression becomes single line }}")
- don't wrap `SANITIZER_OPTION` in double quotes (like it was before)
- something else that i did not find - let me know if you know

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,5 +1,5 @@
 name: Run tests with sanitizers
-# remove me
+
 on:
   schedule:
     - cron: "0 4 * * 1-5"
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,5 +1,5 @@
 name: Run tests with sanitizers
-
+# remove me
 on:
   schedule:
     - cron: "0 4 * * 1-5"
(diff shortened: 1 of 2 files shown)


mathbunnyru (changes requested):
I would probably (if it works):

- use a map
- write it in one-line
- use a validation of input param (I already added this)

Here is how you can create a map in GH Actions: https://stackoverflow.com/a/64405821

mathbunnyru:
Another thing worth trying:
Instead of `SANITIZER_OPTION: |` use `SANITIZER_OPTION: >` or something like that (I don't remember exactly how to do it), but there are some ways to control how multi-line yaml strings are treated

godexsoft:
> Another thing worth trying: Instead of `SANITIZER_OPTION: |` use `SANITIZER_OPTION: >` or something like that (I don't remember exactly how to do it), but there are some ways to control how multi-line yaml strings are treated

I was sure that Claude was hallucinating `|-` because i could not find any trace of it online 😃 But turns out it is a real thing so let's try of course 👍

.github/workflows/sanitizers.yml
```diff
@@ -1,5 +1,5 @@
 name: Run tests with sanitizers
-
+# remove me
```
mathbunnyru: Remove me

mathbunnyru (dismissed):
Should be good to go, after you remove the trigger comment

mathbunnyru (approved):

PR #2103, opened 2025-05-12
refactor: Load balancer forwarding metrics

mathbunnyru:

PR #2128, opened 2025-05-15
feat: Forwarding metrics

kuznetsss:
Port of #2096 and #2103 into `release/2.4.1`.

### DIFF
--- a/src/etl/LoadBalancer.hpp
+++ b/src/etl/LoadBalancer.hpp
@@ -32,6 +32,7 @@
 #include "util/ResponseExpirationCache.hpp"
 #include "util/log/Logger.hpp"
 #include "util/newconfig/ConfigDefinition.hpp"
+#include "util/prometheus/Counter.hpp"
 
 #include <boost/asio.hpp>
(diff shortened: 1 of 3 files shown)


src/etl/LoadBalancer.hpp
```diff
@@ -44,8 +45,10 @@
 #include <xrpl/proto/org/xrpl/rpc/v1/xrp_ledger.grpc.pb.h>
 
 #include <chrono>
+#include <concepts>
 #include <cstdint>
 #include <expected>
+#include <functional>
```
mathbunnyru, line 51: These includes look unrelated to the change

mathbunnyru (approved):
LGTM, only nitpicking

src/etl/LoadBalancer.hpp
kuznetsss: Clangd said they were missing, so I added. I don't think they make any difference.

mathbunnyru: Ok

PR #2129, opened 2025-05-15
fix: Use `git lfs install` and fix verify-commits hook

mathbunnyru:


pre-commit-hooks/lfs/post-checkout
```diff
@@ -1,3 +0,0 @@
-#!/bin/sh
-command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the 'post-checkout' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks').\n"; exit 2; }
-git lfs post-checkout "$@"
```
mathbunnyru, line 3: `git lfs` hooks don't like being run not by git directly, that's why we don't use it as part of `pre-commit` tool.

.pre-commit-config.yaml
```diff
@@ -94,33 +94,7 @@ repos:
       - id: verify-commits
```
mathbunnyru, line 94: There was no stages for this hook, and it didn't work

.pre-commit-config.yaml
```diff
@@ -7,7 +7,7 @@
 # Common tasks
 #
 # - Run on all files:   pre-commit run --all-files
-# - Register git hooks: pre-commit install --install-hooks
```
mathbunnyru, line 10: By default, `pre-commit` only installs `pre-commit` hook stage, but we also need `pre-push`

pre-commit-hooks/verify-commits.sh
```diff
@@ -39,16 +39,14 @@ verify_tag_signed() {
...
-        version=$(git tag --points-at HEAD)
-        echo "Looks like you're trying to push a $version release..."
-        echo "Making sure you've signed and tagged it."
-        if verify_commit_signed && verify_tag && verify_tag_signed ; then
-            : # Ok, I guess you can push
-        else
-            exit 1
-        fi
+# Check some things if we're pushing a branch called "release/"
+if echo "$PRE_COMMIT_REMOTE_BRANCH" | grep ^refs\/heads\/release\/ &> /dev/null ; then
```
mathbunnyru, line 43: This is a env variable provided by `pre-commit` utility

godexsoft (approved):
:rage3:

PR #2130, opened 2025-05-15
fix: Fix some doxygen docs errors

mathbunnyru:


.github/workflows/docs.yml
```diff
@@ -45,6 +51,20 @@ jobs:
           path: build_docs/html
           name: docs-develop
 
+  deploy:
+    needs: build
+    if: github.event_name != 'pull_request' && github.ref == 'refs/heads/develop'
```
mathbunnyru: Unfortunately, GH pages don't support `preview` in PRs (which would be amazing).
But it's still better to build docs in PRs before merging, rather than have an issue later

kuznetsss: I think if doxygen pre-commit check passed than this will work fine. Github pages is a documentation for developers, it is not a big issue if it is outdated for some time if this workflow is broken for some reason. 

The number of CI runners is limited, so I would prefer to not spend them on this. Also I don't remember that we ever had this workflow failed. https://github.com/XRPLF/clio/actions/workflows/docs.yml?query=is%3Afailure

mathbunnyru: 1. `ubuntu-latest` is used to build the docs, so it doesn't affect the number of CI runners.
2. Workflow didn't fail because it was ignoring all the warnings :)

kuznetsss: The number of `ubuntu-latest` running simultaneously is still limited.

mathbunnyru: I don't know the exact number of these runners simultaneously available (I suspect it's quite high though), and all the `ubuntu-latest` runs we have last around 1 minute (except `build_clio_docker_image`, which is quite rare and unlikely to be run in parallel), so this is not really a problem.

kuznetsss (approved):

PR #2131, opened 2025-05-15
feat: add get range feat to export tool

PeterChen13579:


tools/snapshot/internal/parse_args/parse_args.go
```diff
@@ -31,33 +36,43 @@ func Parse() (*Args, error) {
...
+		if *host == "" || *username == "" || *password == "" {
+			return nil, fmt.Errorf("invalid usage: --host, --username, and --password are required for getRange")
 		}
 	} else {
-		return nil, fmt.Errorf("Invalid usage: --export or --server or --range flag is required.")
+		return nil, fmt.Errorf("invalid usage: --export or --server or --range flag is required")
 	}
 
-	return &Args{*exportMode, *seq, *endSeq, *path, *grpcServer, *wsServer, *serverMode, *showRange}, nil
+	return &Args{*exportMode, *seq, *endSeq, *path, *grpcServer, *wsServer, *serverMode, *showRange, *getRange, *host, *username, *password}, nil
```
mathbunnyru: Could you use named arguments here please?

tools/snapshot/internal/util/db.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/gocql/gocql"
+)
+
+func GetLedgerRange(cluster *gocql.ClusterConfig) (uint64, uint64, error) {
+	var (
+		firstLedgerIdx  uint64
+		latestLedgerIdx uint64
+	)
```
mathbunnyru: I would suggest doing the following:

1. Create a new type called Range and return it
2. Rename `latest` -> `last`
3. Return this range by pointer - this will make code idiomatic, when you return `sth, nil` in case everything is ok and `nil, err` if it's not.

tools/snapshot/internal/parse_args/parse_args.go
```diff
@@ -31,33 +36,43 @@ func Parse() (*Args, error) {
 	wsServer := fs.StringP("ws_server", "w", "0.0.0.0:6006", "rippled's gRPC server address")
 	serverMode := fs.BoolP("server", "s", false, "Start server mode")
 	showRange := fs.BoolP("range", "r", false, "Show the range of the snapshot")
+	getRange := fs.Bool("getRange", false, "Show the range of the snapshot")
```
mathbunnyru: Please, update the description

mathbunnyru (changes requested):

tools/snapshot/internal/util/db.go
PeterChen13579: I think `latest` is a better name because the table involves querying `is_latest`. Other than that, changed. 👍

mathbunnyru: Makes sense, thanks for explaining why you named it like this 👍

mathbunnyru (approved):
If it works, LGTM 👍

PR #2135, opened 2025-05-16
feat: Add hadolint docker pre-commit hook

mathbunnyru:


.hadolint.yml
```diff
@@ -0,0 +1,8 @@
+---
```
godexsoft, line 1: At this point, is there any way to have all the multiple linter settings and files in a subdirectory so we don't have to pollute the project root?

mathbunnyru: I think it's possible, but I wouldn't recommend it.

1. I haven't seen other projects do that, and people usually don't have trouble because of multiple config files in the repo root
2. For each tool which config we move, we'll have to run the tool in a custom way in pre-commit
3. If someone uses extensions in the IDE, most of the time they do understand default config location in the root repo, but need to be configured for custom location. Per each tool in IDE and contributor.

So, even though I share your concern about polluting the root of the repo, I think not doing it creates more harm.

godexsoft (approved):

PR #2138, opened 2025-05-18
feat: ETLng publisher

godexsoft:
Fixes concurrency issues in extraction process in ETLng and implements publishing through monitor.
ETLng still Missing:
- Mechanism to become and give up being a writer node
- MPT
- Probably some more small discrepancies, especially around using config values (start seq, end seq, etc.)


src/app/ClioApplication.cpp
```diff
@@ -103,6 +104,10 @@ ClioApplication::run(bool const useNgWebServer)
     // This is not the only io context in the application.
     boost::asio::io_context ioc{threads};
 
+    // Similarly we need a context to run ETLng on
+    // In the future we can remove the raw ioc and use ctx instead
+    util::async::CoroExecutionContext ctx{threads};
```
kuznetsss, line 109: Maybe we should create it inside `makeETLService()`? It looks like a similar problem to dosguard.

src/etl/LoadBalancer.cpp
```diff
@@ -65,6 +65,10 @@ using util::prometheus::Labels;
 
 namespace etl {
 
+namespace {
+std::vector<std::int64_t> const kHISTOGRAM_BUCKETS{1, 2, 5, 10, 20, 50, 100, 200, 500, 700, 1000};
+}  // namespace
+
```
kuznetsss: Looks like this is not used.

src/etl/CacheLoader.hpp
```diff
@@ -46,7 +48,7 @@ namespace etl {
  * @tparam ExecutionContextType The type of the execution context to use
  */
 template <typename ExecutionContextType = util::async::CoroExecutionContext>
-class CacheLoader {
+class CacheLoader : public etlng::CacheLoaderInterface {
```
mathbunnyru, line 51: So, etl depends on etlng, and etlng depends on etl?
That's a circular dependency :(

src/etlng/ETLService.cpp
```diff
@@ -0,0 +1,275 @@
...
+    , publisher_(std::move(publisher))
+    , cacheLoader_(std::move(cacheLoader))
+    , cacheUpdater_(std::move(cacheUpdater))
+    , extractor_(std::move(extractor))
+    , loader_(std::move(loader))
+    , initialLoadObserver_(std::move(initialLoadObserver))
+    , taskManagerProvider_(std::move(taskManagerProvider))
+    , state_(std::move(state))
+{
+    LOG(log_.info()) << "Creating ETLng...";
```
mathbunnyru, line 111:
```suggestion
    LOG(log_.info()) << "Created ETLng...";
```

src/etlng/TaskManagerInterface.hpp
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @brief An interface for the Task Manager
+ */
+struct TaskManagerInterface {
+    /**
+     * @brief Settings for the task manager
+     */
+    struct Settings {
+        size_t numExtractors; /**< number of extraction tasks */
+    };
```
kuznetsss: Are you going to add more parameters here in the future?

src/etlng/impl/LedgerPublisher.hpp
```diff
@@ -0,0 +1,292 @@
...
+    util::Logger log_{"ETL"};
+
+    boost::asio::strand<boost::asio::io_context::executor_type> publishStrand_;
+
+    std::shared_ptr<BackendInterface> backend_;
+    std::shared_ptr<feed::SubscriptionManagerInterface> subscriptions_;
+    std::reference_wrapper<etl::SystemState const> state_;  // shared state for ETL
+
+    std::chrono::time_point<ripple::NetClock> lastCloseTime_;
+    mutable std::shared_mutex closeTimeMtx_;
```
kuznetsss: Maybe use `util::Mutex`? Here and a few lines below.

src/etlng/impl/LedgerPublisher.hpp
```diff
@@ -0,0 +1,292 @@
...
+                std::vector<data::TransactionAndMetadata> transactions =
+                    data::synchronousAndRetryOnTimeout([&](auto yield) {
+                        return backend_->fetchAllTransactionsInLedger(lgrInfo.seq, yield);
+                    });
+
+                auto const ledgerRange = backend_->fetchLedgerRange();
+                ASSERT(ledgerRange.has_value(), "Ledger range must exist");
+
+                std::string const range =
+                    std::to_string(ledgerRange->minSequence) + "-" + std::to_string(ledgerRange->maxSequence);
```
kuznetsss: Would be more efficient to use `fmt::format()`.

src/etlng/impl/LedgerPublisher.hpp
```diff
@@ -0,0 +1,292 @@
...
+                std::ranges::sort(transactions, [](auto const& t1, auto const& t2) {
+                    ripple::SerialIter iter1{t1.metadata.data(), t1.metadata.size()};
+                    ripple::STObject const object1(iter1, ripple::sfMetadata);
+                    ripple::SerialIter iter2{t2.metadata.data(), t2.metadata.size()};
+                    ripple::STObject const object2(iter2, ripple::sfMetadata);
+                    return object1.getFieldU32(ripple::sfTransactionIndex) <
+                        object2.getFieldU32(ripple::sfTransactionIndex);
+                });
+
+                for (auto& txAndMeta : transactions)
```
kuznetsss:
```suggestion
                for (auto const& txAndMeta : transactions)
```

src/etlng/impl/LedgerPublisher.hpp
```diff
@@ -0,0 +1,292 @@
...
+                    subscriptions_->pubTransaction(txAndMeta, lgrInfo);
+
+                subscriptions_->pubBookChanges(lgrInfo, transactions);
+
+                setLastPublishTime();
+                LOG(log_.info()) << "Published ledger " << std::to_string(lgrInfo.seq);
+            } else {
+                LOG(log_.info()) << "Skipping publishing ledger " << std::to_string(lgrInfo.seq);
```
kuznetsss: Do you need `std::to_string()` here?

src/etlng/impl/TaskManager.cpp
```diff
@@ -107,13 +115,24 @@ TaskManager::spawnExtractor(util::async::AnyStrand& strand, PriorityQueue& queue
...
+TaskManager::spawnLoader(TaskQueue& queue)
 {
     return ctx_.execute([this, &queue](auto stopRequested) {
         while (not stopRequested) {
             // TODO (https://github.com/XRPLF/clio/issues/66): does not tell the loader whether it's out of order or not
-            if (auto data = queue.dequeue(); data.has_value())
-                loader_.get().load(*data);
+            if (auto data = queue.dequeue(); data.has_value()) {
+                auto nanos = util::timed<std::chrono::nanoseconds>([this, data = *data] { loader_.get().load(data); });
+                auto const seconds = nanos / 1000000000.0;
```
kuznetsss: Would be good to move it into a named constant and use scientific notation `1.0e9`.

src/etlng/impl/TaskManagerProvider.hpp
```diff
@@ -0,0 +1,74 @@
...
+    {
+    }
+
+    std::unique_ptr<TaskManagerInterface>
+    make(util::async::AnyExecutionContext ctx, std::reference_wrapper<MonitorInterface> monitor, uint32_t seq) override
+    {
+        auto scheduler = impl::makeScheduler(impl::ForwardScheduler{ledgers_, seq});
+        // TODO: add impl::BackfillScheduler{seq - 1, seq - 1000},
+
+        return std::make_unique<TaskManager>(std::move(ctx), std::move(scheduler), *extractor_, *loader_, monitor, seq);
```
kuznetsss, line 70: You should delete move constructors for `extractor_` and `loader_` if you want to use them by reference.

src/etlng/impl/TaskQueue.hpp
```diff
@@ -19,85 +19,93 @@
...
-template <typename T, typename Compare = std::less<T>>
-class StrandedPriorityQueue {
-    util::async::AnyStrand strand_;
+class TaskQueue {
     std::size_t limit_;
-    std::priority_queue<T, std::vector<T>, Compare> queue_;
+    std::uint32_t expectedSequence_;
+    std::uint32_t increment_;
+
+    std::mutex mutex_;
```
kuznetsss: Also maybe use `util::Mutex` here.

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -0,0 +1,356 @@
...
+    std::shared_ptr<MockLoadBalancer> balancer_ = std::make_shared<MockLoadBalancer>();
+    MockNetworkValidatedLedgersPtr ledgers_;
+    std::shared_ptr<MockLedgerPublisher> publisher_ = std::make_shared<MockLedgerPublisher>();
+    std::shared_ptr<MockCacheLoader> cacheLoader_ = std::make_shared<MockCacheLoader>();
+    std::shared_ptr<MockCacheUpdater> cacheUpdater_ = std::make_shared<MockCacheUpdater>();
+    std::shared_ptr<MockExtractor> extractor_ = std::make_shared<MockExtractor>();
+    std::shared_ptr<MockLoader> loader_ = std::make_shared<MockLoader>();
+    std::shared_ptr<MockInitialLoadObserver> initialLoadObserver_ = std::make_shared<MockInitialLoadObserver>();
+    std::shared_ptr<MockTaskManagerProvider> taskManagerProvider_ = std::make_shared<MockTaskManagerProvider>();
+    std::shared_ptr<etl::SystemState> systemState_ = std::make_shared<etl::SystemState>();
```
kuznetsss: Probably they should be `NiceMock` or `StrictMock`.

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+    EXPECT_TRUE(result.contains("etl_sources"));
+    EXPECT_TRUE(result.at("etl_sources").is_array());
+    EXPECT_EQ(result.at("etl_sources").as_array().size(), 1);
+    EXPECT_TRUE(result.contains("is_writer"));
+    EXPECT_TRUE(result.at("is_writer").is_int64());
+    EXPECT_TRUE(result.contains("read_only"));
+    EXPECT_TRUE(result.at("read_only").is_int64());
+    EXPECT_TRUE(result.contains("last_publish_age_seconds"));
+    EXPECT_EQ(result.at("last_publish_age_seconds").as_string(), "42");
```
kuznetsss: Maybe create expected json instead?

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -0,0 +1,356 @@
...
+    return etlng::model::LedgerData{
+        .transactions = {},
+        .objects = {util::createObject(), util::createObject(), util::createObject()},
+        .successors = {},
+        .edgeKeys = {},
+        .header = header,
+        .rawHeader = {},
+        .seq = seq
+    };
+}
```
kuznetsss: Better to move it into fixture or anonymous namespace.

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -0,0 +1,356 @@
...
+        EXPECT_CALL(*backend_, hardFetchLedgerRange(testing::_)).WillOnce(testing::Return(std::nullopt));
+        EXPECT_CALL(*ledgers_, getMostRecent()).WillOnce(testing::Return(mostRecentSeq));
+    }
+
+    void
+    setupPopulatedDatabaseExpectations(uint32_t maxSequence)
+    {
+        EXPECT_CALL(*backend_, hardFetchLedgerRange(testing::_))
+            .WillOnce(testing::Return(data::LedgerRange{.minSequence = 1, .maxSequence = maxSequence}));
+    }
```
kuznetsss: These functions seems to be unused.

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -0,0 +1,356 @@
...
+              cacheLoader_,
+              cacheUpdater_,
+              extractor_,
+              loader_,
+              initialLoadObserver_,
+              taskManagerProvider_,
+              systemState_
+          )
+    {
+    }
```
kuznetsss:
```suggestion
    etlng::ETLService service_{
              ctx_,
              config_,
              backend_,
              balancer_,
              ledgers_,
              publisher_,
              cacheLoader_,
              cacheUpdater_,
              extractor_,
              loader_,
              initialLoadObserver_,
              taskManagerProvider_,
              systemState_
    };
```

tests/unit/etlng/LedgerPublisherTests.cpp
```diff
@@ -0,0 +1,341 @@
...
+    publisher.publish(dummyLedgerHeader);
+
+    // mock fetch fee
+    EXPECT_CALL(*backend_, doFetchLedgerObject(ripple::keylet::fees().key, kSEQ, _))
+        .WillOnce(Return(createLegacyFeeSettingBlob(1, 2, 3, 4, 0)));
+
+    // t1 index > t2 index
+    TransactionAndMetadata t1;
+    t1.transaction = createPaymentTransactionObject(kACCOUNT, kACCOUNT2, 100, 3, kSEQ).getSerializer().peekData();
+    t1.metadata = createPaymentTransactionMetaObject(kACCOUNT, kACCOUNT2, 110, 30, 2).getSerializer().peekData();
```
kuznetsss: Maybe use named constants here?

src/etlng/TaskManagerInterface.hpp
godexsoft: At the start there was also `numLoaders` but i think we no longer need them because at least in the foreseeable future we will only have one forward (and one backfill) loader as they depend on the synchronization in the backend and (today) require to be single threaded

src/app/ClioApplication.cpp
godexsoft: That would require storing it on the heap which i am trying to avoid. The context is also going to be shared by more components in the future so i think it's ok to create it closer to main

kuznetsss: Why would it require to store it on the heap? What if we do something like this:
```cpp
struct ETLServiceData {
  util::async::CoroExecutionContext ctx;
  std::shared_ptr<ETLInterface> etl;

  static ETLServiceData make(...);
};
```
and use it in `ClioApplication`?

src/etlng/ETLService.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+boost::json::object
+ETLService::getInfo() const
+{
+    boost::json::object result;
+
+    result["etl_sources"] = balancer_->toJson();
+    result["is_writer"] = static_cast<int>(state_->isWriting);
+    result["read_only"] = static_cast<int>(state_->isReadOnly);
```
mathbunnyru, line 173: why can't we just have bools here?

src/app/ClioApplication.cpp
godexsoft: Let's do this once we have ETLng working and old ETL removed - it will be more clear how to do this properly and what deps should go where.
https://github.com/XRPLF/clio/issues/2153

kuznetsss (approved):
👍

mathbunnyru (commented):
I started reviewing, but PR merged before I finished 😆

src/etlng/ETLService.cpp
godexsoft: I don't dare to change this and risk breaking someone's scripts or something 😢

src/etl/CacheLoader.hpp
godexsoft: This is irrelevant because `etlng` will merge with `etl` and old `etl` will be removed entirely eventually. Not nice but temporary 😄

PR #2151, opened 2025-05-21
fix: Add Delegate to Ledger types

kuznetsss:
Fix discrepancy with rippled for `account_objects` API.

### DIFF
--- a/tests/unit/util/LedgerUtilsTests.cpp
+++ b/tests/unit/util/LedgerUtilsTests.cpp
@@ -57,7 +57,8 @@ TEST(LedgerUtilsTests, LedgerObjectTypeList)
         JS(permissioned_domain),
         JS(oracle),
         JS(credential),
-        JS(nunl)
+        JS(nunl),
+        JS(delegate)
     };
 
     static_assert(std::size(kTYPES_LIST) == kTYPES.size());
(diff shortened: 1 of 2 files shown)


godexsoft (approved):

tests/unit/util/LedgerUtilsTests.cpp
```diff
@@ -57,7 +57,8 @@ TEST(LedgerUtilsTests, LedgerObjectTypeList)
         JS(permissioned_domain),
         JS(oracle),
         JS(credential),
-        JS(nunl)
+        JS(nunl),
+        JS(delegate)
```
mathbunnyru, line 61: Please, add a trailing comma here

tests/unit/util/LedgerUtilsTests.cpp
```diff
@@ -91,6 +92,7 @@ TEST(LedgerUtilsTests, AccountOwnedTypeList)
         JS(mpt_issuance),
         JS(mptoken),
         JS(permissioned_domain),
+        JS(delegate)
```
mathbunnyru, line 95: Please, add a trailing comma here

mathbunnyru (changes requested):
small comments to make code easier to change the next time

tests/unit/util/LedgerUtilsTests.cpp
kuznetsss: I think this should be forced by tooling, but unfortunately it seems clang-format couldn't do that. Manually check trailing commas each time doesn't worth it if we couldn't make it to be the same style in the whole repository automatically.

mathbunnyru (approved):

PR #2154, opened 2025-05-23
feat: ETLng MPT support

godexsoft:
This PR ports MPT-related ETL functionality into ETLng.


src/etlng/impl/ext/MPT.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+    void
+    onInitialObject(uint32_t seq, model::Object const& obj) const;
+
+    void
+    onInitialData(model::LedgerData const& data) const;
+
+private:
+    void
+    writeMPTHoldersFromTransactions(model::LedgerData const& data) const;
```
mathbunnyru: These methods shouldn't be marked const

src/etlng/impl/ext/MPT.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace etlng::impl {
+
+MPTExt::MPTExt(std::shared_ptr<BackendInterface> backend) : backend_(std::move(backend))
+{
+}
+
+void
+MPTExt::onLedgerData(model::LedgerData const& data) const
+{
+    writeMPTHoldersFromTransactions(data);
```
mathbunnyru, line 49: Maybe log trace here as well?

tests/unit/etlng/ext/MPTTests.cpp
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    ext_.onInitialData(data);
+}
+
+TEST_F(MPTExtTests, OnInitialObjectWritesMPT)
+{
+    auto const data = util::createObjectWithMPT();
+
+    EXPECT_CALL(*backend_, writeMPTHolders).WillOnce([](auto const& holders) { EXPECT_EQ(holders.size(), 1); });
```
mathbunnyru, line 154: Maybe add a test with several holders?

mathbunnyru (commented):
Some nitpicking

src/etlng/impl/ext/Cache.cpp
```diff
@@ -36,14 +36,14 @@ CacheExt::CacheExt(std::shared_ptr<CacheUpdaterInterface> cacheUpdater) : cacheU
 }
 
 void
-CacheExt::onLedgerData(model::LedgerData const& data) const
+CacheExt::onLedgerData(model::LedgerData const& data)
 {
-    cacheUpdater_->update(data);
     LOG(log_.trace()) << "got data. objects cnt = " << data.objects.size();
```
mathbunnyru, line 41: Nice, thanks 👍

mathbunnyru (dismissed):

src/etlng/impl/ext/Cache.cpp
godexsoft: Huh i think i may have deleted a bit too much there

kuznetsss (dismissed):
👍

tests/unit/etlng/ext/MPTTests.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    ext_.onInitialObject(kSEQ, data);
+}
+
+TEST_F(MPTExtTests, OnInitialDataWithMultipleHolders)
+{
+    auto const data = createMultipleHoldersTestData();
+
+    EXPECT_CALL(*backend_, writeMPTHolders).WillOnce([](auto const& holders) {
+        EXPECT_EQ(holders.size(), 3);  // Expect all three AUTHORIZE transactions
```
shawnxie999, line 164: It would be good to assert the actual holder account ID extracted like `EXPECT_EQ(holders[0], "ABCD...")`

Since all three mock transactions have the same metadata (if i'm understanding that correctly), then all three should have the same value as well

shawnxie999 (dismissed):
New logic looks good from my end. One small nit you may decide at your discretion

kuznetsss (approved):

PR #2160, opened 2025-05-27
refactor: Use std::expected instead of std::variant for errors

mathbunnyru:


PeterChen13579 (approved):
LTGM 👍. This is definitely a better change. I noticed that you can probably remove  `include variant ` in a few files like `types.hpp`, and `include expected` in `RPCHelpers.cpp`, or just let clang-tidy handle it later.

mathbunnyru:
> LTGM 👍. This is definitely a better change. I noticed that you can probably remove `include variant ` in a few files like `types.hpp`, and `include expected` in `RPCHelpers.cpp`, or just let clang-tidy handle it later.

Thanks!

Yes, I'll let clang-tidy handle it for me after the merge, as I don't want to be spending time to check where variant is still needed and where not.

godexsoft (approved):
LGTM 👍

Issue #1221, opened 2024-02-28
Check `ledger_index` is available on ETL source during forwarding (Version: 2.1.0)


godexsoft:
## Summary
Currently Clio forwards to a random ETL source. There is no checking that the `ledger_index` requested is available on said ETL source.

## Motivation
The behaviour described above can lead to extra network calls that will result in unnecessary errors.

## Solution
Be smarter about choosing the ETL source to forward to.

mathbunnyru:
This is already done: https://github.com/XRPLF/clio/blob/develop/src/etl/LoadBalancer.cpp#L335

kuznetsss:
This is not done yet. Clio should track whether ETL source is available and its ledger range for each node.

PR #2165, opened 2025-05-28
fix: Use UniformRandomGenerator class to prevent threading issue

mathbunnyru:
Work on: https://github.com/XRPLF/clio/issues/1221
It will allow to not just implement this, but also test it using new mock.

### DIFF
--- a/src/util/Random.hpp
+++ b/src/util/Random.hpp
@@ -26,10 +26,12 @@
 namespace util {
 
 /**
- * @brief Random number generator
+ * @brief Uniform random number generator
  */
-class Random {
+class UniformRandomGenerator {
 public:
+    UniformRandomGenerator();
+
     /**
      * @brief Generate a random number between min and max
      *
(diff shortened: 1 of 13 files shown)


tests/unit/etlng/LoadBalancerTests.cpp
```diff
@@ -144,14 +144,16 @@ struct LoadBalancerConstructorNgTests : util::prometheus::WithPrometheus, MockBa
...
-        return std::make_unique<LoadBalancer>(
+        auto lb = std::make_unique<LoadBalancer>(
             cfg,
             ioContext_,
             backend_,
             subscriptionManager_,
             networkManager_,
             [this](auto&&... args) -> SourcePtr { return sourceFactory_(std::forward<decltype(args)>(args)...); }
         );
+        lb->setSeed(0);
```
kuznetsss: Maybe it would be better to create an interface for `RandomGenerator` and inject mock.

mathbunnyru: I think it's out of scope of this PR and for now overcomplicates things.

mathbunnyru: @kuznetsss I won't be able to use interface here, because `uniform` function is a template

mathbunnyru: Added an interface another way around and moved this implementation to derived class.

src/util/Random.hpp
```diff
@@ -26,10 +26,50 @@
...
+    using SeedType = typename std::mt19937_64::result_type;
+
+    /**
+     * @brief Generate a random number between min and max
+     *
+     * @param min Minimum value
+     * @param max Maximum value
+     * @return Random number between min and max
+     */
+    virtual size_t
```
godexsoft, line 45: nodiscard would be suitable here

mathbunnyru: thanks, done!

godexsoft (approved):
LGTM 👍

PR #2168, opened 2025-05-30
feat: Create releases in CI

### LINKED ISSUE #1779: Create releases in CI

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/1779


.github/scripts/prepare-release-artifacts.sh
```diff
@@ -0,0 +1,18 @@
...
+
+for artifact_name in "${GITHUB_ARTIFACTS}; do
+    pushd "${artifact_name}" || exit 1
+    zip -r "../${artifact_name}.zip" ./${BINARY_NAME}
+    popd || exit 1
+
+    rm "${artifact_name}/${BINARY_NAME}"
+    rm -r "${artifact_name}"
+
+    sha256sum "./${artifact_name}.zip" > "./${artifact_name}.zip.sha256sum"
```
kuznetsss, line 23: Maybe instead we should sign binaries and provide signature files? But this could be done in a separate PR.

.github/workflows/release_impl.yml
```diff
@@ -35,29 +47,53 @@ jobs:
...
-          cp ${{ github.workspace }}/.github/workflows/${{ inputs.notes_header_file }} "${RUNNER_TEMP}/release_notes.md"
-          echo '' >> "${RUNNER_TEMP}/release_notes.md"
-          echo '```' >> "${RUNNER_TEMP}/release_notes.md"
+          echo ${{ inputs.header }} > "${RUNNER_TEMP}/release_notes.md"
+
+      - name: Generate changelog
+        shell: bash
+        if: ${{ inputs.generate_changelog }}
+        run: |
+          git-cliff
```
kuznetsss: Because of our bugfix releases usually I do: `git cliff --tag <current_release_tag> <last_commit_from_previous_release>..`
Otherwise `git-cliff` includes too many changes.

cliff.toml
```diff
@@ -8,12 +8,13 @@
...
 # template for the changelog header
 header = """
-# Changelog\n
-All notable changes to this project will be documented in this file.\n
 """
 # template for the changelog body
 # https://keats.github.io/tera/docs/#introduction
 body = """
+Version {{ version | trim_start_matches(pat="v") }} of Clio, an XRP Ledger API server optimized for HTTP and WebSocket API calls, is now available.
+This release adds new features and bug fixes.
```
kuznetsss: For bugfix releases we use `This release adds bug fixes.`. But I'm not sure whether it is easy to implement a logic analysing version difference here.

.github/scripts/prepare-release-artifacts.sh
mathbunnyru: That's a great idea, and I agree it would be better done separately, and it's also nice to have a script where it can be implemented.

mathbunnyru: https://github.com/XRPLF/clio/issues/2170

.github/workflows/release_impl.yml
mathbunnyru: Done, thanks!

cliff.toml
mathbunnyru: Thanks, I didn't notice this, I think I made it work

cliff.toml
```diff
@@ -8,18 +8,29 @@
...
 body = """
 {% if version %}\
-    ## [{{ version | trim_start_matches(pat="v") }}] - {{ timestamp | date(format="%Y-%m-%d") }}
+  Version {{ version | trim_start_matches(pat="v") }} of Clio, an XRP Ledger API server optimized for HTTP and WebSocket API calls, is now available.
+{% else %}\
+  Clio, an XRP Ledger API server optimized for HTTP and WebSocket API calls, is under active development.
+{% endif %}\
+{% set grouped_commits = commits | filter(attribute="merge_commit", value=false) | group_by(attribute="group") %}\
+{% if "<!-- 0 -->🚀 Features" in grouped_commits %}\
+  This release adds new features and bug fixes.
```
kuznetsss: Sometimes bugfix release still have new features. The only possible logic to determine the type of the release is to check which number in the new version changed.

.github/workflows/release_impl.yml
```diff
@@ -35,29 +47,54 @@ jobs:
...
-            sha256sum ./$archive_name > ./${archive_name}.sha256sum
-            cat ./$archive_name.sha256sum >> "${RUNNER_TEMP}/release_notes.md"
-          done
+      - name: Generate changelog
+        shell: bash
+        if: ${{ inputs.generate_changelog }}
+        run: |
+          LAST_TAG=$(gh release view --json tagName -q .tagName)
+          LAST_TAG_COMMIT=$(git rev-parse $LAST_TAG)
+          git-cliff --tag ${{ inputs.version }} ${LAST_TAG_COMMIT}
```
kuznetsss: It looks similar to what `git-cliff` does by default. I think I had to find the last commit in develop which got into the previous release and use it. But I'm not sure whether it's easy to express in the code.

.github/scripts/prepare-release-artifacts.sh
```diff
@@ -0,0 +1,24 @@
...
+
+for artifact_name in $(ls); do
+    pushd "${artifact_name}" || exit 1
+    zip -r "../${artifact_name}.zip" ./${BINARY_NAME}
+    popd || exit 1
+
+    rm "${artifact_name}/${BINARY_NAME}"
+    rm -r "${artifact_name}"
+
+    sha256sum "./${artifact_name}.zip" > "./${artifact_name}.zip.sha256sum"
```
kuznetsss, line 23: Maybe sha256sum should be applied to the artifact, not to the archive?

cliff.toml
mathbunnyru: Let's keep it like this for now and if needed, and manually adjust later if an error was made and new feature hasn't been marked as feat.
I don't want to implement comparison of version numbers, because it's too much work.

.github/scripts/prepare-release-artifacts.sh
mathbunnyru: That's how we do it now for nightly release.
And, if our artifact is zip file, I think it makes sense to provide the hash for it - because people might not want to even uncompress zip archive if the checksum is not valid.

cliff.toml
kuznetsss: But this is incorrect logic and we may forget to change in draft. 
For example there are new features in [2.4.1](https://github.com/XRPLF/clio/releases/tag/2.4.1), but it is a bugfix release.

Maybe better to add both versions and a reminder to remove one of them before release.

mathbunnyru: Done

.github/workflows/release_impl.yml
mathbunnyru: Done

kuznetsss (approved):

PR #2172, opened 2025-05-30
style: Beautify installation lists in Dockerfile

mathbunnyru:


godexsoft (dismissed):

kuznetsss (approved):

docker/ci/Dockerfile
```diff
@@ -29,9 +29,26 @@ RUN apt-get -qq update \
...
+        ninja-build \
+        python3 \
+        python3-pip \
+        zip \
+    && pip3 install -q --upgrade --no-cache-dir pip \
+    && pip3 install -q --no-cache-dir \
+        cmake==3.31.6 \
+        conan==1.62 \
+        gcovr \
+        pre-commit \
```
legleux, line 51: You can use heredocs in Dockerfiles now (can write to a file also).

You might think it's hideous but...

Since you're using bash you could use an array to comment why you're adding these tools also:
```dockerfile 
RUN  <<EOF
    pkgs=()
    pkgs+=(build-essential)               # required build tools
    ...
    pkgs+=(jq)                            # pretty printing
    pkgs+=(python3-pip python-is-python3) # Python and pip for Conan (and CMake)
    ...
    apt-get update && apt-get install --yes "${pkgs[@]}"
EOF
```
Maybe can lose the `&&` with 
`SHELL ["/bin/bash", "-o", "pipefail", "-o", "errexit", "-c"]`

PR #2178, opened 2025-06-02
ci: Update CI image to use Conan 2

mathbunnyru:
I pushed this image to my fork, and will use it to test everything works


godexsoft (dismissed):
Nice way to handle this with a template. Didn't know we could do this with conan 👍
It wasn't used in the original flags but do you think we may be better off adding options like '-g -O1 -fno-omit-frame-pointer'? Would that give us clearer stack-traces, wdyt? We have them in our cmake file but i noticed we don't specify the same flags for deps, hence the question.

mathbunnyru:
> Nice way to handle this with a template. Didn't know we could do this with conan 👍

Thanks!
Conan 2 has quite nice detection mechanisms and jinja2 templates.

> It wasn't used in the original flags but do you think we may be better off adding options like '-g -O1 -fno-omit-frame-pointer'? Would that give us clearer stack-traces, wdyt? We have them in our cmake file but i noticed we don't specify the same flags for deps, hence the question.

I like the idea, and it should make it better.
I won't do it in this PR, because right now we will keep the same behaviour as conan v1, but I'll try to add these flags after all conan/docker related changes are merged, it should be easy to change it afterwards.

godexsoft (approved):
:rage3:

PR #2180, opened 2025-06-02
chore: Cancellable coroutines

kuznetsss:
Add a wrap for `boost::asio::yield_context` to make async operations cancellable by default. The API could be a bit adjusted when we start switching our code to use it.


src/util/Coroutine.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <boost/asio/spawn.hpp>
+
+#include <cstddef>
+#include <memory>
+#include <utility>
+
+namespace util {
+
+Coroutine::Coroutine(
+    boost::asio::yield_context&& yield,
```
godexsoft: nit: We usually send these by copy elsewhere

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,207 @@
...
+    boost::asio::yield_context yield_;
+    boost::system::error_code error_;
+    boost::asio::cancellation_signal cancellationSignal_;
+    cancellable_yield_context_type cyield_;
+    size_t generation_;
+    std::atomic_bool isCancelled_{false};
+
+    using FamilyCancellationSignal = boost::signals2::signal<void(size_t, boost::asio::cancellation_type_t)>;
+    std::shared_ptr<FamilyCancellationSignal> familySignal_;
+    boost::signals2::connection connection_;
```
godexsoft, line 78: nit: Any reason not to use `scoped_connection`?

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,207 @@
...
+        std::shared_ptr<FamilyCancellationSignal> signal = std::make_shared<FamilyCancellationSignal>(),
+        size_t generation = 0
+    );
+
+public:
+    /**
+     * @brief Destructor for the Coroutine.
+     * Handles cleanup, such as disconnecting from the cancellation signal.
+     */
+    ~Coroutine();
```
godexsoft, line 95: nit: Should not be needed with `scoped_connection`

godexsoft (changes requested):
Looks good but let's add `nodiscard`s where applicable 👍

src/util/Coroutine.hpp
kuznetsss: I think custom destructor expressing more that disconnect should happen before fields deletion. But I'm ok to use `scoped_connection` and leave a comment instead.

src/util/Coroutine.cpp
kuznetsss: Should I change it?

src/util/Coroutine.cpp
```diff
@@ -0,0 +1,100 @@
...
+    std::shared_ptr<FamilyCancellationSignal> signal,
+    size_t generation
+)
+    : yield_(std::move(yield))
+    , cyield_(boost::asio::bind_cancellation_slot(cancellationSignal_.slot(), yield_[error_]))
```
mathbunnyru, line 37: maybe let's explicitly call `error_` constructor, to show it was constructed before `cyield_`?

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,207 @@
...
+     * @param ioContext The I/O execution context on which to spawn the coroutine.
+     * @param fn The function to be executed as the coroutine. It will receive a Coroutine& argument.
+     */
+    template <typename ExecutionContext, CoroutineFunction Fn>
+    static void
+    spawnNew(ExecutionContext& ioContext, Fn&& fn)
+    {
+        boost::asio::spawn(ioContext, [fn = std::forward<Fn>(fn)](boost::asio::yield_context yield) {
+            Coroutine thisCoroutine{std::move(yield)};
+            fn(thisCoroutine);
```
mathbunnyru, line 119: I think you should forward fn here.
And also std::move(thisCoroutine)

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * @brief Returns the error code, if any, associated with the last operation in this coroutine.
+     * @return A boost::system::error_code indicating the status.
+     */
+    boost::system::error_code
+    error() const;
+
+    /**
+     * @brief Cancels this specific coroutine and its direct children.
```
mathbunnyru: `Cancels this specific coroutine` this seems wrong

src/util/Coroutine.cpp
```diff
@@ -0,0 +1,100 @@
...
+        return;
+    familySignal_->operator()(generation_ + 1, cancellationType);
+}
+
+void
+Coroutine::cancelAll(boost::asio::cancellation_type_t cancellationType)
+{
+    if (isCancelled())
+        return;
+    familySignal_->operator()(0, cancellationType);
```
mathbunnyru: Maybe let's add this `0` as a static constexpr var?

src/util/Coroutine.cpp
mathbunnyru: I think yes, because otherwise you won't be able to use yield in the calling context twice easily

kuznetsss: But this constructor is private. The only possible calling context is static method `spawnNew()`.

src/util/Coroutine.cpp
kuznetsss: Our clang-tidy rules don't allow to call constructor if it is not required.

src/util/Coroutine.hpp
kuznetsss: > I think you should forward fn here.

Forward it to where? It is already forwarded into the lambda, which is probably a mistake. It seems it should be always accepted as rvalue and moved into the lambda.

>  And also std::move(thisCoroutine)

`fn`'s signature takes coroutine by reference.

src/util/Coroutine.hpp
kuznetsss: I also thought it is wrong, but looking at Coroutine's constructor I can say that the documentation is correct. Probably I should rename this method to `cancel()`.

src/util/Coroutine.hpp
mathbunnyru: Forward on call `std::forward<Fn>(fn)(std::move(thisCoroutione))`

src/util/Coroutine.cpp
mathbunnyru: Ah, ok, then I don't think it matters that much

src/util/Coroutine.hpp
mathbunnyru: But you pass `generation_ + 1`

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * @brief Returns the error code, if any, associated with the last operation in this coroutine.
+     * @return A boost::system::error_code indicating the status.
+     */
+    [[nodiscard]] boost::system::error_code
+    error() const;
+
+    /**
+     * @brief Cancels its direct children.
```
mathbunnyru:
```suggestion
     * @brief Cancels Coroutine's descendants.
```

mathbunnyru (changes requested):
I think this implementation is wrong.

Imagine following scenario:
1. You create coro A (gen 0)
2. It spawns B and C (gen 1)
3. They do spawn some children and so on (gen >= 2)
4. Then you call B.cancelChildren()

You expect only children of B to be cancelled, but you will also cancel children of C, because you only check the generation number.

After fixing this, test should be added as well

kuznetsss:
Thanks for finding that! 
I decided to remove that functionality for now because it probably will bring some overhead while we may not need it at all.

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * @brief Returns the error code, if any, associated with the last operation in this coroutine.
+     * @return A boost::system::error_code indicating the status.
+     */
+    [[nodiscard]] boost::system::error_code
+    error() const;
+
+    /**
+     * @brief Cancels this coroutine, all its children, and all related coroutines (siblings, parent).
```
mathbunnyru: Maybe it would be better to remove this line?
It's slightly confusing

src/util/Coroutine.hpp
```diff
@@ -0,0 +1,194 @@
...
+     * @return A cancellable_yield_context_type object.
+     */
+    [[nodiscard]] cancellable_yield_context_type
+    yieldContext() const;
+
+    /**
+     * @brief Returns the executor associated with this coroutine's yield context.
+     * @return The executor.
+     */
+    [[nodiscard]] auto
```
mathbunnyru: Why don't we return concrete type here?

kuznetsss: Because I'm not sure which type it is and probably it could be different depending on the executor used to spawn the coroutine.

mathbunnyru (approved):

godexsoft (approved):

PR #2185, opened 2025-06-04
feat: New docker images structure, tools image

mathbunnyru:
This fixes the following things:
- Docker images are now unified in style and easy to read
- they pass `hadolint` pre-commit tool, which verifies good practices for Dockerfiles
- ghcr.io is now the main source of truth for images, and is used for intermediary images (Docker Hub has limits, which we've hit, when working actively on PRs)

The main change - separate `tools` image.
It allows us not to depend on the tools provided by the OS and unifies how we build tools we depend on.
It's also only built when it needs to.
For example, we'll be able to use `mold` linker, which is significantly faster than `ld`
On top of that, if compiler provided by the OS is not able to compile, we'll be able to compile everything with our own compiler with latest features available.

Essentially, we stop depending on the OS, except for `libc` version.


docker/ci/Dockerfile
```diff
@@ -1,6 +1,9 @@
-FROM rippleci/clio_clang:16
+FROM ghcr.io/xrplf/clio-gcc:12 AS clio-gcc
+FROM ghcr.io/xrplf/clio-tools:latest AS clio-tools
+
+FROM ghcr.io/xrplf/clio-clang:16
```
godexsoft, line 4: nit: Maybe add `AS clio-clang` and put the three lines together, without extra space?

docker/ci/Dockerfile
```diff
@@ -49,12 +51,18 @@ RUN apt update -qq \
         conan==1.62 \
```
godexsoft, line 51: Is it too early to move to 2 here? Or perhaps that's in the other PR?

godexsoft (approved):

docker/ci/Dockerfile
mathbunnyru: For now, `ghcr.io/xrplf/clio-clang:16` is used as a base image for the final image, and nothing is copied from it.
It keeps it the way it was before, it might be changed in the future, but likely in a separate PR.

docker/ci/Dockerfile
mathbunnyru: Yep, it's different PR.

PR #2182, opened 2025-06-03
feat: Use new web server by default

kuznetsss:
Fixes #1781.

PR #2187, opened 2025-06-05
chore: Revert "feat: Use new web server by default (#2182)"

kuznetsss:
There is an issue found in the new web server, so we couldn't use it by default for now.
This reverts commit b3f3259b14ebd497563f584a4546481bc0898acf.


godexsoft (approved):
😢 :rage4:

PR #2189, opened 2025-06-05
feat: Pass sanitizer as part of conan_profile

mathbunnyru:
I noticed we don't need `sanitizer` value anymore, so removed it.


godexsoft (approved):
Looks good 👍

kuznetsss (approved):

PR #2199, opened 2025-06-08
feat: Read-write switching in ETLng

### LINKED ISSUE #1597: ETL ng loader

godexsoft:
Fixes #1597 

This PR implements a mechanism for multiple ETL nodes to coordinate who is the writer at any given point.
The mechanism is mirroring the previous ETL's implementation relying on database conflict on write and monitoring the DB for newly added ledgers. Inactivity is detected after certain delay which triggers the node to attempt and become a writer.

This should be the last PR needed for the complete Loader implementation


src/etlng/ETLService.cpp
```diff
@@ -96,9 +98,14 @@ ETLService::ETLService(
     , loader_(std::move(loader))
     , initialLoadObserver_(std::move(initialLoadObserver))
     , taskManagerProvider_(std::move(taskManagerProvider))
+    , monitorProvider_(std::move(monitorProvider))
     , state_(std::move(state))
 {
     LOG(log_.info()) << "Creating ETLng...";
+    state_->isReadOnly = config_.get().get<bool>("read_only");
+    state_->isWriting = false;  // regardless of write/read mode we need to start as a reader
```
mathbunnyru: Maybe we should fill these values few levels above when creating state initially?

src/etlng/ETLService.cpp
```diff
@@ -238,28 +248,64 @@ ETLService::loadInitialLedgerIfNeeded()
...
+    auto rng = backend_->hardFetchLedgerRangeNoThrow();
+    ASSERT(rng.has_value(), "Ledger range can't be null");
+
+    state_->isWriting = true;  // switch to writer
+    LOG(log_.info()) << "Taking over the ETL writer seat";
+    startLoading(rng->maxSequence + 1);
+}
+
+void
+ETLService::giveupWriter()
```
kuznetsss: nit:
```suggestion
ETLService::giveUpWriter()
```

src/etlng/MonitorInterface.hpp
```diff
@@ -54,7 +62,16 @@ class MonitorInterface {
...
+    subscribeToNewSequence(NewSequenceSignalType::slot_type const& subscriber) = 0;
+
+    /**
+     * @brief Allows clients to get notified when no database update is detected for a configured period.
+     *
+     * @param subscriber The slot to connect
+     * @return A connection object that automatically disconnects the subscription once destroyed
+     */
+    [[nodiscard]] virtual boost::signals2::scoped_connection
+    subscribeToNoDbUpdate(NoDbUpdateSignalType::slot_type const& subscriber) = 0;
```
kuznetsss:
```suggestion
    subscribeToDbStale(NoDbUpdateSignalType::slot_type const& subscriber) = 0;
```

src/etlng/impl/Loading.cpp
```diff
@@ -46,29 +49,45 @@ namespace etlng::impl {
...
-            ::util::timed<std::chrono::duration<double>>([&]() { return backend_->finishWrites(data.seq); });
-        LOG(log_.info()) << "Finished writes to DB for " << data.seq << ": " << (success ? "YES" : "NO") << "; took "
-                         << duration;
+        // Only a writer should attempt to commit to DB
+        // This is also where conflicts with other writer nodes will be detected
+        if (state_->isWriting) {
+            auto [success, duration] =
+                ::util::timed<std::chrono::duration<double>>([&]() { return backend_->finishWrites(data.seq); });
+            LOG(log_.info()) << "Finished writes to DB for " << data.seq << ": " << (success ? "YES" : "NO")
+                             << "; took " << duration;
```
kuznetsss: Would be good to add unit, e.g. `ms`.

src/etlng/impl/Loading.cpp
```diff
@@ -46,29 +49,45 @@ namespace etlng::impl {
...
 
-        auto [success, duration] =
-            ::util::timed<std::chrono::duration<double>>([&]() { return backend_->finishWrites(data.seq); });
-        LOG(log_.info()) << "Finished writes to DB for " << data.seq << ": " << (success ? "YES" : "NO") << "; took "
-                         << duration;
+        // Only a writer should attempt to commit to DB
+        // This is also where conflicts with other writer nodes will be detected
+        if (state_->isWriting) {
+            auto [success, duration] =
+                ::util::timed<std::chrono::duration<double>>([&]() { return backend_->finishWrites(data.seq); });
```
kuznetsss: Why double but not milliseconds or microseconds?

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -306,6 +339,142 @@ TEST_F(ETLServiceTests, WaitForValidatedLedgerIsAborted)
...
+    EXPECT_CALL(*monitorProvider_, make(testing::_, testing::_, testing::_, testing::_, testing::_))
+        .WillOnce([&mockMonitor](auto, auto, auto, auto, auto) { return std::move(mockMonitor); });
+
+    EXPECT_CALL(*mockMonitor, subscribeToNewSequence(testing::_)).WillOnce([&capturedCallback](auto&& callback) {
+        capturedCallback = callback;
+        return boost::signals2::scoped_connection{};
+    });
+    EXPECT_CALL(*mockMonitor, subscribeToNoDbUpdate(testing::_))
+        .WillOnce(testing::Return(boost::signals2::scoped_connection{}));
+    EXPECT_CALL(*mockMonitor, run(testing::_));
```
kuznetsss: nit:
```suggestion
    EXPECT_CALL(*mockMonitor, subscribeToNoDbUpdate);
    EXPECT_CALL(*mockMonitor, run);
```

src/etlng/ETLService.cpp
```diff
@@ -238,28 +248,64 @@ ETLService::loadInitialLedgerIfNeeded()
...
-    monitor_ = std::make_unique<impl::Monitor>(ctx_, backend_, ledgers_, seq);
-    monitorSubscription_ = monitor_->subscribe([this](uint32_t seq) {
-        log_.info() << "MONITOR got new seq from db: " << seq;
+    monitor_ = monitorProvider_->make(ctx_, backend_, ledgers_, seq);
+
+    monitorSubscription_ = monitor_->subscribeToNewSequence([this](uint32_t seq) {
+        LOG(log_.info()) << "ETLService (via Monitor) got new seq from db: " << seq;
+
+        if (state_->writeConflict) {
+            LOG(log_.warn()) << "Got a write conflict; Stopping task manager for time being...";
```
mathbunnyru: What do you mean by `Stopping task manager for time being`?

src/etlng/ETLService.cpp
```diff
@@ -238,28 +248,64 @@ ETLService::loadInitialLedgerIfNeeded()
...
+            giveupWriter();
```
mathbunnyru:
```suggestion
            giveUpWriter();
```

src/etlng/ETLService.hpp
```diff
@@ -106,12 +107,14 @@ class ETLService : public ETLServiceInterface {
...
     std::shared_ptr<InitialLoadObserverInterface> initialLoadObserver_;
     std::shared_ptr<etlng::TaskManagerProviderInterface> taskManagerProvider_;
+    std::shared_ptr<etlng::MonitorProviderInterface> monitorProvider_;
     std::shared_ptr<etl::SystemState> state_;
 
     std::unique_ptr<MonitorInterface> monitor_;
     std::unique_ptr<TaskManagerInterface> taskMan_;
 
     boost::signals2::scoped_connection monitorSubscription_;
+    boost::signals2::scoped_connection monitorNoDbUpdateSubscription_;  // New subscription
```
mathbunnyru:
```suggestion
    boost::signals2::scoped_connection monitorNoDbUpdateSubscription_;
```

mathbunnyru (changes requested):

src/etlng/impl/TaskManager.cpp
```diff
@@ -102,29 +103,36 @@ TaskManager::spawnExtractor(TaskQueue& queue)
...
         while (not stopRequested) {
             // TODO (https://github.com/XRPLF/clio/issues/66): does not tell the loader whether it's out of order or not
             if (auto data = queue.dequeue(); data.has_value()) {
-                auto nanos = util::timed<std::chrono::nanoseconds>([this, data = *data] { loader_.get().load(data); });
-                auto const seconds = nanos / kNANO_TO_SECOND;
+                // perhaps this should return an error if conflict happened, then we can stop loading immediately
+                auto [maybeSuccess, nanos] =
+                    util::timed<std::chrono::nanoseconds>([this, data = *data] { return loader_.get().load(data); });
+
+                if (not maybeSuccess) {
```
mathbunnyru:
```suggestion
                if (not maybeSuccess.has_value()) {
```

src/etlng/ETLService.cpp
godexsoft: Maybe the log is confusing 😃 It means that we stop the manager for now as we transition to readonly. Once we transition back to writer we "restart" the task manager by creating a new instance. I'll try to find better wording - suggestions welcome 👍

src/etlng/impl/Loading.cpp
godexsoft: Not sure but let's change 👍

mathbunnyru (changes requested):

src/etl/ETLService.cpp
```diff
@@ -86,13 +87,15 @@ ETLService::makeETLService(
         );
 
         auto state = std::make_shared<etl::SystemState>();
+        state->isReadOnly = config.get<bool>("read_only");
```
mathbunnyru: maybe make it a class and add ctor with bool?

src/etlng/ETLService.cpp
```diff
@@ -96,9 +98,11 @@ ETLService::ETLService(
     , loader_(std::move(loader))
     , initialLoadObserver_(std::move(initialLoadObserver))
     , taskManagerProvider_(std::move(taskManagerProvider))
+    , monitorProvider_(std::move(monitorProvider))
     , state_(std::move(state))
 {
-    LOG(log_.info()) << "Creating ETLng...";
+    ASSERT(not state_->isWriting, "ETL should never start in writer mode");
+    LOG(log_.info()) << "Starting in " << (state_->isReadOnly ? "STRICT READONLY MODE" : "WRITE MODE");
```
mathbunnyru: isReadOnly -> isStrictReadOnly

src/etlng/ETLService.cpp
```diff
@@ -238,28 +245,64 @@ ETLService::loadInitialLedgerIfNeeded()
...
             });
+
             cacheUpdater_->update(seq, diff);
+            backend_->updateRange(seq);
         }
 
         publisher_->publish(seq, {});
     });
+
+    monitorDbStaledSubscription_ = monitor_->subscribeToDbStaled([this]() {
```
mathbunnyru: Probably, Staled -> Stale, because you mean an adjective

src/etlng/ETLService.cpp
```diff
@@ -238,28 +245,64 @@ ETLService::loadInitialLedgerIfNeeded()
...
+        LOG(log_.warn()) << "ETLService received NoDbUpdate signal from Monitor";
```
mathbunnyru:
```suggestion
        LOG(log_.warn()) << "ETLService received DbStale signal from Monitor";
```

src/etlng/MonitorInterface.hpp
```diff
@@ -36,7 +36,8 @@ namespace etlng {
 class MonitorInterface {
 public:
     static constexpr auto kDEFAULT_REPEAT_INTERVAL = std::chrono::seconds{1};
-    using SignalType = boost::signals2::signal<void(uint32_t)>;
+    using NewSequenceSignalType = boost::signals2::signal<void(uint32_t)>;
+    using NoDbUpdateSignalType = boost::signals2::signal<void()>;
```
mathbunnyru: Probably worth searching for `NoDbUpdate` and changing it to `DbStale`

src/etlng/MonitorProviderInterface.hpp
```diff
@@ -0,0 +1,64 @@
...
+namespace etlng {
+
+/**
+ * @brief An interface for providing Monitor instances
+ */
+struct MonitorProviderInterface {
+    /**
+     * @brief The time Monitor should wait before reporting absence of updates to the database
+     */
+    static constexpr auto kDEFAULT_NO_DB_UPDATE_REPORT_DELAY = std::chrono::seconds{10};
```
mathbunnyru: Also, search for `NO_DB_UPDATE`

src/etlng/impl/Loading.cpp
```diff
@@ -46,29 +49,45 @@ namespace etlng::impl {
...
-                         << duration;
+        // Only a writer should attempt to commit to DB
+        // This is also where conflicts with other writer nodes will be detected
+        if (state_->isWriting) {
+            auto [success, duration] =
+                ::util::timed<std::chrono::milliseconds>([&]() { return backend_->finishWrites(data.seq); });
+            LOG(log_.info()) << "Finished writes to DB for " << data.seq << ": " << (success ? "YES" : "NO")
+                             << "; took " << duration << "ms";
+
+            if (not success) {
```
mathbunnyru, line 78: Maybe we should check that our failure is actually a failure because of conflict?
What happens, if the failure is for networking reason?

src/etlng/impl/Loading.cpp
```diff
@@ -46,29 +49,45 @@ namespace etlng::impl {
...
+                state_->writeConflict = true;
+                return std::unexpected("write conflict");
+                LOG(log_.warn()) << "Another node wrote a ledger into the DB - we have a write conflict";
```
mathbunnyru, line 80: Something after return, this shouldn't even compile - could you please add compiler flags so this would be checked?

src/etlng/impl/Loading.cpp
```diff
@@ -46,29 +49,45 @@ namespace etlng::impl {
...
+        // This is also where conflicts with other writer nodes will be detected
+        if (state_->isWriting) {
+            auto [success, duration] =
+                ::util::timed<std::chrono::milliseconds>([&]() { return backend_->finishWrites(data.seq); });
+            LOG(log_.info()) << "Finished writes to DB for " << data.seq << ": " << (success ? "YES" : "NO")
+                             << "; took " << duration << "ms";
+
+            if (not success) {
+                state_->writeConflict = true;
+                return std::unexpected("write conflict");
```
mathbunnyru: Shouldn't be a string, but enum, maybe worth fixing in this PR?

src/etlng/impl/Loading.cpp
```diff
@@ -78,13 +97,32 @@ Loader::onInitialLoadGotMoreObjects(
     std::optional<std::string> lastKey
 )
 {
+    static constexpr std::size_t kLOG_INTERVAL = 1000u;
```
mathbunnyru:
```suggestion
    static constexpr std::size_t kLOG_EVERY = 1000u;
```

I think interval is usually time

src/etlng/impl/Loading.cpp
```diff
@@ -78,13 +97,32 @@ Loader::onInitialLoadGotMoreObjects(
...
+    static auto kINITIAL_LOAD_START_TIME = std::chrono::steady_clock::now();
```
mathbunnyru, line 101: Should be a field

src/etlng/impl/Loading.cpp
```diff
@@ -78,13 +97,32 @@ Loader::onInitialLoadGotMoreObjects(
...
         registry_->dispatchInitialObjects(
             seq,
             data,
             std::move(lastKey).value_or(std::string{})  // TODO: perhaps use optional all the way to extensions?
         );
+
+        initialLoadWrittenObjects_ += data.size();
+        ++initialLoadWrites_;
+        if (initialLoadWrites_ % kLOG_INTERVAL == 0u && initialLoadWrites_ != 0u) {
+            auto totalElapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
```
mathbunnyru: I wouldn't call it `total`

src/etlng/impl/Monitor.cpp
```diff
@@ -80,28 +105,65 @@ Monitor::stop()
...
 void
 Monitor::doWork()
 {
-    if (auto rng = backend_->hardFetchLedgerRangeNoThrow(); rng) {
-        while (rng->maxSequence >= nextSequence_)
+    auto rng = backend_->hardFetchLedgerRangeNoThrow();
+    bool dbProgressedThisCycle = false;
+    auto lck = updateData_.lock();
+
+    if (rng) {
```
mathbunnyru:
```suggestion
    if (rng.has_value()) {
```

mathbunnyru (changes requested):

src/etlng/impl/Loading.cpp
godexsoft: This could be a field but we need to start counting from the first call to this function so it will be awkward to do.. and this is only ever called during initial ledger load which is a one time process, hence the use of a static

src/etlng/impl/Loading.cpp
godexsoft: At this point it is perfectly fine as a string but i added the Error typedef so we can switch easier if it needs to change in the future

src/etlng/impl/Loading.cpp
godexsoft: I must have rearranged something by mistake 🤣

src/etlng/impl/Loading.cpp
godexsoft: It's sort of the same result.. i'll see what i can do

src/etlng/ETLService.cpp
godexsoft: I think the word i was thinking of is actually "stalled", not stale or staled. But we can think about it more

src/etl/ETLService.cpp
godexsoft: I'm not sure what you are trying to achieve with that? `isReadOnly` is a prometheus counter which makes everything a bit more awkward in general. We are also setting and reading this property in few places (including in old ETL for now) so we can't encapsulate the property (technically we can but it makes no sense). I don't see value in making it a class nor in adding a constructor.. but maybe i'm misunderstanding your suggestion

src/etlng/impl/Loading.cpp
godexsoft: Stride is probably the better term here

src/etlng/impl/Loading.cpp
godexsoft: Turns out we did not have the `-Wunreachable-code` option but adding it did not help because `LOG` is a macro and clang simply does not report this thinking it may be a false positive. We should add this option and something for clang-tidy in a separate PR: #2207

src/etlng/impl/Loading.cpp
godexsoft: It appears that the `finishWrites` function retries indefinitely if there is some error other than a write conflict. And there are no exceptions going on there - it's a future with std::expected as result. Judging by the code we can only get here if there was a write conflict

src/etlng/impl/TaskManager.cpp
```diff
@@ -102,29 +103,36 @@ TaskManager::spawnExtractor(TaskQueue& queue)
...
-
     return ctx_.execute([this, &queue](auto stopRequested) {
         while (not stopRequested) {
             // TODO (https://github.com/XRPLF/clio/issues/66): does not tell the loader whether it's out of order or not
             if (auto data = queue.dequeue(); data.has_value()) {
-                auto nanos = util::timed<std::chrono::nanoseconds>([this, data = *data] { loader_.get().load(data); });
-                auto const seconds = nanos / kNANO_TO_SECOND;
+                // perhaps this should return an error if conflict happened, then we can stop loading immediately
+                auto [expectedSuccess, nanos] =
+                    util::timed<std::chrono::nanoseconds>([this, data = *data] { return loader_.get().load(data); });
```
mathbunnyru: It seems that we create an unneeded copy here

cmake/Settings.cmake
```diff
@@ -1,20 +1,20 @@
 set(COMPILER_FLAGS
+    -pedantic
```
mathbunnyru, line 2: Let's not include these cmake changes here at all and deal with them as part of separate PR

src/etlng/impl/Monitor.cpp
```diff
@@ -80,28 +103,65 @@ Monitor::stop()
...
+
+    if (dbProgressedThisCycle) {
+        lck->lastDbProgressTime = std::chrono::steady_clock::now();
+    } else if (std::chrono::steady_clock::now() - lck->lastDbProgressTime > lck->dbStalledReportDelay) {
+        LOG(log_.info()) << "No DB update detected for "
+                         << std::chrono::duration_cast<std::chrono::seconds>(lck->dbStalledReportDelay).count()
+                         << " seconds. Firing dbStalledChannel. Last seen max seq in DB: " << lck->lastSeenMaxSeqInDb
+                         << ". Expecting next: " << nextSequence_;
+        dbStalledChannel_();
+        lck->lastDbProgressTime = std::chrono::steady_clock::now();
```
mathbunnyru: I think `lastDbProgressTime` is not correct, because there was no progress here.
Maybe `lastDbCheckTime`?

src/etlng/impl/TaskManager.cpp
```diff
@@ -102,29 +103,36 @@ TaskManager::spawnExtractor(TaskQueue& queue)
...
+
+                if (not expectedSuccess.has_value()) {
+                    LOG(log_.warn()) << "Immediately stopping loader: " << expectedSuccess.error()
```
mathbunnyru:
```suggestion
                    LOG(log_.warn()) << "Immediately stopping loader with error: " << expectedSuccess.error()
```

src/etlng/impl/TaskManager.cpp
```diff
@@ -102,29 +103,36 @@ TaskManager::spawnExtractor(TaskQueue& queue)
...
+                                     << "; latest ledger cache loaded for " << data->seq;
+                    monitor_.get().notifyWriteConflict(data->seq);
+                    break;
+                }
+
+                auto const seconds = nanos / util::kNANO_PER_SECOND;
```
mathbunnyru, line 135: If it usually takes less than a second, then we should use double.
If not, it would be better to remove constans and use `std::chrono::floor`

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -159,7 +190,7 @@ struct ETLServiceTests : util::prometheus::WithPrometheus, MockBackendTest {
         {"cache.page_fetch_size", ConfigValue{ConfigType::Integer}.defaultValue(512)},
         {"cache.load", ConfigValue{ConfigType::String}.defaultValue("async")}
     };
-    StrictMockSubscriptionManagerSharedPtr subscriptions_;
```
mathbunnyru, line 162: Why not strict?

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -261,62 +295,193 @@ TEST_F(ETLServiceTests, RunWithEmptyDatabase)
...
 
 TEST_F(ETLServiceTests, RunWithPopulatedDatabase)
 {
-    auto mockTaskManager = std::make_unique<testing::NiceMock<MockTaskManager>>();
-
-    EXPECT_CALL(*backend_, hardFetchLedgerRange(testing::_))
+    EXPECT_CALL(*backend_, hardFetchLedgerRange)
         .WillRepeatedly(testing::Return(data::LedgerRange{.minSequence = 1, .maxSequence = kSEQ}));
+    EXPECT_CALL(*monitorProvider_, make(testing::_, testing::_, testing::_, testing::_, testing::_))
+        .WillOnce([](auto, auto, auto, auto, auto) { return std::make_unique<testing::NiceMock<MockMonitor>>(); });
```
mathbunnyru, line 322: nice :)

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -261,62 +295,193 @@ TEST_F(ETLServiceTests, RunWithEmptyDatabase)
...
     service_.run();
 }
 
+TEST_F(ETLServiceTests, HandlesWriteConflictInMonitorSubscription)
+{
+    auto mockMonitor = std::make_unique<testing::NiceMock<MockMonitor>>();
+    std::function<void(uint32_t)> capturedCallback;
+
+    EXPECT_CALL(*monitorProvider_, make(testing::_, testing::_, testing::_, testing::_, testing::_))
+        .WillOnce([&mockMonitor](auto, auto, auto, auto, auto) { return std::move(mockMonitor); });
```
mathbunnyru: Probably UB: you're moving from mockMonitor, but then you're checking EXPECT_CALL on it.

You can probably save reference to the created object and use it for your `EXPECT_CALL`s.

The same is true for other tests as well.

mathbunnyru (changes requested):

tests/unit/etlng/ETLServiceTests.cpp
godexsoft: It's not needed

tests/unit/etlng/ETLServiceTests.cpp
godexsoft: Strongly doubt it is actually UB but it is easy enough to create a ref to be sure

src/etlng/impl/Loading.cpp
mathbunnyru: Ok

mathbunnyru (approved):

PR #2204, opened 2025-06-10
fix: Only set compile flag for grpc

mathbunnyru:
I built all conan packages locally and this flag is only required for grpc, so let's only set it for grpc.
This is better - it's explicit, and we'll know that if we update grpc recipe, we can remove this.

I also uploaded all rebuilt packages to the artifactory.

### DIFF
--- a/docs/build-clio.md
+++ b/docs/build-clio.md
@@ -37,7 +37,7 @@ compiler.version=16
 os=Macos
 
 [conf]
-tools.build:cxxflags+=["-Wno-missing-template-arg-list-after-template-kw"]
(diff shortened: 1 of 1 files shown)


godexsoft (approved):
Good idea. How can we make sure that we won't forget to update/remove this when we update the version? Do we have an issue for this?

mathbunnyru:
> Good idea. How can we make sure that we won't forget to update/remove this when we update the version? Do we have an issue for this?

Decided to not create an issue for this change, even if we forget to remove from docs for some time - there will be no harm.

PR #2212, opened 2025-06-11
feat: Build GCC natively and then merge the image

mathbunnyru:
Checked in my fork, that there is a merged image with expected tags


godexsoft (approved):
👍

PR #2217, opened 2025-06-12
ci: Upload conan deps for all profiles

mathbunnyru:


.github/workflows/upload_conan_deps.yml
```diff
@@ -0,0 +1,81 @@
...
+      - name: Run conan and cmake
+        uses: ./.github/actions/generate
+        with:
+          conan_profile: ${{ env.CONAN_PROFILE }}
+          conan_cache_hit: "false"
+          build_type: Release
+
+      - name: Login to Conan
+        if: github.event_name != 'pull_request'
+        run: conan remote login -p ${{ secrets.CONAN_PASSWORD }} ripple ${{ secrets.CONAN_USERNAME }}
```
godexsoft, line 77: This is not going to plaintext the secret in any shape in the log right?

mathbunnyru: GitHub does it automatically: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#redacting-secrets-from-workflow-run-logs

godexsoft (approved):

PR #2220, opened 2025-06-12
feat: Add conan lockfile

mathbunnyru:


godexsoft (approved):

conan.lock
```diff
@@ -0,0 +1,57 @@
...
+        ],
+        "lz4/1.9.4": [
+            "lz4/1.10.0"
+        ],
+        "sqlite3/3.44.2": [
+            "sqlite3/3.47.0"
+        ]
+    },
+    "config_requires": []
+}
```
godexsoft, line 57: Please add newline to make gh happy 🙏

godexsoft (changes requested):

conan.lock
mathbunnyru: I don't like it too.
Unfortunately, that's how Conan generates this file 😑 
I usually prefer not to change auto generated files, to make changes easier (we will definitely need to update this file from time to time).

godexsoft (approved):

PR #2225, opened 2025-06-13
feat: Build all possible conan configurations in CI

mathbunnyru:


.github/workflows/upload_conan_deps.yml
```diff
@@ -70,7 +65,7 @@ jobs:
         with:
           conan_profile: ${{ env.CONAN_PROFILE }}
           conan_cache_hit: "false"
-          build_type: Release
```
mathbunnyru, line 73: This was a bug, because initially I didn't notice we have a debug build, so added this to matrix later and forgot to use the value here

godexsoft (approved):
👍

PR #2232, opened 2025-06-17
fix: ETLng cleanup and graceful shutdown

godexsoft:
This PR implements graceful shutdown during initial ledger load. This is a bit of an edge case but good to have for completeness and robustness of the implementation.
Also in this PR some missing tests are added and general cleanup of the new ETL is performed.


src/etlng/ETLService.cpp
```diff
@@ -223,13 +230,21 @@ ETLService::loadInitialLedgerIfNeeded()
...
-
-                    // TODO: this should be interruptible for graceful shutdown
-                    return loader_->loadInitialLedger(data);
-                });
+                return extractor_->extractLedgerOnly(seq).and_then(
+                    [this, seq](auto&& data) -> std::optional<ripple::LedgerHeader> {
+                        // TODO: loadInitialLedger in balancer should be called fetchEdgeKeys or similar
+                        auto res = balancer_->loadInitialLedger(seq, *initialLoadObserver_);
+                        if (not res.has_value() and res.error() == InitialLedgerLoadError::Cancel) {
+                            LOG(log_.info()) << "Initial ledger load got cancelled";
```
mathbunnyru: Maybe warn here as well?

src/etlng/LoadBalancer.cpp
```diff
@@ -210,30 +210,34 @@ LoadBalancer::LoadBalancer(
...
-
-            if (!res) {
-                LOG(log_.error()) << "Failed to download initial ledger."
-                                  << " Sequence = " << sequence << " source = " << source->toString();
-            } else {
-                response = std::move(data);
+            auto res = source->loadInitialLedger(sequence, downloadRanges_, loadObserver);
+
+            if (not res.has_value()) {
+                if (res.error() == InitialLedgerLoadError::Error) {
```
mathbunnyru: Let's do a switch without `default` here, because we might add new cases, and compiler will tell us we forgot about them

mathbunnyru: and consider it for other places

src/etlng/LoadBalancerInterface.hpp
```diff
@@ -39,6 +39,20 @@
 
 namespace etlng {
 
+/**
+ * @brief Represents possible errors for initial ledger load
+ */
+enum class InitialLedgerLoadError {
+    Cancel, /*< Indicating the initial load got cancelled by user */
+    Error,  /*< Indicating some error happened during initial ledger load */
```
mathbunnyru: Maybe we could put comments as part of doxygen?
Or will they be generated already?

src/etlng/LoadBalancerInterface.hpp
```diff
@@ -39,6 +39,20 @@
...
+};
+
+/**
+ * @brief The result type of the initial ledger load
+ * @note The value channel represents edge keys
```
mathbunnyru: channel?

src/etlng/LoaderInterface.hpp
```diff
@@ -25,11 +25,16 @@
...
 #include <optional>
-#include <string>
 
 namespace etlng {
 
-using Error = std::string;
+/**
+ * @brief Enumeration of possible errors that can occur during loading operations
+ */
+enum class LoaderError {
```
mathbunnyru, line 34: Nice, that's what I was talking about int the previous PR 🙂

src/etlng/impl/TaskManager.cpp
```diff
@@ -121,15 +121,21 @@ TaskManager::spawnLoader(TaskQueue& queue)
...
+                        LOG(log_.warn()) << "Immediately stopping loader on write conflict"
+                                         << "; latest ledger cache loaded for " << data->seq;
+                        monitor_.get().notifyWriteConflict(data->seq);
+                        break;
+                    }
+
+                    if (expectedSuccess.error() == LoaderError::AmendmentBlocked) {
+                        LOG(log_.warn()) << "Immediately stopping loader on amendment block";
+                        break;
+                    }
```
mathbunnyru: definitely switch here

mathbunnyru (changes requested):
Overall, looks good

src/etlng/ETLService.cpp
godexsoft: I was considering to actually lower this log to debug not to pollute the log with similar messages

src/etlng/LoadBalancer.cpp
godexsoft: But there is only one interesting case here. I don't see the benefit in this case.
In the other one switch aint gonna work because we use `break` there, hence no switch

src/etlng/LoadBalancerInterface.hpp
godexsoft: These are doxygen comments.

src/etlng/LoadBalancerInterface.hpp
godexsoft: Expected got value and error, they are commonly called "channels".

src/etlng/LoaderInterface.hpp
godexsoft: It made sense to change from string because of the difference in how we treat these two 👍

src/etlng/LoadBalancerInterface.hpp
mathbunnyru: Cool 👍

src/etlng/impl/TaskManager.cpp
godexsoft: Feel free to propose how to change it so it's a switch 😃

src/etlng/ETLService.cpp
mathbunnyru: I think it's reasonable

src/etlng/impl/TaskManager.cpp
mathbunnyru: Yes, you're right, no good way right now

src/etlng/ETLService.cpp
```diff
@@ -223,13 +230,21 @@ ETLService::loadInitialLedgerIfNeeded()
...
-                    data.edgeKeys = balancer_->loadInitialLedger(seq, *initialLoadObserver_);
-
-                    // TODO: this should be interruptible for graceful shutdown
-                    return loader_->loadInitialLedger(data);
-                });
+                return extractor_->extractLedgerOnly(seq).and_then(
+                    [this, seq](auto&& data) -> std::optional<ripple::LedgerHeader> {
+                        // TODO: loadInitialLedger in balancer should be called fetchEdgeKeys or similar
+                        auto res = balancer_->loadInitialLedger(seq, *initialLoadObserver_);
+                        if (not res.has_value() and res.error() == InitialLedgerLoadError::Cancel) {
```
kuznetsss: Shouldn't switch-case here be used to make sure we have all the possible errors covered?

Also maybe error name could be `Cancelled` or `OperationAborted`?

src/etlng/LoadBalancer.cpp
mathbunnyru: Got it, I think you can combine 2 if statements in 1 here though

src/etlng/LoadBalancerInterface.hpp
kuznetsss: Channel is confusing for me as well. Usually channel is a thread safe queue.

src/etlng/impl/GrpcSource.cpp
```diff
@@ -131,9 +136,9 @@ GrpcSource::loadInitialLedger(
         ASSERT(tag != nullptr, "Tag can't be null.");
         auto ptr = static_cast<AsyncGrpcCall*>(tag);
 
-        if (!ok) {
-            LOG(log_.error()) << "loadInitialLedger - ok is false";
-            return {{}, false};  // cancelled
+        if (not ok or initialLoadShouldStop_->load()) {
```
kuznetsss: Shouldn't we check the flag at the beginning of the method as well if somehow `loadInitialLedger()` would be called after `stop()`? Or is it not possible?

src/etlng/impl/TaskManager.cpp
kuznetsss: I agree with @mathbunnyru, if you are using an `enum`, it should be processed with a `switch`. It will save us from forgetting to cover new enum's variants in the future.

src/etlng/ETLService.cpp
godexsoft: We just had the same convo with @mathbunnyru - we are only interested in one case here so it does not make sense to use switch imo. (if the logic changes we can always adjust).

src/etlng/LoadBalancerInterface.hpp
godexsoft: This is not golang 😃

godexsoft: On a more serious note, what would you call it so i can adjust? @kuznetsss @mathbunnyru

src/etlng/impl/GrpcSource.cpp
godexsoft: No harm in adding it next to the stub check i guess

src/etlng/LoadBalancerInterface.hpp
mathbunnyru: The successful value represents edge keys

src/etlng/impl/TaskManager.cpp
godexsoft: @kuznetsss i challenge you to implement it then 😃

godexsoft: Fine, let's use immediately invoked lambda with a switch inside. not super clean but addresses your concerns

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -488,9 +489,7 @@ TEST_F(ETLServiceTests, GiveUpWriterAfterWriteConflict)
     EXPECT_FALSE(systemState_->writeConflict);  // and removes write conflict flag
 }
 
-struct ETLServiceAssertTests : common::util::WithMockAssert, ETLServiceTests {};
```
mathbunnyru, line 491: Why did you remove this?

src/etlng/ETLService.cpp
```diff
@@ -136,7 +139,11 @@ ETLService::run()
             return;
         }
 
-        ASSERT(rng.has_value(), "Ledger range can't be null");
+        if (not rng.has_value()) {
+            LOG(log_.warn()) << "Initial ledger download got cancelled - stopping ETL service";
```
mathbunnyru, line 143: Don't we have to check the error here?

src/etlng/impl/GrpcSource.cpp
```diff
@@ -131,9 +140,9 @@ GrpcSource::loadInitialLedger(
         ASSERT(tag != nullptr, "Tag can't be null.");
         auto ptr = static_cast<AsyncGrpcCall*>(tag);
 
-        if (!ok) {
-            LOG(log_.error()) << "loadInitialLedger - ok is false";
-            return {{}, false};  // cancelled
+        if (not ok or *initialLoadShouldStop_) {
```
mathbunnyru, line 143: I don't like having `ok` var here, because it looks like we're just hiding errors, maybe there is a better name?

tests/unit/etlng/ETLServiceTests.cpp
```diff
@@ -501,10 +500,10 @@ TEST_F(ETLServiceAssertTests, FailToLoadInitialLedger)
     EXPECT_CALL(*loader_, loadInitialLedger).Times(0);
     EXPECT_CALL(*taskManagerProvider_, make).Times(0);
 
-    EXPECT_CLIO_ASSERT_FAIL({ service_.run(); });
```
mathbunnyru, line 504: Why this change?

mathbunnyru (approved):
Left some questions - if you think these places are ok, don't want to prevent you from merging, so 👍

tests/unit/etlng/ETLServiceTests.cpp
godexsoft: Because these asserts are gone too now that we can potentially stop the loading of initial ledger 👍

src/etlng/ETLService.cpp
godexsoft: The only reason to get here is cancellation - there is an assert below checking for invalid logic. Any error inside should trigger retry until success or the end of the universe, whichever comes first

tests/unit/etlng/ETLServiceTests.cpp
godexsoft: Assert is gone

src/etlng/impl/GrpcSource.cpp
godexsoft: I don't like the entire grpc way of async downloading and frankly idk what this 'ok' is - it copied over from the old logic and i don't know enough about it to confidently change it atm

PR #2242, opened 2025-06-18
feat: Add init_conan script and improve docs

mathbunnyru:
This should make life of a developer so much easier


.github/scripts/conan/init.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+set -ex
+
+CURRENT_DIR=$(dirname "$0")
+REPO_DIR=$(cd "$CURRENT_DIR/../../../" && pwd)
```
kuznetsss: Are you sure it is 3 `..`, not 2?

mathbunnyru: Yes, `.github/scripts/conan/` is 3 levels of dirs.
And the script would break later when copying file if the wrong dir were chosen (since `-e` flag)

docs/build-clio.md
```diff
@@ -19,6 +19,10 @@
 
 ### Conan Configuration
 
+> [!TIP]
+> To setup Conan automatically, you can run `.github/scripts/conan/init.sh`.
+> This will delete ~/.conan2 directory, if it exists, set up profiles and add Artifactory remote
```
godexsoft: This makes me think that using default profile is perhaps not a very nice UX in some cases - if someone already uses conan to build some other projects for example

mathbunnyru: Well, I'm saying explicitly what this does, so if you're already using Conan, you will know, that this script is not for you.

Script also supports $CONAN_HOME env dir, which allows to share/not share between projects. I will mention it.

mathbunnyru: Added it and also divided conan to sections, to make it easier to follow

godexsoft (approved):
Should be good with the extra explanation about `$CONAN_HOME` 👍

Issue #2047, opened 2025-04-27
Update CI to GCC 15.2


mathbunnyru:
GCC 15.1 has been released.
https://gcc.gnu.org/gcc-15/changes.html

mathbunnyru:
Waiting for GCC 15.2 release, since there are bugs in GCC 15.1, which are not nice to fix (see the PR).

mathbunnyru:
Fixed in:
- https://github.com/XRPLF/clio/pull/2426
- https://github.com/XRPLF/clio/pull/2428

PR #2246, opened 2025-06-19
refactor: Refactor GCC image to make upgrades easier

mathbunnyru:
Work on: https://github.com/XRPLF/clio/issues/2047


docker/compilers/gcc/Dockerfile
```diff
@@ -1,10 +1,19 @@
-FROM ubuntu:20.04 AS build
+ARG UBUNTU_VERSION=20.04
+
+ARG GCC_MAJOR_VERSION=12
```
godexsoft, line 3: Why is this separated from the rest of gcc args?

docker/compilers/gcc/Dockerfile
```diff
@@ -53,38 +65,54 @@ RUN /gcc-$GCC_VERSION/configure \
...
-    && make install-strip DESTDIR=/gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION \
+        --enable-checking=release
+
+RUN make -j "$(nproc)"
+
+RUN make install-strip DESTDIR=/gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION \
     && mkdir -p /gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION/usr/share/gdb/auto-load/usr/lib64 \
-    && mv /gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION/usr/lib64/libstdc++.so.6.0.30-gdb.py /gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION/usr/share/gdb/auto-load/usr/lib64/libstdc++.so.6.0.30-gdb.py
+    && mv \
+        /gcc-$GCC_VERSION-$BUILD_VERSION-ubuntu-$UBUNTU_VERSION/usr/lib64/libstdc++.so.6.0.30-gdb.py \
```
godexsoft, line 75: This probably will change with newer versions of gcc? should we have this as arg too then?

docker/compilers/gcc/control.m4
```diff
@@ -2,5 +2,5 @@ Package: gcc-12-ubuntu-UBUNTUVERSION
 Version: VERSION
 Architecture: TARGETARCH
 Maintainer: Alex Kremer <akremer@ripple.com>
```
godexsoft, line 4: You should at least add yourself here

godexsoft (changes requested):

docker/compilers/gcc/Dockerfile
mathbunnyru: Because it affects both build and final image.

docker/compilers/gcc/Dockerfile
mathbunnyru: Honestly, I don't know. If it's going to change, I will put it in a variable when updating the image

docker/compilers/gcc/Dockerfile
godexsoft: Ah, got it, thanks

docker/compilers/gcc/control.m4
mathbunnyru: Only 1 maintainer is allowed to be specified here, but I added to uploaders

godexsoft (approved):

PR #2249, opened 2025-06-19
fix: Make update-libxrpl-version work with lockfile

mathbunnyru:


.github/workflows/check_libxrpl.yml
```diff
@@ -27,7 +27,13 @@ jobs:
       - name: Update libXRPL version requirement
         shell: bash
         run: |
-          ./.github/scripts/update-libxrpl-version ${{ github.event.client_payload.version }}
+          sed -i.bak -E "s|'xrpl/[a-zA-Z0-9\\.\\-]+'|'xrpl/${{ github.event.client_payload.version }}'|g" conanfile.py
```
godexsoft, line 30: I assume this works on all supported platforms

godexsoft (approved):

.github/workflows/check_libxrpl.yml
mathbunnyru: Yes

PR #2251, opened 2025-06-19
ci: Don't use dynamic names when they are not needed

mathbunnyru:


godexsoft (commented):
Why remove it i wonder?

mathbunnyru:
> Why remove it i wonder? 

It doesn't look nice when the job is not run.

And in my opinion it doesn't bring much value, because we **always** run Linux in containers and Mac natively.

It would make sense if we were running the same thing several ways, but currently it's not the case.

godexsoft (approved):

PR #2257, opened 2025-06-23
fix: Fix multiple async writes to socket

kuznetsss:
There was a problem in the new web server where subscriptions may schedule multiple concurrent asynchronous sending on web socket connection which is UB. I added an async mutex to fix this issue.


src/util/AsyncMutex.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+AsyncMutexLock
+AsyncMutex::lock(boost::asio::yield_context yield)
+{
+    while (locked_) {
+        boost::system::error_code error;
+        timer_.async_wait(yield[error]);
+    }
+    locked_ = true;
```
mathbunnyru, line 57: How is this going to work?
`locked_` is declared as simple bool: `bool locked_ = false;`

mathbunnyru (changes requested):
I'm not sure the current implementation is correct, please take a look

src/util/AsyncMutex.cpp
godexsoft: This is supposed to be called only through a strand. It's not very obvious and probably can be misused

mathbunnyru:
Please, make sure tests run fine under thread sanitizer.

kuznetsss:
I'm closing this PR, because such design will cancel scheduled sendings in O(N) while in the old web server does it in O(1). I will rewrite sending part of `WsConnection` to improve performance.

PR #2261, opened 2025-06-26
fix: Only set package_id:confs for sanitized builds

mathbunnyru:


mathbunnyru:
I want to add mold linker later.
It shouldn't affect package_id, but it currently will.
Let's only affect package_id in sanitizers.

Also, if someone forgets to install global.conf, nothing bad will happen.

docker/ci/conan/sanitizer_template.profile
```diff
@@ -10,11 +11,13 @@
...
-boost/*:without_stacktrace = True
+boost/*:extra_b2_flags="cxxflags=\"{{ sanitizer_build_flags_str }}\" linkflags=\"{{ sanitizer_link_flags_str }}\""
+boost/*:without_stacktrace=True
 
 [conf]
-tools.build:cflags += {{ sanitizer_build_flags }}
-tools.build:cxxflags += {{ sanitizer_build_flags }}
-tools.build:exelinkflags += {{ sanitizer_link_flags }}
-tools.build:sharedlinkflags += {{ sanitizer_link_flags }}
+tools.build:cflags+={{ sanitizer_build_flags }}
```
godexsoft, line 18: Out of curiosity, why remove the spaces?

mathbunnyru: https://docs.conan.io/2/reference/config_files/profiles.html#buildenv
`avoid using extra spaces around = in profiles, use the syntax shown above.`

mathbunnyru: It doesn't always apply, but let's be uniform and have it everywhere in the same style.

godexsoft (dismissed):

godexsoft (approved):

PR #2262, opened 2025-06-26
style: Fix JSON style in C++ code

mathbunnyru:


mathbunnyru:
I wrote a simple python script which looks for R"JSON strings in C++ code and adds some spaces.

I won't be adding this script and hook for now, because:

- it's not well written
- it won't work in case there are json's with escaping quotes (currently it works though)
- in some places I had to revert the file change, because test checks for exact string match

I'm also pretty sure we will have less such problems, because copy-pasting won't create more of these issues and auto-generation will understand where to put spaces.

godexsoft (approved):

PR #2266, opened 2025-06-27
style: Add pre-commit hook to fix  JSON style in C++ code

mathbunnyru:


.pre-commit-config.yaml
```diff
@@ -55,15 +55,30 @@ repos:
...
+  # Running some C++ hooks before clang-format
+  # to ensure that the style is consistent.
   - repo: local
     hooks:
+      - id: json-in-cpp
+        name: Fix JSON style in C++
+        entry: pre-commit-hooks/json_in_cpp.py
+        types: [c++]
+        language: python
+        exclude: |
```
godexsoft, line 67: Why we need to exclude some files?

.pre-commit-config.yaml
```diff
@@ -55,15 +55,30 @@ repos:
...
 
-  # Running fix-local-includes before clang-format
-  # to ensure that the include order is correct.
+  # Running some C++ hooks before clang-format
+  # to ensure that the style is consistent.
   - repo: local
     hooks:
+      - id: json-in-cpp
+        name: Fix JSON style in C++
+        entry: pre-commit-hooks/json_in_cpp.py
```
godexsoft, line 64: This is sad, we prefer golang for tooling :)

.pre-commit-config.yaml
mathbunnyru: These tests are written in a way that they require exact string representation of the response.
In the future we can rewrite them, but I think we're good for now.

.pre-commit-config.yaml
mathbunnyru: String manipulation and regular expressions are really nice in Python.
I also don't think we'll have to change this script a lot.

godexsoft: It's more about having only one language for tooling in the codebase.

godexsoft: It's ok let's keep this. You already sneaked a python script before so no-python is not a thing anymore either way :D

godexsoft (approved):

PR #2273, opened 2025-06-30
fix: Add sending queue to ng web server

kuznetsss:


src/web/ng/impl/WsConnection.hpp
```diff
@@ -57,14 +60,17 @@ class WsConnectionBase : public Connection {
...
+    sendShared(std::shared_ptr<std::string> message, boost::asio::yield_context yield) = 0;
 };
 
 template <typename StreamType>
 class WsConnection : public WsConnectionBase {
     boost::beast::websocket::stream<StreamType> stream_;
     boost::beast::http::request<boost::beast::http::string_body> initialRequest_;
     bool closed_{false};
+    std::queue<std::variant<Response, std::shared_ptr<std::string>>> sendingQueue_;
+    bool isSending_{false};
```
godexsoft: nit: Consider moving both bools to the bottom

src/web/ng/impl/WsConnection.hpp
```diff
@@ -167,6 +166,37 @@ class WsConnection : public WsConnectionBase {
             })
         );
     }
+
+    std::optional<Error>
+    sendImpl(std::variant<Response, std::shared_ptr<std::string>> message, boost::asio::yield_context yield)
```
godexsoft: This looks very similar to the http implementation. Consider adding an abstraction so that if we change one place we don't have to adjust the other each time

tests/unit/web/ng/impl/HttpConnectionTests.cpp
```diff
@@ -207,6 +209,69 @@ TEST_F(HttpConnectionTests, SendMultipleTimes)
     });
 }
 
+TEST_F(HttpConnectionTests, SendMultipleTimesFromMultipleCoroutines)
+{
+    Request const request{request_};
+    Response const response{http::status::ok, "some response data", request};
+
+    boost::asio::spawn(ctx_, [this, response = response](boost::asio::yield_context yield) mutable {
+        auto maybeError = httpClient_.connect("localhost", httpServer_.port(), yield, std::chrono::milliseconds{100});
```
godexsoft: nit: const

tests/unit/web/ng/impl/HttpConnectionTests.cpp
```diff
@@ -207,6 +209,69 @@ TEST_F(HttpConnectionTests, SendMultipleTimes)
...
+        }
+    });
+
+    runSpawn([this, &response](boost::asio::yield_context yield) {
+        auto connection = acceptConnection(yield);
+
+        util::CoroutineGroup group{yield};
+        for ([[maybe_unused]] auto i : std::ranges::iota_view{0, 3}) {
+            group.spawn(yield, [&response, &connection](boost::asio::yield_context innerYield) {
+                auto maybeError = connection.send(response, innerYield);
```
godexsoft: nit: const

tests/unit/web/ng/impl/HttpConnectionTests.cpp
```diff
@@ -207,6 +209,69 @@ TEST_F(HttpConnectionTests, SendMultipleTimes)
...
+        }
+        group.asyncWait(yield);
+    });
+}
+
+TEST_F(HttpConnectionTests, SendMultipleTimesClientDisconnected)
+{
+    Response const response{http::status::ok, "some response data", Request{request_}};
+    boost::asio::spawn(ctx_, [this, response = response](boost::asio::yield_context yield) mutable {
+        auto maybeError = httpClient_.connect("localhost", httpServer_.port(), yield, std::chrono::milliseconds{1});
```
godexsoft: nit: const here and other `maybeError` where applicable

tests/unit/web/ng/impl/WsConnectionTests.cpp
```diff
@@ -178,6 +197,35 @@ TEST_F(WebWsConnectionTests, MultipleSend)
...
+            [&]() { ASSERT_TRUE(expectedMessage.has_value()) << expectedMessage.error().message(); }();
+            EXPECT_EQ(expectedMessage.value(), response.message());
+        }
+    });
+
+    runSpawn([this, &response](boost::asio::yield_context yield) {
+        auto wsConnection = acceptConnection(yield);
+
+        util::CoroutineGroup group{yield};
+        for ([[maybe_unused]] auto unused : std::ranges::iota_view{0, 3}) {
```
godexsoft: nit: Consider calling this `i` or calling other `i`s `unused` :)

godexsoft (changes requested):
Leaving a few nits. Overall should be good 👍

src/web/ng/impl/HttpConnection.hpp
```diff
@@ -75,6 +76,13 @@ class HttpConnection : public UpgradableConnection {
     StreamType stream_;
     std::optional<boost::beast::http::request<boost::beast::http::string_body>> request_;
     std::chrono::steady_clock::duration timeout_{kDEFAULT_TIMEOUT};
+
+    using MessageType = boost::beast::http::response<boost::beast::http::string_body>;
+    SendingQueue<MessageType> sendingQueue_;
+    // std::queue<MessageType> sendingQueue_;
+    // boost::system::error_code sendingError_;
+    // bool isSending_{false};
```
mathbunnyru: This was debug?

src/web/ng/impl/WsConnection.hpp
```diff
@@ -57,13 +60,17 @@ class WsConnectionBase : public Connection {
     using Connection::Connection;
 
     virtual std::optional<Error>
```
mathbunnyru, line 62: Maybe make it std::expected<void, Error>

src/web/ng/SubscriptionContext.cpp
```diff
@@ -50,24 +51,31 @@ SubscriptionContext::SubscriptionContext(
 {
 }
 
+SubscriptionContext::~SubscriptionContext()
+{
+    ASSERT(disconnected_, "SubscriptionContext must be disconnected before destroying");
```
mathbunnyru, line 56: That's UB if the exception is thrown, but I think we should be fine.

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace web::ng::impl {
+
+template <typename T>
+class SendingQueue {
+public:
+    using Sender = std::function<void(T const&, boost::asio::basic_yield_context<boost::asio::any_io_executor>)>;
+
+private:
+    std::queue<T> queue_;
+    boost::system::error_code error_;
```
mathbunnyru:
```suggestion
    Error error_;
```

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -0,0 +1,73 @@
...
+    bool isSending_{false};
+    Sender sender_;
+
+public:
+    SendingQueue(Sender sender) : sender_{std::move(sender)}
+    {
+    }
+
+    std::optional<Error>
```
mathbunnyru, line 50: Here also maybe std::expected<void, Error>?

src/web/ng/impl/WsConnection.hpp
```diff
@@ -77,10 +84,30 @@ class WsConnection : public WsConnectionBase {
         : WsConnectionBase(std::move(ip), std::move(buffer), tagDecoratorFactory)
         , stream_(std::move(stream))
         , initialRequest_(std::move(initialRequest))
+        , sendingQueue_{[this](MessageType const& message, auto&& yield) {
+            boost::asio::const_buffer const buffer = std::visit(
+                util::OverloadSet{
+                    [](Response const& r) { return r.asWsResponse(); },
```
mathbunnyru: Let's add a return type here as well

mathbunnyru (commented):
Overall, looks good, only left some minor comments

src/web/ng/impl/WsConnection.hpp
kuznetsss: I added it with this return type to match other methods of `Connection`.  Created an issue for refactoring https://github.com/XRPLF/clio/issues/2292

src/web/ng/SubscriptionContext.cpp
kuznetsss: We throw exceptions from `ASSERT()` only in tests. I don't think there is a better way to check this, but I will think how to get rid of exceptions in tests for `ASSERT()`.

src/web/ng/impl/SendingQueue.hpp
kuznetsss: Again it is to match the return type of other methods in `Connection`.

src/web/ng/SubscriptionContext.cpp
kuznetsss: I added `WithMockAssertNoThrow` fixture for this case in tests.

godexsoft (approved):

PR #2281, opened 2025-07-01
chore: Update fmt to 11.2.0

mathbunnyru:
Seems that new clang-tidy fails with our old fmt lib


src/data/CassandraBackend.hpp
```diff
@@ -40,7 +40,7 @@
 #include <boost/uuid/string_generator.hpp>
 #include <boost/uuid/uuid.hpp>
 #include <cassandra.h>
-#include <fmt/core.h>
```
mathbunnyru, line 43: > https://github.com/fmtlib/fmt/blob/master/include/fmt/core.h

// This file is only provided for compatibility and may be removed in future
// versions. Use fmt/base.h if you don't need fmt::format and fmt/format.h
// otherwise.

src/util/config/ConfigConstraints.hpp
```diff
@@ -24,8 +24,8 @@
 #include "util/config/Types.hpp"
 #include "util/log/Logger.hpp"
 
-#include <fmt/core.h>
 #include <fmt/format.h>
+#include <fmt/ranges.h>
```
mathbunnyru, line 28: This is needed for fmt::join

godexsoft (approved):
🔥

Issue #1952, opened 2025-03-10
Update compiler to Clang 20


kuznetsss:
LLVM 20.1.0 was released, we should switch to it.

mathbunnyru:
I tried to update to clang20, but unfortunately, it seems that we can't build boost with clang20.
It doesn't find some libraries, builds boost without some parts of it, and then errors on packaging:

```
boost/1.83.0: WARN: Boost component 'locale' is missing libraries. Try building boost with '-o boost:without_locale'. (Option is not guaranteed to exist)
boost/1.83.0: WARN: Boost component 'stacktrace_addr2line' is missing libraries. Try building boost with '-o boost:without_stacktrace_addr2line'. (Option is not guaranteed to exist)
boost/1.83.0: WARN: Boost component 'stacktrace_backtrace' is missing libraries. Try building boost with '-o boost:without_stacktrace_backtrace'. (Option is not guaranteed to exist)
ERROR: boost/1.83.0: Error in package_info() method, line 2015
	raise ConanException(f"These libraries were expected to be built, but were not built: {non_built}")
	ConanException: These libraries were expected to be built, but were not built: {'boost_locale', 'boost_stacktrace_addr2line', 'boost_stacktrace_backtrace'}
```

We will likely need to update boost to the latest version first, and then to update the compiler.

mathbunnyru:
Done in https://github.com/XRPLF/clio/pull/3098/changes, now it's clang-22

PR #2291, opened 2025-07-02
chore: Don't hardcode versions in Dockerfiles and workflows

mathbunnyru:
Work on: https://github.com/XRPLF/clio/issues/1952

### DIFF
--- a/docker/compilers/clang/Dockerfile
+++ b/docker/compilers/clang/Dockerfile
@@ -8,8 +8,6 @@ SHELL ["/bin/bash", "-c"]
 USER root
 WORKDIR /root
 
-ARG CLANG_VERSION=16
-
 RUN apt-get update \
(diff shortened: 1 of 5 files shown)


docker/compilers/clang/Dockerfile
```diff
@@ -18,13 +16,15 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+ARG CLANG_MAJOR_VERSION
```
godexsoft: Don't you want to default this like in gcc?

mathbunnyru:
```suggestion
ARG CLANG_MAJOR_VERSION=invalid
```

mathbunnyru: Thanks, done

godexsoft (approved):

PR #2297, opened 2025-07-07
refactor: Simplify CMakeLists.txt for sanitizers

mathbunnyru:


CMakeLists.txt
```diff
@@ -69,15 +69,14 @@ endif ()
...
     message(FATAL_ERROR "Error: Unsupported sanitizer '${san}'. Supported values are: ${SUPPORTED_SANITIZERS}.")
   endif ()
 
-  target_compile_options(
-    clio_options INTERFACE # Sanitizers recommend minimum of -O1 for reasonable performance
-                           $<$<CONFIG:Debug>:-O1> ${SAN_FLAG} -fno-omit-frame-pointer
-  )
+  # Sanitizers recommend minimum of -O1 for reasonable performance so we enable it for debug builds
+  target_compile_options(clio_options INTERFACE $<$<CONFIG:Debug>:-O1>)
+  target_compile_options(clio_options INTERFACE ${SAN_FLAG} -fno-omit-frame-pointer)
```
kuznetsss: I think one `target_compile_options()` with longer list of options is more readable

mathbunnyru: In my opinion it's not readable at all, because one of them contains an implicit if statement.
In any programming language besides cmake it would have been a separate statement under `if` 😄

mathbunnyru: Also, the comment now applies to the whole next line as it should.
Before it applied to a prefix of the next line (not cool)

mathbunnyru: Rewrote as if statement, and created a task to have CMake profiles: https://github.com/XRPLF/clio/issues/2321

kuznetsss (approved):

PR #2304, opened 2025-07-08
feat: Use mold linker on Linux

### LINKED ISSUE #1242: Use mold linker on Linux in CI
We could try to use mold linker on Linux to speed up build a bit.

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/1242

Depends on: https://github.com/XRPLF/clio/pull/2329

### DIFF
--- a/cmake/Linker.cmake
+++ b/cmake/Linker.cmake
@@ -1,0 +1,8 @@
+if (use_mold)
+  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    message(STATUS "Using Mold linker")
+    set(CMAKE_LINKER_TYPE MOLD)
+  else ()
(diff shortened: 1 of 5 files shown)


cmake/Linker.cmake
```diff
@@ -0,0 +1,3 @@
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
kuznetsss: Shouldn't this be under an option? I think building Clio should by default require as few dependencies as possible

godexsoft: I agree, definitely should not be the default. Or at least use it by default only if it is available

godexsoft (approved):

PR #2306, opened 2025-07-08
fix: ASAN stack-buffer-overflow in NFTHelpersTest_NFTDataFromLedgerObject

godexsoft:
Fixes a stack-buffer-overflow error in one NFTHelpers test.


kuznetsss (approved):

src/etl/NFTHelpers.cpp
```diff
@@ -359,14 +360,18 @@ getNFTDataFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
 std::vector<NFTsData>
 getNFTDataFromObj(std::uint32_t const seq, std::string const& key, std::string const& blob)
 {
-    std::vector<NFTsData> nfts;
-    ripple::STLedgerEntry const sle =
+    // https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0020-non-fungible-tokens#tokenpage-id-format
+    ASSERT(key.size() == ripple::uint256::size(), "The size of the key (token) is expected to fit uint256 exactly");
+
+    auto const sle =
```
mathbunnyru, line 366: Maybe simply construct it as ripple::STLedgerEntry sle(...)?

The same comment about `auto const blob` below

godexsoft: Nah, we prefer `auto` :)

PR #2307, opened 2025-07-08
feat: Add workflow to build multiarch images

mathbunnyru:


mathbunnyru:
This should be implemented as a reusable workflow in `XRPLF/actions`, and here just simply use it

PR #2310, opened 2025-07-09
fix: ASAN heap-buffer-overflow issue in DBHelpers

godexsoft:
Fixes an issue in `isDirNode` where it could be passed an object with less than 3 bytes/items but the code would check index 1 and 2 effectively reading past the container's memory.

### DIFF
--- a/src/data/DBHelpers.hpp
+++ b/src/data/DBHelpers.hpp
@@ -199,39 +199,6 @@ struct MPTHolderData {
 };
 
 /**
- * @brief Check whether the supplied object is an offer.
- *
- * @param object The object to check
- * @return true if the object is an offer; false otherwise
- */
-template <typename T>
-inline bool
-isOffer(T const& object)
-{
-    static constexpr short kOFFER_OFFSET = 0x006f;
-    static constexpr short kSHIFT = 8;
-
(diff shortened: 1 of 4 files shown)


src/data/DBHelpers.hpp
```diff
@@ -241,6 +241,10 @@ template <typename T>
 inline bool
 isDirNode(T const& object)
 {
+    static constexpr auto kMIN_SIZE_REQUIRED = 3;
+    if (std::size(object) < kMIN_SIZE_REQUIRED)
+        return false;
```
mathbunnyru, line 213: There are other functions in this file with the same problem.
Let's fix them as well

mathbunnyru (changes requested):

src/data/DBHelpers.hpp
godexsoft: Yeah, they did not show up in sanitizer reports but i agree we should use this opportunity to fix them

kuznetsss (dismissed):

src/data/DBHelpers.hpp
godexsoft: I fixed them but then realised they were all dead code. So the real fix is to delete those 👍

mathbunnyru (approved):

PR #2312, opened 2025-07-09
test: Fix prometheus tests

kuznetsss:


tests/common/util/MockPrometheus.hpp
```diff
@@ -258,20 +258,24 @@ struct WithMockPrometheus : virtual ::testing::Test {
 /**
  * @note this class should be the first in the inheritance list
  */
-struct WithPrometheus : virtual ::testing::Test {
-    WithPrometheus()
+template <bool IsEnabled = true>
+struct WithPrometheusImpl : virtual ::testing::Test {
+    WithPrometheusImpl()
```
mathbunnyru, line 263: Maybe let's put this to impl namespace, remove `true` by default and inherit both `WithPrometheus` and `WithPrometheusDisabled` from this impl class?

mathbunnyru (commented):
I think it's good, but we could make it slightly better

tests/common/util/MockPrometheus.hpp
godexsoft: The way it's done now is a common patter we use elsewhere in the codebase - let's leave it as is 👍

godexsoft (approved):

mathbunnyru (approved):

PR #2313, opened 2025-07-09
fix: Link with boost libraries explicitly

mathbunnyru:


mathbunnyru:
We will need to link to boost libs explicitly after https://github.com/XRPLF/rippled/pull/5546
Now we just link to every boost library, which is not great and violates ODR (explained in the mentioned PR).

src/util/CMakeLists.txt
```diff
@@ -42,12 +42,15 @@ target_sources(
...
-  target_link_libraries(clio_util PUBLIC Boost::stacktrace_backtrace dl libbacktrace::libbacktrace)
+  target_link_libraries(clio_util PUBLIC Boost::stacktrace_backtrace)
 endif ()
 
 target_link_libraries(
   clio_util
   PUBLIC Boost::headers
+         Boost::iostreams
+         Boost::log
+         Boost::log_setup
```
kuznetsss, line 53: I thought `Boost::headers` includes all the header only libraries

mathbunnyru: It does, these are not header-only

kuznetsss: But how did it work before?

mathbunnyru: Because rippled conanfile was telling rippled was linking to all boost libraries, please see the comment above and the PR it mentions.

kuznetsss (approved):

PR #2318, opened 2025-07-10
ci: Use cache in PRs where makes sense

mathbunnyru:


.github/workflows/nightly.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       run_unit_tests: true
       run_integration_tests: true
       upload_clio_server: true
-      disable_cache: true
+      disable_cache: ${{ github.event_name != 'pull_request' }}
```
kuznetsss, line 64: It is theoretically possible that because of this condition workflow will pass CI in PR but fail when running without cache

mathbunnyru: ccache is not ideal, but quite good, so it's highly unlikely

mathbunnyru:
I thought about this a bit more - unfortunately, we don't run these workflows on every commit in develop, so the cache won't work.

For example, for nightly workflow, we need either:
- run for every commit in develop
- better find commit where it was last run in develop
- run nightly and use the cache from the last nightly run
- create cache key which won't have info about commit (reuse the same cache key if it's possible)

In any case, it's not easy, and there are pros and cons to each solution, so closing this for now

PR #2324, opened 2025-07-10
fix: ASAN SEGV in cluster communication

godexsoft:
Fixed by @kuznetsss

### DIFF
--- a/src/cluster/ClusterCommunicationService.cpp
+++ b/src/cluster/ClusterCommunicationService.cpp
@@ -66,7 +66,7 @@ ClusterCommunicationService::run()
     boost::asio::spawn(strand_, [this](boost::asio::yield_context yield) {
         auto timer = std::make_shared<boost::asio::steady_timer>(yield.get_executor());
         boost::signals2::scoped_connection connection =
-            stopSignal_.connect([yield, &timer]() { boost::asio::spawn(yield, [timer](auto&&) { timer->cancel(); }); });
+            stopSignal_.connect([yield, timer]() { boost::asio::spawn(yield, [timer](auto&&) { timer->cancel(); }); });
         boost::system::error_code error;
         while (not stopped_) {
             timer->expires_after(readInterval_);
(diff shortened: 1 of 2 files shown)


kuznetsss (dismissed):

src/cluster/ClusterCommunicationService.cpp
```diff
@@ -63,20 +64,27 @@ void
...
 {
     boost::asio::spawn(strand_, [this](boost::asio::yield_context yield) {
-        boost::asio::steady_timer timer(yield.get_executor());
-        while (true) {
-            timer.expires_after(readInterval_);
-            timer.async_wait(yield);
-            doRead(yield);
+        auto timer = std::make_shared<boost::asio::steady_timer>(yield.get_executor());
+        boost::signals2::scoped_connection connection =
+            stopSignal_.connect([yield, &timer]() { boost::asio::spawn(yield, [timer](auto&&) { timer->cancel(); }); });
```
kuznetsss:
```suggestion
            stopSignal_.connect([yield, timer]() { boost::asio::spawn(yield, [timer](auto&&) { timer->cancel(); }); });
```

kuznetsss (approved):

mathbunnyru (commented):
Would be great, if you told what the bug was and how this fixes it

godexsoft:
> Would be great, if you told what the bug was and how this fixes it

We are not 100% certain that this is not a false positive that got masked by this 'fix'. 
Will try to create a minimal reproducible example and take it from there. 

Essentially the idea was that it's no good to stop the context while some coroutines are still running. At least that's what we are seeing from ASAN and also that's what AI is suggesting. But the docs of asio say it's perfectly fine to stop the context and that the ongoing coroutines will just be abandoned. So it may well be a false positive.

mathbunnyru (changes requested):
The minimal example doesn't fail with boost 1.88, so let's not merge it, because we won't probably be fixing anything, but making code more complicated

PR #2333, opened 2025-07-15
fix: Fix writing into dead variables in BlockingCache

kuznetsss:


mathbunnyru (changes requested):
Could you please add a test, which fails on develop, but works in this branch? (and check it in develop)

src/util/BlockingCache.hpp
```diff
@@ -197,22 +197,31 @@ class BlockingCache {
     std::expected<ValueType, ErrorType>
     wait(boost::asio::yield_context yield, Updater updater, Verifier verifier)
     {
-        boost::asio::steady_timer timer{yield.get_executor(), boost::asio::steady_timer::duration::max()};
+        struct SharedContext {
```
mathbunnyru, line 200: Do we declare structs inside functions inside our codebase?
Nothing against it, just wanted to know our codestyle rules here.

kuznetsss:
The bug was happening because `result` and `timer` were accessed for writing when they are already destroyed. But this only happens when multiple coroutines are accessing the cache simultaneously with the right timing. I don't think we can reproduce this in test.

src/util/BlockingCache.hpp
kuznetsss: @godexsoft, WDYT?

godexsoft: I have no issue with it as long as they are small - it's sort of similar to writing a lambda imho.

godexsoft (approved):
I agree with @mathbunnyru that it would be nice to add a test for this fix. But i also recall it was not very easy to even reproduce, right?

mathbunnyru:
> I agree with @mathbunnyru that it would be nice to add a test for this fix. But i also recall it was not very easy to even reproduce, right?

@kuznetsss could you please answer here?
Just for the history, why we're not adding the test

kuznetsss:
I already answered:

> I don't think we can reproduce this in test.

mathbunnyru (approved):

PR #2334, opened 2025-07-15
chore: Add trufflehog security tool

mathbunnyru:
Tool works locally and doesn't require internet connection, if `--no-verification` is passed.
I also tried to add `secret.cpp` with github PAT and the tool detected it.


kennyzlei (approved):
thanks for checking internet connection on this tool

PR #2338, opened 2025-07-17
fix: add trustline locking flag to account_info

### LINKED ISSUE #2323: Add TrustlineLocking flag to account_info RPC
## Summary
Add allowTrustLineLocking flag to account_info RPC.
Rippled PR:  https://github.com/XRPLF/rippled/pull/5525

PeterChen13579:
fixes #2323


src/rpc/handlers/AccountInfo.hpp
```diff
@@ -83,6 +85,7 @@ class AccountInfoHandler {
             ripple::STLedgerEntry sle,
             bool isDisallowIncomingEnabled,
             bool isClawbackEnabled,
+            bool isTokenEscrowEnabled,
```
mathbunnyru, line 88: Do we need this ctor at all?

Maybe we can keep it as a struct without ctor, and also use designated initializers to prevent errors.

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -328,6 +328,7 @@ TEST_F(RPCAccountInfoHandlerTest, SignerListsInvalid)
     ON_CALL(*backend_, doFetchLedgerObject(signersKey, 30, _))
         .WillByDefault(Return(createLegacyFeeSettingBlob(1, 2, 3, 4, 0)));
     EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(_, Amendments::DisallowIncoming, _)).WillOnce(Return(false));
+    EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(_, Amendments::TokenEscrow, _)).WillOnce(Return(false));
```
mathbunnyru: Please, keep the order:
1. DisallowIncoming
2. Clawback
3. TokenEscrow

And in many places below

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -742,14 +748,14 @@ TEST_F(RPCAccountInfoHandlerTest, DisallowIncoming)
     });
 }
 
-TEST_F(RPCAccountInfoHandlerTest, Clawback)
+TEST_F(RPCAccountInfoHandlerTest, AmendmentsEnabled)
```
mathbunnyru, line 751: Why did you change this test and didn't add a new one?

mathbunnyru (changes requested):

tests/unit/rpc/handlers/AccountInfoTests.cpp
PeterChen13579: This test was initially testing 2 boolean flags when set to true, the output json contains these flags. The `trustlinelocking` is simply another boolean flag getting added to the response. I don't think it's worth having a new test for every boolean flag we may add in.

src/rpc/handlers/AccountInfo.hpp
PeterChen13579: Ah, I think this is a great suggestion! Additionally, I think this could be applied to other handler classes `Output` as well. I will create an issue for this because I think it'll be out of scope for this PR to refactor every other class

PeterChen13579: https://github.com/XRPLF/clio/issues/2452

mathbunnyru: May I ask to try to fix it after this PR if it doesn't take too much time?
AI could help here easily, I suppose

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -723,6 +728,7 @@ TEST_F(RPCAccountInfoHandlerTest, DisallowIncoming)
     ON_CALL(*backend_, doFetchLedgerObject(accountKk, 30, _))
         .WillByDefault(Return(accountRoot.getSerializer().peekData()));
     EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(_, Amendments::DisallowIncoming, _)).WillOnce(Return(true));
+    EXPECT_CALL(*mockAmendmentCenterPtr_, isEnabled(_, Amendments::TokenEscrow, _)).WillOnce(Return(false));
```
mathbunnyru: missed the order here

mathbunnyru (changes requested):

src/rpc/handlers/AccountInfo.hpp
PeterChen13579: Yea, sounds good. It should be a quick PR anyways 🙏

mathbunnyru: Thanks ☺️

mathbunnyru (approved):

PR #2349, opened 2025-07-18
fix: Handle logger exceptions

kuznetsss:


src/util/log/Logger.cpp
```diff
@@ -65,6 +71,30 @@
...
+    std::reference_wrapper<util::prometheus::CounterInt> exceptionCounter_ =
+        PrometheusService::counterInt("logger_exceptions_total_number", util::prometheus::Labels{});
+
+public:
+    using result_type = void;
+
+    LoggerExceptionHandler()
+    {
+        // Check that prometheus is already initialized
+        PrometheusService::instance();
```
godexsoft: This feels a bit like a hack :rage4:  
Maybe add a static member function that returns whether the instance exists or not without ASSERTing and use an ASSERT here and in `instance()` itself using the new function?

src/util/log/Logger.cpp
```diff
@@ -65,6 +71,30 @@
...
+
+    void
+    operator()(std::exception const& e) const
+    {
+        std::cerr << fmt::format("Exception in logger: {}", e.what()) << std::endl;
+        exceptionCounter_.get() += 1;
+    }
+};
+
+}  // namespace
```
godexsoft, line 96: Add some space?

mathbunnyru, line 96: nit: need an empty line after this

godexsoft (changes requested):
Leaving a few nits 👍

src/main/Main.cpp
```diff
@@ -52,6 +53,7 @@ try {
             if (not app::parseConfig(run.configPath))
                 return EXIT_FAILURE;
 
+            PrometheusService::init(gClioConfig);
```
mathbunnyru, line 56: Prometheus init doesn't depend on logger, right?

src/util/log/Logger.cpp
```diff
@@ -65,6 +71,30 @@
...
+    {
+        // Check that prometheus is already initialized
+        PrometheusService::instance();
+    }
+
+    void
+    operator()(std::exception const& e) const
+    {
+        std::cerr << fmt::format("Exception in logger: {}", e.what()) << std::endl;
+        exceptionCounter_.get() += 1;
```
mathbunnyru:
```suggestion
        ++exceptionCounter_.get();
```

src/util/log/Logger.cpp
```diff
@@ -65,6 +71,30 @@
 
 namespace util {
 
+namespace {
+
+class LoggerExceptionHandler {
+    std::reference_wrapper<util::prometheus::CounterInt> exceptionCounter_ =
+        PrometheusService::counterInt("logger_exceptions_total_number", util::prometheus::Labels{});
```
mathbunnyru, line 78: Maybe `logger_thrown_exceptions`?

src/util/log/Logger.cpp
mathbunnyru: Do you need this? You init prometheus before logger, so I don't see a case where this is needed.

mathbunnyru (commented):
Overall, looks good, please:
- check if issue still persists with boost 1.88
- create an issue to remove this when fixed in the boost or by using spdlog

src/util/log/Logger.cpp
godexsoft: I assume this is to prevent potential future breakage if someone moves the init again

mathbunnyru: It seems so, and I don't like two globals depending on each other :(

src/util/log/Logger.cpp
kuznetsss: Prometheus [naming conventions](https://prometheus.io/docs/practices/naming/) recommends to add `_total` suffix for counters. While it recommends to use just `_total` for metrics without units, we use `_total_number` everywhere in such cases.
I don't think `thrown` adds any value because there is no other state for an exception rather than being thrown.

mathbunnyru: Got it, thanks

src/main/Main.cpp
kuznetsss: No, I doesn't depend on logger.

mathbunnyru (approved):
👍

godexsoft (approved):
LGTM :rage3:

PR #2354, opened 2025-07-22
ci: Pin docker image to commit hash

mathbunnyru:


kuznetsss:
Would it be possible to specify this docker image in one place and reuse it everywhere to avoid changing 12 files each time we update docker?

mathbunnyru:
> Would it be possible to specify this docker image in one place and reuse it everywhere to avoid changing 12 files each time we update docker?

I thought about it and asked gemini and don't know a good way.
The problem is we need to define the image before running steps.

1. We can use variables, but they are stored as part of settings, so won't work for us
2. We can define a job in each workflow which will print the image in GITHUB_OUTPUT (the version might be stored in a callable workflow or .github/actions), and we will add `need` statement and use this value. This won't look nice in the UI and in the workflow files too
3. We can reduce a little bit duplication and define image in env part of each workflow. But the number of files changing will stay the same

If you have other ideas, please let me know, but I don't think there is a really good solution here

godexsoft (approved):

kuznetsss (approved):

Issue #2331, opened 2025-07-14
Update to latest boost release


mathbunnyru:
The old one prevents us from enabling address sanitizer, as boost asio crashes

mathbunnyru:
Will probably fix these:
- https://github.com/XRPLF/clio/issues/1611
- https://github.com/XRPLF/clio/issues/1610
- https://github.com/XRPLF/clio/issues/1001

mathbunnyru:
This was updated in the upstream, and now we all use `boost/1.91.0`

PR #2355, opened 2025-07-22
chore: Update boost usages to match 1.88

godexsoft:
This PR reuses most of #2331 
It turns out most of the changes for boost 1.88 are totally fine under 1.83 too. This is an easier path to migration - first we fix the code without changing the version, then a much smaller PR to just upgrade to newer version of the boost libraries should do the trick.

Changes in this PR include:
- Add a `util::spawn` proxy function for `boost::asio::spawn` that preserves the exception propagating behaviour our codebase is built upon
- Replace all `boost::asio::spawn` calls that are not explicitly designed to be used with `boost::asio::detached` with `util::spawn`
- Remove any use of `io_service` in favour of `io_context`
- Replace deprecated calls like `reset` to their new versions (like `restart`)
- Use `boost::asio::post` instead of `context.post(...)` which no longer compiles
- Use the newer `make_address` API instead of deprecated `from_string` calls


mathbunnyru (commented):
Overall, LGTM

Please, resolve conflict.
And also please add types of changes you did in the PR description

tests/common/util/AsioContextTestFixture.hpp
```diff
@@ -106,10 +106,14 @@ struct SyncAsioContextTest : virtual public NoLoggerFixture {
         if (allowMockLeak)
             testing::Mock::AllowLeak(&call);
 
-        spawn(ctx_, [&, _ = make_work_guard(ctx_)](yield_context yield) {
-            f(yield);
-            call.Call();
-        });
+        spawn(
```
mathbunnyru: As I mentioned earlier, please, use `boost::asio::spawn` everywhere. You can just search for ` spawn(` and replace.

(and probably you'll be able to remove using namespace above)

tests/common/util/AsioContextTestFixture.hpp
```diff
@@ -151,22 +163,22 @@ struct SyncAsioContextTest : virtual public NoLoggerFixture {
     runContext()
     {
         ctx_.run();
-        ctx_.reset();
+        ctx_.restart();
```
mathbunnyru, line 150: Is is the name change, or something actually changed here?

godexsoft: ```Deprecated: Use restart().) Reset the io_context in preparation for a subsequent run() invocation.```

mathbunnyru: Cool, thanks

mathbunnyru (dismissed):

src/app/Stopper.cpp
```diff
@@ -36,7 +37,7 @@ Stopper::~Stopper()
 void
 Stopper::setOnStop(std::function<void(boost::asio::yield_context)> cb)
 {
-    boost::asio::spawn(ctx_, std::move(cb));
+    boost::asio::spawn(ctx_, std::move(cb), boost::asio::detached);
```
kuznetsss: Using `detached` everywhere may (as far as I know) drop uncaught exceptions in coroutines. I think it is better to have crash when an exception we didn't expect was thrown.

src/util/Coroutine.hpp
```diff
@@ -114,10 +115,14 @@ class Coroutine {
     static void
     spawnNew(ExecutionContext& ioContext, Fn fn)
     {
-        boost::asio::spawn(ioContext, [fn = std::move(fn)](boost::asio::yield_context yield) {
-            Coroutine thisCoroutine{std::move(yield)};
-            fn(thisCoroutine);
-        });
+        boost::asio::spawn(
+            ioContext,
```
kuznetsss: Should we explicitly create strand everywhere we do `spawn()` on context?

godexsoft: I have created `util::spawn` for this, otherwise we would end up maintaining a huge list of those manual strands. In the future it would be nice to hide this under some more abstractions and only use them in user code but today it's not a feasible change imo.

src/app/Stopper.cpp
godexsoft: Right 😢 It is passing our tests but i guess we just did not check for exception propagation if/where needed. It's easier to preserve the old behaviour through a proxy function.

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+namespace util {
+
+template <typename T>
+struct IsStrandType : std::false_type {};
+
+template <typename Executor>
+struct IsStrandType<boost::asio::strand<Executor>> : std::true_type {};
+
+template <typename T>
+concept IsStrand = IsStrandType<std::decay_t<T>>::value;
```
mathbunnyru: It would be better to have (or reuse if we have one available already) a generic `IsInstantiationOf` concept

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+     * @param ePtr The exception that was caught on the coroutine
+     */
+    template <typename... R>
+    void
+    operator()(std::exception_ptr ePtr, R...)
+    {
+        if (ePtr)
+            std::rethrow_exception(ePtr);
+    }
+} kPROPAGATE_EXCEPTIONS;
```
mathbunnyru: Maybe let's make it a lambda and put it to an impl namespace?
This is not meant to be used outside of this header

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+ * This is intended to be passed as the third argument to `boost::asio::spawn` so that exceptions are not ignored but
+ * propagated to `io_context.run()` call site.
+ */
+inline constexpr struct PropagatingCompletionHandler {
+    /**
+     * @brief The completion handler
+     * @tparam R Return type (omitted for functions returning `void`)
+     * @param ePtr The exception that was caught on the coroutine
+     */
+    template <typename... R>
```
mathbunnyru: Do we need this R at all?

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+ */
+template <typename Ctx, typename F>
+auto
+spawn(Ctx&& ctx, F&& func)
+{
+    if constexpr (IsStrand<Ctx>) {
+        return boost::asio::spawn(std::forward<Ctx>(ctx), std::forward<F>(func), kPROPAGATE_EXCEPTIONS);
+    } else {
+        return boost::asio::spawn(
+            boost::asio::make_strand(ctx.get_executor()), std::forward<F>(func), kPROPAGATE_EXCEPTIONS
```
mathbunnyru:
```suggestion
            boost::asio::make_strand(std::forward<Ctx>(ctx).get_executor()), std::forward<F>(func), kPROPAGATE_EXCEPTIONS
```

src/web/ng/Server.cpp
```diff
@@ -247,28 +248,28 @@ Server::run()
...
-                    continue;
-                }
-                boost::asio::spawn(
-                    ctx_.get(),
-                    [this, socket = std::move(socket)](boost::asio::yield_context yield) mutable {
-                        handleConnection(std::move(socket), yield);
-                    },
-                    boost::asio::detached
-                );
+    util::spawn(ctx_.get(), [this, acceptor = std::move(acceptor).value()](boost::asio::yield_context yield) mutable {
```
mathbunnyru, line 251: Unfortunate to see huge formatting diff here, anything we could do about it?

tests/unit/util/SpawnTests.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+        boost::asio::spawn(
+            io,
+            [](boost::asio::yield_context) { throw std::runtime_error("Test exception in coroutine"); },
+            util::kPROPAGATE_EXCEPTIONS
+        );
+
+        io.run();
+    }());
+}
```
mathbunnyru: Do we need these 2?
It seems they're checking the same as above

src/util/Spawn.hpp
godexsoft: This is used outside of the header even right now (in tests). I'm not sure i prefer a lambda for this either

src/util/Spawn.hpp
godexsoft: Yes. It's used for completion handlers that return non-void, as per their docs (it also does not compile without it for anything but void)

src/util/Spawn.hpp
godexsoft: Yeah, fair point 👍

src/web/ng/Server.cpp
godexsoft: Yea.. ignore it :rage2:

tests/unit/util/SpawnTests.cpp
godexsoft: They are checking direct usage of util::kPROPAGATE_EXCEPTIONS (yes it's the same as util::spawn does inside, except for the strand wrapping)

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @tparam Ctx The type of the context/strand
+ * @tparam F The type of the function to execute
+ * @param ctx The execution context
+ * @param func The function to execute
+ * @return Propagated from underlying `boost::asio::spawn` call
+ */
+template <typename Ctx, typename F>
+auto
+spawn(Ctx&& ctx, F&& func)
```
kuznetsss, line 69: I think `boost::asio::spawn()` always takes the first parameter as a reference. Maybe here it should be `&` instead of `&&` for `Ctx`?

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+ * @note This uses kPROPAGATE_EXCEPTIONS to force asio to propagate exceptions through `io_context`
+ * @note Since implicit strand was removed from boost::asio::spawn this helper function adds the strand back
+ *
+ * @tparam Ctx The type of the context/strand
+ * @tparam F The type of the function to execute
+ * @param ctx The execution context
+ * @param func The function to execute
+ * @return Propagated from underlying `boost::asio::spawn` call
+ */
+template <typename Ctx, typename F>
```
kuznetsss, line 66: `F`'s signature is always `void(boost::asio::yiled_context)` so maybe it would be good to add a concept?

godexsoft: `F`'s signature is `R(boost::asio::yield_context)` where `R` is the return type. It would just complicate things, don't think it's needed. wdyt?

src/util/Spawn.hpp
godexsoft: Perfect forwarding is better. `ctx` here can be a `boost::asio::make_strand(ioc)` for example.

src/util/Spawn.hpp
godexsoft: Imo not really worth for such a small usecase

src/util/Spawn.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <exception>
+#include <type_traits>
+
+namespace util {
+
+template <typename T>
+struct IsStrandType : std::false_type {};
+
+template <typename Executor>
+struct IsStrandType<boost::asio::strand<Executor>> : std::true_type {};
```
kuznetsss: Couldn't you do something like:
```cpp
template <typename T, typename Executor>
concept IsStrand = std::is_same<T, boost::asio::strand<Executor>>;
```

src/util/Spawn.hpp
kuznetsss: I think with `kPROPAGATE_EXCEPTIONS` completion handler `F` couldn't return any value.

godexsoft: Fine, let's lock it to `void`

src/web/ng/Server.cpp
mathbunnyru: Ok :)

mathbunnyru (dismissed):

src/util/Spawn.hpp
godexsoft: As a consequence of locking to `void` some of the things @mathbunnyru wanted are now also done. Hopefully this version looks ok to everyone now 👍

tests/unit/util/SpawnTests.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+TEST(SpawnTest, SpawnOnCoroutine)
+{
+    EXPECT_ANY_THROW([] {
+        boost::asio::io_context io;
+        boost::asio::spawn(io, [](boost::asio::yield_context yield) {
```
godexsoft: Damn, this will probably break on boost 1.88 :rage1:

kuznetsss (dismissed):

kuznetsss (approved):

mathbunnyru (approved):

PR #2361, opened 2025-07-24
chore: Improve RC release notes

mathbunnyru:


kuznetsss (approved):
LGTM
Btw, we could now remove publishing SHA in release notes because github now does it automatically. But it could be a separate PR.

PR #2365, opened 2025-07-25
chore: Verify version of the binary on release

mathbunnyru:


.github/workflows/build_impl.yml
```diff
@@ -183,6 +189,20 @@ jobs:
         if: ${{ inputs.code_coverage }}
         uses: ./.github/actions/code_coverage
 
+      - name: Verify expected version
```
kuznetsss, line 192: Did you do it here to not have to download artefact in `release.yml`?

mathbunnyru: I didn't want to download it there, yes.
Also, I would have to duplicate strategy, which is also not nice.
And, I'm also not able to verify inside release_impl, because it runs on Linux, but we also want to verify version for mac.

kuznetsss (approved):

PR #2367, opened 2025-07-25
ci: Pass expected_version to build_impl.yml

mathbunnyru:


.github/workflows/build_and_test.yml
```diff
@@ -77,6 +77,7 @@ jobs:
       upload_clio_server: ${{ inputs.upload_clio_server }}
       targets: ${{ inputs.targets }}
       analyze_build_time: false
+      expected_version: ${{ inputs.expected_version }}
```
mathbunnyru, line 80: Forgot to pass it here, so it was always working

.github/workflows/build_impl.yml
```diff
@@ -195,12 +195,10 @@ jobs:
...
       - name: Verify expected version
         if: ${{ inputs.expected_version != '' }}
+        shell: bash
         run: |
+          set -e
           EXPECTED_VERSION="clio-${{ inputs.expected_version }}"
-          if [[ ! -f build/clio_server ]]; then
-            echo "clio_server binary not found"
-            exit 1
-          fi
```
mathbunnyru, line 203: Don't need this, better to just `set -e` and fail on the line where we call `--version`

kuznetsss (approved):

PR #2369, opened 2025-07-28
chore: Do not generate source files and use target_compile_definition…

mathbunnyru:
…s to set version

Advantages:
- not modifying src dir in build time (`git clean` will also remove less)
- simpler codecov (if we decide to move it to separate stage in the future)
- The cpp file is perfectly valid, so it will be treated as a C++ files by all the tooling
- No need to use CMAKE_PROJECT_INCLUDE_BEFORE, should work fine without it
- Instead of generating + recompiling, we will just be doing recompilation of 1 file

### DIFF
--- a/src/util/build/Build.cpp
+++ b/src/util/build/Build.cpp
@@ -23,19 +23,22 @@
 
 namespace util::build {
 
-static constexpr char versionString[] = "@CLIO_VERSION@"; // NOLINT(readability-identifier-naming)
+#ifndef CLIO_VERSION
+#error "CLIO_VERSION must be defined"
+#endif
+static constexpr char versionString[] = CLIO_VERSION;
 
 std::string const&
 getClioVersionString()
 {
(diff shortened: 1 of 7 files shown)


src/util/build/Build.cpp
```diff
@@ -23,19 +23,22 @@
...
+#ifndef CLIO_VERSION
+#error "CLIO_VERSION must be defined"
+#endif
+static constexpr char versionString[] = CLIO_VERSION;
 
 std::string const&
 getClioVersionString()
 {
-    static std::string const value = versionString; // NOLINT(readability-identifier-naming)
+    static std::string const value = versionString;  // NOLINT(readability-identifier-naming)
```
mathbunnyru, line 34: For example, clang-format started treating this as all other source files and there should be 2 spaces before comment

godexsoft (approved):
Looks better than before 👍

PR #2371, opened 2025-07-28
chore: Move config definition to cpp file

mathbunnyru:
Having such a long init list in almost every translation unit is definitely not nice.
It makes it more difficult to work on custom logger (because I need to adjust log_format), so let's move it to cpp file


mathbunnyru:
This also allows to use ClioConfigDefinition when constructing other globals.

godexsoft (approved):
Alright 👍 Let's see if @kuznetsss and @PeterChen13579 are happy with this as well.

kuznetsss (approved):

PR #2397, opened 2025-08-05
fix: Fix default bool value print in config definition

### LINKED ISSUE #2396: Wrong default boolean values in `config-description.md`
It seems that config definition inverts boolean variables default values in the generated markdown

Code:

https://github.com/XRPLF/clio/blob/cfffbfba9dabcfa758d76f523b9bbb2d08218283/src/util/config/ConfigDefinition.cpp#L366

The generated `config-description.md`:

https://github.com/XRPLF/clio/blob/cfffbfba9dabcfa758d76f523b9bbb2d08218283/docs/config-description.md?plain=1#L454

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2396

### DIFF
--- a/docs/config-description.md
+++ b/docs/config-description.md
@@ -155,7 +155,7 @@ This document provides a list of all available Clio configuration properties in 
 
 - **Required**: True
 - **Type**: boolean
-- **Default value**: `True`
+- **Default value**: `False`
 - **Constraints**: None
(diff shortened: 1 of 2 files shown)


PeterChen13579 (approved):
Thank you for the fix 😄

kuznetsss (approved):

Issue #1852, opened 2025-01-28
AsyncFramework support sleeping through context


godexsoft:
## Summary
It is useful for certain long running tasks to be able to pause for arbitrary amounts of time. 
Different contexts would implement this differently (e.g. CoroContext could use an async timer and yield instead of actually pausing the thread).

## Motivation
ETL ng task manager needs a better way to sleep between attempts when there is nothing to do.

## Solution
The stopToken passed to client functions by async framework is currently used just as a stop token. Let's upgrade it into a `context` object that can act as a stop token but also provides a way to pause the task.  

Possibly combine with: https://github.com/XRPLF/clio/issues/1380

PR #2398, opened 2025-08-06
fix: Remove tight loop in ETLng

godexsoft:
This may be a temporary solution which will likely change once we implement #1852


kuznetsss (approved):

src/etlng/impl/TaskQueue.hpp
```diff
@@ -126,6 +141,17 @@ class TaskQueue {
...
     }
+
+    /**
+     * @brief Awaits for the queue to become non-empty
+     * @note This function blocks until there is a task or the queue is being destroyed
+     */
+    void
+    awaitTask()
+    {
+        auto lock = data_.lock<std::unique_lock>();
```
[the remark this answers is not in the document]
godexsoft: Can we stop using this nonsense reviewer? seems completely not useful to me.

src/etlng/impl/TaskQueue.hpp
```diff
@@ -75,6 +81,13 @@ class TaskQueue {
     {
     }
 
+    ~TaskQueue()
+    {
+        // unblock all waiters
+        stopping_ = true;
+        cv_.notify_all();
```
mathbunnyru, line 88: This won't work, because you're relying on some values from destructed object

src/etlng/impl/TaskQueue.hpp
```diff
@@ -57,6 +61,8 @@ class TaskQueue {
     };
 
     util::Mutex<Data> data_;
+    std::condition_variable cv_;
+    boost::atomics::atomic_bool stopping_ = false;
```
mathbunnyru, line 65: No need to use boost here

PR #2399, opened 2025-08-06
fix: ETLng more cpu usage fixes 

godexsoft:
- Fixes race condition 
- Removes boost/atomic use (was meant to be std::atomic)


kuznetsss (approved):

mathbunnyru (approved):
I don't understand this, but approve

PR #2404, opened 2025-08-07
ci: Build benchmark code in CI

mathbunnyru:


tools/rebuild_conan.py
```diff
@@ -27,7 +27,7 @@ def rebuild():
                     f"--output-folder=build_{profile}_{build_type}",
                     "-s", f"build_type={build_type}",
                     "-o", "&:tests=True",
-                    "-o", "&:integration_tests=True",
```
godexsoft, line 30: Is this meant to be replaced or added to the list?

mathbunnyru: This was a bug, and benchmark dependencies were not built when running this script (they should)

godexsoft (approved):

PR #2405, opened 2025-08-07
chore: Update trufflehog max depth

godexsoft:
Max depth appears to be necessary, otherwise trufflehog is triggered by git history


mathbunnyru (approved):
I'm good with this change if you checked it works as expected

PR #2408, opened 2025-08-07
fix: remove 'count' from logger level

PeterChen13579:


src/util/log/Logger.cpp
```diff
@@ -221,7 +221,7 @@ getMinSeverity(config::ClioConfigDefinition const& config, Severity defaultSever
     for (auto it = overrides.begin<util::config::ObjectView>(); it != overrides.end<util::config::ObjectView>(); ++it) {
         auto const& channelConfig = *it;
         auto const name = channelConfig.get<std::string>("channel");
-        if (std::ranges::count(Logger::kCHANNELS, name) == 0) {  // TODO: use std::ranges::contains when available
+        if (!std::ranges::contains(Logger::kCHANNELS, name)) {
```
godexsoft: Let's use 'not' here?

godexsoft (changes requested):

mathbunnyru (approved):

godexsoft (approved):
@mathbunnyru please check in context of `spdlog` update, you may have to do the same thing 👍

PR #2409, opened 2025-08-08
perf: Pass const reference Input into handlers

### LINKED ISSUE #2265: Move or pass be reference input into handlers
## Summary

Looks like we could move `Input` here: https://github.com/XRPLF/clio/blob/develop/src/rpc/common/impl/Processors.hpp#L47-L48

Or change all (where possible) handlers to get `Input` by const reference.

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/2265

### DIFF
--- a/src/rpc/common/Concepts.hpp
+++ b/src/rpc/common/Concepts.hpp
@@ -69,9 +69,10 @@ concept SomeProcessor = (SomeRequirement<T> or SomeModifier<T>);
  * @brief A process function that expects both some Input and a Context.
  */
 template <typename T>
-concept SomeContextProcessWithInput = requires(T a, typename T::Input in, typename T::Output out, Context const& ctx) {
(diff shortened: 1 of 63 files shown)


kuznetsss (changes requested):
@emreariyurek, thanks for fixing this!
Could you please change handler concept to make sure that all handlers are taking `Input` as constant reference?
https://github.com/XRPLF/clio/blob/develop/src/rpc/common/Concepts.hpp

emreariyurek:
> @emreariyurek, thanks for fixing this! Could you please change handler concept to make sure that all handlers are taking `Input` as constant reference? https://github.com/XRPLF/clio/blob/develop/src/rpc/common/Concepts.hpp

Done

godexsoft (approved):

kuznetsss (approved):
LGTM, thanks!

PR #2433, opened 2025-08-14
test: Write more data to fill socket buffer

mathbunnyru:


kuznetsss (dismissed):

tests/unit/util/requests/WsConnectionTests.cpp
```diff
@@ -171,7 +171,7 @@ TEST_F(WsConnectionTests, WriteTimeout)
         // Write is success even if the other side is not reading.
         // It seems we need to fill some socket buffer before the timeout occurs.
         while (not error.has_value()) {
-            error = connection->write("hello", yield, std::chrono::milliseconds{1});
+            error = connection->write(std::string(100'000, 'a'), yield, std::chrono::milliseconds{1});
```
kuznetsss, line 175: Maybe we should limit the allowed number of writes here?

mathbunnyru: Thanks, done 👍

kuznetsss (approved):

PR #2437, opened 2025-08-14
fix: Support canonical names for type in account_objects

### LINKED ISSUE #2275: Clio doesn't accept canonical names for type in account_objects
## Issue Description

For a value of `type` in `account_objects`, Clio accepts only `Short Name` but not `Canonical Name` listed in the [document](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_objects),  whereas rippled does both.

## Steps to Reproduce

Sample Request:
```
{
    "method": "account_objects",
    "params": [
        {
            "account": "rrhHkGLpow1GxEVJhi1LKzBRqzifJEaka5",
            "deletion_blockers_only": false,
            "limit": 20,
            "validated": true,
            "type": "Escrow"
        }
    ]
}
```

## Expected Result

Rippled response:
```
{
    "result": {
        "account": "rrhHkGLpow1GxEVJhi1LKzBRqzifJEaka5",
        "account_objects": [
            {
                "Account": "rPuijb9MYVFLdv4YjRc4gzUDDrHWNAWrb6",
                "Amount": {
                    "currency": "USD",
                    "issuer": "rhtTyZXorjvoz1JCMSYNLcY7x6LL7mwC2d",
                    "value": "10000"
                },
                "CancelAfter": 804783958,
                "Destination": "rrhHkGLpow1GxEVJhi1LKzBRqzifJEaka5",
                "DestinationNode": "0",
                "FinishAfter": 804783932,
                "Flags": 0,
                "IssuerNode": "0",
                "LedgerEntryType": "Escrow",
                "OwnerNode": "0",
                "PreviousTxnID": "FF857C63ECCBCF8D27B625497A064AE7BA2E22022BE4DAF9DBDB143600A5E7A4",
                "PreviousTxnLgrSeq": 166853,
                "index": "1E89B5CEC00DED60972861C31CA13FE2BD2EFCF414C6DB06ABB35A333895AFBA"
            }
        ],
        "ledger_current_index": 172498,
        "status": "success",
        "validated": false
    }
}
```

## Actual Result

Clio response:
```
{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "Invalid field 'type'.",
        "status": "error",
        "type": "response",
        "request": {
            "method": "account_objects",
            "params": [
                {
                    "account": "rrhHkGLpow1GxEVJhi1LKzBRqzifJEaka5",
                    "deletion_blockers_only": false,
                    "limit": 20,
                    "validated": true,
                    "type": "Escrow"
                }
            ]
        }
    },
    "warnings": [
        {
            "id": 2001,
        }
    ]
}
```

emreariyurek:
Fix: https://github.com/XRPLF/clio/issues/2275


src/util/LedgerUtils.hpp
```diff
@@ -50,30 +50,36 @@ class LedgerTypeAttribute {
...
 
-    constexpr LedgerTypeAttribute(char const* name, ripple::LedgerEntryType type, LedgerCategory category)
-        : type_(type), name_(name), category_(category)
+    constexpr LedgerTypeAttribute(
+        char const* name,
+        char const* rpcName,
+        ripple::LedgerEntryType type,
+        LedgerCategory category
+    )
+        : type_(type), name_(name), rpcName_(rpcName), category_(category)
```
PeterChen13579: Please use `{} `instead of `()` here.

src/rpc/handlers/AccountObjects.hpp
```diff
@@ -116,9 +114,7 @@ class AccountObjectsHandler {
              validation::Type<uint32_t>{},
              validation::Min(1u),
              modifiers::Clamp<int32_t>(kLIMIT_MIN, kLIMIT_MAX)},
-            {JS(type),
-             validation::Type<std::string>{},
-             validation::OneOf<std::string>(accountOwnedTypes.cbegin(), accountOwnedTypes.cend())},
+            {JS(type), validation::CustomValidators::accountTypeValidator},
```
PeterChen13579, line 116: The same change would need to be applied for `ledger_data` RPC, as in this [document](https://github.com/XRPLF/rippled/issues/4393) ledger_data response also includes a "type" field, which has been updated to use canonical names.

PeterChen13579 (changes requested):
Thank you so much for adding this feature to Clio! Appreciate your time and it's almost there. Just a couple of changes needed 🙌

tests/unit/util/LedgerUtilsTests.cpp
```diff
@@ -139,3 +104,85 @@ TEST(LedgerUtilsTests, DeletionBlockerTypes)
...
+    std::string input;
+    ripple::LedgerEntryType expected;
+};
+
+class AccountOwnedLedgerTypeFromStrTest : public ::testing::TestWithParam<LedgerTypeParam> {};
+
+TEST_P(AccountOwnedLedgerTypeFromStrTest, Test)
+{
+    auto const& param = GetParam();
+    auto result = util::LedgerTypes::getAccountOwnedLedgerTypeFromStr(param.input);
```
godexsoft: nit: Can be `const`

godexsoft (changes requested):
Overall this looks good and Thank you for working on it 👍 
Let's address the nits we left for you and we should be able to get this merged 🙇

src/util/LedgerUtils.hpp
kuznetsss: Why? I think it doesn't make a difference here.

tests/unit/util/LedgerUtilsTests.cpp
kuznetsss: Fixed

src/rpc/handlers/AccountObjects.hpp
kuznetsss: Added, please have a look.

kuznetsss:
@emreariyurek, we want to include this into 2.6.0 release so I fixed review comments. Hope that's ok for you.

emreariyurek:
> @emreariyurek, we want to include this into 2.6.0 release so I fixed review comments. Hope that's ok for you.

Sure, it works for me! Sorry, slow reply — summer mode limited my access :)

src/rpc/handlers/AccountObjects.hpp
emreariyurek: As far as I understand from the document [ledger_data](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_data#:~:text=response%20left%20off.-,type,-String) & [account_objects](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_objects#:~:text=response%20left%20off.-,type,-String) since ledger_data accepts all kinds of ledger categories, just using an account-type validator won’t fully cover this case (but I might be mistaken)

src/util/LedgerUtils.hpp
PeterChen13579: It doesn't make a difference here but just like to use `{}` for consistency. It's fine to leave it as is

emreariyurek: Done

src/rpc/handlers/AccountObjects.hpp
PeterChen13579: @emreariyurek , should be enough, I tested a few `types` locally and it seems to work fine. Are you thinking of any particular cases?

emreariyurek: No, I mean in the original code, ledger_data uses [getLedgerEntryTypeStrList](https://github.com/XRPLF/clio/blob/0a2930d861825c2b7f04fe8149e20e0b7a612b1f/src/rpc/handlers/LedgerData.hpp#L116) in the spec, which means it accepts all ledger entry types. However, we now ignore type for the chain ledger category like amendments, fee, etc. due to using the same validator of account_objects & this [filter](https://github.com/XRPLF/clio/blob/b37fe11e1d512282092b0fb5280eb2ba6753124d/src/util/LedgerUtils.cpp#L49-L51) If that’s not an issue, then it should be fine

PeterChen13579: @emreariyurek , I tested and you're right. Thanks for the catch! @kuznetsss , for ledger_data, you will need to create another accountTypeValidator which does not have this [filter](https://github.com/XRPLF/clio/blob/b37fe11e1d512282092b0fb5280eb2ba6753124d/src/util/LedgerUtils.cpp#L49-L51. Maybe have 2 functions, one calls the other one which doesn't have the filter.

kuznetsss:
@emreariyurek, thanks for your contribution! I added a search optimisation in `LedgerUtils`, could you please have a look when you have time?

PeterChen13579 (approved):
LGTM

emreariyurek:
> @emreariyurek, thanks for your contribution! I added a search optimisation in `LedgerUtils`, could you please have a look when you have time?

Thank you, LGTM!

PR #2440, opened 2025-08-15
refactor: Put log options in log section in config

### LINKED ISSUE #2417: Move all log configuration options into separate `log` section in config
For example, instead of `spdlog_format`, `log_channels` in the global scope, it will look like:
```json
{
  "log": {
    "format": "...",
    "channels": "..."
  }
}
```

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2417

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,12 +1,13 @@
 # Logging
 
-Clio provides several logging options, which all are configurable via the config file. These are detailed in the following sections.
+Clio provides several logging options, which all are configurable via the config file under the `log` section.
+These are detailed in the following sections.
 
-## `log_level`
+## `log.level`
 
 The minimum level of severity at which the log message will be outputted by default. Severity options are `trace`, `debug`, `info`, `warning`, `error`, `fatal`. Defaults to `info`.
 
-## `spdlog_format`
+## `log.format`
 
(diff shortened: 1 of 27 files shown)


docs/logging.md
```diff
@@ -1,12 +1,13 @@
 # Logging
 
-Clio provides several logging options, which all are configurable via the config file. These are detailed in the following sections.
+Clio provides several logging options, which all are configurable via the config file under the `log` section.
+These are detailed in the following sections.
 
-## `log_level`
+## `log.level`
```
kuznetsss, line 6: Do we need this file at all? Looks like it duplicates `config-description.md`.

mathbunnyru: This one is human written and is specifically about logging.
I think it's not difficult to support it, so let's keep it

kuznetsss: Config description was created specifically to generate documentation for each config entry. I don't think there is a value of this file containing exactly the same information in some other place.

mathbunnyru: Ok, I'll do it after merging this PR.
https://github.com/XRPLF/clio/issues/2447

kuznetsss (approved):

PeterChen13579 (approved):
LGTM

PR #2448, opened 2025-08-18
docs: Update information about minimum compilers

mathbunnyru:


docs/build-clio.md
```diff
@@ -11,6 +11,11 @@
 - [**Optional**] [GCovr](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html): needed for code coverage generation
 - [**Optional**] [CCache](https://ccache.dev/): speeds up compilation if you are going to compile Clio often
 
+We use our Docker image `ghcr.io/XRPLF/clio-ci` to build `Clio`, see [Building Clio with Docker](#building-clio-with-docker).
+You can find information about exact compiler versions and tools in the [image's README](https://github.com/XRPLF/clio/blob/develop/docker/ci/README.md).
+
+Compilers below the following versions are unlikely to work without drastic changes to the codebase:
+
```
kuznetsss, line 19: Probably we should update the table below to our current compiler versions.

mathbunnyru: Well, we need to decide what we want to specify here.

For now, a phrase `Compilers below the following versions are unlikely to work without drastic changes to the codebase:` means that for example GCC 12.2 won't work easily at all, but also doesn't guarantee that GCC 12.3 will work without some minor changes.
And it reflects the current state.

kuznetsss: Maybe we should put there our current compilers and say something like: 
```
The following compiler version are guarantied to work. Any compiler with lower version may not be able to build Clio.
```

mathbunnyru: Done, please take a look

docs/build-clio.md
```diff
@@ -11,11 +11,17 @@
 - [**Optional**] [GCovr](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html): needed for code coverage generation
 - [**Optional**] [CCache](https://ccache.dev/): speeds up compilation if you are going to compile Clio often
 
+We use our Docker image `ghcr.io/XRPLF/clio-ci` to build `Clio`, see [Building Clio with Docker](#building-clio-with-docker).
+You can find information about exact compiler versions and tools in the [image's README](https://github.com/XRPLF/clio/blob/develop/docker/ci/README.md).
+
+The following compiler version are guaranteed to work.
+Any compiler with lower version may not be able to build Clio.:
```
kuznetsss:
```suggestion
Any compiler with lower version may not be able to build Clio:
```

mathbunnyru: You set me up 😃

kuznetsss (approved):

PR #2449, opened 2025-08-18
fix: Fix default severity in Logger

mathbunnyru:
`data.severity` was uninitialized at all after changing to spdlog 😮 
I renamed it to `data.defaultSeverity` to make it more clear and initialized from config and used it


PeterChen13579 (approved):
Tested logger with rippled using example-config and works nicely 👍

kuznetsss (approved):
👍 
Please create a ticket to refactor this part to avoid such bug in the future.

mathbunnyru:
In the long term we should think about more robust solution: https://github.com/XRPLF/clio/issues/2451

PR #2454, opened 2025-08-18
feat: Support Keyspace

PeterChen13579:
Support AWS Keyspace queries


docs/config-description.md
```diff
@@ -87,6 +87,14 @@ This document provides a list of all available Clio configuration properties in
 - **Constraints**: The minimum value is `1`. The maximum value is `4294967295`.
 - **Description**: Represents the number of threads that will be used for database operations.
 
+### database.cassandra.provider
+
+- **Required**: True
+- **Type**: string
+- **Default value**: `scylladb`
```
kuznetsss: I think using the value `scylladb` when running with Cassandra may be a bit confusing. Maybe `cassandra` would be better instead?

src/data/CassandraBackend.hpp
```diff
@@ -514,44 +515,79 @@ class BasicCassandraBackend : public BackendInterface {
     ) const override
     {
         NFTsAndCursor ret;
+        std::vector<ripple::uint256> nftIDs;
+
+        // --- A specific taxon is requested ---
+        if (taxon.has_value()) {
```
kuznetsss: Maybe move this if-else section into 3 separate functions?

src/data/cassandra/impl/Batch.cpp
```diff
@@ -38,7 +38,7 @@ namespace data::cassandra::impl {
 
 // TODO: Use an appropriate value instead of CASS_BATCH_TYPE_LOGGED for different use cases
 Batch::Batch(std::vector<Statement> const& statements)
-    : ManagedObject{cass_batch_new(CASS_BATCH_TYPE_LOGGED), kBATCH_DELETER}
+    : ManagedObject{cass_batch_new(CASS_BATCH_TYPE_UNLOGGED), kBATCH_DELETER}
```
kuznetsss, line 50: Should the TODO above be removed? Please leave a comment in the code why this value should be used.

src/data/cassandra/impl/Cluster.cpp
```diff
@@ -60,6 +60,11 @@ Cluster::Cluster(Settings const& settings) : ManagedObject{cass_cluster_new(), k
     cass_cluster_set_connect_timeout(*this, settings.connectionTimeout.count());
     cass_cluster_set_request_timeout(*this, settings.requestTimeout.count());
 
+    // TODO: AWS keyspace reads should be local_one to save cost
+    if (auto const rc = cass_cluster_set_consistency(*this, CASS_CONSISTENCY_LOCAL_QUORUM); rc != CASS_OK) {
```
kuznetsss: Does it affect Scylla/Cassandra?

docs/config-description.md
PeterChen13579: Hmm, Keyspace actually uses Cassandra under the hood. If you think of a better way to provide distinction, please let me know

kuznetsss: I'm saying that for users running Cassandra put `scylladb` here is confusing.

kuznetsss (commented):
LGTM, but let's wait for an approval from @godexsoft

src/data/CassandraBackend.hpp
```diff
@@ -225,8 +225,11 @@ class BasicCassandraBackend : public BackendInterface {
     {
         waitForWritesToFinish();
 
+        // !range means the table 'ledger_range' is not populated it; This would be first write to the table
+        // In this case, insert both min_sequence/max_sequence range into the table
         if (!range_) {
-            executor_.writeSync(schema_->updateLedgerRange, ledgerSequence_, false, ledgerSequence_);
+            executor_.writeSync(schema_->insertLedgerRange, false, ledgerSequence_);
+            executor_.writeSync(schema_->insertLedgerRange, true, ledgerSequence_);
```
godexsoft: Why was this not needed before, do you know?

docs/config-description.md
```diff
@@ -89,6 +89,14 @@ This document provides a list of all available Clio configuration properties in
 - **Constraints**: The minimum value is `1`. The maximum value is `4294967295`.
 - **Description**: Represents the number of threads that will be used for database operations.
 
+### database.cassandra.provider
+
+- **Required**: True
+- **Type**: string
+- **Default value**: `cassandra`
+- **Constraints**: The value must be one of the following: `cassandra`, `aws_keyspace`.
```
godexsoft, line 97: May as well add `scylladb` to the mix then.. if we are using the same backend for multiple DBs already

src/data/cassandra/Schema.hpp
```diff
@@ -347,6 +348,71 @@ class Schema {
...
+                           AND (taxon, token_id) > ?
+                        ORDER BY taxon ASC, token_id ASC
+                         LIMIT ?
+                        )",
+                            qualifiedTableName(settingsProvider_.get(), "issuer_nf_tokens_v2")
+                        )
+                    );
+                }();
+                // AWS_keyspace supported queries
+            } else if (settingsProvider_.get().getSettings().provider == "aws_keyspace") {
```
godexsoft: Please can you explain why we need so many changes and what exactly is not supported so we had to do this?

src/data/cassandra/Schema.hpp
```diff
@@ -526,14 +592,25 @@ class Schema {
...
+        }();
+
         PreparedStatement updateLedgerRange = [this]() {
             return handle_.get().prepare(
                 fmt::format(
                     R"(
                 UPDATE {}
                    SET sequence = ?
                  WHERE is_latest = ?
-                    IF sequence IN (?, null)
```
godexsoft, line 536: This change was suggested by scylla team before IIRC.. this gave us a good speedup - let's not go back to the slower old version

godexsoft (commented):
I have not thoroughly reviewed the code because i don't yet understand why we need so many changes if keyspaces is supposed to be compatible with cassandra?
In general i'm not a fan of adding custom stuff to the cassandra backend.. if we need to have custom behaviour maybe we should create a separate backend and inherit where we can, modify where we need.

PeterChen13579:
@godexsoft For some context, Keyspace is compatible with Cassandra but there's a few limitations. Specifically the places where we had: 

```
IF ... IN clause 
PER PARTITION LIMIT 1 
Tuple comparisons, e.g.  (taxon, token_id) > ? 
```
To tackle this, I split the queries that contains the above into 2 queries. ie, 2 queries will be equivalent to the 1 from before. 

The goal of this PR is to get Clio to run with Keyspace while those running cassandra/scylladb would work the exact same. 
Also, keyspace is trying to support the above statements that they don't support right now. Hopefully once they get it supported, I'm happy to revert this PR.

src/data/CassandraBackend.hpp
PeterChen13579: Because before, the UpdateLedgerRange would insert the first range that it finds, let's say rippled started at ledger 256; this would insert 256, and the `false` means it is **not**  the latest ledger in our `ledger_range` table. 
Since the `updateLedgerRange` can't be used in Keyspace, I simplified it to both keyspace and scylla using this insertLedgerRange schema, as I felt like it is easier to understand the logic behind it. (ie, first time clio loads up without a ledger range table, insert both the latest and non latest range into the table)

src/data/cassandra/Schema.hpp
PeterChen13579: I can revert the logic so that scylladb uses this 👍

src/data/cassandra/Schema.hpp
PeterChen13579: Explained in my other comment 👍

docs/config-description.md
PeterChen13579: I think the reason we didn't add `scylladb` is because `scylladb` and `cassandra` would be the same as it uses fully the same schemas

docs/config-description.md
```diff
@@ -89,6 +89,14 @@ This document provides a list of all available Clio configuration properties in
...
+- **Description**: The specific database backend provider we are using. Currently we only support cassandra, or aws_keyspace.
```
mathbunnyru:
```suggestion
- **Description**: The specific database backend provider we are using.
```

This is already written as a constraint just one line above, so let's not duplicate?

src/data/CassandraBackend.hpp
```diff
@@ -225,8 +225,11 @@ class BasicCassandraBackend : public BackendInterface {
     {
         waitForWritesToFinish();
 
+        // !range means the table 'ledger_range' is not populated it; This would be first write to the table
+        // In this case, insert both min_sequence/max_sequence range into the table
         if (!range_) {
```
mathbunnyru:
```suggestion
        // !range_.has_value() means the table 'ledger_range' is not populated; This would be the first write to the table
        // In this case, insert both min_sequence/max_sequence range into the table
        if (!range_.has_value()) {
```

src/data/CassandraBackend.hpp
```diff
@@ -513,80 +516,16 @@ class BasicCassandraBackend : public BackendInterface {
...
-
-        for (auto i = 0u; i < nftIDs.size(); i++) {
-            if (auto const maybeRow = nftInfos[i].template get<uint32_t, ripple::AccountID, bool>(); maybeRow) {
-                auto [seq, owner, isBurned] = *maybeRow;
-                NFT nft(nftIDs[i], seq, owner, isBurned);
-                if (auto const maybeUri = nftUris[i].template get<ripple::Blob>(); maybeUri)
-                    nft.uri = *maybeUri;
-                ret.nfts.push_back(nft);
-            }
+        // --- A specific taxon is requested ---
```
mathbunnyru: This comment should be inside if statement (and it will look symmetrical with the next comment)

Or, remove both comments

src/data/CassandraBackend.hpp
```diff
@@ -1116,6 +1056,139 @@ class BasicCassandraBackend : public BackendInterface {
...
+    std::vector<ripple::uint256>
+    fetchNFTIDsWithoutTaxon(
+        ripple::AccountID const& issuer,
+        std::uint32_t const limit,
+        std::optional<ripple::uint256> const& cursorIn,
+        boost::asio::yield_context yield
+    ) const
+    {
+        std::vector<ripple::uint256> nftIDs;
+        if (settingsProvider_.getSettings().provider == "aws_keyspace") {
```
mathbunnyru: Let's have 2 private methods, and here just call one or another under if

src/data/cassandra/Schema.hpp
```diff
@@ -347,6 +348,86 @@ class Schema {
         Statements(SettingsProviderType const& settingsProvider, Handle const& handle)
             : settingsProvider_{settingsProvider}, handle_{std::cref(handle)}
         {
+            // initialize scylladb supported queries
+            if (settingsProvider_.get().getSettings().provider == "scylladb") {
```
mathbunnyru: Let's also have 2 private methods here

src/data/cassandra/Schema.hpp
```diff
@@ -654,6 +746,10 @@ class Schema {
             );
         }();
 
+        /*
```
mathbunnyru: This creates dead uncompiled code. Let's move it to method, and call it from here, but commented.

src/data/cassandra/Schema.hpp
```diff
@@ -953,6 +1004,15 @@ class Schema {
...
         }();
+
+        // For ScyllaDB / Cassandra ONLY
+        std::optional<PreparedStatement> selectAccountFromBeginningScylla;
+        std::optional<PreparedStatement> selectAccountFromTokenScylla;
+        std::optional<PreparedStatement> selectNFTsByIssuerScylla;
+
+        // For AWS Keyspaces ONLY
+        // NOTE: AWS keyspace is not able to load cache with accounts
+        std::optional<PreparedStatement> selectNFTsAfterTaxonKeyspaces;
```
mathbunnyru: Should be `std::variant<CassandraStatements, AWSKeyspacesStatements>`, and each one is a struct of needed statements, you will probably be able to get rid of std::optional as well

src/data/cassandra/impl/Cluster.cpp
```diff
@@ -60,6 +60,13 @@ Cluster::Cluster(Settings const& settings) : ManagedObject{cass_cluster_new(), k
     cass_cluster_set_connect_timeout(*this, settings.connectionTimeout.count());
     cass_cluster_set_request_timeout(*this, settings.requestTimeout.count());
 
+    // TODO: AWS keyspace reads should be local_one to save cost
```
mathbunnyru, line 63: Let's create a Jira issue?

src/data/cassandra/impl/Cluster.hpp
```diff
@@ -45,6 +45,7 @@ struct Settings {
     static constexpr uint32_t kDEFAULT_MAX_WRITE_REQUESTS_OUTSTANDING = 10'000;
     static constexpr uint32_t kDEFAULT_MAX_READ_REQUESTS_OUTSTANDING = 100'000;
     static constexpr std::size_t kDEFAULT_BATCH_SIZE = 20;
+    static constexpr std::string kDEFAULT_PROVIDER = "scylladb";
```
mathbunnyru: Let's remove this line, since we already have a default value specified in config.

Also, this default value is incorrect, because it's not one of the 2 we support.

src/data/cassandra/impl/Cluster.hpp
```diff
@@ -88,6 +89,9 @@ struct Settings {
     /** @brief Size of batches when writing */
     std::size_t writeBatchSize = kDEFAULT_BATCH_SIZE;
 
+    /** @brief Provider to know if we are using scylladb or keyspace */
+    std::string provider = kDEFAULT_PROVIDER;
```
mathbunnyru: Let's create an enum for the provider.
I know that the config definition doesn't support the enumeration, but I think we shouldn't use raw strings and raw strings comparison in the code itself, because it's error-prone (see my comment above for incorrect `kDEFAULT_PROVIDER` value)

tests/integration/data/BackendFactoryTests.cpp
```diff
@@ -53,6 +53,7 @@ struct BackendCassandraFactoryTest : SyncAsioContextTest, util::prometheus::With
         {"database.cassandra.secure_connect_bundle", ConfigValue{ConfigType::String}.optional()},
         {"database.cassandra.port", ConfigValue{ConfigType::Integer}.optional()},
         {"database.cassandra.keyspace", ConfigValue{ConfigType::String}.defaultValue(kKEYSPACE)},
+        {"database.cassandra.provider", ConfigValue{ConfigType::String}.defaultValue("cassandra")},
```
mathbunnyru: To keep it as a line above, let's create `kPROVIDER` value?

mathbunnyru (dismissed):
Left some comments to mostly make the code more readable and less error-prone

src/data/cassandra/Schema.hpp
mathbunnyru: Why was this marked as resolved?

src/data/cassandra/Schema.hpp
mathbunnyru: Why was this marked resolved?

src/data/CassandraBackend.hpp
mathbunnyru: Why was this marked as resolved?

PeterChen13579: After discussing with @kuznetsss and @godexsoft , we are completely changing the structure of the backend, (ie, adding a cassandraFamily level where CassandraBackend and KeyspaceBackend is inheriting from) there's not going to be anymore checking `provider` in backend, so I marked it as resolved.

mathbunnyru: Ok, please, don't just "resolve" the issue, without any comment 😂

mathbunnyru: Keep it open until it's actually resolved, even if it's resolved not the way that reviewer (in this case me) suggested

src/data/cassandra/Schema.hpp
PeterChen13579: Marking as resolved now as no changes in `schema.hpp `

src/data/cassandra/Schema.hpp
PeterChen13579: Changed back to before 👍

src/data/cassandra/Schema.hpp
PeterChen13579: Marked as resolved as no changes in `schema.hpp`

src/data/CassandraBackend.hpp
PeterChen13579: Marked as resolve now as no changes in `schema.hpp` 😄

docs/config-description.md
godexsoft: I don't mind but what i meant is we could have both "scylladb" and "cassandra" be a valid string that both point to our cassandra implementation.

src/data/BackendFactory.hpp
```diff
@@ -55,9 +56,15 @@ makeBackend(util::config::ClioConfigDefinition const& config, data::LedgerCacheI
 
     if (boost::iequals(type, "cassandra")) {
         auto const cfg = config.getObject("database." + type);
-        backend = std::make_shared<data::cassandra::CassandraBackend>(
-            data::cassandra::SettingsProvider{cfg}, cache, readOnly
-        );
+        if (cfg.getValueView("provider").asString() == "aws_keyspace") {
```
godexsoft: There was a suggestion to use an `enum` for this which i too still think should be used. You can probably use your `Provider` enum here in some way given `to_string` for it exists

src/data/BackendFactory.hpp
```diff
@@ -55,9 +56,15 @@ makeBackend(util::config::ClioConfigDefinition const& config, data::LedgerCacheI
 
     if (boost::iequals(type, "cassandra")) {
```
godexsoft, line 57: Similarly this is using a magic string that could be updated once constants are added.

src/data/CassandraBackend.hpp
```diff
@@ -80,156 +67,43 @@ template <
...
+
+    // protected because CassandraMigrationBackend inherits from this class
 protected:
-    Handle handle_;
-
-    // have to be mutable because BackendInterface constness :(
-    mutable ExecutionStrategyType executor_;
-    // TODO: move to interface level
-    mutable FetchLedgerCacheType ledgerCache_{};
+    using DefaultCassandraFamily::executor_;
```
godexsoft, line 82: This is the same as `using CassandraBackendFamily::executor_;`. I don't think you need `DefaultCassandraFamily` above.

src/data/CassandraBackend.hpp
```diff
@@ -80,156 +67,43 @@ template <
...
-                "node with write access to DB should be started first.",
-                ex.what(),
-                readOnly
-            );
-            LOG(log_.error()) << error;
-            throw std::runtime_error(error);
-        }
-        LOG(log_.info()) << "Created (revamped) CassandraBackend";
-    }
+    using DefaultCassandraFamily::DefaultCassandraFamily;
```
godexsoft, line 92: Isn't the constructor already public? you only need this if the constructor was protected or private and you want to expose it publicly through the child class.

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,330 @@
...
+template <
+    SomeSettingsProvider SettingsProviderType,
+    SomeExecutionStrategy ExecutionStrategyType,
+    typename FetchLedgerCacheType = FetchLedgerCache>
+class BasicKeyspaceBackend : public CassandraBackendFamily<
+                                 SettingsProviderType,
+                                 ExecutionStrategyType,
+                                 KeyspaceSchema<SettingsProviderType>,
+                                 FetchLedgerCacheType> {
+    using DefaultCassandraFamily = CassandraBackendFamily<
```
godexsoft, line 73: Same story here

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,330 @@
...
+    std::vector<ripple::uint256>
+    fetchAccountRoots(
+        [[maybe_unused]] std::uint32_t number,
+        [[maybe_unused]] std::uint32_t pageSize,
+        [[maybe_unused]] std::uint32_t seq,
+        [[maybe_unused]] boost::asio::yield_context yield
+    ) const override
+    {
+        LOG(log_.error()) << "Fetching account roots is not supported by the Keyspaces backend.";
+        throw std::runtime_error("Fetching account roots is not supported by the Keyspaces backend.");
```
godexsoft: Shouldn't this be an ASSERT instead?

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,969 @@
...
+        return executor_.isTooBusy();
+    }
+
+    boost::json::object
+    stats() const override
+    {
+        return executor_.stats();
+    }
+
+private:
```
godexsoft: I think you may want to reuse this in the implementations so this probably should be `protected` now?

src/data/cassandra/KeyspaceSchema.hpp
```diff
@@ -0,0 +1,941 @@
...
+
+    /**
+     * @brief Recreates the prepared statements.
+     *
+     * @param handle The handle to the DB
+     */
+    void
+    prepareStatements(Handle const& handle)
+    {
+        LOG(log_.info()) << "Preparing keyspace cassandra statements";
```
godexsoft: I guess "aws keyspaces" statements or something

src/data/cassandra/impl/Cluster.hpp
```diff
@@ -31,10 +31,29 @@
...
 #include <string_view>
 #include <thread>
+#include <utility>
 #include <variant>
 
 namespace data::cassandra::impl {
 
+namespace {
+
+enum class Provider { CASSANDRA, KEYSPACE };
```
godexsoft: We need to be careful with caps-only names - they are usually reserved for macros. Let's use the same style we use elsewhere

src/data/CassandraBackend.hpp
godexsoft: I'm not sure why you need to re-expose something that is already `protected` in the base class.. maybe we can remove all of this and it should still work fine

godexsoft (changes requested):
I left a few comments, nothing major, should be easy enough to fix.
For the Schema inheritance idea let's discuss. I think we can still do slightly better than having two copies of the schema with modifications

godexsoft (dismissed):
Looks good 👍

godexsoft (approved):

src/data/BackendFactory.hpp
```diff
@@ -55,9 +56,15 @@ makeBackend(util::config::ClioConfigDefinition const& config, data::LedgerCacheI
...
         auto const cfg = config.getObject("database." + type);
-        backend = std::make_shared<data::cassandra::CassandraBackend>(
-            data::cassandra::SettingsProvider{cfg}, cache, readOnly
-        );
+        if (cfg.getValueView("provider").asString() == toString(cassandra::impl::Provider::Keyspace)) {
```
mathbunnyru, line 59: It's better to do it vice-versa, have a function to convert string to a enum, and then compare enum values

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+#include <vector>
+
+namespace data::cassandra {
+
+/**
+ * @brief Implements @ref CassandraBackendFamily for Keyspace
+ *
+ * @tparam SettingsProviderType The settings provider type to use
+ * @tparam ExecutionStrategyType The execution strategy type to use
+ * @tparam FetchLedgerCacheType The ledger header cache type to use
```
mathbunnyru, line 62: Let's remove `to use` in the end, it's not helpful

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+    bool
+    doFinishWrites() override
+    {
+        this->waitForWritesToFinish();
+
+        // !range_.has_value() means the table 'ledger_range' is not populated;
+        // This would be the first write to the table.
+        // In this case, insert both min_sequence/max_sequence range into the table.
+        if (not(range_.has_value())) {
```
mathbunnyru, line 104: Well, using `not` this way ( `not(x)` instead of `not x` ) is something new, let's not do it 😆

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+            executor_.writeSync(schema_->insertLedgerRange, false, ledgerSequence_);
+            executor_.writeSync(schema_->insertLedgerRange, true, ledgerSequence_);
+        }
+
+        if (not this->executeSyncUpdate(schema_->updateLedgerRange.bind(ledgerSequence_, true, ledgerSequence_ - 1))) {
```
mathbunnyru, line 109: I would probably add inline comments to a boolean variable like `executor_.writeSync(schema_->insertLedgerRange, /* some_opt_name = */ false, ledgerSequence_);`

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+        std::optional<ripple::uint256> const& cursorIn,
+        boost::asio::yield_context yield
+    ) const override
+    {
+        std::vector<ripple::uint256> nftIDs;
+        if (taxon.has_value()) {
+            // Keyspace and ScyllaDB uses the same logic for taxon-filtered queries
+            nftIDs = fetchNFTIDsByTaxon(issuer, *taxon, limit, cursorIn, yield);
+        } else {
+            // --- Amazon Keyspaces Workflow for non-taxon queries ---
```
mathbunnyru, line 133: No need for `---` in the comment, I don't think we do that

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+            auto const startTokenID = cursorIn.value_or(ripple::uint256(0));
+
+            Statement firstQuery = schema_->selectNFTIDsByIssuerTaxon.bind(issuer);
+            firstQuery.bindAt(1, startTaxon);
+            firstQuery.bindAt(2, startTokenID);
+            firstQuery.bindAt(3, Limit{limit});
+
+            auto const firstRes = executor_.read(yield, firstQuery);
+            if (firstRes) {
+                for (auto const [nftID] : extract<ripple::uint256>(firstRes.value()))
```
mathbunnyru, line 144: first line - `.has_value()`
second one - simply `*firstRes`

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+            if (nftIDs.size() < limit) {
+                auto const remainingLimit = limit - nftIDs.size();
+                Statement secondQuery = schema_->selectNFTsAfterTaxonKeyspaces.bind(issuer);
+                secondQuery.bindAt(1, startTaxon);
+                secondQuery.bindAt(2, Limit{remainingLimit});
+
+                auto const secondRes = executor_.read(yield, secondQuery);
+                if (secondRes) {
+                    for (auto const [nftID] : extract<ripple::uint256>(secondRes.value()))
```
mathbunnyru, line 156: first line - `.has_value()`
second one - simply `*secondRes`

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+                        nftIDs.push_back(nftID);
+                }
+            }
+        }
+        return populateNFTsAndCreateCursor(nftIDs, ledgerSequence, limit, yield);
+    }
+
+    /**
+     * @brief (Unsupported in Keyspaces) Fetches account root object indexes by page.
+     * * @note Loading the cache by enumerating all accounts is currently unsupported by the AWS Keyspaces backend.
```
mathbunnyru, line 166: Do we need extra start before note?

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+    {
+        std::vector<ripple::uint256> nftIDs;
+        Statement statement = schema_->selectNFTIDsByIssuerTaxon.bind(issuer);
+        statement.bindAt(1, taxon);
+        statement.bindAt(2, cursorIn.value_or(ripple::uint256(0)));
+        statement.bindAt(3, Limit{limit});
+
+        auto const res = executor_.read(yield, statement);
+        if (res && res.value().hasRows()) {
+            for (auto const [nftID] : extract<ripple::uint256>(res.value()))
```
mathbunnyru, line 207: same as above

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+        auto const startTokenID = cursorIn.value_or(ripple::uint256(0));
+
+        Statement firstQuery = schema_->selectNFTIDsByIssuerTaxon.bind(issuer);
+        firstQuery.bindAt(1, startTaxon);
+        firstQuery.bindAt(2, startTokenID);
+        firstQuery.bindAt(3, Limit{limit});
+
+        auto const firstRes = executor_.read(yield, firstQuery);
+        if (firstRes) {
+            for (auto const [nftID] : extract<ripple::uint256>(firstRes.value()))
```
mathbunnyru, line 233: same as above

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+
+        if (nftIDs.size() < limit) {
+            auto const remainingLimit = limit - nftIDs.size();
+            Statement secondQuery = schema_->selectNFTsAfterTaxonKeyspaces.bind(issuer);
+            secondQuery.bindAt(1, startTaxon);
+            secondQuery.bindAt(2, Limit{remainingLimit});
+
+            auto const secondRes = executor_.read(yield, secondQuery);
+            if (secondRes) {
+                for (auto const [nftID] : extract<ripple::uint256>(secondRes.value()))
```
mathbunnyru, line 245: same as above

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+    ) const
+    {
+        if (nftIDs.empty()) {
+            LOG(log_.debug()) << "No rows returned";
+            return {};
+        }
+
+        NFTsAndCursor ret;
+        if (nftIDs.size() == limit)
+            ret.cursor = nftIDs.back();
```
mathbunnyru, line 270: Can someone provide 0 as a limit and crash our app?

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+                return schema_->selectNFTURI.bind(nftID, ledgerSequence);
+            }
+        );
+
+        auto const nftInfos = executor_.readEach(yield, selectNFTStatements);
+        auto const nftUris = executor_.readEach(yield, selectNFTURIStatements);
+
+        // Combine the results into final NFT objects.
+        for (auto i = 0u; i < nftIDs.size(); ++i) {
+            if (auto const maybeRow = nftInfos[i].template get<uint32_t, ripple::AccountID, bool>(); maybeRow) {
```
mathbunnyru, line 294: `.has_value()`

src/data/KeyspaceBackend.hpp
```diff
@@ -0,0 +1,308 @@
...
+                auto [seq, owner, isBurned] = *maybeRow;
+                NFT nft(nftIDs[i], seq, owner, isBurned);
+                if (auto const maybeUri = nftUris[i].template get<ripple::Blob>(); maybeUri)
+                    nft.uri = *maybeUri;
```
mathbunnyru, line 298: same as above

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,1077 @@
...
+
+/**
+ * @brief Implements @ref BackendInterface for Cassandra/ScyllaDB/Keyspace.
+ *
+ * Note: This is a safer and more correct rewrite of the original implementation of the backend.
+ *
+ * @tparam SettingsProviderType The settings provider type to use
+ * @tparam ExecutionStrategyType The execution strategy type to use
+ * @tparam SchemaType The Schema type to use
+ * @tparam FetchLedgerCacheType The ledger header cache type to use
```
mathbunnyru, line 76: remove `to use`

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,1077 @@
...
+    mutable ExecutionStrategyType executor_;
+    // TODO: move to interface level
+    mutable FetchLedgerCacheType ledgerCache_{};
+
+public:
+    /**
+     * @brief Create a new cassandra/scylla backend instance.
+     *
+     * @param settingsProvider The settings provider to use
+     * @param cache The ledger cache to use
```
mathbunnyru, line 104: remove `to use`

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,1077 @@
...
+        , settingsProvider_{std::move(settingsProvider)}
+        , schema_{settingsProvider_}
+        , handle_{settingsProvider_.getSettings()}
+        , executor_{settingsProvider_.getSettings(), handle_}
+    {
+        if (auto const res = handle_.connect(); not res)
+            throw std::runtime_error("Could not connect to database: " + res.error());
+
+        if (not readOnly) {
+            if (auto const res = handle_.execute(schema_.createKeyspace); not res) {
```
mathbunnyru, line 118: same as above

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,1077 @@
...
+                // on datastax, creation of keyspaces can be configured to only be done thru the admin
+                // interface. this does not mean that the keyspace does not already exist tho.
+                if (res.error().code() != CASS_ERROR_SERVER_UNAUTHORIZED)
+                    throw std::runtime_error("Could not create keyspace: " + res.error());
+            }
+
+            if (auto const res = handle_.executeEach(schema_.createSchema); not res)
```
mathbunnyru, line 125: same as above

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -0,0 +1,1077 @@
...
+    /**
+     * @copydoc BackendInterface::fetchLatestLedgerSequence
+     */
+    std::optional<std::uint32_t>
+    fetchLatestLedgerSequence(boost::asio::yield_context yield) const override
+    {
+        if (auto const res = executor_.read(yield, schema_->selectLatestLedger); res) {
+            if (auto const& result = res.value(); result) {
+                if (auto const maybeValue = result.template get<uint32_t>(); maybeValue)
+                    return maybeValue;
```
mathbunnyru, line 251: same as above

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -93,6 +93,7 @@ class BackendCassandraTestBase : public SyncAsioContextTest, public WithPromethe
         {"database.cassandra.port", ConfigValue{ConfigType::Integer}.optional()},
         {"database.cassandra.keyspace",
          ConfigValue{ConfigType::String}.defaultValue(TestGlobals::instance().backendKeyspace)},
+        {"database.cassandra.provider", ConfigValue{ConfigType::String}.defaultValue("cassandra")},
```
mathbunnyru, line 96: Let's move it to a variable (and in places below as well)

src/data/cassandra/impl/Cluster.hpp
```diff
@@ -31,10 +31,29 @@
...
 #include <variant>
 
 namespace data::cassandra::impl {
 
+namespace {
+
+enum class Provider { Cassandra, Keyspace };
+
+inline std::string
+toString(Provider provider)
```
mathbunnyru, line 44: You will have a function providerFromString instead, and this function could be removed

mathbunnyru:
Left some comments, mostly about style

When you use optional/expected, I suggest using this pattern (`has_value()` + `*`):

```cpp
if (my_opt.has_value()) {
  use_it(*my_opt);
}
```

src/data/KeyspaceBackend.hpp
PeterChen13579: This isn't possible because in settings, we have a min value of 1. https://github.com/XRPLF/clio/blob/dc5f8b9c23ea494693358f717875f7212f33126a/src/rpc/handlers/NFTsByIssuer.hpp#L48

PR #2456, opened 2025-08-19
refactor: output struct

### LINKED ISSUE #2452: Refactor Output Struct Initialization to Use Designated Initializers
Currently in accountInfo class, we construct the output by: 
`Output myOutput(ledgerId, "hash...", mySle, false, true .... );`

It is better to use Designated Initializers, such that it will be:

```cpp
Output myOutput{
    .ledgerIndex = ledgerId,
    .ledgerHash = "hash...",
    .accountData = std::move(mySle),
    .isDisallowIncomingEnabled = false,
}
```

PeterChen13579:
fixes #2452


src/rpc/handlers/GetAggregatePrice.hpp
```diff
@@ -59,8 +59,8 @@ class GetAggregatePriceHandler {
      * @brief A struct to hold the statistics
      */
     struct Stats {
-        ripple::STAmount avg;
-        ripple::Number sd;  // standard deviation
+        ripple::STAmount avg{};  // NOLINT(readability-redundant-member-init)
+        ripple::Number sd{};     // NOLINT(readability-redundant-member-init). standard deviation
```
mathbunnyru: maybe move `// standard deviation` on it's own line above this field declaration?

src/rpc/handlers/AccountInfo.cpp
```diff
@@ -112,27 +113,28 @@ AccountInfoHandler::process(AccountInfoHandler::Input const& input, Context cons
...
+        return Output{
+            .ledgerIndex = lgrInfo.seq,
+            .ledgerHash = ripple::strHex(lgrInfo.hash),
+            .accountData = sle,
+            .isDisallowIncomingEnabled = isDisallowIncomingEnabled,
+            .isClawbackEnabled = isClawbackEnabled,
+            .isTokenEscrowEnabled = isTokenEscrowEnabled,
+            .apiVersion = ctx.apiVersion,
+            .signerLists = signerList
+        };
```
mathbunnyru: Can we only do one return in the end maybe?
As far as I see, the only difference is `.signerLists`: we can probably move vector out of this `if` and in the end write ternary.

PeterChen13579: yea that's a good idea

src/rpc/handlers/AccountInfo.cpp
```diff
@@ -109,30 +121,11 @@ AccountInfoHandler::process(AccountInfoHandler::Input const& input, Context cons
             if (!signersKey.check(sleSigners))
                 return Error{Status{RippledError::rpcDB_DESERIALIZATION}};
 
-            signerList.push_back(sleSigners);
+            out.signerLists = std::vector<ripple::STLedgerEntry>{sleSigners};
```
mathbunnyru, line 123: I don't see where you deleted a vector

mathbunnyru (changes requested):

src/rpc/handlers/AccountInfo.cpp
PeterChen13579: I never deleted a vector.. This is now an assignment operator instead of `push_back` because SignarList is initialized as nullopt.

mathbunnyru: > I never deleted a vector.

That's exactly the problem I'm talking about :)
You now have `signerList` vector above, which is created, but unused

mathbunnyru (approved):

PR #2457, opened 2025-08-19
feat: Validate unexpected config values

mathbunnyru:


src/util/config/ConfigDefinition.cpp
```diff
@@ -237,6 +237,12 @@ ClioConfigDefinition::parse(ConfigFileInterface const& config)
         });
     }
 
+    for (auto const& key : config.getAllKeys()) {
+        if (!map_.contains(key) && !arrayPrefixesToKeysMap.contains(key)) {
+            listOfErrors.emplace_back("Unknown key: " + key);
+        }
+    }
+
```
kuznetsss, line 245: Maybe you should add tests for this logic?

mathbunnyru: Done

kuznetsss (approved):

PR #2460, opened 2025-08-20
fix: Print error and help on unknown arg

### LINKED ISSUE #2458: Write error and help on `clio_server --unknown-arg `
If I run `./clio_server --unknown-arg`, it fails and doesn't produce any output.

We should:
1. Write a reason why it failed (unknown arg)
2. Write help output as well

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2458


src/app/CliArgs.cpp
```diff
@@ -57,8 +57,14 @@ CliArgs::parse(int argc, char const* argv[])
...
 
     po::variables_map parsed;
-    po::store(po::command_line_parser(argc, argv).options(description).positional(positional).run(), parsed);
-    po::notify(parsed);
+    try {
+        po::store(po::command_line_parser(argc, argv).options(description).positional(positional).run(), parsed);
+        po::notify(parsed);
+    } catch (po::error const& e) {
+        std::cerr << "Error: " << e.what() << std::endl << std::endl;
+        std::cout << "Clio server " << util::build::getClioFullVersionString() << "\n\n" << description;
```
kuznetsss: This should be moved to a local lambda to avoid duplication

kuznetsss (approved):

PR #2464, opened 2025-08-20
fix: Shutdown LogService after exceptions are printed

mathbunnyru:


src/main/Main.cpp
```diff
@@ -76,10 +72,22 @@ try {
...
-    LOG(util::LogService::fatal()) << "Exit on exception: unknown";
-    return EXIT_FAILURE;
+}
+
+int
+main(int argc, char const* argv[])
+{
+    util::setTerminationHandler();
+
+    util::ScopeGuard const loggerShutdownGuard{[]() { util::LogService::shutdown(); }};
```
godexsoft: We can drop the `()` in the lambda now

godexsoft (changes requested):

godexsoft (dismissed):
:rage3:

godexsoft (approved):

PR #2469, opened 2025-08-21
fix: Fix package issues

mathbunnyru:


CMakeLists.txt
```diff
@@ -40,11 +40,6 @@ if (verbose)
   set(CMAKE_VERBOSE_MAKEFILE TRUE)
 endif ()
 
-if (packaging)
-  add_definitions(-DPKG=1)
```
godexsoft, line 44: @legleux this is no longer needed, right? 
I assume it was previously used in the external project where the cmake file for packaging was included from?

legleux: Nope, no longer needed.

legleux (approved):

godexsoft (approved):

PR #2472, opened 2025-08-22
chore: Cleanup packages installed on macOS runner

mathbunnyru:


mathbunnyru:
I don't think any of these are needed on our macOS runners, after removing them we'll be able to make `prepare_runner` an action which is used by both `rippled` and `clio`

godexsoft (approved):
:rage1:

PR #2473, opened 2025-08-22
ci: Use XRPLF/prepare_actions

mathbunnyru:


godexsoft (approved):
Nice! In the future it would be extra cool to have dependabot somehow detect newer versions of our actions and suggest to update in Clio/rippled 🔥

PR #2480, opened 2025-08-27
fix: ASAN issues in CommunicationService

godexsoft:
This PR fixes ASAN issues from CommunicationService and its tests.
While the code was already correct according to Asio docs and was passing tests, ASAN is a bit needy and if it can't see through the entire process it will detect issues that it probably should not. 
The way to make ASAN happy is to make sure that the thread pool cleanly exits after all spawned coroutines were gracefully finished. I suspect we will have to add similar fixes elsewhere too.

### DIFF
--- a/src/cluster/ClusterCommunicationService.cpp
+++ b/src/cluster/ClusterCommunicationService.cpp
@@ -47,7 +47,7 @@
 #include <vector>
 
 namespace {
-auto constexpr kTOTAL_WORKERS = 2uz;  // 1 reading and 1 writing worker (coroutines)
+constexpr auto kTOTAL_WORKERS = 2uz;  // 1 reading and 1 writing worker (coroutines)
 }  // namespace
 
 namespace cluster {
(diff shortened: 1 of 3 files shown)


kuznetsss (dismissed):
👍

src/cluster/ClusterCommunicationService.cpp
```diff
@@ -36,11 +40,16 @@
...
 #include <chrono>
 #include <ctime>
+#include <latch>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
+namespace {
+auto constexpr kTOTAL_WORKERS = 2uz;  // 1 reading and 1 writing worker (coroutines)
```
mathbunnyru:
```suggestion
constexpr auto kTOTAL_WORKERS = 2uz;  // 1 reading and 1 writing worker (coroutines)
```

mathbunnyru (approved):
Great job 👍

PR #2485, opened 2025-08-28
style: Fix JSON indent style in C++ code

mathbunnyru:


tests/unit/app/VerifyConfigTests.cpp
```diff
@@ -38,7 +38,7 @@ TEST(VerifyConfigTest, ValidConfig)
 {
     // used to Verify Config test
     static constexpr auto kVALID_JSON_DATA = R"JSON({
-         "server": {
+        "server": {
```
mathbunnyru, line 41: This file I changed manually, because the formatting was really bad before

mathbunnyru:
Note: this is not making json beautiful, it was written in a bad way, but I'm not sure it can be done easily at all.

What might:

1. clang-format might change the indentation of the first line where `JSON(` started
2. It can't and won't change next lines inside json, since it's not allowed to change the meaning of the code (and changing spaces is actually changing string literals, so changing code as well)
3. My tool will detect, that indentantion is not expected and fix it

godexsoft:
> 1. clang-format might change the indentation of the first line where `JSON(` started
> 2. It can't and won't change next lines inside json, since it's not allowed to change the meaning of the code (and changing spaces is actually changing string literals, so changing code as well)
> 3. My tool will detect, that indentantion is not expected and fix it

My concern here is that a dev will add some JSON, pre-commit will "fix" the first line using clang-format, then your tool will fix it again to what we like to see. Then all pass and the dev will push the changes. Then on CI clang-format check will fail.. or am i misunderstanding something?

mathbunnyru:
> > 1. clang-format might change the indentation of the first line where `JSON(` started
> > 2. It can't and won't change next lines inside json, since it's not allowed to change the meaning of the code (and changing spaces is actually changing string literals, so changing code as well)
> > 3. My tool will detect, that indentantion is not expected and fix it
> 
> My concern here is that a dev will add some JSON, pre-commit will "fix" the first line using clang-format, then your tool will fix it again to what we like to see. Then all pass and the dev will push the changes. Then on CI clang-format check will fail.. or am i misunderstanding something?

Let me explain the main property of how `clang-format` works and what it means for strings.
It is not allowed to change the behaviour of the program in any way.
So, it doesn't care if it's json or not json inside the string and won't in any way change the string itself.
It might change the first line before the string starts, yes, but not the content.

My tool changes only content, and it doesn't change anything before first line of the json string itself.

That's why I don't think there will be a situation, where they will break each other infinitely.

For example, in this PR, clang-format didn't break even once.

godexsoft (approved):

PR #2489, opened 2025-08-29
chore: Use boost 1.88 and xrpl/2.6.0

godexsoft:
xrpl 2.6.0 got released but did not include https://github.com/XRPLF/rippled/commit/1506e65558818ea73a2233d2a55c1d1441d5fb10 so this PR uses a custom build of 2.6.0 with the cherry-picked boost 1.88 commit. This allows Clio to move over to boost 1.88 and use the latest xrpl at the same time


src/data/AmendmentCenter.hpp
```diff
@@ -153,6 +155,7 @@ struct Amendments {
     REGISTER(fixNFTokenNegOffer);
 
     // Retired amendments
+    REGISTER(OwnerPaysFee);  // Removed in xrpl 2.6.0 (https://github.com/XRPLF/rippled/pull/5435)
```
godexsoft, line 158: This is probably fine but we need to make sure we can still use it for history in some meaningful way (i.e. check this does not break anything).
I had to change it in the test below because once it is retired (supported=no) we assert in accessor code so it fails which suggests that it might break something else too.

mathbunnyru (approved):
Best change ever 👍

kuznetsss:
Shouldn't we wait until libxrpl is switched to boost 1.88 instead of using a custom build?

PR #2490, opened 2025-08-29
feat: Proxy support

kuznetsss:
Add client IP resolving support in case when there is a proxy in front of Clio.


docs/config-description.md
```diff
@@ -329,6 +329,22 @@ This document provides a list of all available Clio configuration properties in
 - **Constraints**: The minimum value is `1`. The maximum value is `4294967295`.
 - **Description**: Maximum queue size for sending subscription data to clients. This queue buffers data when a client is slow to receive it, ensuring delivery once the client is ready.
 
+### server.proxy_ips.[]
```
mathbunnyru:
```suggestion
### server.proxy.ips.[]
```

I think we should put `proxy` in it's own section - this way config will be better structured

src/app/ClioApplication.cpp
```diff
@@ -178,7 +178,9 @@ ClioApplication::run(bool const useNgWebServer)
         }
         auto const adminVerifier = std::move(expectedAdminVerifier).value();
 
-        auto httpServer = web::ng::makeServer(config_, OnConnectCheck{dosGuard}, DisconnectHook{dosGuard}, ioc);
+        auto httpServer = web::ng::makeServer(
+            config_, OnConnectCheck{dosGuard}, IpChangeHook{dosGuard}, DisconnectHook{dosGuard}, ioc
```
mathbunnyru, line 182: Let's create a class `ServerHooks` and have all 3 there - this way there will be less params, and if new hook is needed, there will be much less code as well.

src/app/WebHandlers.hpp
```diff
@@ -64,6 +65,16 @@ class OnConnectCheck {
...
 };
 
+class IpChangeHook {
+    std::reference_wrapper<web::dosguard::DOSGuardInterface> dosguard_;
+
+public:
+    IpChangeHook(web::dosguard::DOSGuardInterface& dosguard);
+
+    void
+    operator()(std::string const& oldIp, std::string const& newIp);
```
mathbunnyru, line 90: maybe 2 `string_view`s?

src/util/Shasum.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <string_view>
+
+namespace util {
+
+ripple::uint256
+sha256sum(std::string_view s)
+{
+    ripple::sha256_hasher hasher;
+    hasher(s.data(), s.size());
+    auto const d = static_cast<ripple::sha256_hasher::result_type>(hasher);
```
mathbunnyru: Maybe let's not use single-char name here?

src/util/Shasum.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <xrpl/protocol/digest.h>
+
+#include <cstring>
+#include <string>
+#include <string_view>
+
+namespace util {
+
+ripple::uint256
+sha256sum(std::string_view s)
```
mathbunnyru, line 32: Let's add a test for this file

src/web/AdminVerificationStrategy.cpp
```diff
@@ -43,12 +44,7 @@ IPAdminVerificationStrategy::isAdmin(RequestHeader const&, std::string_view ip)
 
 PasswordAdminVerificationStrategy::PasswordAdminVerificationStrategy(std::string const& password)
 {
-    ripple::sha256_hasher hasher;
-    hasher(password.data(), password.size());
-    auto const d = static_cast<ripple::sha256_hasher::result_type>(hasher);
-    ripple::uint256 sha256;
-    std::memcpy(sha256.data(), d.data(), d.size());
-    passwordSha256_ = ripple::to_string(sha256);
+    passwordSha256_ = util::sha256sumString(password);
```
mathbunnyru: Maybe do `passwordSha256_(util::toUpper(util::sha256sumString(password)))` in the initialization list and leave body empty?

src/web/ProxyIpResolver.hpp
```diff
@@ -0,0 +1,102 @@
...
+ * @brief Resolves the client's IP address, considering proxy servers.
+ *
+ * This class is designed to determine the original IP address of a client when the connection
+ * is forwarded through a proxy server. It uses a configurable list of trusted proxy IPs
+ * and proxy tokens to decide whether to trust the `Forwarded` HTTP header.
+ */
+class ProxyIpResolver {
+    std::unordered_set<std::string> proxyIps_;
+    // ripple::uint256 doesn't have hash implementation
+    std::vector<ripple::uint256> proxyTokens_;
```
mathbunnyru: I think it would be better to add our own hasher class for their type and use it here, it will make proxyIps and proxyTokens symmetrical (and ctor will be easier)

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (auto it = proxyIps_.find(connectionIp); it != proxyIps_.end()) {
+        return extractClientIp(headers).value_or(connectionIp);
+    }
+
+    if (auto it = headers.find(kPROXY_TOKEN_HEADER); it != headers.end()) {
+        auto const tokenHash = util::sha256sum(it->value());
+        if (std::ranges::find(proxyTokens_, tokenHash) != proxyTokens_.end()) {
+            return extractClientIp(headers).value_or(connectionIp);
+        }
+    }
```
mathbunnyru: we can probably use `contains` here everywhere

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+std::optional<std::string>
+ProxyIpResolver::extractClientIp(HttpHeaders const& headers)
+{
+    auto const it = headers.find(boost::beast::http::field::forwarded);
+    if (it == headers.end()) {
+        return std::nullopt;
+    }
+
+    auto const headerValue = util::toLower(it->value());
```
mathbunnyru, line 95: why do you need to lower it?

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+    static constexpr std::string_view kFOR_PREFIX = "for=";
+    auto const startPos = headerValue.find(kFOR_PREFIX);
+    if (startPos == std::string::npos) {
+        return std::nullopt;
+    }
+
```
mathbunnyru, line 103: I suggest creating a string_view here, which will start right after `kFOR_PREFIX`, it should make implementation later more clear, I think

mathbunnyru (changes requested):
Left some comments how I think we can improve code and config.

Also, I found there is [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) header, maybe we should use it?

src/app/WebHandlers.hpp
kuznetsss: DosGuard takes `std::string const&`

src/web/ProxyIpResolver.cpp
kuznetsss: Because `forwarded` header is case insensitive: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded#using_the_forwarded_header

mathbunnyru: Wow, thanks
Let's maybe add a link to this doc here, because it isn't obvious at all :(

src/app/WebHandlers.hpp
mathbunnyru: 👌

kuznetsss:
> Also, I found there is [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For) header, maybe we should use it?

`X-Forwarded-For` is not standardised while `Forwarded` is a part of http specification. If needed we could add fallback to `X-Forwarded-For` in the future.

docs/config-description.md
```diff
@@ -329,6 +329,22 @@ This document provides a list of all available Clio configuration properties in
 - **Constraints**: The minimum value is `1`. The maximum value is `4294967295`.
 - **Description**: Maximum queue size for sending subscription data to clients. This queue buffers data when a client is slow to receive it, ensuring delivery once the client is ready.
 
+### server.proxy.ips.[]
+
+- **Required**: True
+- **Type**: string
+- **Default value**: None
+- **Constraints**: None
+- **Description**: List of proxy ip addresses. When Clio receives a request from proxy it will use `Forwarded` value (if any) as client ip. When this option is used together with `server.proxy_tokens` Clio will identify proxy by ip or by token.
```
mathbunnyru: Please, search for occurrences of `server.proxy_tokens` and `server.proxy_ips` and replace with new names

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,114 @@
...
+ProxyIpResolver
+ProxyIpResolver::fromConfig(util::config::ClioConfigDefinition const& config)
+{
+    using util::config::ValueView;
+
+    std::unordered_set<std::string> ips;
+    auto const ipsFromConfig = config.getArray("server.proxy.ips");
+    for (auto it = ipsFromConfig.begin<ValueView>(); it != ipsFromConfig.end<ValueView>(); ++it) {
+        ips.insert((*it).asString());
```
mathbunnyru, line 57:
```suggestion
        ips.insert(it->asString());
```

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    std::unordered_set<std::string> tokens;
+    auto const tokensFromConfig = config.getArray("server.proxy.tokens");
+    for (auto it = tokensFromConfig.begin<ValueView>(); it != tokensFromConfig.end<ValueView>(); ++it) {
+        tokens.insert((*it).asString());
```
mathbunnyru, line 63:
```suggestion
        tokens.insert(it->asString());
```

src/web/ProxyIpResolver.cpp
kuznetsss: This custom iterator doesn't support it

src/web/ProxyIpResolver.cpp
```diff
@@ -0,0 +1,113 @@
...
+    if (startPos == std::string::npos) {
+        return std::nullopt;
+    }
+    auto value = it->value().substr(startPos + kFOR_PREFIX.size());
+
+    static constexpr char kDELIMITER = ';';
+    auto const endPos = value.find(kDELIMITER);
+    auto const ip = value.substr(0, endPos);
+    if (ip.starts_with('"')) {
+        return ip.substr(1, ip.size() - 2);
```
mathbunnyru, line 114: Don't we need to check for size here?

mathbunnyru: More general idea - if we can get data from user here, it should be validated

kuznetsss: Added size check for `ip`

mathbunnyru (dismissed):

mathbunnyru (approved):

PR #2505, opened 2025-09-03
feat: Add network id to ledger feed

### LINKED ISSUE #2350: Add `network_id` to subscriptions output
## Summary

Verify https://github.com/XRPLF/rippled/pull/5579 `network_id` is mirrored in Clio correctly.

## Motivation

While this is an `extra field` which is not considered a breaking change it is nice to be able to have this as an officially supported and documented field so Clio should add it as well.

kuznetsss:
Fixes #2350.


src/feed/impl/LedgerFeed.cpp
```diff
@@ -44,7 +44,8 @@ LedgerFeed::makeLedgerPubMessage(
     ripple::LedgerHeader const& lgrInfo,
     ripple::Fees const& fees,
     std::string const& ledgerRange,
-    std::uint32_t const txnCount
+    std::uint32_t const txnCount,
```
mathbunnyru, line 47: nit: let's remove `std::` here too

src/feed/impl/LedgerFeed.cpp
```diff
@@ -93,9 +96,10 @@ LedgerFeed::pub(
...
```
mathbunnyru, line 99: and here too

src/feed/impl/LedgerFeed.cpp
mathbunnyru:
```suggestion
    uint32_t const txnCount,
```

src/feed/impl/LedgerFeed.cpp
mathbunnyru:
```suggestion
    uint32_t const txnCount,
```

src/feed/impl/LedgerFeed.hpp
```diff
@@ -76,15 +77,17 @@ class LedgerFeed : public SingleFeedBase {
     pub(ripple::LedgerHeader const& lgrInfo,
         ripple::Fees const& fees,
         std::string const& ledgerRange,
-        std::uint32_t txnCount);
+        std::uint32_t txnCount,
```
mathbunnyru:
```suggestion
        uint32_t txnCount,
```

tests/unit/feed/SubscriptionManagerTests.cpp
```diff
@@ -271,6 +273,8 @@ TEST_F(SubscriptionManagerTest, BookChangesTest)
 TEST_F(SubscriptionManagerTest, LedgerTest)
 {
     backend_->setRange(10, 30);
+    subscriptionManagerPtr_->setNetworkID(networkID_);
```
mathbunnyru, line 276: Shouldn't we do it in ctor?

mathbunnyru (commented):
Left some comments

PeterChen13579 (dismissed):
👍

tests/unit/feed/SubscriptionManagerTests.cpp
PeterChen13579: No because we need to make sure [ETL state](https://github.com/XRPLF/clio/blob/d9faf7a83350241443add86700556b9828f71ce3/src/etl/ETLService.cpp#L139) is available first:

mathbunnyru (approved):

PR #2521, opened 2025-09-04
chore: Use mold when available without option

mathbunnyru:


cmake/Linker.cmake
```diff
@@ -1,8 +1,10 @@
...
-    message(FATAL_ERROR "Mold linker is only supported on Linux.")
-  endif ()
+if (DEFINED CMAKE_LINKER_TYPE)
+  message(STATUS "Custom linker is already set: ${CMAKE_LINKER_TYPE}")
+endif ()
+
+find_program(MOLD_PATH mold)
+
+if (MOLD_PATH AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  message(STATUS "Using Mold linker")
```
kuznetsss:
```suggestion
  message(STATUS "Using Mold linker: ${MOLD_PATH}")
```

cmake/Linker.cmake
```diff
@@ -1,8 +1,10 @@
-if (use_mold)
-  if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    message(STATUS "Using Mold linker")
-    set(CMAKE_LINKER_TYPE MOLD)
-  else ()
-    message(FATAL_ERROR "Mold linker is only supported on Linux.")
-  endif ()
+if (DEFINED CMAKE_LINKER_TYPE)
+  message(STATUS "Custom linker is already set: ${CMAKE_LINKER_TYPE}")
```
kuznetsss, line 2: Don't you need to add return here?

kuznetsss (approved):

PR #2549, opened 2025-09-08
fix: Match ledger_entry error codes with rippled

### LINKED ISSUE #2540: Match ledger_entry error codes with rippled
## Issue Description

Some ledger_entry RPC error code changed on Rippled side. Clio will need to match it: 
https://github.com/XRPLF/rippled/pull/5237

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/2540

Precondition: Ensure that the changes from 2540 are applied to xrpl.


src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -64,6 +64,8 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input const& input, Context cons
 
     if (input.index) {
         key = ripple::uint256{std::string_view(*(input.index))};
+        if (key.isZero())
```
godexsoft, line 68: Can you please tell me where this came from? is that something they changed on `rippled` side too or was it missing all along in Clio?

godexsoft (commented):
Thank you for working on this 👍 I just have one question. Otherwise it looks good and I believe it matches the `rippled` PR 👍

src/rpc/handlers/LedgerEntry.cpp
emreariyurek: It comes from [rippled side - PR5237](https://github.com/mvadari/rippled/blob/308b0bb63508e7deeb034ed1fe10e6571568f8e9/src/xrpld/rpc/handlers/LedgerEntry.cpp#L796-L799)

godexsoft:
As @emreariyurek suggested (and is evident from failing CI) this requires a newer libXRPL than what we currently have. Putting this on hold till then.

PeterChen13579:
@godexsoft  @emreariyurek Ah, so I just checked. The [PR](https://github.com/XRPLF/rippled/pull/5237) to update ledger entry error codes is not included in [libxrpl 2.6.1](https://github.com/XRPLF/rippled/commits/2.6.1/), we can move this to next release of clio 👍

src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -220,20 +222,20 @@ LedgerEntryHandler::process(LedgerEntryHandler::Input const& input, Context cons
 
     if (!ledgerObject || ledgerObject->empty()) {
         if (not input.includeDeleted)
-            return Error{Status{ClioError::RpcEntryNotFound}};
+            return Error{Status{RippledError::rpcENTRY_NOT_FOUND}};
```
PeterChen13579, line 225: Since Rippled has its own error code now, would you mind changing `ClioError::RpcEntryNotFound` to `RippledError::rpcENTRY_NOT_FOUND`, in all parts of clio and removing the Clio error code completely? Or feel free to create another issue for this. Thanks! 😃

PeterChen13579 (changes requested):
Thanks for doing this task @emreariyurek ! Just one quick comment on my side

godexsoft:
> @godexsoft @emreariyurek Ah, so I just checked. The [PR](https://github.com/XRPLF/rippled/pull/5237) to update ledger entry error codes is not included in [libxrpl 2.6.1](https://github.com/XRPLF/rippled/commits/2.6.1/), we can move this to next release of clio 👍

Yeah, that was expected since this was not in the beta nor in the rc we tried with.. next release it is 👍

src/rpc/handlers/LedgerEntry.cpp
emreariyurek: Done here: https://github.com/XRPLF/clio/pull/2661

mathbunnyru (approved):

PeterChen13579 (approved):
👍

Issue #2451, opened 2025-08-18
Better Data initialization in LogService


mathbunnyru:
1. Data has no constructor nor default values, so the values are undefined by default
2. Users might call LogService::registerLogger before init

PR #2557, opened 2025-09-08
fix: Remove global static loggers

mathbunnyru:


godexsoft (approved):
Not as many of them as i thought there would be 👍

src/data/BackendInterface.cpp
```diff
@@ -43,11 +42,6 @@
 #include <utility>
 #include <vector>
 
-// local to compilation unit loggers
-namespace {
-util::Logger gLog{"Backend"};
-}  // namespace
```
kuznetsss, line 49: Removing this will not save us from writing similar code somewhere else in the future. I think we should change implementation of `Logger`

mathbunnyru: https://github.com/XRPLF/clio/issues/2451

PR #2565, opened 2025-09-09
refactor: Use expected<void, error> instead of optional<error>

### LINKED ISSUE #2292: Use expected<void, error> instead of optional<error>
## Summary

We have a few places (at least in the new web server but maybe somewhere else) where we return result as `std::optional<error>`. It would be better to use `std::expected<void, error>` in such case.

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/2292

### DIFF
--- a/src/web/ng/Server.cpp
+++ b/src/web/ng/Server.cpp
@@ -148,8 +148,8 @@ makeConnection(
         );
         sslConnection->setTimeout(std::chrono::seconds{10});
         auto const maybeError = sslConnection->sslHandshake(yield);
-        if (maybeError.has_value())
-            return std::unexpected{fmt::format("SSL handshake error: {}", maybeError->message())};
+        if (not maybeError.has_value())
+            return std::unexpected{fmt::format("SSL handshake error: {}", maybeError.error().message())};
 
         connection = std::move(sslConnection);
     } else {
--- a/src/web/ng/Server.cpp
+++ b/src/web/ng/Server.cpp
@@ -147,9 +147,9 @@ makeConnection(
             tagDecoratorFactory
         );
         sslConnection->setTimeout(std::chrono::seconds{10});
-        auto const maybeError = sslConnection->sslHandshake(yield);
-        if (not maybeError.has_value())
-            return std::unexpected{fmt::format("SSL handshake error: {}", maybeError.error().message())};
+        auto const expectedSuccess = sslConnection->sslHandshake(yield);
+        if (not expectedSuccess.has_value())
+            return std::unexpected{fmt::format("SSL handshake error: {}", expectedSuccess.error().message())};
 
         connection = std::move(sslConnection);
     } else {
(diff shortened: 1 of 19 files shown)


tests/unit/web/ng/SubscriptionContextTests.cpp
```diff
@@ -67,7 +67,7 @@ TEST_F(NgSubscriptionContextTests, Send)
 
         EXPECT_CALL(connection_, sendShared).WillOnce([&message](std::shared_ptr<std::string> sendingMessage, auto&&) {
             EXPECT_EQ(sendingMessage, message);
-            return std::nullopt;
+            return std::expected<void, web::ng::Error>{};
```
mathbunnyru: Maybe we can add this to a lambda return type and make it `return {};` in cases like this?

src/web/ng/Server.cpp
```diff
@@ -148,8 +148,8 @@ makeConnection(
         );
         sslConnection->setTimeout(std::chrono::seconds{10});
         auto const maybeError = sslConnection->sslHandshake(yield);
-        if (maybeError.has_value())
-            return std::unexpected{fmt::format("SSL handshake error: {}", maybeError->message())};
+        if (not maybeError.has_value())
```
mathbunnyru: `maybeError` becomes a bad name, when using `std::expected<void, Error>`.
Usually, we call `maybeSomething` where `Something` - is the expected (first param) type.

I would suggest renaming it to `expectedSuccess`. If you have better naming suggestion, feel free to share them here.
And then this line would look more readable `if (not expectedSuccess.has_value())`

tests/unit/web/ng/ServerTests.cpp
```diff
@@ -436,13 +438,15 @@ TEST_F(ServerHttpTest, OnDisconnectHook)
 
         auto maybeError =
             client.connect("127.0.0.1", std::to_string(serverPort), yield, std::chrono::milliseconds{100});
-        [&]() { ASSERT_FALSE(maybeError.has_value()) << maybeError->message(); }();
+        [&]() { ASSERT_TRUE(maybeError.has_value()) << maybeError.error().message(); }();
 
-        client.send(
+        // Have to send a request here because the server does async_detect_ssl() which waits for some data to appear
+        maybeError = client.send(
```
mathbunnyru: This probably should have been a separate PR, but I'm ok keeping it here, just letting other reviewers know, there is a slightly unrelated change.

mathbunnyru (changes requested):
I like this 👍 

Left a few comments, please, take a look

tests/unit/web/ng/SubscriptionContextTests.cpp
emreariyurek: done

src/web/ng/Server.cpp
emreariyurek: `expectedSuccess` works for me as well, updated :)

tests/unit/web/ng/ServerTests.cpp
emreariyurek: Since the return type was not checked here, as it is in the other functions so I added this for the sake of consistency

mathbunnyru: Sure, I understand the reason

mathbunnyru (approved):
LGTM, thank you! 👍

godexsoft (approved):

PR #2594, opened 2025-09-11
fix: Workaround large number validation in RPC

godexsoft:
Fixes #2586

PR #2608, opened 2025-09-12
fix: Workaround large number validation and parsing

### LINKED ISSUE #2586: Internal error with large limit values (Version: mainnet)
## Issue Description

If I provide a super large `limit` value to Clio in an RPC that accepts a `limit` value, I get an `internal` error. With slightly smaller values, it's a bit more flaky - sometimes I get a ton of results back, and sometimes I get an `internal` error.

I assume this applies to other RPCs that take numbers, but I didn't explore too much.

## Steps to Reproduce

https://xrpl.org/resources/dev-tools/websocket-api-tool?server=wss%3A%2F%2Fs1.ripple.com%2F&req=%7B%20%20%22ledger_hash%22%3A%20%22842B57C1CC0613299A686D3E9F310EC0422C84D3911E5056389AA7E5808A93C8%22%2C%0A%20%20%22command%22%3A%20%22ledger_data%22%2C%0A%20%20%22limit%22%3A%2099999999999%2C%0A%20%20%22binary%22%3A%20true%0A%7D

## Expected Result

Either a clean error or actual responses

## Actual Result

```
{
  "error": "internal",
  "error_code": 73,
  "error_message": "Internal error.",
  "status": "error",
  "type": "response",
  "request": {
    "ledger_hash": "842B57C1CC0613299A686D3E9F310EC0422C84D3911E5056389AA7E5808A93C8",
    "command": "ledger_data",
    "limit": 99999999999,
    "binary": true
  },
  "warnings": [
    {
      "id": 2001,
    }
  ]
}
```

## Environment

Whatever is currently running on mainnet

godexsoft:
Fixes #2586 

Alternative approach to #2594

### DIFF
--- a/src/rpc/common/ValidationHelpers.hpp
+++ b/src/rpc/common/ValidationHelpers.hpp
@@ -34,7 +34,7 @@ namespace impl {
 
 template <std::unsigned_integral Expected>
 static void
-clampUnsigned(boost::json::value& value)
+clampAs(boost::json::value& value)
 {
     if (value.is_uint64()) {
         auto const valueUint = value.as_uint64();
@@ -49,7 +49,7 @@ clampUnsigned(boost::json::value& value)
 
 template <std::signed_integral Expected>
 static void
-clampSigned(boost::json::value& value)
+clampAs(boost::json::value& value)
 {
     if (value.is_uint64()) {
         auto const valueUint = value.as_uint64();
@@ -125,13 +125,8 @@ checkTypeAndClamp(boost::json::value& value)
     if (not checkType<Expected>(value))
         return false;  // fails basic type check
 
-    if constexpr (std::is_integral_v<Expected> and not std::is_same_v<Expected, bool>) {
-        if constexpr (std::is_unsigned_v<Expected>) {
-            impl::clampUnsigned<Expected>(value);
-        } else {
-            impl::clampSigned<Expected>(value);
-        }
-    }
+    if constexpr (std::is_integral_v<Expected> and not std::is_same_v<Expected, bool>)
+        impl::clampAs<Expected>(value);
 
     return true;
 }
--- a/src/rpc/common/ValidationHelpers.hpp
+++ b/src/rpc/common/ValidationHelpers.hpp
@@ -113,7 +113,7 @@ checkTypeAndClamp(boost::json::value& value)
                 if (v > static_cast<int64_t>(std::numeric_limits<Expected>::max())) {
                     value = std::numeric_limits<Expected>::max();
                 } else if (v < static_cast<int64_t>(std::numeric_limits<Expected>::min())) {
-                    value = std::numeric_limits<Expected>::max();
+                    value = std::numeric_limits<Expected>::min();
                 }
             } else {
                 hasError = true;
--- a/src/rpc/common/ValidationHelpers.hpp
+++ b/src/rpc/common/ValidationHelpers.hpp
@@ -33,7 +33,7 @@ namespace rpc::validation {
 namespace impl {
 
 template <std::unsigned_integral Expected>
-static void
+void
 clampAs(boost::json::value& value)
 {
     if (value.is_uint64()) {
@@ -48,7 +48,7 @@ clampAs(boost::json::value& value)
 }
 
 template <std::signed_integral Expected>
-static void
+void
 clampAs(boost::json::value& value)
 {
     if (value.is_uint64()) {
@@ -75,7 +75,7 @@ clampAs(boost::json::value& value)
  * @return true if convertible; false otherwise
  */
 template <typename Expected>
-[[nodiscard]] static bool
+[[nodiscard]] bool
 checkType(boost::json::value const& value)
 {
     auto hasError = false;
@@ -119,7 +119,7 @@ checkType(boost::json::value const& value)
  * @return true if convertible; false otherwise
  */
 template <typename Expected>
-[[nodiscard]] static bool
+[[nodiscard]] bool
 checkTypeAndClamp(boost::json::value& value)
 {
     if (not checkType<Expected>(value))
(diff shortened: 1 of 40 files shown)


src/util/JsonUtils.hpp
```diff
@@ -86,4 +88,26 @@ removeSecret(boost::json::object const& object)
...
+ * @note This conversion can possibly cause wrapping around or UB. Use with caution.
+ *
+ * @tparam Type The type to cast to
+ * @param value The JSON value to cast
+ * @return Value casted to the requested type
+ * @throws logic_error if the underlying number is neither int64 nor uint64
+ */
+template <std::integral Type>
+Type
+integralValueFrom(boost::json::value const& value)
```
mathbunnyru: Maybe call it `integralValueAs`?

src/rpc/common/ValidationHelpers.hpp
```diff
@@ -68,4 +70,58 @@ checkType(boost::json::value const& value)
...
+ * order to avoid Min, Max and other validators throw "not exact" error from Boost.Json library if the value does not
+ * fit in the specified type.
+ *
+ * @tparam Expected The expected type that value should be convertible to
+ * @param value The json value to check the type of
+ * @return true if convertible; false otherwise
+ */
+template <typename Expected>
+[[nodiscard]] static bool
+checkTypeAndClamp(boost::json::value& value)
```
mathbunnyru, line 123: Let's put this function (maybe without `checkType` call) to it's own file and add a test for this function

src/rpc/common/ValidationHelpers.hpp
```diff
@@ -68,4 +70,58 @@ checkType(boost::json::value const& value)
...
+{
+    auto hasError = false;
+
+    if (not checkType<Expected>(value))
+        return false;  // fails basic type check
+
+    if constexpr (std::is_integral_v<Expected> and not std::is_same_v<Expected, bool>) {
+        if constexpr (std::is_unsigned_v<Expected>) {
```
mathbunnyru: It would be more readable, if you had 2 functions in impl namespace, one for unsigned, and one for signed

mathbunnyru (changes requested):

src/rpc/common/ValidationHelpers.hpp
godexsoft: Don't think it needs/deserves its own file but will see what tests i can add 👍

src/rpc/common/ValidationHelpers.hpp
```diff
@@ -23,10 +23,49 @@
...
+#include <limits>
 #include <string>
+#include <type_traits>
 
 namespace rpc::validation {
+namespace impl {
+
+template <std::unsigned_integral Expected>
+static void
+clampUnsigned(boost::json::value& value)
```
kuznetsss: Maybe `clampSigned` and `clampUnsigned` could be named both as `clamp`? I think compiler should be able to resolve it.

kuznetsss (dismissed):

src/rpc/common/ValidationHelpers.hpp
godexsoft: Yeah you are right. Good call 👍

src/rpc/common/ValidationHelpers.hpp
```diff
@@ -23,10 +23,49 @@
...
 #include <cstdint>
+#include <limits>
 #include <string>
+#include <type_traits>
 
 namespace rpc::validation {
+namespace impl {
+
+template <std::unsigned_integral Expected>
+static void
```
mathbunnyru: static is not needed here (and below)

src/rpc/common/ValidationHelpers.hpp
```diff
@@ -23,10 +23,49 @@
...
 #include <boost/json/value.hpp>
 
+#include <concepts>
 #include <cstdint>
+#include <limits>
 #include <string>
+#include <type_traits>
 
 namespace rpc::validation {
+namespace impl {
```
mathbunnyru, line 33: nit: empty line between

src/rpc/common/ValidationHelpers.hpp
mathbunnyru: Ok

src/rpc/common/Validators.hpp
```diff
@@ -148,13 +148,13 @@ struct Type final {
      * @return `RippledError::rpcINVALID_PARAMS` if validation failed; otherwise no error is returned
      */
     [[nodiscard]] MaybeError
-    verify(boost::json::value const& value, std::string_view key) const
+    verify(boost::json::value& value, std::string_view key) const
```
mathbunnyru, line 153: Don't we want to name it `verifyAndClamp` or sth like that?

godexsoft: No, this is supposed to be used with only validators (non-modifying) but we make the exception for clamping ints because we had to. Let's not introduce unnecessary changes

src/rpc/common/ValidationHelpers.hpp
godexsoft: Why? while we have a few examples with a space line between, most cases are like this without the space. Frankly i think this is one place where the empty line does not make it better

mathbunnyru: Because you have some code after `impl` closes, and it doesn't look symmetrical

mathbunnyru: Not a big deal at all, you can leave it as it is right now :)

src/rpc/common/Validators.hpp
mathbunnyru: Maybe let's add a comment here why we (possibly) modify value?

src/rpc/common/ValidationHelpers.hpp
godexsoft: Yeah i agree, result of copypasta 👍

src/rpc/common/ValidationHelpers.hpp
godexsoft: Nah i like to fix it, so now all examples with extra line are gone 😝

mathbunnyru (dismissed):

mathbunnyru (approved):

PR #2614, opened 2025-09-12
fix: Keep spdlog loggers valid between tests

### LINKED ISSUE #2451: Better Data initialization in LogService

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2451

### DIFF
--- a/src/util/log/Logger.cpp
+++ b/src/util/log/Logger.cpp
@@ -367,10 +367,7 @@ LogService::init(config::ClioConfigDefinition const& config)
 void
 LogService::shutdown()
 {
-    if (!initialized_) {
-        throw std::logic_error("LogService not initialized");
-    }
-    if (isAsync_) {
+    if (initialized_ && isAsync_) {
         // We run in async mode in production, so we need to make sure all logs are flushed before shutting down
         spdlog::shutdown();
     }
--- a/src/util/log/Logger.cpp
+++ b/src/util/log/Logger.cpp
@@ -266,6 +266,9 @@ LogServiceState::init(bool isAsync, Severity defaultSeverity, std::vector<spdlog
 void
 LogServiceState::reset()
 {
+    if (!initialized_) {
+        throw std::logic_error("LogService is not initialized");
+    }
     isAsync_ = true;
     defaultSeverity_ = Severity::NFO;
     sinks_.clear();
@@ -276,7 +279,7 @@ std::shared_ptr<spdlog::logger>
 LogServiceState::registerLogger(std::string const& channel, std::optional<Severity> severity)
 {
     if (not initialized_) {
-        throw std::logic_error("LogService not initialized");
+        throw std::logic_error("LogService is not initialized");
     }
 
     std::shared_ptr<spdlog::logger> existingLogger = spdlog::get(channel);
--- a/src/util/log/Logger.cpp
+++ b/src/util/log/Logger.cpp
@@ -85,21 +85,6 @@ toSpdlogLevel(Severity sev)
     return spdlog::level::info;
 }
 
-std::string_view
-toString(Severity sev)
-{
-    static constexpr std::array<std::string_view, 6> kLABELS = {
-        "TRC",
-        "DBG",
-        "NFO",
-        "WRN",
-        "ERR",
-        "FTL",
-    };
-
-    return kLABELS.at(static_cast<int>(sev));
-}
-
 }  // namespace
 
 /**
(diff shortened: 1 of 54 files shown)


src/util/log/Logger.cpp
```diff
@@ -238,46 +241,77 @@ getMinSeverity(config::ClioConfigDefinition const& config, Severity defaultSever
...
+    sinks_ = sinks;
+    initialized_ = true;
+
+    spdlog::apply_all([](std::shared_ptr<spdlog::logger> logger) {
+        logger->set_level(toSpdlogLevel(defaultSeverity_));
+    });
+
+    if (isAsync) {
+        // Create a thread pool with 8192 queue size and 1 thread
+        spdlog::init_thread_pool(8192, 1);
```
godexsoft: nit: Instead of comments you could have two constants here

src/util/log/Logger.cpp
```diff
@@ -56,7 +56,10 @@
 
 namespace util {
 
-LogService::Data LogService::data{};
+bool LogServiceState::isAsync_{true};
```
godexsoft, line 59: nit: I don't really mind but the naming is inconsistent with `initialized_`

src/util/log/Logger.cpp
```diff
@@ -238,46 +241,77 @@ getMinSeverity(config::ClioConfigDefinition const& config, Severity defaultSever
...
+    if (isAsync) {
+        // Create a thread pool with 8192 queue size and 1 thread
+        spdlog::init_thread_pool(8192, 1);
+    }
+}
+
+void
+LogServiceState::reset()
+{
+    if (!initialized_) {
```
godexsoft: nit: You now have a function for this

src/util/log/Logger.cpp
```diff
@@ -307,20 +357,23 @@ LogService::init(config::ClioConfigDefinition const& config)
...
     return {};
 }
 
 void
 LogService::shutdown()
 {
-    spdlog::shutdown();
+    if (initialized_ && isAsync_) {
+        // We run in async mode in production, so we need to make sure all logs are flushed before shutting down
+        spdlog::shutdown();
```
godexsoft, line 382: nit: Does it hurt to do that regardless?

src/util/log/Logger.hpp
```diff
@@ -229,26 +232,68 @@ class Logger final {
...
+     * @brief Register a new logger for the specified channel.
+     *
+     * Creates and registers a new spdlog logger instance for the given channel
+     * with the specified or default severity level.
+     *
+     * @param channel The name of the logging channel
+     * @param severity Optional severity level override; uses default if not specified
+     * @return Shared pointer to the registered spdlog logger
+     */
     static std::shared_ptr<spdlog::logger>
```
godexsoft, line 293: Do we want this to be nodiscard or not?

tests/common/util/LoggerFixtures.hpp
```diff
@@ -19,37 +19,38 @@
...
+
+    /**
+     * @brief Sets up spdlog loggers for each channel. Should be called once before using any loggers.
+     * Simulates the `util::LogService::init(config)` call
+     */
+    static void
+    init();
 
 protected:
     std::string
```
godexsoft, line 48: nit: This probably should be nodiscard

godexsoft (changes requested):
Leaving some nits. Overall looks good 👍 
Not too sure about having `public virtual` everywhere - most places don't need it. But it's not a blocker.

src/util/log/Logger.cpp
mathbunnyru: we can't do it, because all loggers will be destroyed

src/util/log/Logger.cpp
mathbunnyru: agreed not to rename

src/util/log/Logger.cpp
mathbunnyru: move initialized to state, and used it, thanks 👍

src/util/log/Logger.hpp
mathbunnyru: mostly, it's not needed, so won't add here

tests/common/util/LoggerFixtures.hpp
mathbunnyru: added

godexsoft (approved):
:rage2:

PR #2618, opened 2025-09-14
fix: Include TestGlobals.hpp is using <> instead of quotations

### LINKED ISSUE #1642: #Include TestGlobals.hpp is using "<>" instead of quotations
## Issue Description
The correct include structure is `#Include "TestGlobals.hpp"` , but Git prehook is converting it into `#Include <TestGlobals.hpp>` for all files

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/1642


tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -35,11 +35,11 @@
 #include "util/Random.hpp"
 #include "util/Spawn.hpp"
 #include "util/StringUtils.hpp"
+#include "util/TestGlobals.hpp"
```
mathbunnyru: I think putting code in `tests/integration/util/` creates confusion, because these 3 lines look similar, but the `Spawn.hpp` is from another directory.

Maybe we should fix [fix-local-includes.sh](https://github.com/XRPLF/clio/blob/develop/pre-commit-hooks/fix-local-includes.sh) instead?

godexsoft: Fixing the script is not gonna be enough as each time you save in IDE it will update it back as far as i remember.

mathbunnyru: > Fixing the script is not gonna be enough as each time you save in IDE it will update it back as far as i remember.

@godexsoft I checked in both "VS Code" and "Zed" and they both didn't update it back

godexsoft: You are right, it was the pre-commit script that updated it back. Definitely need to fix the script..

emreariyurek: This happens [here](https://github.com/XRPLF/clio/blob/c3c8b2d79627f0db4d9440a7fef007e91456d4d1/tests/integration/migration/cassandra/CassandraMigrationManagerTests.cpp#L35) as well @mathbunnyru. Btw, I’ve updated the script and please take a look

pre-commit-hooks/fix-local-includes.sh
```diff
@@ -16,6 +16,16 @@ else
...
 
+# Get a list of all header filenames in src/ and tests/ (recursively, unique)
+all_local_headers=$(find ./src ./tests -type f \( -name '*.hpp' -o -name '*.h' \) -exec basename {} \; | sort -u)
+
+# Generate a sed script for batch replacements
+sed_script=$(mktemp)
+for header in $all_local_headers; do
+    escaped_header=$(printf '%s\n' "$header" | sed 's/[\/&]/\\&/g')
+    echo "s|#include[[:space:]]*[\"<]$escaped_header[\">]|#include \"$header\"|g" >> "$sed_script"
+done
```
mathbunnyru, line 27: Is it possible to make replacements directly, instead of generating code to make replacements?

pre-commit-hooks/fix-local-includes.sh
```diff
@@ -16,6 +16,16 @@ else
     main_src_dirs=$(find ./src -maxdepth 1 -type d -exec basename {} \; | paste -sd '|' | sed 's/|/\\|/g')
 fi
 
+# Get a list of all header filenames in src/ and tests/ (recursively, unique)
+all_local_headers=$(find ./src ./tests -type f \( -name '*.hpp' -o -name '*.h' \) -exec basename {} \; | sort -u)
```
mathbunnyru, line 20: We don't have `*.h` files

mathbunnyru, line 20: Do we need 2 implementations?
I mean using both `main_src_dirs` and `all_local_headers`?

mathbunnyru (changes requested):

tests/integration/data/cassandra/BackendTests.cpp
mathbunnyru: Thanks!

PR #2637, opened 2025-09-17
ci: Make separate download/upload options for ccache

mathbunnyru:


.github/workflows/build.yml
```diff
@@ -76,7 +78,8 @@ jobs:
       container: '{ "image": "ghcr.io/xrplf/clio-ci:384e79cd32f5f6c0ab9be3a1122ead41c5a7e67d" }'
       conan_profile: gcc
       build_type: Debug
-      disable_cache: false
+      download_ccache: true
```
kuznetsss, line 81: Maybe we should make it to wait for the cache to be ready? Then it will be able to build everything very quick. Otherwise cache will likely not be ready when this workflow start. Same for package

mathbunnyru: In PRs it won't be useful, because jobs are waiting for develop (not for each other).
And we actually want to start as soon as possible (all jobs as parallel as possible) to make dev experience better.

We could save a bit of resources in develop, but this would require duplicating jobs, with `needs` and without `needs` - it looks bad in a workflow file, and will look bad in status checks, as we will always have more skipped jobs.

So, overall, I don't think it's worth doing

kuznetsss (approved):

PR #2648, opened 2025-09-25
chore: Replace manual forward declarations with spdlog/fwd.h

emreariyurek:


mathbunnyru:
I did it like this on purpose - I wanted the spdlog target (and includes) be private, so no one can easily use them.
That's why your builds are failing 🙂

emreariyurek:
> I did it like this on purpose - I wanted the spdlog target (and includes) be private, so no one can easily use them. That's why your builds are failing 🙂

Okay, thanks for letting me know, so closing this one :)

PR #2661, opened 2025-10-04
fix: Remove RpcEntryNotFound from ClioError

emreariyurek:
Since rippled has its own error codes, all occurrences have been replaced except for [PR2549](https://github.com/XRPLF/clio/pull/2549), and the entry has also been removed entirely from Clio’s error enum. This modification follows the feedback provided in this comment: https://github.com/XRPLF/clio/pull/2549#discussion_r2394840688

Pending compatible libxrpl — temporarily on hold


src/rpc/Errors.hpp
```diff
@@ -43,7 +43,6 @@ enum class ClioError {
     RpcFieldNotFoundTransaction = 5006,
     RpcMalformedOracleDocumentId = 5007,
     RpcMalformedAuthorizedCredentials = 5008,
-    RpcEntryNotFound = 5009,
```
mathbunnyru, line 46: I would keep this line, but commented out and with a comment why.

Otherwise, in the future we will definitely forget this code was used, and reusing them is not great (because people might rely on the old code too much)

mathbunnyru (dismissed):

kuznetsss:
Looks good, thanks! 
But why don't we have to change any tests?  I think it would be good to add some tests (in this PR or create an issue and do it in a separate PR).

mathbunnyru (approved):

godexsoft (approved):
LGTM :rage1:

PR #2667, opened 2025-10-06
style: Rename workflows to use dash and show reusable

mathbunnyru:
This matches the style used in `rippled` and makes it more unified:
- using dash in workflow names
- show reusable workflows by adding `reusable-` prefix


.github/workflows/update-docker-ci.yml
```diff
@@ -94,7 +94,7 @@ jobs:
 
       - name: Get changed files
         id: changed-files
-        uses: tj-actions/changed-files@24d32ffd492484c1d75e0c0b894501ddb9d30d62 # v47.0.0
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
```
mathbunnyru, line 97: node24 is not supported on our arm runners, had to downgrade here

.github/workflows/reusable-upload-coverage-report.yml
```diff
@@ -1,7 +1,6 @@
 name: Upload report
 
 on:
-  workflow_dispatch:
```
mathbunnyru, line 4: this is not meant to be run manually, so removed it from here

godexsoft (approved):

PR #2680, opened 2025-10-07
feat: Support account_mptoken_issuances and account_mptokens

yinyiqian1:
`account_mptoken_issuances`: get all the `MPTokenIssuance` created by the specified account.
`account_mptokens`: get all the `MPToken` held by the specified account.


src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,247 @@
...
+    ripple::AccountID const& account
+)
+{
+    MPTokenIssuanceResponse issuance;
+
+    issuance.issuer = ripple::to_string(account);
+    issuance.sequence = sle.getFieldU32(ripple::sfSequence);
+    auto const flags = sle.getFieldU32(ripple::sfFlags);
+
+    if ((flags & ripple::lsfMPTLocked) != 0u)
```
godexsoft: nit: Could use a lambda to make this line read better: `if (hasFlag(ripple::lsfMPTCanLock))` or similar

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,247 @@
...
+{
+    using boost::json::value_from;
+
+    auto obj = boost::json::object{
+        {JS(account), output.account},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(ledger_index), output.ledgerIndex},
+        {JS(validated), output.validated},
+        {JS(limit), output.limit},
+        {"mpt_issuances", value_from(output.issuances)},
```
godexsoft, line 191: nit: There is no JS string for this, right?

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,247 @@
...
+    boost::json::value& jv,
+    AccountMPTokenIssuancesHandler::MPTokenIssuanceResponse const& issuance
+)
+{
+    auto obj = boost::json::object{
+        {JS(issuer), issuance.issuer},
+        {JS(sequence), issuance.sequence},
+    };
+
+    if (issuance.transferFee)
```
godexsoft: nit: For clarity we usually use `.has_value()` on optionals

src/rpc/handlers/AccountMPTokenIssuances.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+namespace rpc {
+
+/**
+ * @brief The account_mptoken_issuances method returns information about all MPTokenIssuance objects the account has
+ * created.
+ */
+class AccountMPTokenIssuancesHandler {
+    // dependencies
+    std::shared_ptr<BackendInterface> const sharedPtrBackend_;
```
godexsoft: Not a big deal as you likely copied this from `AccountChannels` but let's avoid unnecessary `const` members 👍

src/rpc/handlers/AccountMPTokenIssuances.hpp
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    using Result = HandlerReturnType<Output>;
+
+    /**
+     * @brief Construct a new AccountMPTokenIssuancesHandler object.
+     *
+     * @param sharedPtrBackend The backend to use.
+     */
+    AccountMPTokenIssuancesHandler(std::shared_ptr<BackendInterface> const& sharedPtrBackend)
```
godexsoft: We usually pass `shared_ptr` by value and `std::move` it

src/rpc/handlers/AccountMPTokenIssuances.hpp
```diff
@@ -0,0 +1,199 @@
...
+                 validation::CustomValidators::accountValidator, Status(RippledError::rpcACT_MALFORMED)
+             }},
+            {JS(ledger_hash), validation::CustomValidators::uint256HexStringValidator},
+            {JS(limit),
+             validation::Type<uint32_t>{},
+             validation::Min(1u),
+             modifiers::Clamp<int32_t>{kLIMIT_MIN, kLIMIT_MAX}},
+            {JS(ledger_index), validation::CustomValidators::ledgerIndexValidator},
+            {JS(marker), validation::CustomValidators::accountMarkerValidator},
+            {JS(ledger), check::Deprecated{}},
```
godexsoft, line 143: Is 'ledger' really deprecated in this brand new API?

src/rpc/handlers/AccountMPTokenIssuances.hpp
```diff
@@ -0,0 +1,199 @@
...
+     */
+    Result
+    process(Input const& input, Context const& ctx) const;
+
+private:
+    /**
+     * @brief Populates an MPTokenIssuanceResponse from a ledger entry and adds it to the list.
+     */
+    static void
+    addMPTokenIssuance(
```
godexsoft, line 161: Can make it a simple function in an anonymous namespace in the cpp file - it's private anyway so we are not going to be testing it directly

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,185 @@
...
+    auto const flags = sle.getFieldU32(ripple::sfFlags);
+
+    token.account = ripple::to_string(sle.getAccountID(ripple::sfAccount));
+    token.MPTokenIssuanceID = ripple::strHex(sle.getFieldH192(ripple::sfMPTokenIssuanceID));
+    token.MPTAmount = sle.getFieldU64(ripple::sfMPTAmount);
+
+    if (sle.isFieldPresent(ripple::sfLockedAmount))
+        token.lockedAmount = sle.getFieldU64(ripple::sfLockedAmount);
+
+    if ((flags & ripple::lsfMPTLocked) != 0u)
```
godexsoft: If this is a recurring pattern, maybe a helper function would be nice to have

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,185 @@
...
+    if (!expectedLgrInfo.has_value())
+        return Error{expectedLgrInfo.error()};
+
+    auto const& lgrInfo = expectedLgrInfo.value();
+    auto const accountID = accountFromStringStrict(input.account);
+    auto const accountLedgerObject =
+        sharedPtrBackend_->fetchLedgerObject(ripple::keylet::account(*accountID).key, lgrInfo.seq, ctx.yield);
+
+    if (not accountLedgerObject)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
```
godexsoft: I think the second parameter is not needed here - please try

src/rpc/handlers/AccountMPTokens.hpp
```diff
@@ -0,0 +1,182 @@
...
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace rpc {
+
+/**
+ * @brief The account_mptokens method returns information about the MPTokens the account currently holds.
+ *
```
godexsoft: Extra empty line can be removed

src/rpc/handlers/AccountMPTokens.hpp
```diff
@@ -0,0 +1,182 @@
...
+ */
+class AccountMPTokensHandler {
+    // dependencies
+    std::shared_ptr<BackendInterface> const sharedPtrBackend_;
```
godexsoft: Let's remove `const` here as well

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,185 @@
...
+void
+tag_invoke(boost::json::value_from_tag, boost::json::value& jv, AccountMPTokensHandler::MPTokenResponse const& mptoken)
+{
+    auto obj = boost::json::object{
+        {JS(account), mptoken.account},
+        {JS(mpt_issuance_id), mptoken.MPTokenIssuanceID},
+        {JS(mpt_amount), mptoken.MPTAmount},
+    };
+
+    if (mptoken.lockedAmount)
```
godexsoft: nit: Same story about `has_value()`

src/rpc/handlers/AccountMPTokens.hpp
```diff
@@ -0,0 +1,182 @@
...
+                 validation::CustomValidators::accountValidator, Status(RippledError::rpcACT_MALFORMED)
+             }},
+            {JS(ledger_hash), validation::CustomValidators::uint256HexStringValidator},
+            {JS(limit),
+             validation::Type<uint32_t>{},
+             validation::Min(1u),
+             modifiers::Clamp<int32_t>{kLIMIT_MIN, kLIMIT_MAX}},
+            {JS(ledger_index), validation::CustomValidators::ledgerIndexValidator},
+            {JS(marker), validation::CustomValidators::accountMarkerValidator},
+            {JS(ledger), check::Deprecated{}},
```
godexsoft, line 129: Again a deprecated field in a new API. Please double check this is needed 🔢

src/rpc/handlers/AccountMPTokens.hpp
```diff
@@ -0,0 +1,182 @@
...
+     */
+    Result
+    process(Input const& input, Context const& ctx) const;
+
+private:
+    /**
+     * @brief Populates an MPTokenResponse from a ledger entry and adds it to the list.
+     */
+    static void
+    addMPToken(std::vector<MPTokenResponse>& mpts, ripple::SLE const& sle);
```
godexsoft, line 147: nit: Same suggestion about anonymous namespace (or keep it as is if you prefer)

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
```diff
@@ -0,0 +1,843 @@
...
+constexpr uint64_t kISSUANCE1_MAX_AMOUNT = 10000;
+constexpr uint64_t kISSUANCE1_OUTSTANDING_AMOUNT = 5000;
+constexpr uint8_t kISSUANCE1_ASSET_SCALE = 2;
+
+// unique values for issuance2
+constexpr uint64_t kISSUANCE2_MAX_AMOUNT = 20000;
+constexpr uint64_t kISSUANCE2_OUTSTANDING_AMOUNT = 800;
+constexpr uint64_t kISSUANCE2_LOCKED_AMOUNT = 100;
+constexpr uint16_t kISSUANCE2_TRANSFER_FEE = 5;
+std::string const kISSUANCE2_METADATA = "test-meta";
```
godexsoft: Let's only keep `constexpr auto` for string constants

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
```diff
@@ -0,0 +1,843 @@
...
+        EXPECT_EQ(err.at("error_message").as_string(), "ledgerNotFound");
+    });
+}
+
+// ledger not found via string index
+TEST_F(RPCAccountMPTokenIssuancesHandlerTest, NonExistLedgerViaLedgerStringIndex)
+{
+    // mock fetchLedgerBySequence return empty
+    ON_CALL(*backend_, fetchLedgerBySequence).WillByDefault(Return(std::optional<ripple::LedgerHeader>{}));
+    EXPECT_CALL(*backend_, fetchLedgerBySequence).Times(1);
```
godexsoft: Times(1) is the default - no need to specify.
Also, please don't use ON_CALL with EXPECT_CALL - use just EXPECT_CALL. See how to use it in other tests (WillOnce, WillRedpeatedly)

tests/unit/rpc/handlers/LedgerEntryTests.cpp
```diff
@@ -3744,7 +3744,6 @@ TEST_F(RPCLedgerEntryTest, SyntheticMPTIssuanceID)
             "Issuer": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
             "LedgerEntryType": "MPTokenIssuance",
             "MPTokenMetadata": "6D65746164617461",
-            "MaximumAmount": "0",
```
godexsoft, line 3759: What happened to this field, is there a document somewhere that describes why it's removed?

godexsoft (changes requested):
Left several suggestions and questions. Thanks 👍

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: Yes, this is clio only. We do not have `mpt_issuances` in rippled.

tests/unit/rpc/handlers/LedgerEntryTests.cpp
yinyiqian1: Because I made some changes to the `TestObject.cpp::createMptIssuanceObject`, this field was modified to `std::optional<std::uint64_t>`. Previously it was set to 0 anyway. But in rippled, it is `soeOPTIONAL`.

src/rpc/handlers/AccountMPTokenIssuances.hpp
yinyiqian1: I think so. Like `account_info`, `account_lines`, `account_offers`, the users are encouraged to use ledger_index/ledger_hash. If they use `ledger`, they will get warning. And we keep the consistency with the existing apis.

src/rpc/handlers/AccountMPTokenIssuances.hpp
yinyiqian1: Can we keep it as it is now? I need to use `rpc::AccountMPTokenIssuancesHandler::MPTokenIssuanceResponse` if it's in an anonymous namespace. The private static pattern is used by other handlers as well. We can refactor them together in a separate PR.

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: modified to
```
auto const setFlag = [&](std::optional<bool>& field, std::uint32_t mask) {
        if ((flags & mask) != 0u)
            field = true;
    };
```

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: I also added a lambda:
```
auto const setIfPresent = [&](boost::json::string_view field, auto const& value) {
        if (value.has_value()) {
            obj[field] = *value;
        }
    };
```

src/rpc/handlers/AccountMPTokens.cpp
yinyiqian1: Added the lambda with `has_value()`
```
auto const setIfPresent = [&](boost::json::string_view field, auto const& value) {
        if (value.has_value()) {
            obj[field] = *value;
        }
    };
```

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
yinyiqian1: Updated these places.

godexsoft (dismissed):
LGTM 👍

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <optional>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace rpc {
+
+void
+AccountMPTokenIssuancesHandler::addMPTokenIssuance(
+    std::vector<MPTokenIssuanceResponse>& issuances,
```
mathbunnyru, line 54: Usually, we try not no modify input params.
Maybe we can call this function something like `getMPTokenIssuanceResponse`, and push back on the call-side?

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,237 @@
...
+        issuance.outstandingAmount = sle.getFieldU64(ripple::sfOutstandingAmount);
+
+    if (sle.isFieldPresent(ripple::sfLockedAmount))
+        issuance.lockedAmount = sle.getFieldU64(ripple::sfLockedAmount);
+
+    if (sle.isFieldPresent(ripple::sfMPTokenMetadata))
+        issuance.mptokenMetadata = ripple::strHex(sle.getFieldVL(ripple::sfMPTokenMetadata));
+
+    if (sle.isFieldPresent(ripple::sfDomainID))
+        issuance.domainID = ripple::strHex(sle.getFieldH256(ripple::sfDomainID));
```
mathbunnyru, line 97: Maybe we can have a lambda for all these fields as well?
Let's call it `setField`, it will be templated.

I'm not convinced if it would be better and if it's a good idea

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    if (!expectedLgrInfo.has_value())
+        return Error{expectedLgrInfo.error()};
+
+    auto const& lgrInfo = expectedLgrInfo.value();
+    auto const accountID = accountFromStringStrict(input.account);
+    auto const accountLedgerObject =
+        sharedPtrBackend_->fetchLedgerObject(ripple::keylet::account(*accountID).key, lgrInfo.seq, ctx.yield);
+
+    if (not accountLedgerObject)
```
mathbunnyru:
```suggestion
    if (not accountLedgerObject.has_value())
```

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    auto const expectedNext = traverseOwnedNodes(
+        *sharedPtrBackend_, *accountID, lgrInfo.seq, input.limit, input.marker, ctx.yield, addToResponse
+    );
+
+    if (!expectedNext.has_value())
+        return Error{expectedNext.error()};
+
+    auto const nextMarker = expectedNext.value();
```
mathbunnyru:
```suggestion
    auto const& nextMarker = expectedNext.value();
```

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,191 @@
...
+    if (nextMarker.isNonZero())
+        response.marker = nextMarker.toString();
+
+    return response;
+}
+
+AccountMPTokensHandler::Input
+tag_invoke(boost::json::value_to_tag<AccountMPTokensHandler::Input>, boost::json::value const& jv)
+{
+    auto input = AccountMPTokensHandler::Input{};
```
mathbunnyru:
```suggestion
    AccountMPTokensHandler::Input input{};
```

src/rpc/handlers/AccountMPTokenIssuances.hpp
```diff
@@ -0,0 +1,196 @@
...
+    struct MPTokenIssuanceResponse {
+        std::string issuer;
+        uint32_t sequence{};
+
+        std::optional<uint16_t> transferFee{};
+        std::optional<uint8_t> assetScale{};
+
+        std::optional<std::uint64_t> maximumAmount;
+        std::optional<std::uint64_t> outstandingAmount;
+        std::optional<std::uint64_t> lockedAmount;
```
mathbunnyru, line 72: Why is this field optional and not just `std::uint64_t`?

tests/common/util/TestObject.cpp
```diff
@@ -1442,38 +1442,69 @@ createLptCurrency(std::string_view assetCurrency, std::string_view asset2Currenc
...
+    std::string_view accountId,
+    std::uint32_t seq,
+    std::optional<std::string_view> metadata,
+    std::uint32_t flags,
+    std::uint64_t outstandingAmount,
+    std::optional<std::uint16_t> transferFee,
+    std::optional<std::uint8_t> assetScale,
+    std::optional<std::uint64_t> maxAmount,
+    std::optional<std::uint64_t> lockedAmount,
+    std::optional<std::string_view> domainId
```
mathbunnyru, line 1455: Is it true that these fields are optional or non-optional independently?

Because if it's not, it would be better to have a design, where we have a struct with (some of) these fields, and have `optional<STRUCT>`

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    if (!expectedLgrInfo.has_value())
+        return Error{expectedLgrInfo.error()};
+
+    auto const& lgrInfo = expectedLgrInfo.value();
+    auto const accountID = accountFromStringStrict(input.account);
+    auto const accountLedgerObject =
+        sharedPtrBackend_->fetchLedgerObject(ripple::keylet::account(*accountID).key, lgrInfo.seq, ctx.yield);
+
+    if (not accountLedgerObject)
```
mathbunnyru:
```suggestion
    if (not accountLedgerObject.has_value())
```

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,237 @@
...
+        {JS(account), output.account},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(ledger_index), output.ledgerIndex},
+        {JS(validated), output.validated},
+        {JS(limit), output.limit},
+        {"mpt_issuances", value_from(output.issuances)},
+    };
+
+    if (output.marker)
+        obj[JS(marker)] = output.marker.value();
```
mathbunnyru:
```suggestion
    if (output.marker.has_value())
        obj[JS(marker)] = *output.marker;
```

src/rpc/handlers/AccountMPTokens.cpp
```diff
@@ -0,0 +1,191 @@
...
+    auto obj = boost::json::object{
+        {JS(account), output.account},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(ledger_index), output.ledgerIndex},
+        {JS(validated), output.validated},
+        {JS(limit), output.limit},
+        {"mptokens", boost::json::value_from(output.mpts)},
+    };
+
+    if (output.marker)
```
mathbunnyru:
```suggestion
    if (output.marker.has_value())
```

tests/common/util/TestObject.cpp
```diff
@@ -1442,38 +1442,69 @@ createLptCurrency(std::string_view assetCurrency, std::string_view asset2Currenc
...
+        mptIssuance.setFieldU8(ripple::sfAssetScale, *assetScale);
+    if (maxAmount)
+        mptIssuance.setFieldU64(ripple::sfMaximumAmount, *maxAmount);
+    if (lockedAmount)
+        mptIssuance.setFieldU64(ripple::sfLockedAmount, *lockedAmount);
+    if (metadata) {
+        ripple::Slice const sliceMetadata(metadata->data(), metadata->size());
+        mptIssuance.setFieldVL(ripple::sfMPTokenMetadata, sliceMetadata);
+    }
+    if (domainId)
```
mathbunnyru: Could you please add `.has_value()` in all these statements above?

tests/common/util/TestObject.cpp
```diff
@@ -1442,38 +1442,69 @@ createLptCurrency(std::string_view assetCurrency, std::string_view asset2Currenc
...
     mptoken.setFieldU16(ripple::sfLedgerEntryType, ripple::ltMPTOKEN);
-    mptoken.setFieldU32(ripple::sfFlags, 0);
+    mptoken.setFieldU32(ripple::sfFlags, flags);
     mptoken.setFieldU64(ripple::sfOwnerNode, 0);
     mptoken.setFieldH256(ripple::sfPreviousTxnID, ripple::uint256{});
     mptoken.setFieldU32(ripple::sfPreviousTxnLgrSeq, 0);
 
     if (mptAmount != 0u)
         mptoken.setFieldU64(ripple::sfMPTAmount, mptAmount);
+    if (lockedAmount)
```
mathbunnyru:
```suggestion
    if (lockedAmount.has_value())
```

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -0,0 +1,237 @@
...
+    if (jsonObject.contains(JS(ledger_hash)))
+        input.ledgerHash = boost::json::value_to<std::string>(jv.at(JS(ledger_hash)));
+
+    if (jsonObject.contains(JS(ledger_index))) {
+        if (!jsonObject.at(JS(ledger_index)).is_string()) {
+            input.ledgerIndex = util::integralValueAs<uint32_t>(jv.at(JS(ledger_index)));
+        } else if (jsonObject.at(JS(ledger_index)).as_string() != "validated") {
+            input.ledgerIndex = std::stoi(boost::json::value_to<std::string>(jv.at(JS(ledger_index))));
+        }
+    }
```
mathbunnyru, line 175: This seems to be an extremely duplicated pattern for different RPCs, may I ask you to create an issue to have a function to do this?

mathbunnyru (changes requested):
Overall, LGTM, left some (mostly nitpicking) comments

src/rpc/handlers/AccountMPTokenIssuances.cpp
godexsoft: I did not suggest to do exactly that because of `sle.getField***` which are different for each line and it will just make things more cumbersome than already are

src/rpc/handlers/AccountMPTokenIssuances.cpp
godexsoft: We have other RPC with same pattern, unfortunately. We probably should change it in a separate PR all at once

mathbunnyru: Ok, then could you please create an issue for this as well?

src/rpc/handlers/AccountMPTokenIssuances.hpp
yinyiqian1: Because this field is optional in the ledger object `MPTokenIssuance` in rippled

tests/common/util/TestObject.cpp
yinyiqian1: In the header file, these values have default value as std::nullopt. So it's flexible to keep the signature as it is.

for example, in the following callers, the rest parameters not provided will be the default value. It's cleaner and easier to just leave them out.
```
createMptIssuanceObject(kACCOUNT, kSEQ, "metadata");
createMptIssuanceObject(kACCOUNT, 1, 1000, "meta", ripple::lsfMPTCanTrade);
createMptIssuanceObject(
        kACCOUNT,
        2,
        kISSUANCE2_METADATA,
        ripple::lsfMPTLocked | ripple::lsfMPTCanLock | ripple::lsfMPTCanClawback,
        kISSUANCE2_OUTSTANDING_AMOUNT);
```

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: Yes. These getField methods are different. So I did not use lambda for this block of code.

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: There's another comment related to refactor. I can create an issue and list these two items in it. There might be other places need to refactor in the RPC handlers.

mathbunnyru: I think it's 2 separate problems, so let's create 2 separate issues for them.

src/rpc/handlers/AccountMPTokenIssuances.cpp
mathbunnyru: Got it, thanks 👍

mathbunnyru (approved):

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: Just created: https://github.com/XRPLF/clio/issues/2732

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: Just created: https://github.com/XRPLF/clio/issues/2733

PR #2684, opened 2025-10-08
refactor: Keyspace comments

PeterChen13579:


src/data/cassandra/impl/Cluster.hpp
```diff
@@ -40,16 +40,10 @@ namespace {
...
+providerFromString(std::string const& provider)
 {
-    switch (provider) {
-        case Provider::Cassandra:
-            return "cassandra";
-        case Provider::Keyspace:
-            return "aws_keyspace";
-    }
-    std::unreachable();
+    return provider == "cassandra" ? Provider::Cassandra : Provider::Keyspace;
```
mathbunnyru, line 48: Should `ASSERT` that the value is one of expected values, otherwise someone writing `casandra` (with one `s`) will get Keyspace 😆

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -233,9 +233,9 @@ class CassandraBackendFamily : public BackendInterface {
     std::optional<std::uint32_t>
     fetchLatestLedgerSequence(boost::asio::yield_context yield) const override
     {
-        if (auto const res = executor_.read(yield, schema_->selectLatestLedger); res) {
-            if (auto const& result = res.value(); result) {
-                if (auto const maybeValue = result.template get<uint32_t>(); maybeValue)
+        if (auto const res = executor_.read(yield, schema_->selectLatestLedger); res.has_value()) {
+            if (auto const& result = *res; result) {
```
mathbunnyru: `result` inside `res`, could we do better with naming here?

src/data/KeyspaceBackend.hpp
```diff
@@ -203,8 +203,8 @@ class BasicKeyspaceBackend : public CassandraBackendFamily<
         statement.bindAt(3, Limit{limit});
 
         auto const res = executor_.read(yield, statement);
-        if (res && res.value().hasRows()) {
-            for (auto const [nftID] : extract<ripple::uint256>(res.value()))
+        if (res.has_value() && (*res).hasRows()) {
```
mathbunnyru:
```suggestion
        if (res.has_value() && res->hasRows()) {
```

mathbunnyru (changes requested):
Only small changes, overall, looks good, thank you for fixing this!

mathbunnyru (approved):

PR #2685, opened 2025-10-09
chore: Publish RCs as non-draft

### LINKED ISSUE #2672: Publish releases in non-draft mode
Now, we publish releases always as drafts, but the release process is now quite stable, so we should consider creating releases in non-draft mode

mathbunnyru:
Fix: https://github.com/XRPLF/clio/issues/2672


godexsoft (approved):
:rage1:

PR #2686, opened 2025-10-09
fix: Drop dynamic loggers to fix memory leak

mathbunnyru:


src/util/log/Logger.cpp
```diff
@@ -427,8 +429,19 @@ LogServiceState::replaceSinks(std::vector<std::shared_ptr<spdlog::sinks::sink>>
...
+{
+}
+
+Logger::~Logger()
 {
+    if (logger_ == nullptr) {
+        return;
+    }
+    bool const isDynamic = !std::ranges::contains(kCHANNELS, logger_->name());
+    if (isDynamic && logger_.use_count() == 2) {
```
godexsoft: nit: Can we make a constant for this '2' value? it's not clear

godexsoft: nit: Also, `and` would fit perfectly here :)

godexsoft: It may never happen but it may be worth doing <= instead == just to be sure

tests/unit/util/log/LoggerTests.cpp
```diff
@@ -21,11 +21,23 @@
...
 #include <cstddef>
 #include <string>
 using namespace util;
 
+namespace {
+size_t
+loggersNum()
+{
+    size_t counter = 0;
+    spdlog::apply_all([&counter](std::shared_ptr<spdlog::logger>) { ++counter; });
```
godexsoft, line 36: nit: `auto` maybe?

tests/unit/util/log/LoggerTests.cpp
```diff
@@ -71,3 +83,27 @@ TEST_F(LoggerTest, LOGMacro)
...
 #endif
+
+TEST_F(LoggerTest, ManyDynamicLoggers)
+{
+    static constexpr size_t kNUM_LOGGERS = 10'000;
+
+    auto initialLoggers = loggersNum();
+
+    for (size_t i = 0; i < kNUM_LOGGERS; ++i) {
+        Logger log{"DynamicLogger" + std::to_string(i)};
```
godexsoft: nit: Can create a constant for the strings and reuse down below

godexsoft (changes requested):
Leaving a few nits. I think overall this should fix the issue and it looks like a good solution 👍

src/util/log/Logger.cpp
mathbunnyru: Made 2 a constant with explanation why

I don't think `<=` makes sense, it should happen only once

tests/unit/util/log/LoggerTests.cpp
mathbunnyru: With `auto` it's not very clear from first glance what `all` means in `spdlog::apply_*all*`

tests/unit/util/log/LoggerTests.cpp
mathbunnyru: Thanks, done

godexsoft (approved):
:rage2:

src/util/log/Logger.cpp
```diff
@@ -284,13 +284,15 @@ LogServiceState::reset()
 }
 
 std::shared_ptr<spdlog::logger>
-LogServiceState::registerLogger(std::string const& channel, std::optional<Severity> severity)
+LogServiceState::registerLogger(std::string_view channel, std::optional<Severity> severity)
```
kuznetsss, line 287: Using `std::string` here would allow user of the method to move `channel` if it is possible

PR #2697, opened 2025-10-16
ci: Add libatomic as dependency for pre-commit image

mathbunnyru:
Same as in https://github.com/XRPLF/ci/commit/a8c7be18944ad85c0a2c6d6459b9efcf2f080c2a


mathbunnyru:
CI failure is fine, it will be fixed when we start using this newer image

godexsoft (approved):

Issue #1702, opened 2024-10-24
Handle Prometheus requests in Work Queue


kuznetsss:
## Summary

Currently Clio is using web server's threads to handle Prometheus requests. It would be better to leave these threads only for web server and post a job on a work queue instead.

kuznetsss:
Also move other things into work queue to speed up web server.

PR #2721, opened 2025-10-25
feat: WorkQueue priorities

godexsoft:
This PR adds normal/high priority support to WorkQueue. The current implementation is a fair share round robin approach.
Some tests need to be updated to use WorkQueue Mock instead as using the real one is cumbersome in tests.

Related to #1702


src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,138 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+                }
+
+                if (not state->normal.empty()) {
+                    batch.push_back(std::move(state->normal.front()));
+                    state->normal.pop();
+                }
+            }
+        }
+
+        if (not stopping_ and shouldWait) {
```
kuznetsss:
```suggestion
        if (not stopping_ and batch.empty()) {
```

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -99,10 +107,61 @@ TEST_F(WorkQueueTest, NonWhitelistedPreventSchedulingAtQueueLimitExceeded)
...
     EXPECT_TRUE(unblocked);
 }
 
+struct WorkQueuePriorityTest : WithPrometheus, RPCWorkQueueTestBase {
+    WorkQueuePriorityTest() : RPCWorkQueueTestBase(1, 100)
+    {
+    }
+};
+
+TEST_F(WorkQueuePriorityTest, HighPriorityTasks)
```
mathbunnyru, line 118: This test failed, so there is some bug somewhere

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+        std::vector<std::function<void(boost::asio::yield_context)>> batch;
+
+        {
+            auto state = dispatcherState_.lock();
+
+            if (state->empty()) {
+                state->isIdle = true;
+            } else {
+                auto highPrioCount = 0uz;
+                while (highPrioCount < kHIGH_PRIO_RATIO and not state->high.empty()) {
```
mathbunnyru: `kHIGH_PRIO_RATIO` - it's not exactly ratio, maybe let's call it `kTAKE_HIGH_PRIO`?

mathbunnyru: This can be written in one `for` loop, and won't need to increment `highPrioCount` in the end of loop

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+        if (not stopping_ and batch.empty()) {
+            waitTimer_.expires_at(std::chrono::steady_clock::time_point::max());
+            boost::system::error_code ec;
+            waitTimer_.async_wait(yield[ec]);
+        } else {
+            for (auto task : std::move(batch)) {
+                util::spawn(
+                    ioc_, [this, start = std::chrono::system_clock::now(), task = std::move(task)](auto yield) mutable {
+                        auto const run = std::chrono::system_clock::now();
+                        auto const wait = std::chrono::duration_cast<std::chrono::microseconds>(run - start).count();
```
mathbunnyru:
```suggestion
                    ioc_, [this, spawnedAt = std::chrono::system_clock::now(), task = std::move(task)](auto yield) mutable {
                        auto const takenAt = std::chrono::system_clock::now();
                        auto const waited = std::chrono::duration_cast<std::chrono::microseconds>(takenAt - spawnedAt).count();
```

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+
+                        ++queued_.get();
+                        durationUs_.get() += wait;
```
mathbunnyru: I know it's not your code, but probably should be named waitedUs_

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+        boost::asio::post(strand_, [this] { waitTimer_.cancel(); });
+
+    return true;
 }
 
 void
-WorkQueue::stop(std::function<void()> onQueueEmpty)
+WorkQueue::dispatcherLoop(boost::asio::yield_context yield)
+{
+    LOG(log_.debug()) << "WorkQueue dispatcher starting";
```
mathbunnyru: maybe info here?

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+
+            boost::asio::post(ioc_.get_executor(), yield);  // yield back to avoid hijacking the thread
+        }
+    }
+
+    LOG(log_.info()) << "WorkQueue dispatcher shutdown requested - time to execute onTasksComplete";
+
+    auto onTasksComplete = onQueueEmpty_.lock();
+    ASSERT(onTasksComplete->operator bool(), "onTasksComplete must be set when stopping is true.");
+    onTasksComplete->operator()();
```
mathbunnyru: Maybe let's create a section here, since last LOG station doesn't need to be under mutex?

src/rpc/WorkQueue.hpp
```diff
@@ -19,33 +19,100 @@
...
 #include <cstddef>
 #include <cstdint>
 #include <functional>
 #include <limits>
+#include <queue>
 
 namespace rpc {
 
+namespace impl {
+class OneTimeCallable {
```
mathbunnyru: Let's maybe make it non-copyable?

mathbunnyru: And let's test this class?

src/rpc/WorkQueue.hpp
```diff
@@ -19,33 +19,100 @@
...
+        bool isIdle = false;
+
+        void
+        push(Priority priority, auto&& task)
+        {
+            if (priority == Priority::High) {
+                high.push(std::forward<decltype(task)>(task));
+            } else {
+                normal.push(std::forward<decltype(task)>(task));
+            }
```
mathbunnyru: nit: you can take a reference to a right queue using a ternary operator, and then push to it

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +81,145 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
     }
+
+    if (needsWakeup)
+        boost::asio::post(strand_, [this] { waitTimer_.cancel(); });
+}
+
+void
+WorkQueue::stop()
+{
+    if (not stopping_)
```
mathbunnyru: Is this an attempt to prevent from calling twice? If yes, it's not how it should be done: if 2 threads call stop simultaneously, and execute `if (not stopping_)`, they will both go to `requestStop()`

src/rpc/WorkQueue.hpp
```diff
@@ -19,33 +19,100 @@
...
+     * @brief Represents a task scheduling priority
+     */
+    enum class Priority : uint8_t {
+        High,
+        Default,
+    };
+
+private:
+    struct DispatcherState {
+        std::queue<std::function<void(boost::asio::yield_context)>> high;
```
mathbunnyru: Let's add a `using` for this `std::function<void(boost::asio::yield_context)>` and use it everywhere?

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -30,24 +30,35 @@
...
+              {"workers", ConfigValue{ConfigType::Integer}.defaultValue(workers)},
+          }
+        , queue{WorkQueue::makeWorkQueue(cfg)}
+    {
+    }
 };
 
-struct WorkQueueTest : WithPrometheus, RPCWorkQueueTestBase {};
+struct WorkQueueTest : WithPrometheus, RPCWorkQueueTestBase {
+    WorkQueueTest() : RPCWorkQueueTestBase(4, 2)
```
mathbunnyru: probably don't need to specify 4 and 2 here, since they are default

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -99,10 +107,61 @@ TEST_F(WorkQueueTest, NonWhitelistedPreventSchedulingAtQueueLimitExceeded)
...
+    std::vector<WorkQueue::Priority> executionOrder;
+    std::mutex mtx;
+
+    for (int i = 0; i < kTOTAL; ++i) {
+        queue.postCoro(
+            [&](auto) {
+                std::lock_guard const lock(mtx);
+                executionOrder.push_back(WorkQueue::Priority::High);
+            },
+            true,
```
mathbunnyru: Please, add a comment what this `true` means like `/* whiteListed =/* true`

and below as well

Even better would be to have a enum class for that, but might be too much effort

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -99,10 +107,61 @@ TEST_F(WorkQueueTest, NonWhitelistedPreventSchedulingAtQueueLimitExceeded)
...
+    // with 1 worker, the execution order is deterministic
+    // we should see 4 high prio tasks, then 1 normal prio task, until high prio tasks are depleted
+    std::vector<WorkQueue::Priority> const expectedOrder = {
+        WorkQueue::Priority::High,    WorkQueue::Priority::High,    WorkQueue::Priority::High,
+        WorkQueue::Priority::High,    WorkQueue::Priority::Default, WorkQueue::Priority::High,
+        WorkQueue::Priority::High,    WorkQueue::Priority::High,    WorkQueue::Priority::High,
+        WorkQueue::Priority::Default, WorkQueue::Priority::High,    WorkQueue::Priority::High,
+        WorkQueue::Priority::Default, WorkQueue::Priority::Default, WorkQueue::Priority::Default,
+        WorkQueue::Priority::Default, WorkQueue::Priority::Default, WorkQueue::Priority::Default,
+        WorkQueue::Priority::Default, WorkQueue::Priority::Default,
```
mathbunnyru, line 155: Please, write a lambda/func to give this expected order for given amount of high / total number of tasks

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -99,10 +107,61 @@ TEST_F(WorkQueueTest, NonWhitelistedPreventSchedulingAtQueueLimitExceeded)
...
+    };
+
+    ASSERT_EQ(executionOrder.size(), expectedOrder.size());
+    for (auto i = 0uz; i < executionOrder.size(); ++i) {
+        EXPECT_EQ(executionOrder[i], expectedOrder[i]) << "Mismatch at index " << i;
+    }
```
mathbunnyru, line 161: There is a ContainerEq matcher https://google.github.io/googletest/reference/matchers.html#container-matchers

mathbunnyru (changes requested):

tests/unit/rpc/WorkQueueTests.cpp
godexsoft: Yes, with enough attempts i was able to reproduce even on the mac. Had to add means to postpone task processing to fix this test. To be fair this mechanism might come in handy some day so i'm not opposed adding it eitherway

tests/unit/rpc/WorkQueueTests.cpp
godexsoft: The whole point is that it's visible in plain text here. writing a lambda means "replicate" the inner algorithm of the work queue - i wanted to avoid that specifically

tests/unit/rpc/WorkQueueTests.cpp
godexsoft: Since i specify it in the other one I decided to leave it in. Maybe i should remove the default instead to force clients to set these

src/rpc/WorkQueue.cpp
godexsoft: It's a "best effort" check. The `requestStop` function can already be called multiple times but i think you are still right because the empty callable here will overwrite whatever was set by `requestStop` and potentially call the empty one internally instead of the user-provided one. I think we can make it safe with `.exchange(true)` though

src/rpc/WorkQueue.cpp
godexsoft: That's the name of the metric.. let's keep as is, maybe revisit later

src/rpc/WorkQueue.cpp
godexsoft: I have no strong opinion on this so i'll set back to info

src/rpc/WorkQueue.cpp
godexsoft: You will have to explain it again :)

src/rpc/WorkQueue.hpp
godexsoft: Not a fan of ternary for this but maybe an immediately invoked lambda

tests/unit/rpc/WorkQueueTests.cpp
godexsoft: I was excited for a minute.. it does not work unfortunately. It does not include the index (despite doc says so) and its output is worse than the custom loop we have

src/rpc/WorkQueue.cpp
mathbunnyru: You don't need to have onQueueEmpty_ locked to make a last LOG, so we can slightly decrease the scop of lock

src/rpc/WorkQueue.cpp
mathbunnyru: 👍

src/rpc/WorkQueue.hpp
mathbunnyru: There are more places with `std::function<void(boost::asio::yield_context)>`

tests/unit/rpc/WorkQueueTests.cpp
mathbunnyru: I agree, let's remove the defaults, but add comments on the caller side what these values are for

tests/unit/rpc/WorkQueueTests.cpp
mathbunnyru: That's unfortunate:(

src/rpc/WorkQueue.cpp
mathbunnyru: Ok

kuznetsss (dismissed):

tests/unit/rpc/WorkQueueTests.cpp
mathbunnyru: Ok

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +57,151 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+    LOG(log_.info()) << "WorkQueue dispatcher finished";
 }
 
 void
-WorkQueue::stop(std::function<void()> onQueueEmpty)
+WorkQueue::requestStop(std::function<void()> onQueueEmpty)
 {
     auto handler = onQueueEmpty_.lock();
-    handler->setCallable(std::move(onQueueEmpty));
+    handler->operator=(std::move(onQueueEmpty));
```
mathbunnyru:
```suggestion
    *handler = std::move(onQueueEmpty);
```

Maybe we can rewrite this in a clearer way?

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +57,151 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
-        handler->operator()();
+    auto needsWakeup = false;
+
+    {
+        auto state = dispatcherState_.lock();
+        if (state->isIdle) {
+            needsWakeup = true;
+            state->isIdle = false;
+        }
     }
```
mathbunnyru, line 197:
```suggestion

    bool needsWakeup;
    {
        auto state = dispatcherState_.lock();
        needsWakeup = std::exchange(state->isIdle, false);
}
```

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +57,151 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+    ++curSize_.get();
+    auto needsWakeup = false;
+
+    {
+        auto state = dispatcherState_.lock();
+
+        if (state->isIdle) {
+            needsWakeup = true;
+            state->isIdle = false;
+        }
```
mathbunnyru: Can use `std::exchange` here

mathbunnyru (dismissed):
Overall, LGTM
Left some minor suggestions - feel free to ignore

src/rpc/WorkQueue.cpp
godexsoft: @kuznetsss did you prefer the longer syntax or you don't care?

kuznetsss: I think it doesn't really matter, both are readable fine for me

src/rpc/WorkQueue.cpp
kuznetsss: I think this suggestion decreases readability because the current version is clear, but with exchange it is a bit tricky

src/rpc/WorkQueue.cpp
```diff
@@ -69,25 +57,151 @@ WorkQueue::WorkQueue(std::uint32_t numWorkers, uint32_t maxSize)
...
+WorkQueue::dispatcherLoop(boost::asio::yield_context yield)
+{
+    LOG(log_.info()) << "WorkQueue dispatcher starting";
+
+    // all ongoing tasks must be completed before stopping fully
+    while (not stopping_ or size() > 0) {
+        if (not enabled_) {
+            boost::asio::post(ioc_.get_executor(), yield);  // yield back to avoid hijacking the thread
+            continue;
+        }
```
kuznetsss: I think it would be better to remove this `enabled_` flag and make enable() method to spawn dispatcher loop on strand if it is not running yet. And add a flag 
```cpp
std::optional<DontStartProcessing> disabled = std::nullopt
```
to the constructor.
This would help to separate starting logic and processing loop.

src/rpc/WorkQueue.cpp
godexsoft: I agree, the code was simpler but with `exchange` it's clear enough (you do need to know what `std::exchange` is tho)

src/rpc/WorkQueue.cpp
mathbunnyru: Agree, but let's call method `startProcessing` maybe?
and call flag `isProcessing`

mathbunnyru (dismissed):

kuznetsss (dismissed):

mathbunnyru (approved):

PR #2729, opened 2025-10-28
test: Fix flaky test

### LINKED ISSUE #2700: WebWsConnectionTests.DisconnectClientOnInactivity flaky or incorrect
## Issue Description

The test is either flaky or does not work correctly. 
With increased client timer (50ms) it was able to pass 7500 iterations but then it still failed:

```
/clio/tests/unit/web/ng/impl/WsConnectionTests.cpp:136: Failure
Expected: (end - start) < (std::chrono::milliseconds{50}), actual: 59301292ns vs 50ms
```

Test is disabled in #2699 

## Steps to Reproduce

Run the test enough times

## Expected Result

The test should always pass

## Actual Result

Test sometimes fails

## Environment

I'm using x86 docker on an arm Mac. This is slow so the test shows the flakiness (or that it's not really correct to start with)

kuznetsss:
Fixes #2700.


tests/unit/web/ng/impl/WsConnectionTests.cpp
```diff
@@ -108,32 +110,38 @@ TEST_F(WebWsConnectionTests, WasUpgraded)
...
+    bool finished{false};
+
+    util::spawn(clientCtx, [&](boost::asio::yield_context yield) {
         auto expectedSuccess =
             wsClient_.connect("localhost", httpServer_.port(), yield, std::chrono::milliseconds{100});
         [&]() { ASSERT_TRUE(expectedSuccess.has_value()) << expectedSuccess.error().message(); }();
-        boost::asio::steady_timer timer{yield.get_executor(), std::chrono::milliseconds{5}};
-        timer.async_wait(yield);
+        std::unique_lock lock{mutex};
+        auto const gotNotified = cv.wait_for(lock, std::chrono::seconds{2}, [&finished]() { return finished; });
```
mathbunnyru, line 129: Just `wait` will also work if the test succeeds.
But I do understand the desire to limit the test execution.

Let's maybe add a comment why we use time limit here?

mathbunnyru (dismissed):

mathbunnyru (approved):

PR #2745, opened 2025-10-30
ci: Update docker images

mathbunnyru:


.pre-commit-config.yaml
```diff
@@ -11,7 +11,10 @@
 #
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-exclude: ^(docs/doxygen-awesome-theme/|conan\.lock$)
+exclude: |
+  (?x)^(
+    docs/doxygen-awesome-theme/.*
+  )$
```
mathbunnyru, line 17: This change and below are not mistakes - now `conan lock` adds newline to the end of file (as it should), so there is no reason to exclude it from pre-commit.
Also rewrote pre-commit so it's easier to extend when new files/dirs are needed to be added (also matches how rippled does it)

kuznetsss (approved):

PR #2746, opened 2025-10-30
fix: Empty signer list

### LINKED ISSUE #2730: Empty signer_lists field missing in response (Version: 2.6.0)
## Issue Description

Issue shared from Mattermost:
> looks like clio has braking change and do not return empty signer lists in response anymore

## Reproduction Steps

if I send request like this, then there is no signer_lists on any response, in case it is empty
```
{
    "command": "account_info",
    "account": "rsuUjfWxrACCAwGQDsNeZUhpzXf1n1NK5Z",
    "signer_lists": true
}
{
    "command": "account_info",
    "account": "rsuUjfWxrACCAwGQDsNeZUhpzXf1n1NK5Z",
    "ledger_index": "61770679",
    "signer_lists": true
}
```
looks like it was broken here, there is no any handling if signers is not exist

https://github.com/XRPLF/clio/commit/c780ef8a0b0a3e735eceea796c89213079c0e812
btw, this works, but it should be rippled response, not clio
```
{
    "command": "account_info",
    "account": "rsuUjfWxrACCAwGQDsNeZUhpzXf1n1NK5Z",
    "ledger_index": "current",
    "signer_lists": true
}
```

## Supporting Files

https://github.com/XRPLF/clio/commit/c780ef8a0b0a3e735eceea796c89213079c0e812

PeterChen13579:
fixes #2730


src/rpc/handlers/AccountInfo.cpp
```diff
@@ -122,8 +123,9 @@ AccountInfoHandler::process(AccountInfoHandler::Input const& input, Context cons
             if (!signersKey.check(sleSigners))
                 return Error{Status{RippledError::rpcDB_DESERIALIZATION}};
 
-            out.signerLists = std::vector<ripple::STLedgerEntry>{sleSigners};
+            signerList.push_back(sleSigners);
         }
+        out.signerLists = std::move(signerList);
```
kuznetsss: We could avoid this by creating vector inside `out` at the line 116:
```cpp
out.signerList = std::vector<ripple::STLedgerEntry>();
```
And push into it on the line 126: 
```cpp
out.signerList->push_back(signerList);
```

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -655,6 +655,49 @@ TEST_F(RPCAccountInfoHandlerTest, IdentAndSignerListsFalse)
     });
 }
 
+TEST_F(RPCAccountInfoHandlerTest, EmptySignerLists)
+{
+    auto const ledgerHeader = createLedgerHeader(kLEDGER_HASH, 30);
+    EXPECT_CALL(*backend_, fetchLedgerBySequence).Times(1);
+    ON_CALL(*backend_, fetchLedgerBySequence).WillByDefault(Return(ledgerHeader));
```
kuznetsss: I know that this is probably copied from another test, but let's use one `EXPECT_CALL().WillOnce()` where possible instead of `EXPECT_CALL()` and `ON_CALL()`

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -208,9 +207,8 @@ TEST_F(RPCAccountInfoHandlerTest, LedgerNonExistViaIntSequence)
 
 TEST_F(RPCAccountInfoHandlerTest, LedgerNonExistViaStringSequence)
 {
-    EXPECT_CALL(*backend_, fetchLedgerBySequence).Times(1);
     // return empty ledgerHeader
-    ON_CALL(*backend_, fetchLedgerBySequence(30, _)).WillByDefault(Return(std::nullopt));
```
mathbunnyru, line 213: I think you removed the check of expected input params, let's add it back in these tests?

PeterChen13579: Ah, you're right. This has an extra check that `fetchLedgerBySequence` must be called with param value of `30`

mathbunnyru: Yeah, but I think you can have a better syntax and a check for input param as well.

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -655,6 +647,48 @@ TEST_F(RPCAccountInfoHandlerTest, IdentAndSignerListsFalse)
...
+TEST_F(RPCAccountInfoHandlerTest, EmptySignerLists)
+{
+    auto const ledgerHeader = createLedgerHeader(kLEDGER_HASH, 30);
+    EXPECT_CALL(*backend_, fetchLedgerBySequence).WillOnce(Return(ledgerHeader));
+
+    auto const account = getAccountIdWithString(kACCOUNT);
+    auto const accountKk = ripple::keylet::account(account).key;
+    auto const accountRoot = createAccountRootObject(kACCOUNT, 0, 2, 200, 2, kINDEX1, 2);
+    ON_CALL(*backend_, doFetchLedgerObject(accountKk, 30, _))
+        .WillByDefault(Return(accountRoot.getSerializer().peekData()));
```
kuznetsss, line 657: Is it expected to be called more than once? I think it would be better to put a specific number of expected calls here because we should fetch data from db as few times as possible.

PeterChen13579: It’s only called once, but EXPECT_CALL is not applicable here because this line isn’t verifying behavior, it’s defining mock behavior.
Here we’re just setting up the mock so that fetching the account keylet returns the accountRoot data. The number of times it’s called isn’t what we’re testing in this case.

kuznetsss: I think it would still better to check that we don't query DB more times than it's actually needed.

kuznetsss (dismissed):
👍

tests/unit/rpc/handlers/AccountInfoTests.cpp
PeterChen13579: Added; Sorry, originally I thought you meant change `on_call` to `expect_call` behavior 😂

mathbunnyru (approved):
👍

PR #2751, opened 2025-11-03
feat: Async framework `submit` on strand/ctx

godexsoft:
This adds a simple way to run one shot tasks that don't require a handle to retrieve the result (and no yield context).
It will come in handy for the publisher upgrade to AsyncContext.

### DIFF
--- a/src/util/async/AnyStrand.hpp
+++ b/src/util/async/AnyStrand.hpp
@@ -155,7 +155,7 @@ public:
 
     /**
      * @brief Schedule an operation on the execution context without expectations of a result
-     * @note Errors are caught internally and logged as errors
+     * @note Exceptions are caught internally and `ASSERT`ed on
      *
      * @param fn The block of code to execute
      */
--- a/src/util/async/AnyStrand.hpp
+++ b/src/util/async/AnyStrand.hpp
@@ -68,12 +68,12 @@ public:
         static_assert(not std::is_same_v<RetType, std::any>);
 
         return AnyOperation<RetType>(  //
-            pimpl_->execute([fn = std::forward<decltype(fn)>(fn)]() -> std::any {
+            pimpl_->execute([fn = std::forward<decltype(fn)>(fn)] mutable -> std::any {
                 if constexpr (std::is_void_v<RetType>) {
-                    std::invoke(std::move(fn));
+                    std::invoke(std::forward<decltype(fn)>(fn));
                     return {};
                 } else {
-                    return std::make_any<RetType>(std::invoke(std::move(fn)));
+                    return std::make_any<RetType>(std::invoke(std::forward<decltype(fn)>(fn)));
                 }
             })
         );
@@ -146,8 +146,8 @@ public:
 
         auto const millis = std::chrono::duration_cast<std::chrono::milliseconds>(interval);
         return AnyOperation<RetType>(  //
-            pimpl_->executeRepeatedly(millis, [fn = std::forward<decltype(fn)>(fn)] -> std::any {
-                std::invoke(std::move(fn));
+            pimpl_->executeRepeatedly(millis, [fn = std::forward<decltype(fn)>(fn)] mutable -> std::any {
+                std::invoke(std::forward<decltype(fn)>(fn));
                 return {};
             })
         );
--- a/src/util/async/AnyStrand.hpp
+++ b/src/util/async/AnyStrand.hpp
@@ -92,12 +92,12 @@ public:
         static_assert(not std::is_same_v<RetType, std::any>);
 
         return AnyOperation<RetType>(  //
-            pimpl_->execute([fn = std::forward<decltype(fn)>(fn)](auto stopToken) -> std::any {
+            pimpl_->execute([fn = std::forward<decltype(fn)>(fn)](auto stopToken) mutable -> std::any {
                 if constexpr (std::is_void_v<RetType>) {
-                    fn(std::move(stopToken));
+                    std::invoke(std::forward<decltype(fn)>(fn), std::move(stopToken));
                     return {};
                 } else {
-                    return std::make_any<RetType>(fn(std::move(stopToken)));
+                    return std::make_any<RetType>(std::invoke(std::forward<decltype(fn)>(fn), std::move(stopToken)));
                 }
             })
         );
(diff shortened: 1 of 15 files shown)


src/util/async/impl/ErrorHandling.hpp
```diff
@@ -50,6 +51,22 @@ struct DefaultErrorHandler {
...
+        return [fn = std::forward<decltype(fn)>(fn)] {
+            try {
+                std::invoke(fn);
+            } catch (std::exception const& e) {
+                util::Logger log("AsyncFramework");
+                LOG(log.error()) << "Exception silenced: " << e.what() << " on thread " << std::this_thread::get_id();
+            } catch (...) {
+                util::Logger log("AsyncFramework");
+                LOG(log.error()) << "Unknown exception silenced on thread " << std::this_thread::get_id();
+            }
```
kuznetsss, line 65: Maybe it would be better to add `ASSERT()` here instead of silencing exceptions

mathbunnyru (changes requested):
There were some docs for async, let's update if necessary

src/util/async/context/impl/Execution.hpp
```diff
@@ -42,9 +42,9 @@ struct SpawnDispatchStrategy {
              fn = std::forward<decltype(fn)>(fn)](auto yield) mutable {
                 if constexpr (SomeStoppableOutcome<OutcomeType>) {
                     auto& stopSource = outcome.getStopSource();
-                    fn(outcome, stopSource, stopSource[yield]);
+                    std::invoke(fn, outcome, stopSource, stopSource[yield]);
```
kuznetsss: Probably you should forward `fn` here and in the other invoke calls

tests/unit/util/async/AsyncExecutionContextTests.cpp
```diff
@@ -38,7 +39,7 @@ using namespace util::async;
 using ::testing::Types;
 
 template <typename T>
-struct ExecutionContextTests : public ::testing::Test {
+struct ExecutionContextTests : public common::util::WithMockAssertNoThrow {
```
kuznetsss: Nit:
```suggestion
struct ExecutionContextTests : common::util::WithMockAssertNoThrow {
```

src/util/async/context/impl/Execution.hpp
```diff
@@ -42,15 +42,22 @@ struct SpawnDispatchStrategy {
...
                 }
             }
         );
 
         return op;
     }
+
+    template <typename ContextType>
+    static void
+    post(ContextType& ctx, auto&& fn)
```
mathbunnyru: Let's create a typename for `fn` above as well, because now it's not consistent (you have type for `ctx`)
And then use this type in `std::forward` below

src/util/async/context/impl/Execution.hpp
```diff
@@ -42,15 +42,22 @@ struct SpawnDispatchStrategy {
...
+    {
+        util::spawn(ctx.getExecutor(), [fn = std::forward<decltype(fn)>(fn)](auto) mutable { std::invoke(fn); });
```
mathbunnyru: Ideally, `fn` should also be forward'ed (and would be great if you fixed in `std::invoke` below as well)

src/util/async/context/impl/Execution.hpp
```diff
@@ -65,15 +72,22 @@ struct PostDispatchStrategy {
...
                 }
             }
         );
 
         return op;
     }
+
+    template <typename ContextType>
+    static void
+    post(ContextType& ctx, auto&& fn)
```
mathbunnyru: Here let's also add a type for `fn`

src/util/async/context/impl/Strand.hpp
```diff
@@ -108,12 +108,12 @@ class BasicStrand {
             context_,
             impl::outcomeForHandler<StopSourceType>(fn),
             ErrorHandlerType::wrap([fn = std::forward<decltype(fn)>(fn)](auto& outcome) mutable {
-                using FnRetType = std::decay_t<decltype(fn())>;
+                using FnRetType = std::decay_t<decltype(std::invoke(fn))>;
```
mathbunnyru: There is `invoke_result_t`

src/util/async/context/impl/Utils.hpp
```diff
@@ -66,7 +66,7 @@ outcomeForHandler(auto&& fn)
 
         return StoppableOutcome<RetType, StopSourceType>();
     } else {
-        using FnRetType = decltype(fn());
+        using FnRetType = decltype(std::invoke(fn));
```
mathbunnyru: Also `invoke_result_t`

src/util/async/context/BasicExecutionContext.hpp
```diff
@@ -350,17 +351,29 @@ class BasicExecutionContext {
             context_,
             impl::outcomeForHandler<StopSourceType>(fn),
             ErrorHandlerType::wrap([fn = std::forward<decltype(fn)>(fn)](auto& outcome) mutable {
-                using FnRetType = std::decay_t<decltype(fn())>;
+                using FnRetType = std::decay_t<decltype(std::invoke(fn))>;
```
mathbunnyru: `invoke_result_t`

src/util/async/AnyStrand.hpp
```diff
@@ -64,16 +64,16 @@ class AnyStrand {
     [[nodiscard]] auto
     execute(SomeHandlerWithoutStopToken auto&& fn)
     {
-        using RetType = std::decay_t<decltype(fn())>;
+        using RetType = std::decay_t<decltype(std::invoke(fn))>;
```
mathbunnyru: also `invoke_result_t`

tests/common/util/MockAssert.hpp
```diff
@@ -64,11 +63,17 @@ class WithMockAssertNoThrow : virtual public testing::Test {
             common::util::WithMockAssert::MockAssertException                                                 \
         );                                                                                                    \
     } else if (dynamic_cast<common::util::WithMockAssertNoThrow*>(this) != nullptr) {                         \
-        testing::StrictMock<testing::MockFunction<void(std::string_view)>> callMock;                          \
-        ::util::impl::OnAssert::setAction([&callMock](std::string_view m) { callMock.Call(m); });             \
-        EXPECT_CALL(callMock, Call(testing::ContainsRegex(message_regex)));                                   \
+        struct MockGuard {                                                                                    \
```
mathbunnyru: Let's put this out of macros

src/util/async/context/impl/Execution.hpp
```diff
@@ -30,68 +30,87 @@
...
         );
 
         return op;
     }
+
+    template <typename ContextType, typename FnType>
+    static void
+    post(ContextType& ctx, FnType&& fn)
+    {
+        util::spawn(ctx.getExecutor(), [fn = std::forward<FnType>(fn)](auto) mutable { std::invoke(std::move(fn)); });
```
mathbunnyru: std::move -> std::forward

src/util/async/AnyStrand.hpp
mathbunnyru: This is still true

tests/common/data/cassandra/FakesAndMocks.hpp
```diff
@@ -124,6 +124,6 @@ struct FakeRetryPolicy {
     void
     retry(Fn&& fn)
     {
-        fn();
+        std::invoke(fn);
```
mathbunnyru: forward here

mathbunnyru (changes requested):
3 places left as far as I see

mathbunnyru (approved):

mathbunnyru:
🥇

PR #2752, opened 2025-11-03
feat: New ETL by default

godexsoft:
Removes old ETL and uses ETLng by default.
This does not remove `ioc` use from ETL (LedgerPublisher) and it requires #2751 to do it easily.
After merging this and #2751 I will prepare another PR to address that last bit.


src/app/CMakeLists.txt
```diff
@@ -2,13 +2,5 @@ add_library(clio_app)
 target_sources(clio_app PRIVATE CliArgs.cpp ClioApplication.cpp Stopper.cpp WebHandlers.cpp)
 
 target_link_libraries(
-  clio_app
```
mathbunnyru, line 5: nit: I think there was a way to keep it nice here by adding a comment?

godexsoft: I tried a few things. According to the docs we should be able to sort things automatically and use `#cmake-format: sortable` to hint if needed.. none of that worked in practice so that leaves one working option - `#cmake-format: off` :rage4:

godexsoft: alternatively we can add some dummy comment in the list itself to force it but it's finicky

godexsoft: another, possibly better way is to set `max_pargs_hwrap: 5` in the config yaml. it changes this and one more case where it looks better vertical anyway 👍

.cmake-format.yaml
```diff
@@ -54,7 +54,7 @@ format:
   _help_max_pargs_hwrap:
     - If a positional argument group contains more than this many
     - arguments, then force it to a vertical layout.
-  max_pargs_hwrap: 6
+  max_pargs_hwrap: 5
```
godexsoft, line 57: This only affects 2 files in the project and both changes look good imo

mathbunnyru (approved):
🥇

mathbunnyru:
Please, grep "etlng", and if there are no more occurrences, I think this is ready to be merged

src/app/CMakeLists.txt
mathbunnyru: LGTM

.cmake-format.yaml
mathbunnyru: LGTM

PR #2755, opened 2025-11-04
refactor: Duplicate ledger_index pattern for RPC handlers

### LINKED ISSUE #2733: Refactor: Duplicate pattern for RPC handers
Please refer to: https://github.com/XRPLF/clio/pull/2680#discussion_r2452149557

Different RPC handlers have duplicate pattern in `tag_invoke` when filling the jsonObject. We should create a common function to do the common part.

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/2733


mathbunnyru (dismissed):
Nice, thank you! 👍

tests/unit/JsonUtilTests.cpp
```diff
@@ -88,3 +88,23 @@ TEST(JsonUtils, integralValueAs)
     auto const stringJson = boost::json::value("not a number");
     EXPECT_THROW(util::integralValueAs<int>(stringJson), std::logic_error);
 }
+
+TEST(JsonUtils, getLedgerIndex)
+{
+    auto const emptyJson = boost::json::value();
+    EXPECT_THROW(util::getLedgerIndex(emptyJson), std::logic_error);
```
kuznetsss: Maybe it would be better to not throw but return `nullopt` in this case? @godexsoft, @mathbunnyru wdyt?

tests/unit/JsonUtilTests.cpp
```diff
@@ -88,3 +88,23 @@ TEST(JsonUtils, integralValueAs)
...
+    EXPECT_EQ(ledgerIndex.value(), 12345u);
+
+    auto const stringJson = boost::json::value("12345");
+    ledgerIndex = util::getLedgerIndex(stringJson);
+    EXPECT_TRUE(ledgerIndex.has_value());
+    EXPECT_EQ(ledgerIndex.value(), 12345u);
+
+    auto const validatedJson = boost::json::value("validated");
+    ledgerIndex = util::getLedgerIndex(validatedJson);
+    EXPECT_FALSE(ledgerIndex.has_value());
```
kuznetsss, line 115: Could you please add test cases where `getLedgerIndex()` is called with bool and with some invalid string?

tests/unit/JsonUtilTests.cpp
mathbunnyru: That's a good question, but I think it's a bit out of scope for this PR, because this is a pure refactoring PR, so with no changes to behaviour at all.

That being said, I agree `nullopt` is better - it's better not to give an easy way to cause exceptions because of the user-input.

godexsoft: Maybe an `expected` then?

src/util/JsonUtils.hpp
```diff
@@ -110,4 +112,26 @@ integralValueAs(boost::json::value const& value)
...
 
+/**
+ * @brief Extracts ledger index from a JSON value which can be either a number or a string.
+ *
+ * @param value The JSON value to extract ledger index from
+ * @return An optional containing the ledger index if it is a number; std::nullopt otherwise
+ * @throws logic_error comes from integralValueAs if the underlying number is neither int64 nor uint64
+ * @throws std::invalid_argument or std::out_of_range if the string cannot be converted to a number
+ */
+inline std::optional<uint32_t>
```
godexsoft: Probably worthy of a `[[nodiscard]]` 👍

mathbunnyru (approved):

PR #2761, opened 2025-11-07
feat: Read and write LedgerCache to file

### LINKED ISSUE #2413: Implement Cache Writing/Reading from Disk

kuznetsss:
Fixes #2413.


pre-commit-hooks/json_in_cpp.py
```diff
@@ -44,7 +44,13 @@ def replace_json(match):
 
 
 def fix_indentation(cpp_content: str) -> str:
+    # Handle edge case: if content doesn't contain JSON patterns, return unchanged
```
kuznetsss: I will move this and clang-tidy change to a separate commit

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,205 @@
...
+        if (not file.read(header)) {
+            return std::unexpected{"Error reading cache header"};
+        }
+        if (header.version != kVERSION) {
+            return std::unexpected{
+                fmt::format("Cache has wrong version: expected {} found {}", kVERSION, header.version)
+            };
+        }
+        result.latestSeq = header.latestSeq;
+        // TODO: check datetime or remove it. Maybe check the latestSeq
```
kuznetsss: Datetime could be removed but need to add sequence verification somewhere to ensure the cache is not too old.

src/app/ClioApplication.cpp
```diff
@@ -110,7 +111,23 @@ ClioApplication::run(bool const useNgWebServer)
     auto const dosguardWeights = web::dosguard::Weights::make(config_);
     auto dosGuard = web::dosguard::DOSGuard{config_, whitelistHandler, dosguardWeights};
     auto sweepHandler = web::dosguard::IntervalSweepHandler{config_, ioc, dosGuard};
+
     auto cache = data::LedgerCache{};
+    appStopper_.setOnStop([&cache, this](auto&&) {
+        // TODO(kuznetsss): move this into Stopper::makeOnStopCallback()
```
kuznetsss, line 117: This will be fixed in a separate PR

pre-commit-hooks/json_in_cpp.py
kuznetsss: Created #2766

src/data/impl/LedgerCacheFile.cpp
kuznetsss: Fixed.

docs/config-description.md
```diff
@@ -441,6 +441,22 @@ This document provides a list of all available Clio configuration properties in
...
 
+### cache.file.path
+
+- **Required**: False
+- **Type**: string
+- **Default value**: None
+- **Constraints**: None
+- **Description**: The path to a file where cache will be saved to on shutdown and loaded from on startup. If the file couldn't be read Clio will load cache as usual (from DB or from rippled).
+
+### cache.file.max_sequence_lag
```
godexsoft: nit: Maybe we call it `max_age` or something? while age is not the most descriptive name it is what we use elsewhere when referring to the same thing

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+using Separator = std::array<char, 16>;
+static constexpr Separator kSEPARATOR = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
+
+namespace {
+
+std::expected<std::pair<ripple::uint256, LedgerCache::CacheEntry>, std::string>
+readCacheEntry(InputFile& file, size_t i)
+{
+    ripple::uint256 key;
+    if (not file.readRaw(reinterpret_cast<char*>(key.data()), ripple::base_uint<256>::bytes)) {
```
mathbunnyru, line 49: Can't we use `.read` here?

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+        return std::unexpected(fmt::format("Failed to read sequence at index {}", i));
+    }
+
+    size_t blobSize{};
+    if (not file.read(blobSize)) {
+        return std::unexpected(fmt::format("Failed to read blob size at index {}", i));
+    }
+
+    Blob blob;
+    blob.resize(blobSize);
```
mathbunnyru: Maybe Blob has a ctor which takes size?

tests/common/util/TmpFile.hpp
```diff
@@ -63,5 +66,10 @@ struct TmpFile {
     {
         if (not path.empty())
             std::filesystem::remove(path);
+
+        // Some code, e.g. LedgerCacheFile may add prefix .new to the path
+        auto const pathWithNewPrefix = fmt::format("{}.new", path);
```
godexsoft: This does not feel like the right place for this as we did not create this filepath from this class.. wdyt?

tests/unit/data/LedgerCacheTests.cpp
```diff
@@ -18,14 +18,41 @@
...
+    explicit TimeMeasurement(std::string op) : operation(std::move(op))
+    {
+        start = std::chrono::high_resolution_clock::now();
+    }
+
+    ~TimeMeasurement()
+    {
+        auto end = std::chrono::high_resolution_clock::now();
+        auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+        std::cout << operation << " took: " << duration.count() << " ms" << std::endl;
```
godexsoft: Is `cout` desired here?

godexsoft (commented):
Left a few suggestions

docs/config-description.md
```diff
@@ -441,6 +441,22 @@ This document provides a list of all available Clio configuration properties in
 - **Constraints**: The value must be one of the following: `sync`, `async`, `none`.
 - **Description**: The strategy used for Cache loading.
 
+### cache.file.path
+
+- **Required**: False
+- **Type**: string
+- **Default value**: None
+- **Constraints**: None
+- **Description**: The path to a file where cache will be saved to on shutdown and loaded from on startup. If the file couldn't be read Clio will load cache as usual (from DB or from rippled).
```
mathbunnyru, line 450:
```suggestion
- **Description**: The path to a file where cache will be loaded from on startup and saved to on shutdown. If the file couldn't be read Clio will load cache as usual (from DB or from rippled).
```

src/etl/CacheLoaderSettings.hpp
```diff
@@ -40,6 +43,9 @@ struct CacheLoaderSettings {
     size_t numCacheCursorsFromAccount = 0; /**< number of cursors to fetch from account_tx */
 
     LoadStyle loadStyle = LoadStyle::ASYNC; /**< how to load the cache */
+    std::optional<std::string> cacheFilePath =
+        std::nullopt; /**< optional path to the file to load cache from on start and save cache to on shutdown */
+    uint32_t cacheFileMaxLag = 5000; /**< max difference between latest sequence in cache file and DB */
```
mathbunnyru: I think a bit better design would be to add a struct for cache params, and make it optional.

This would better reflect that these params are connected, and it would also mimic the json

src/etl/CacheLoader.hpp
```diff
@@ -98,6 +102,10 @@ class CacheLoader : public CacheLoaderInterface {
             return;
         }
 
+        if (loadCacheFromFile()) {
+            return;
+        }
+
         std::shared_ptr<impl::BaseCursorProvider> provider;
```
mathbunnyru, line 109: Let's put everything below in a separate private method as well?

src/etl/CacheLoader.hpp
```diff
@@ -149,6 +157,35 @@ class CacheLoader : public CacheLoaderInterface {
...
             loader_->wait();
     }
+
+private:
+    bool
+    loadCacheFromFile()
+    {
+        if (not settings_.cacheFilePath.has_value()) {
+            return false;
+        }
```
mathbunnyru: maybe let's check this if outside?
Then the function will only return `false` if it actually failed (maybe even worth returning `std::expected<void, void>`)

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+#include <exception>
+#include <filesystem>
+#include <string>
+#include <utility>
+
+namespace data::impl {
+
+using Hash = ripple::uint256;
+using Separator = std::array<char, 16>;
+static constexpr Separator kSEPARATOR = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
```
mathbunnyru, line 41: We calculate hash, so I don't think this separator gives much more safety for broken files

src/data/impl/InputFile.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    return file_.is_open();
+}
+
+bool
+InputFile::readRaw(char* data, size_t size)
+{
+    file_.read(data, size);
+    shasum_.update(data, size);
+    return not file_.fail();
```
mathbunnyru, line 48: I think we should actually use exceptions here.
It's an exceptional situation where we can't read a file, which we wrote ourselves.

And it's only done once on startup, so there will be no performance overhead at all, but the code will be much more simple.

Almost everywhere instead of
```
Sth sth;
if (not file.read(sth)) {
   handle_error()
}
```

it will be just:
```
auto sth = file.read<Sth>()
```

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+        .latestSeq = dataView.latestSeq, .mapSize = dataView.map.size(), .deletedSize = dataView.deleted.size()
+    };
+    file.write(header);
+    file.write(kSEPARATOR);
+
+    for (auto const& [k, v] : dataView.map) {
+        file.write(k.data(), decltype(k)::bytes);
+        file.write(v.seq);
+        file.write(v.blob.size());
+        file.writeRaw(reinterpret_cast<char const*>(v.blob.data()), v.blob.size());
```
mathbunnyru, line 104: I think a better design would be to have one more class like `FileWriter` and `FileReader`, which supports needed types natively.

They can even support big structures like `dataView.map`

Now the logic how to read and write types is duplicated, and there are lots of `reinterpret_cast`s, we can get rid of that - for example the code for `dataView.deleted` is the same as for `dataView.map`

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+        if (auto verificationResult = verifySeparator(separator); not verificationResult.has_value()) {
+            return std::unexpected{std::move(verificationResult).error()};
+        }
+
+        for (size_t i = 0; i < header.mapSize; ++i) {
+            auto cacheEntryExpected = readCacheEntry(file, i);
+            if (not cacheEntryExpected.has_value()) {
+                return std::unexpected{std::move(cacheEntryExpected).error()};
+            }
+            result.map.insert(result.map.end(), std::move(cacheEntryExpected).value());
```
mathbunnyru, line 167: This is not obvious, why it's not simple assignment, let's add a comment

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -0,0 +1,209 @@
...
+            }
+            result.deleted.insert(result.deleted.end(), std::move(cacheEntryExpected).value());
+        }
+
+        if (not file.readRaw(separator.data(), separator.size())) {
+            return std::unexpected{"Error reading separator"};
+        }
+        if (auto verificationResult = verifySeparator(separator); not verificationResult.has_value()) {
+            return std::unexpected{std::move(verificationResult).error()};
+        }
```
mathbunnyru, line 190: This also duplicates statement above

src/data/impl/LedgerCacheFile.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace data::impl {
+
+class LedgerCacheFile {
+public:
+    struct Header {
+        uint32_t version = kVERSION;
+        uint32_t latestSeq{};
+        uint64_t mapSize{};
+        uint64_t deletedSize{};
```
mathbunnyru, line 42: Do they need to be in a header?
I think it's fine to have them before maps

mathbunnyru (changes requested):

docs/config-description.md
kuznetsss: What is the difference?

src/data/impl/InputFile.cpp
kuznetsss: I think exceptions add invisible interface to a function which could be found only after reading either implementation or documentation. In my opinion we should avoid exceptions as much as possible.

src/data/impl/LedgerCacheFile.cpp
kuznetsss: I agree, but I wanted to add some extra checks to make sure file is not corrupted. The overhead is only 48 bytes, so I don't think it makes any difference.

src/data/impl/LedgerCacheFile.cpp
kuznetsss: No because `ripple::uint256` is a class with an internal array.

docs/config-description.md
mathbunnyru: Well, you first start the application, and then shut it down, so makes sense to write in this order :)

src/data/impl/LedgerCacheFile.cpp
kuznetsss: I like that currently the structure of the output file is defined by only this function. Splitting the logic into multiple files will make the output file structure less clear.

src/data/impl/LedgerCacheFile.cpp
kuznetsss: I don't see how this duplication could be reduced.

src/data/impl/LedgerCacheFile.hpp
kuznetsss: If you mean without struct, I think having a struct makes the code and the structure of the file more clean and logically separated.

src/etl/CacheLoader.hpp
kuznetsss: Why?

src/etl/CacheLoader.hpp
kuznetsss: I don't think it will make it any better. This function has functionality to load cache from file and it returns the outcome of loading operation: whether the cache was loaded or not.

docs/config-description.md
kuznetsss: First save cache on shutdown and then load it on the next start. It is not possible to load cache without having a file :)

mathbunnyru: Ok, makes sense, let's keep it this way :)

src/data/impl/LedgerCacheFile.cpp
mathbunnyru: And, why is it a problem?

src/etl/CacheLoader.hpp
mathbunnyru: It will be more symmetrical - one method to read from the file, and one from db.

src/data/impl/LedgerCacheFile.cpp
mathbunnyru: That's what abstractions are for :)
I don't think you can avoid code duplication another way.

src/data/impl/LedgerCacheFile.cpp
mathbunnyru: By having `FileWriter` and `FileReader`

src/data/impl/LedgerCacheFile.cpp
mathbunnyru: Another option is not to try to implement it (de)serialization manually at all, and use a library
https://github.com/felixguendling/cista looks like a good choice

src/data/impl/LedgerCacheFile.hpp
mathbunnyru: Yes, but at the same time, it's unclear why you need to store the map size and content in different places.
I don't think you need to access sizes before reading maps.

I thought that usually you first store the size, then the content.
And in your case, it's going to be size1, size2, content1, content2.

src/data/impl/LedgerCacheFile.cpp
kuznetsss: There could be any other private fields there or implementation may change to move data on the heap so deserialization will break. I don't think it is a good idea to rely on implementation of something instead of public methods.

src/etl/CacheLoader.hpp
kuznetsss: I think it is out of scope of this PR

src/data/impl/LedgerCacheFile.cpp
kuznetsss: I've looked at cista before implementing this feature. It doesn't support ordered map.

src/data/impl/LedgerCacheFile.hpp
kuznetsss: I don't think the order makes any difference.

src/data/impl/LedgerCacheFile.cpp
```diff
@@ -163,6 +162,8 @@ LedgerCacheFile::read(uint32_t minLatestSequence)
             if (not cacheEntryExpected.has_value()) {
                 return std::unexpected{std::move(cacheEntryExpected).error()};
             }
+            // Using insert with hint here to dectrease insert operation complexity to the amortized constant instead of
```
godexsoft: typo: "decrease"

kuznetsss: Interesting why cspell allowed this

mathbunnyru (approved):

src/data/impl/InputFile.cpp
```diff
@@ -0,0 +1,58 @@
...
+bool
+InputFile::isOpen() const
+{
+    return file_.is_open();
+}
+
+bool
+InputFile::readRaw(char* data, size_t size)
+{
+    file_.read(data, size);
```
emreariyurek, line 46: null check before reading?

kuznetsss: It will crash anyway if `nullptr` is passed here.  Or do you mean something else?

PR #2771, opened 2025-11-11
chore: Change default `max_queue_size` to 1000

godexsoft:
This PR fixes the value of `max_queue_size`. It was 0 by default before we accidentally changed it to 1. 1 does not make sense because that makes the work queue essentially single-threaded. 0 is too permissive as well.

### DIFF
--- a/docs/config-description.md
+++ b/docs/config-description.md
@@ -293,7 +293,7 @@ This document provides a list of all available Clio configuration properties in 
 
 - **Required**: True
 - **Type**: int
-- **Default value**: `1`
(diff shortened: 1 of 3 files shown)


mathbunnyru:
Needs updating the cpp file with the same description

godexsoft:
> Needs updating the cpp file with the same description

Not sure i understand what you want to update there. The description is still correct as it is. The incorrect portion was in the example config only. Let me know 👍

mathbunnyru (approved):

PR #2774, opened 2025-11-12
chore: Use ucontext with ASAN 

godexsoft:
This forces usage of ucontext with ASAN.


CMakeLists.txt
```diff
@@ -76,8 +76,13 @@ if (san)
...
   target_compile_definitions(
-    clio_options INTERFACE $<$<STREQUAL:${san},address>:SANITIZER=ASAN> $<$<STREQUAL:${san},thread>:SANITIZER=TSAN>
-                           $<$<STREQUAL:${san},memory>:SANITIZER=MSAN> $<$<STREQUAL:${san},undefined>:SANITIZER=UBSAN>
+    clio_options
+    INTERFACE $<$<STREQUAL:${san},address>:SANITIZER=ASAN
+              BOOST_USE_ASAN=1
+              BOOST_USE_UCONTEXT=1>
+              $<$<STREQUAL:${san},thread>:SANITIZER=TSAN>
+              $<$<STREQUAL:${san},memory>:SANITIZER=MSAN>
+              $<$<STREQUAL:${san},undefined>:SANITIZER=UBSAN>
```
mathbunnyru: I've searched the source code, and we never use `SANITIZER`.
Let's remove this, and then under if for san==address add 2 definitions

mathbunnyru (changes requested):

CMakeLists.txt
godexsoft: From what i found it is copied from rippled and in rippled they use it in BuildInfo. In our case should be safe to remove i guess. let's try then.

mathbunnyru (approved):

PR #2775, opened 2025-11-12
ci: Force ucontext in ASAN builds

godexsoft:
These are the changes to docker CI image. Tested through #2774

### DIFF
--- a/docker/ci/conan/sanitizer_template.profile
+++ b/docker/ci/conan/sanitizer_template.profile
@@ -1,7 +1,10 @@
 {% set compiler, sani = profile_name.split('.') %}
 
 {% set sanitizer_opt_map = {"asan": "address", "tsan": "thread", "ubsan": "undefined"} %}
+{% set sanitizer_b2_flags_map = {"asan": "context-impl=ucontext address-sanitizer=on", "tsan": "thread-sanitizer=on", "ubsan": "undefined-sanitizer=on"} %}
+
 {% set sanitizer = sanitizer_opt_map[sani] %}
+{% set sanitizer_b2_flags_str = sanitizer_b2_flags_map[sani] %}
 
 {% set sanitizer_build_flags_str = "-fsanitize=" ~ sanitizer ~ " -g -O1 -fno-omit-frame-pointer" %}
 {% set sanitizer_build_flags = sanitizer_build_flags_str.split(' ') %}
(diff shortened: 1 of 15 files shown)


docker/ci/conan/sanitizer_template.profile
```diff
@@ -1,7 +1,10 @@
 {% set compiler, sani = profile_name.split('.') %}
 
 {% set sanitizer_opt_map = {"asan": "address", "tsan": "thread", "ubsan": "undefined"} %}
+{% set sanitizer_b2_flags_map = {"asan": "context-impl=ucontext address-sanitizer=on", "tsan": "thread-sanitizer=on", "ubsan": "undefined-sanitizer=on"} %}
+
 {% set sanitizer = sanitizer_opt_map[sani] %}
+{% set sanitizer_b2_flags_str = sanitizer_b2_flags_map[sani] %}
```
mathbunnyru:
```suggestion
{% set sanitizer = sanitizer_opt_map[sani] %}

{% set sanitizer_b2_flags_map = {"address": "context-impl=ucontext address-sanitizer=on", "thread": "thread-sanitizer=on", "undefined": "undefined-sanitizer=on"} %}
{% set sanitizer_b2_flags_str = sanitizer_b2_flags_map[sanitizer] %}
```

I probably wouldn't use `sani` here
I had to use it to not change conan profile names already created, but in the implementation we don't need it

godexsoft: I don't mind either way. initially left it with sani so that the map keys are the same but both versions look ok to me

kuznetsss:
I think it should be `chore` or `ci`. `feat` is usually for a new feature in Clio itself.

mathbunnyru (dismissed):

docker/ci/conan/sanitizer_template.profile
```diff
@@ -3,6 +3,9 @@
 {% set sanitizer_opt_map = {"asan": "address", "tsan": "thread", "ubsan": "undefined"} %}
 {% set sanitizer = sanitizer_opt_map[sani] %}
 
+{% set sanitizer_b2_flags_map = {"address": "define=BOOST_USE_ASAN=1 context-impl=ucontext address-sanitizer=on", "thread": "thread-sanitizer=on", "undefined": "undefined-sanitizer=on"} %}
```
godexsoft, line 6: Turns out adding this definition fixes the remaining false positives on both clang and gcc

mathbunnyru: That's great! Do we maybe need to add definition for ucontext as well?

godexsoft: Does not look like we need to - i think the context-impl parameter takes care of that. Everything seems to work without issues with the current setup. We can change later if we still find some issues but i tried locally on docker with both clang and gcc and gtest_repeat=-1 for a long time - it does not seem to fail 👍

mathbunnyru: Ok, let's keep it like this. Its great you found a way to not have false positives at all 👍👍👍

mathbunnyru (approved):

PR #2780, opened 2025-11-13
ci: Improve cache implementation

mathbunnyru:


.github/workflows/build.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       container: '{ "image": "ghcr.io/xrplf/clio-ci:62369411404eb32b0140603a785ff05e1dc36ce8" }'
       conan_profile: gcc
       build_type: Debug
-      download_ccache: true
+      download_ccache: false
```
kuznetsss: Why don't we need cache for code coverage?

mathbunnyru: You can see on the next line, upload_ccache is (and was) false, so it was never actually uploaded

AI told me that running the coverage creates `.gcno` files, which are not cached by ccache.

The part about having these files is true, but it looks like ccache is aware of these files, so let's try to enable coverage ccache.

mathbunnyru: Reverted deleting coverage parts and enabled upload - let's try if it works

.github/actions/cache-key/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    description: Conan profile name
+    required: true
+  build_type:
+    description: Current build type (e.g. Release, Debug)
+    required: true
+    default: Release
+  code_coverage:
+    description: Whether code coverage is on
+    required: true
+    default: "false"
```
godexsoft, line 15: Does this have to be a string?

godexsoft (approved):
:rage1:

.github/actions/cache-key/action.yml
mathbunnyru: Yes, input params in actions are always treated as strings, unfortunately:(

PR #2784, opened 2025-11-14
chore: Add defines for asan/tsan to conan profile

godexsoft:
We found out that building TSAN with ucontext helps to uncover lots of issues. Many of them are probably solved by one fix but either way to run TSAN in CI this change should be necessary.
Also defining flags right in the profile will allow us to clean up our CMakeLists.txt with regards to sanitizers.
`norecover` setting suggested by @mathbunnyru - this will force the suite to stop once any tsan/asan error is detected.


mathbunnyru (dismissed):
🥇

mathbunnyru (dismissed):
🥇

mathbunnyru (approved):

PR #2788, opened 2025-11-15
refactor: Make getLedgerIndex return std::expected instead of throwing

### LINKED ISSUE #2767: Make `getLedgerIndex` not throw exceptions and return `std::expected`
Discussion: https://github.com/XRPLF/clio/pull/2755#discussion_r2494150092

emreariyurek:
Fix https://github.com/XRPLF/clio/issues/2767


src/rpc/handlers/AccountChannels.cpp
```diff
@@ -154,8 +154,11 @@ tag_invoke(boost::json::value_to_tag<AccountChannelsHandler::Input>, boost::json
     if (jsonObject.contains(JS(destination_account)))
         input.destinationAccount = boost::json::value_to<std::string>(jv.at(JS(destination_account)));
 
-    if (jsonObject.contains(JS(ledger_index)))
-        input.ledgerIndex = util::getLedgerIndex(jv.at(JS(ledger_index)));
+    if (jsonObject.contains(JS(ledger_index))) {
+        auto const expectedLedgerIndex = util::getLedgerIndex(jv.at(JS(ledger_index)));
+        if (expectedLedgerIndex.has_value())
+            input.ledgerIndex = expectedLedgerIndex.value();
```
mathbunnyru:
```suggestion
            input.ledgerIndex = *expectedLedgerIndex;
```

We're under `if`, so we can safely use faster operation

src/rpc/handlers/AccountChannels.cpp
```diff
@@ -154,8 +154,11 @@ tag_invoke(boost::json::value_to_tag<AccountChannelsHandler::Input>, boost::json
     if (jsonObject.contains(JS(destination_account)))
         input.destinationAccount = boost::json::value_to<std::string>(jv.at(JS(destination_account)));
 
-    if (jsonObject.contains(JS(ledger_index)))
-        input.ledgerIndex = util::getLedgerIndex(jv.at(JS(ledger_index)));
+    if (jsonObject.contains(JS(ledger_index))) {
+        auto const expectedLedgerIndex = util::getLedgerIndex(jv.at(JS(ledger_index)));
+        if (expectedLedgerIndex.has_value())
```
mathbunnyru, line 159:
```suggestion
        if (auto const expectedLedgerIndex = util::getLedgerIndex(jv.at(JS(ledger_index))); expectedLedgerIndex.has_value())
```

This can be written as one-line.
But please, check, if after clang-format it's going to be 2 lines, it's not worth it

src/util/JsonUtils.hpp
```diff
@@ -116,22 +117,23 @@ integralValueAs(boost::json::value const& value)
...
+            return util::integralValueAs<uint32_t>(value);
+        else if (value.as_string() != "validated")
+            return std::stoi(value.as_string().c_str());
+        else
+            return std::unexpected("'validated' ledger index is requested");
+    } catch (std::exception const& ex) {
+        return std::unexpected(ex.what());
+    } catch (...) {
+        return std::unexpected("Unknown error in getLedgerIndex");
     }
```
mathbunnyru: This is still dealing with exceptions based on user-provided input, which we would like to avoid.

I suggest adding `tryIntegralValueAs` function which will return `std::expected`
And instead of `std::stoi` using `std::from_chars`

mathbunnyru (changes requested):

src/util/JsonUtils.hpp
```diff
@@ -113,6 +114,27 @@ integralValueAs(boost::json::value const& value)
...
 
+/**
+ * @brief Detects the type of number stored in value and casts it back to the requested Type.
+ * @note This conversion can possibly cause wrapping around or UB. Use with caution.
+ *
+ * @tparam Type The type to cast to
+ * @param value The JSON value to cast
+ * @return Value casted to the requested type or an error message
+ */
+template <std::integral Type>
```
mathbunnyru, line 126: Nice usage of concepts 👍

src/rpc/handlers/AccountChannels.cpp
emreariyurek: It takes 2 lines

src/util/JsonUtils.hpp
```diff
@@ -112,26 +114,53 @@ integralValueAs(boost::json::value const& value)
...
+        auto last = value.as_string().data() + value.as_string().size();
+        if (first != last && *first == '+')
+            ++first;
+
+        uint32_t ledgerIndex{};
+        auto const ret = std::from_chars(first, last, ledgerIndex);
+        if (ret.ec == std::errc() && ret.ptr == last)
+            return ledgerIndex;
+        else
+            return std::unexpected("Invalid ledger index string");
```
mathbunnyru: Maybe we should just use `LexicalCast` from rippled?
It does the same thing, and we will be following their behaviour

mathbunnyru (changes requested):

mathbunnyru (dismissed):

src/util/JsonUtils.hpp
```diff
@@ -112,26 +115,47 @@ integralValueAs(boost::json::value const& value)
     throw std::logic_error("Value neither uint64 nor int64");
```
kuznetsss: Could you please use `tryIntegralValueAs()` here and throw exception if it returned unexpected?

kuznetsss (changes requested):

godexsoft (dismissed):
Nice 👍

src/util/JsonUtils.hpp
emreariyurek: done

mathbunnyru (dismissed):

src/util/JsonUtils.hpp
```diff
@@ -121,19 +120,18 @@ integralValueAs(boost::json::value const& value)
...
-        return static_cast<Type>(value.as_uint64());
+    auto const expectedResult = tryIntegralValueAs<Type>(value);
+    if (expectedResult.has_value())
+        return *expectedResult;
 
-    if (value.is_int64())
-        return static_cast<Type>(value.as_int64());
-
-    return std::unexpected("Value neither uint64 nor int64");
+    throw std::logic_error("Value neither uint64 nor int64");
```
kuznetsss: Let's reuse error message:
```suggestion
    auto expectedResult = tryIntegralValueAs<Type>(value);
    if (expectedResult.has_value())
        return *expectedResult;

    throw std::logic_error(std::move(expectedResult).value());
```

kuznetsss (dismissed):
Thanks!

kuznetsss (approved):

PR #2790, opened 2025-11-18
feat: Handle prometheus requests in WorkQueue

mathbunnyru:


src/app/ClioApplication.cpp
```diff
@@ -194,7 +194,7 @@ ClioApplication::run(bool const useNgWebServer)
             return EXIT_FAILURE;
         }
 
-        httpServer->onGet("/metrics", MetricsHandler{adminVerifier});
+        httpServer->onGet("/metrics", MetricsHandler{adminVerifier, workQueue});
         httpServer->onGet("/health", HealthCheckHandler{});
         httpServer->onGet("/cache_state", CacheStateHandler{cache});
```
kuznetsss, line 187: Probably all of these handles should post processing to WorkQueue with high priority

src/app/WebHandlers.cpp
```diff
@@ -86,19 +89,41 @@ MetricsHandler::operator()(
...
+            // notify the coroutine group that the foreign task is done
+            onTaskComplete();
+        },
+        /* isWhiteListed= */ true,
+        rpc::WorkQueue::Priority::High
     );
-    ASSERT(maybeResponse.has_value(), "Got unexpected request for Prometheus");
-    return web::ng::Response{std::move(maybeResponse).value(), request};
+
+    ASSERT(postSuccessful, "Posting Prometheus request to WorkQueue failed");
```
kuznetsss: This could crash Clio when work queue is full or is stopping.

src/app/ClioApplication.cpp
godexsoft: Seems like out of scope of this PR. Let's create new issue for the rest please 👍

kuznetsss: Sure, but I think it affects the design of the changes in the current PR: we will need some abstraction for requests handler which takes business logic, wraps it into a task and pushes it into `WorkQueue` with high priority.

kuznetsss: Thinking about it a bit more, it looks like I was wrong in my original comment. `/health` and `/cache_state` handlers are tiny, they almost don't have any work so using `WorkQueue` for them will slow down processing.

src/app/WebHandlers.cpp
mathbunnyru: Good call, returning `too_many_requests` now

src/app/WebHandlers.cpp
```diff
@@ -86,19 +90,45 @@ MetricsHandler::operator()(
...
+            boost::asio::yield_context
+        ) mutable {
+            auto const maybeHttpRequest = request.asHttpRequest();
+            ASSERT(maybeHttpRequest.has_value(), "Got not a http request in Get");
+            auto const& httpRequest = maybeHttpRequest->get();
+
+            auto maybeResponse = util::prometheus::handlePrometheusRequest(
+                httpRequest, adminVerifier_->isAdmin(httpRequest, connectionMetadata.ip())
+            );
+            ASSERT(maybeResponse.has_value(), "Got unexpected request for Prometheus");
```
kuznetsss, line 112: This also may crash Clio if user is not admin for example

kuznetsss: No, sorry, I'm wrong, this assert should never be triggered.

mathbunnyru: This assert has been present before, you sure?

godexsoft (approved):
:rage2:

kuznetsss (approved):

PR #2792, opened 2025-11-19
ci: Don't download ccache on develop branch

mathbunnyru:
We don't want to download it on develop, otherwise:
- it will constantly grow
- we actually want to check in develop that clio builds fine from source


godexsoft (approved):
:rage1:

PR #2793, opened 2025-11-19
ci: Always upload cache on develop

mathbunnyru:


.github/workflows/reusable-build.yml
```diff
@@ -160,17 +159,14 @@ jobs:
...
         run: |
-          ccache -s > /tmp/ccache.stats
-          miss_rate=$(cat /tmp/ccache.stats | grep 'Misses' | head -n1 | sed 's/.*(\(.*\)%).*/\1/')
-          echo "miss_rate=${miss_rate}" >> $GITHUB_OUTPUT
-          cat /tmp/ccache.stats
+          ccache --show-stats
+          ccache --zero-stats
 
       - name: Save ccache cache
-        if: ${{ inputs.upload_ccache && github.ref == 'refs/heads/develop' && (steps.restore_cache.outputs.cache-hit != 'true' || steps.ccache_stats.outputs.miss_rate == '100.0') }}
```
mathbunnyru, line 173: I don't think this is needed, GitHub handles cache eviction for you.

.github/workflows/reusable-build.yml
```diff
@@ -160,17 +159,14 @@ jobs:
...
+      - name: Show ccache's statistics and zero it
         if: ${{ inputs.download_ccache }}
-        id: ccache_stats
         run: |
-          ccache -s > /tmp/ccache.stats
-          miss_rate=$(cat /tmp/ccache.stats | grep 'Misses' | head -n1 | sed 's/.*(\(.*\)%).*/\1/')
-          echo "miss_rate=${miss_rate}" >> $GITHUB_OUTPUT
-          cat /tmp/ccache.stats
+          ccache --show-stats
+          ccache --zero-stats
```
mathbunnyru, line 166: This is needed, otherwise next time when we download cache, it will include stats from the time when cache was actually saved, which makes it much more difficult to reason how many hits there were actually

godexsoft (approved):
:rage2:

PR #2802, opened 2025-11-21
chore: Try less delay in taskman

godexsoft:
There was a report that ETL became somewhat slower to pick up a new ledger. The only thing i could find that may impact this is task manager's 100ms delay between attempts to take the next sequence to extract. If there is no next ledger yet we may wait 100ms to try again. This seems like it can pile up and become 200-300 ms of delay. Let's try with 10ms and see if it changes anything. If it does, it will confirm that this is the problematic place and we can think of better solutions for it.


godexsoft:
So this did help and it confirms the general location of the fix to come.

godexsoft:
Let's merge this so that the current ETL works as expected. We will change the approach to a more reactive one in #2838

src/etl/impl/TaskManager.cpp
```diff
@@ -87,8 +87,8 @@ TaskManager::run(std::size_t numExtractors)
 util::async::AnyOperation<void>
 TaskManager::spawnExtractor(TaskQueue& queue)
 {
-    // TODO: these values may be extracted to config later and/or need to be fine-tuned on a realistic system
-    static constexpr auto kDELAY_BETWEEN_ATTEMPTS = std::chrono::milliseconds{100u};
+    // TODO 2838: the approach should be changed to a reactive one instead
```
kuznetsss: Nit: maybe link to the issue would be better.

kuznetsss (dismissed):

kuznetsss (approved):

PR #2803, opened 2025-11-21
ci: Run clang-tidy 3 times to make sure we don't have to fix again

mathbunnyru:
There is a way to run clang-tidy on diff (or only some files), but it won't work when changes are in header files, so we have to run it again.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -66,6 +66,7 @@ jobs:
         continue-on-error: true
         id: clang_tidy
         run: |
+          # We run clang-tidy several times, because some fixes may enable new fixes in subsequent runs.
           CLANG_TIDY_COMMAND="run-clang-tidy-${{ env.LLVM_TOOLS_VERSION }} -p build -j ${{ steps.nproc.outputs.nproc }} -fix -quiet"
           ${CLANG_TIDY_COMMAND} ||
           ${CLANG_TIDY_COMMAND} ||
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -64,6 +64,7 @@ jobs:
 
       - name: Run clang-tidy (several times)
         continue-on-error: true
+        id: clang_tidy
         run: |
           CLANG_TIDY_COMMAND="run-clang-tidy-${{ env.LLVM_TOOLS_VERSION }} -p build -j ${{ steps.nproc.outputs.nproc }} -fix -quiet"
           ${CLANG_TIDY_COMMAND} ||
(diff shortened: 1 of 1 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -62,27 +62,28 @@ jobs:
...
-      - name: Print issues found
-        if: ${{ steps.run_clang_tidy.outcome != 'success' }}
-        run: |
-          sed -i '/error\||/!d' ./output.txt
-          cat output.txt
-          rm output.txt
-
       - name: Create an issue
-        if: ${{ steps.run_clang_tidy.outcome != 'success' && github.event_name != 'pull_request' }}
+        if: ${{ steps.files_changed.outcome != 'success' && github.event_name != 'pull_request' }}
```
kuznetsss: We should always create an issue. Even if clang-tidy couldn't fix it.

mathbunnyru: Ah, dang, I will have capture rc there as well :(

mathbunnyru: Good call

.github/workflows/clang-tidy.yml
```diff
@@ -62,27 +62,28 @@ jobs:
...
-      - name: Run clang-tidy
+      - name: Run clang-tidy (several times)
         continue-on-error: true
-        id: run_clang_tidy
         run: |
-          run-clang-tidy-${{ env.LLVM_TOOLS_VERSION }} -p build -j "${{ steps.nproc.outputs.nproc }}" -fix -quiet 1>output.txt
+          CLANG_TIDY_COMMAND="run-clang-tidy-${{ env.LLVM_TOOLS_VERSION }} -p build -j ${{ steps.nproc.outputs.nproc }} -fix -quiet"
+          ${CLANG_TIDY_COMMAND} ||
+          ${CLANG_TIDY_COMMAND} ||
+          ${CLANG_TIDY_COMMAND}
```
kuznetsss, line 73: Could you explain why do we need to do this? And why exactly 3 times?

mathbunnyru: Take a look at last 3 commits in develop - we were not able to fix it on the first try. And on the second. Going 3 times through approval and merging is not nice, so I would rather let it work at max 1 extra time than have to verify diff again and again.

There is no good number of how many times to run, 3 is a good compromise and we have a real example, that it was needed.

.github/workflows/clang-tidy.yml
mathbunnyru: Done, please, take a look

kuznetsss (approved):

PR #2809, opened 2025-11-24
chore: WebServerAdminTestsSuit TSAN issues

godexsoft:
These are issues caused by the test itself. Keep in mind that the old webserver will be removed along with those tests sometime in the future.

### DIFF
--- a/tests/unit/web/ServerTests.cpp
+++ b/tests/unit/web/ServerTests.cpp
@@ -221,12 +221,6 @@ public:
 
 namespace {
 
-struct SyncData {
-    std::mutex mtx;
-    std::condition_variable cv;
-    bool ready = false;
-};
-
 template <class Executor>
 std::shared_ptr<web::HttpServer<Executor>>
 makeServerSync(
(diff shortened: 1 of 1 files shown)


tests/unit/web/ServerTests.cpp
```diff
@@ -232,22 +238,21 @@ makeServerSync(
...
-    boost::asio::dispatch(ioc.get_executor(), [&]() mutable {
-        server = web::makeHttpServer(config, ioc, dosGuard, handler, cache);
-        {
-            std::lock_guard const lk(m);
-            ready = true;
+    auto syncData = std::make_shared<SyncData>();
+
+    boost::asio::dispatch(
+        ioc.get_executor(), [&server, &config, &ioc, &dosGuard, &handler, &cache, syncData]() mutable {
+            server = web::makeHttpServer(config, ioc, dosGuard, handler, cache);
```
kuznetsss: `boost::asio::spawn(..., boost::asio::use_future).wait()` would be more simple.
Also why couldn't we create server synchronously?

godexsoft: Actually, no clue why we did not already have it synchronous but you are right - this whole thing can be simplified to one line 👍

kuznetsss (approved):

tests/unit/web/ServerTests.cpp
```diff
@@ -231,25 +231,7 @@ makeServerSync(
...
-            ready = true;
-        }
-        cv.notify_one();
-    });
-    {
-        std::unique_lock lk(m);
-        cv.wait(lk, [&] { return ready; });
-    }
-    return server;
+    return web::makeHttpServer(config, ioc, dosGuard, handler, cache);
```
mathbunnyru, line 234: Why not remove `makeServerSync` completely and just call `web::makeHttpServer` in this test file?

godexsoft: Yeah, crossed my mind too. I'm going to leave it as is in case someone needs to do something extra in the factory function in the future.

PR #2812, opened 2025-11-25
chore: Use mold to build 3rd party tools

mathbunnyru:


mathbunnyru:
Doesn't seem to help, linking is still slow :(

PR #2820, opened 2025-11-26
feat: Add DynamicMPT in account_mptoken_issuances

yinyiqian1:
- Support DynamicMPT for the  account_mptoken_issuances handler.

Related commit: https://github.com/XRPLF/clio/commit/eed757e0c48fcab17db7693ae830d7644925a6e0

The original spec for `DynamicMPT` can be found here: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT

- fix: added mpt_issuance_id to account_mptoken_issuances and added mpt_id to account_mptokens.

### DIFF
--- a/src/rpc/handlers/AccountMPTokenIssuances.cpp
+++ b/src/rpc/handlers/AccountMPTokenIssuances.cpp
@@ -56,6 +56,7 @@ AccountMPTokenIssuancesHandler::addMPTokenIssuance(
 {
     MPTokenIssuanceResponse issuance;
 
+    issuance.MPTokenIssuanceID = ripple::strHex(sle.key());
     issuance.issuer = ripple::to_string(account);
     issuance.sequence = sle.getFieldU32(ripple::sfSequence);
     auto const flags = sle.getFieldU32(ripple::sfFlags);
@@ -219,6 +220,7 @@ tag_invoke(
 )
 {
     auto obj = boost::json::object{
+        {JS(mpt_issuance_id), issuance.MPTokenIssuanceID},
         {JS(issuer), issuance.issuer},
         {JS(sequence), issuance.sequence},
     };
--- a/src/rpc/handlers/AccountMPTokenIssuances.cpp
+++ b/src/rpc/handlers/AccountMPTokenIssuances.cpp
@@ -73,6 +73,24 @@ AccountMPTokenIssuancesHandler::addMPTokenIssuance(
     setFlag(issuance.mptCanTransfer, ripple::lsfMPTCanTransfer);
     setFlag(issuance.mptCanClawback, ripple::lsfMPTCanClawback);
 
+    if (sle.isFieldPresent(ripple::sfMutableFlags)) {
+        auto const mutableFlags = sle.getFieldU32(ripple::sfMutableFlags);
+
+        auto const setMutableFlag = [&](std::optional<bool>& field, std::uint32_t mask) {
+            if ((mutableFlags & mask) != 0u)
+                field = true;
+        };
+
+        setMutableFlag(issuance.mptCanMutateCanLock, ripple::lsmfMPTCanMutateCanLock);
+        setMutableFlag(issuance.mptCanMutateRequireAuth, ripple::lsmfMPTCanMutateRequireAuth);
+        setMutableFlag(issuance.mptCanMutateCanEscrow, ripple::lsmfMPTCanMutateCanEscrow);
+        setMutableFlag(issuance.mptCanMutateCanTrade, ripple::lsmfMPTCanMutateCanTrade);
+        setMutableFlag(issuance.mptCanMutateCanTransfer, ripple::lsmfMPTCanMutateCanTransfer);
+        setMutableFlag(issuance.mptCanMutateCanClawback, ripple::lsmfMPTCanMutateCanClawback);
+        setMutableFlag(issuance.mptCanMutateMetadata, ripple::lsmfMPTCanMutateMetadata);
+        setMutableFlag(issuance.mptCanMutateTransferFee, ripple::lsmfMPTCanMutateTransferFee);
+    }
+
     if (sle.isFieldPresent(ripple::sfTransferFee))
         issuance.transferFee = sle.getFieldU16(ripple::sfTransferFee);
 
@@ -226,6 +244,15 @@ tag_invoke(
     setIfPresent("mpt_can_trade", issuance.mptCanTrade);
     setIfPresent("mpt_can_transfer", issuance.mptCanTransfer);
     setIfPresent("mpt_can_clawback", issuance.mptCanClawback);
+
+    setIfPresent("mpt_can_mutate_can_lock", issuance.mptCanMutateCanLock);
+    setIfPresent("mpt_can_mutate_require_auth", issuance.mptCanMutateRequireAuth);
+    setIfPresent("mpt_can_mutate_can_escrow", issuance.mptCanMutateCanEscrow);
+    setIfPresent("mpt_can_mutate_can_trade", issuance.mptCanMutateCanTrade);
+    setIfPresent("mpt_can_mutate_can_transfer", issuance.mptCanMutateCanTransfer);
+    setIfPresent("mpt_can_mutate_can_clawback", issuance.mptCanMutateCanClawback);
+    setIfPresent("mpt_can_mutate_metadata", issuance.mptCanMutateMetadata);
+    setIfPresent("mpt_can_mutate_transfer_fee", issuance.mptCanMutateTransferFee);
 
     jv = std::move(obj);
 }
--- a/src/rpc/handlers/AccountMPTokenIssuances.hpp
+++ b/src/rpc/handlers/AccountMPTokenIssuances.hpp
@@ -80,6 +80,15 @@ public:
         std::optional<bool> mptCanTrade;
         std::optional<bool> mptCanTransfer;
         std::optional<bool> mptCanClawback;
+
+        std::optional<bool> mptCanMutateCanLock;
+        std::optional<bool> mptCanMutateRequireAuth;
+        std::optional<bool> mptCanMutateCanEscrow;
+        std::optional<bool> mptCanMutateCanTrade;
+        std::optional<bool> mptCanMutateCanTransfer;
+        std::optional<bool> mptCanMutateCanClawback;
+        std::optional<bool> mptCanMutateMetadata;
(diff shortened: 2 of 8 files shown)


src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -73,6 +73,24 @@ AccountMPTokenIssuancesHandler::addMPTokenIssuance(
     setFlag(issuance.mptCanTransfer, ripple::lsfMPTCanTransfer);
     setFlag(issuance.mptCanClawback, ripple::lsfMPTCanClawback);
 
+    if (sle.isFieldPresent(ripple::sfMutableFlags)) {
+        auto const mutableFlags = sle.getFieldU32(ripple::sfMutableFlags);
+
+        auto const setMutableFlag = [&](std::optional<bool>& field, std::uint32_t mask) {
+            if ((mutableFlags & mask) != 0u)
+                field = true;
```
kuznetsss, line 82: nit:

```suggestion
            field = (mutableFlags & mask) != 0u;
```

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
```diff
@@ -838,3 +839,120 @@ TEST_F(RPCAccountMPTokenIssuancesHandlerTest, EmptyResult)
...
+    auto accountKk = ripple::keylet::account(account).key;
+    auto owneDirKk = ripple::keylet::ownerDir(account).key;
+    ON_CALL(*backend_, doFetchLedgerObject(accountKk, _, _)).WillByDefault(Return(Blob{'f', 'a', 'k', 'e'}));
+
+    ripple::STObject const ownerDir = createOwnerDirLedgerObject(
+        {ripple::uint256{kISSUANCE_INDEX1}, ripple::uint256{kISSUANCE_INDEX2}}, kISSUANCE_INDEX1
+    );
+    ON_CALL(*backend_, doFetchLedgerObject(owneDirKk, _, _)).WillByDefault(Return(ownerDir.getSerializer().peekData()));
+    EXPECT_CALL(*backend_, doFetchLedgerObject).Times(2);
+
```
kuznetsss, line 873: It would be better to do it in one line:
```cpp
EXPECT_CALL(...).Times(2).WillRepeatedly(...);
```

src/rpc/handlers/AccountMPTokenIssuances.cpp
```diff
@@ -73,6 +73,24 @@ AccountMPTokenIssuancesHandler::addMPTokenIssuance(
...
+        };
+
+        setMutableFlag(issuance.mptCanMutateCanLock, ripple::lsmfMPTCanMutateCanLock);
+        setMutableFlag(issuance.mptCanMutateRequireAuth, ripple::lsmfMPTCanMutateRequireAuth);
+        setMutableFlag(issuance.mptCanMutateCanEscrow, ripple::lsmfMPTCanMutateCanEscrow);
+        setMutableFlag(issuance.mptCanMutateCanTrade, ripple::lsmfMPTCanMutateCanTrade);
+        setMutableFlag(issuance.mptCanMutateCanTransfer, ripple::lsmfMPTCanMutateCanTransfer);
+        setMutableFlag(issuance.mptCanMutateCanClawback, ripple::lsmfMPTCanMutateCanClawback);
+        setMutableFlag(issuance.mptCanMutateMetadata, ripple::lsmfMPTCanMutateMetadata);
+        setMutableFlag(issuance.mptCanMutateTransferFee, ripple::lsmfMPTCanMutateTransferFee);
```
kuznetsss, line 92: Maybe it would be good to add a parametrised test to check output for each single flag presented in `mutableFlags`. What do you think?

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
```diff
@@ -838,3 +839,120 @@ TEST_F(RPCAccountMPTokenIssuancesHandlerTest, EmptyResult)
...
+    std::vector<Blob> bbs;
```
godexsoft: nit: This can probably be `const` and use braced initialization. 
Something like this:
```cpp
auto const bbs = std::vector<Blob>{
    createMptIssuanceObject(...).getSerializer().peekData(), 
    ...,
};
```

Or if you prefer you can use an immediately invoked lambda to initialize it.

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
```diff
@@ -838,3 +839,120 @@ TEST_F(RPCAccountMPTokenIssuancesHandlerTest, EmptyResult)
...
+    uint32_t const mutableFlags1 = ripple::lsmfMPTCanMutateCanLock | ripple::lsmfMPTCanMutateRequireAuth |
+        ripple::lsmfMPTCanMutateCanEscrow | ripple::lsmfMPTCanMutateCanTrade;
+
+    uint32_t const mutableFlags2 = ripple::lsmfMPTCanMutateCanTransfer | ripple::lsmfMPTCanMutateCanClawback |
+        ripple::lsmfMPTCanMutateMetadata | ripple::lsmfMPTCanMutateTransferFee;
+
+    auto ledgerHeader = createLedgerHeader(kLEDGER_HASH, 30);
+    EXPECT_CALL(*backend_, fetchLedgerBySequence).WillOnce(Return(ledgerHeader));
+
+    auto account = getAccountIdWithString(kACCOUNT);
```
godexsoft: nit: Let's make everything we can `const` here

godexsoft (changes requested):
I second @kuznetsss comments and add small suggestions on top 👍

src/rpc/handlers/AccountMPTokenIssuances.cpp
yinyiqian1: field is optional, so when it is false, we want to omit it instead of setting it to false.

tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp
yinyiqian1: The whole block of code is:
```
   ON_CALL(*backend_, doFetchLedgerObject(accountKk, _, _)).WillByDefault(Return(Blob{'f', 'a', 'k', 'e'}));
...
    ON_CALL(*backend_, doFetchLedgerObject(owneDirKk, _, _)).WillByDefault(Return(ownerDir.getSerializer().peekData()));
    EXPECT_CALL(*backend_, doFetchLedgerObject).Times(2);
```
it has two different calls, so we can not simply use `WillRepeatedly `

kuznetsss: It would be better to specify each in one line:
```cpp
 EXPECT_CALL(*backend_, doFetchLedgerObject(accountKk, _, _)).WillOnce(Return(Blob{'f', 'a', 'k', 'e'}));
 ...
 EXPECT_CALL(*backend_, doFetchLedgerObject(owneDirKk, _, _)).WillOnce(Return(ownerDir.getSerializer().peekData()));
 ```
 Or use `.Times().WillRepeatedly()` instead of `WillOnce()` if needed.

File: src/rpc/handlers/AccountMPTokenIssuances.cpp (hunk 1 above)
Comment (kuznetsss) [reply src/rpc/handlers/AccountMPTokenIssuances.cpp:82]: You are right, sorry I thought it is just `bool`.

File: src/rpc/handlers/AccountMPTokenIssuances.cpp (hunk 3 above)
Comment (yinyiqian1) [reply src/rpc/handlers/AccountMPTokenIssuances.cpp:92]: Added single flag test for both `Flag` and `MutableFlag`: https://github.com/XRPLF/clio/pull/2820/commits/b3f12e56226255672b4720d64de1a40b699db0b8

File: tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp (hunk 4 above)
Comment (yinyiqian1) [reply tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp]: refactored in https://github.com/XRPLF/clio/pull/2820/commits/b3f12e56226255672b4720d64de1a40b699db0b8

File: tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp (hunk 5 above)
Comment (yinyiqian1) [reply tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp]: modified in:  https://github.com/XRPLF/clio/pull/2820/commits/0f69b27d7b76fe02215eb125ec26bf6c09be8b44

File: tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp (hunk 2 above)
Comment (yinyiqian1) [reply tests/unit/rpc/handlers/AccountMPTokenIssuancesTests.cpp:873]: Updated in: https://github.com/XRPLF/clio/pull/2820/commits/4d825c29cde2d29301a0d16d10bf577599366c1c

**Comment by yinyiqian1** [comment]
@godexsoft @kuznetsss I have a commit https://github.com/XRPLF/clio/pull/2820/commits/e684fde3e8d91e3d86d42998142de7c0d38c6142, which adds the `mpt_issuance_id` and `mpt_id` to the response.
Please let me know if I can just keep the fix in this PR or I need a separate PR.

File: src/rpc/handlers/AccountMPTokenIssuances.hpp [hunk 6]
```diff
@@ -61,6 +61,7 @@ class AccountMPTokenIssuancesHandler {
      * @brief A struct to hold data for one MPTokenIssuance response.
      */
     struct MPTokenIssuanceResponse {
+        std::string MPTokenIssuanceID;
```
godexsoft, line 64: I'm surprised this did not trigger our naming check as this looks like a class/struct name. I think this should be `mptokenIssuanceId` or similar. I guess it is easier to fix these in a separate PR later though as i found other examples already in the codebase 🔢

godexsoft (approved):
Looks good, thanks for fixing the nits 👍 
I noticed a naming issue that already slipped through so left a note about it. We should also check why our tooling is not picking it up because i think it should/could.

godexsoft:
> @godexsoft @kuznetsss I have a commit [e684fde](https://github.com/XRPLF/clio/commit/e684fde3e8d91e3d86d42998142de7c0d38c6142), which adds the `mpt_issuance_id` and `mpt_id` to the response. Please let me know if I can just keep the fix in this PR or I need a separate PR.

It's ok to leave it in this PR imo.

kuznetsss (approved):

PR #2827, opened 2025-12-01
feat: Add delegate filter param's to account_tx RPC

PeterChen13579:
This PR adds the ability to filter account_tx results based on transaction delegation. This is useful for scenarios where one account (the delegatee) signs and submits a transaction on behalf of another account (the delegator).

New Request Parameter:
```
"delegate": {
     "delegate_filter": "actor" or "authorizor",
     "counter_party": "account_address" 
}

```
 
Example: If User A submits a transaction to C on behalf of B:
User B (The account holder) can query with delegate_filter: "actor" to find transactions signed by others (like User A).
User A (The signer) can query with delegate_filter: "authorizer" to find transactions they signed for others (like User B).
The counter_party field can be used in either case to filter by a specific valid address


src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -203,9 +212,63 @@ class CassandraBackendFamily : public BackendInterface {
...
+                LOG(log_.warn()) << "Invalid counterparty account in filter";
+                return {.txns = {}, .cursor = {}};
+            }
+        }
+
+        std::vector<TransactionAndMetadata> resultTxns;
+        if (delegateFilter.has_value()) {
+            resultTxns.reserve(txns.size());
+
+            for (auto& txn : txns) {
```
godexsoft: Looks like a good place to try and use ranges? 🙂

src/rpc/RPCHelpers.cpp
```diff
@@ -1567,4 +1567,48 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
...
 
+std::optional<DelegateFilter::Role>
+parseDelegateType(boost::json::value const& delegateType)
+{
+    if (not delegateType.is_string())
+        return {};
+
+    auto const& type = delegateType.as_string();
+
+    if (type == "delegator")
```
godexsoft, line 1664: Magic strings everywhere 😃 these should either be JS(...) if possible or some sort of constants 🔢

src/rpc/common/Types.hpp
```diff
@@ -194,6 +194,27 @@ struct AccountCursor {
...
 
+/**
+ * @brief A delegate object used filter account_tx by specific delegate accounts
+ */
+struct DelegateFilter {
+    /**
+     * @brief A delegate type used in delegate filter
+     */
+    enum class Role {
+        // This account is the *active* sender, acting on behalf of another party.
```
godexsoft: This could be part of doxygen, perhaps with a trailing comment style

PeterChen13579:
Will wait for Rippled [PR](https://github.com/XRPLF/rippled/pull/6126) to be merged first before merging this one 🫡

src/rpc/filters/impl/DelegateTransactionsFilter.cpp
```diff
@@ -0,0 +1,85 @@
+//------------------------------------------------------------------------------
```
godexsoft: We no longer add these banners

src/rpc/filters/impl/DelegateTransactionsFilter.hpp
```diff
@@ -0,0 +1,54 @@
+//------------------------------------------------------------------------------
```
godexsoft: Same here

src/rpc/filters/TransactionFilter.hpp
```diff
@@ -0,0 +1,54 @@
+//------------------------------------------------------------------------------
```
godexsoft: And here

src/rpc/handlers/AccountTx.cpp
```diff
@@ -298,6 +328,10 @@ tag_invoke(boost::json::value_to_tag<AccountTxHandler::Input>, boost::json::valu
             boost::json::value_to<std::string>(jsonObject.at("tx_type"));
     }
 
+    if (jsonObject.contains(JS(delegate))) {
```
godexsoft: nit: Probably can do without the curlies here?

tests/unit/rpc/filters/impl/DelegateTransactionsFilterTests.cpp
```diff
@@ -0,0 +1,212 @@
+//------------------------------------------------------------------------------
```
godexsoft: This is also not needed

godexsoft (changes requested):
Leaving some small things. Overall PR looks good! 
One more thing that you can try to do is to find and remove/simplify comments that are too verbose or unnecessary - i have seen some on the way through this PR.

src/rpc/RPCHelpers.cpp
```diff
@@ -1653,4 +1653,48 @@ toJsonWithBinaryTx(data::TransactionAndMetadata const& txnPlusMeta, std::uint32_
     return obj;
 }
 
+std::optional<DelegateFilter::Role>
```
mathbunnyru, line 1656: std::expected<DelegateFilter::Role, void> better shows what you're trying to achieve.
Below as well

src/rpc/filters/impl/DelegateTransactionsFilter.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    // Filter by "authorizer" ie. the queried account is the actor (signer) and the user wants to
+    // find the authorizer (owner) it acted for.
+    if (delegateFilter_.delegateType == rpc::DelegateFilter::Role::Authorizer) {
+        if (*txDelegate == queriedAccount_) {
+            if (!counterparty_ || *counterparty_ == txAccount)
+                return {.shouldInclude = true, .relevantAccount = txAccount};
+        }
+    }
+
```
mathbunnyru, line 52: I highly suggest using switch instead of if/else when comparing to rpc::DelegateFilter::Role possible values

src/rpc/common/Types.hpp
```diff
@@ -175,6 +175,25 @@ struct AccountCursor {
...
 };
 
+/**
+ * @brief A delegate object used filter account_tx by specific delegate accounts
+ */
+struct DelegateFilter {
+    /**
+     * @brief A delegate type used in delegate filter
+     */
+    enum class Role {
```
mathbunnyru: `: std::uint8_t`

src/rpc/common/Types.hpp
```diff
@@ -175,6 +175,25 @@ struct AccountCursor {
...
+        Actor, /**< This account is the *active* sender, acting on behalf of another party.
+                * e.g., Account A in "A sends payment to B on behalf of C." */
+
+        Authorizer /**< This account is the *passive* party whose funds are being moved from.
+                    * e.g., Account C in "A sends payment to B on behalf of C." */
+    };
+
+    Role delegateType = Role::Actor;
```
mathbunnyru: Do we need do add a default value here?
I would probably avoid it if possible, to not to forget setting the right one when needed

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -616,6 +616,104 @@ TEST_F(RPCHelpersTest, FetchAndCheckAnyFlagExists_TrustLineIsFrozenAndCheckFreez
...
+        EXPECT_EQ(result->delegateType, DelegateFilter::Role::Authorizer);
+        EXPECT_FALSE(result->counterParty.has_value());
+        // NOLINTEND(bugprone-unchecked-optional-access)
+    }
+
+    // delegate agent + counter_party is valid
+    {
+        auto const json = boost::json::parse(R"JSON({
+            "delegate_filter": "actor",
+            "counter_party": "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun"
```
mathbunnyru: Can we reuse some existing acounts here, or at least put it into a variable?

src/rpc/filters/TransactionFilter.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <optional>
+
+namespace rpc {
+
+/**
+ * @brief Result of a filter check.
+ */
+struct FilterResult {
+    bool shouldInclude = false;
+    std::optional<xrpl::AccountID> relevantAccount;
```
mathbunnyru: You have both bool and std::optional - you sure we need both?

mathbunnyru (changes requested):

src/rpc/RPCHelpers.cpp
PeterChen13579: I just realized the `void` in `std::expected<DelegateFilter::Role, void>` is not valid code. I'd probably leave it as is as `customValidator` checks for if parsing failed or not ( `has_value()`)

src/rpc/filters/TransactionFilter.hpp
PeterChen13579: yes, you are right

src/rpc/common/Types.hpp
PeterChen13579: This is needed as otherwise clang-tidy complains; Role is an enum class with no default initialization

src/rpc/RPCHelpers.cpp
mathbunnyru: void partial specialization is absolutely valid: https://en.cppreference.com/cpp/utility/expected

src/rpc/common/Types.hpp
mathbunnyru: It's not about clang-tidy. Can we require users of this class to pass Role in ctor?
Yes, this class won't be default constructible, but do we need it to be?

mathbunnyru (changes requested):
Please, take a look at comment I left

src/rpc/RPCHelpers.cpp
PeterChen13579: @mathbunnyru it is 
```
template< class T, class E >
    requires std::is_void_v<T>
class expected<T, E>;
```
So something like expected<void, error> works but not expected<ValueType, void>.
and since none of the callers need error detail, I think leaving as std::optional is better.

src/rpc/common/Types.hpp
PeterChen13579: Thanks, Yea this is better

src/rpc/RPCHelpers.cpp
mathbunnyru: Ah, yeah, sorry, let's leave this as is

mathbunnyru (dismissed):

src/rpc/common/Validators.cpp
```diff
@@ -383,4 +385,34 @@ CustomValidator CustomValidators::authorizeCredentialValidator =
         return MaybeError{};
     }};
 
+CustomValidator CustomValidators::delegateValidator =
+    CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
+        if (!value.is_object())
```
godexsoft: nit: `not value.is_object()` reads nicer 😛

src/rpc/filters/impl/DelegateTransactionsFilter.cpp
```diff
@@ -0,0 +1,63 @@
...
+    std::optional<xrpl::AccountID> txDelegate;
+    if (sttx.isFieldPresent(xrpl::sfDelegate))
+        txDelegate = sttx.getAccountID(xrpl::sfDelegate);
+
+    // Transactions without an sfDelegate field are not delegated; exclude them immediately.
+    if (not txDelegate.has_value())
+        return std::nullopt;
+
+    switch (delegateFilter_.delegateType) {
+        case rpc::DelegateFilter::Role::Authorizer:
```
godexsoft: nit: probably don't need the `rpc::` here?

src/rpc/filters/impl/DelegateTransactionsFilter.hpp
```diff
@@ -0,0 +1,35 @@
...
+     * @param filter The filter parameters from the JSON request (role, counterparty string)
+     * @param queriedAccount The account currently being queried in account_tx (input from
+     * account_tx handler)
+     */
+    DelegateTransactionFilter(rpc::DelegateFilter filter, xrpl::AccountID queriedAccount);
+
+    [[nodiscard]] std::optional<xrpl::AccountID>
+    check(data::TransactionAndMetadata const& txnPlusMeta) const override;
+
+private:
```
godexsoft: nit: we usually put data members above interface in the implicit private block

src/rpc/filters/TransactionFilter.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @brief Interface for filtering transactions.
+ */
+class TransactionFilter {
+public:
+    virtual ~TransactionFilter() = default;
+
+    /**
+     * @brief Check if a transaction blob matches the filter criteria.
```
godexsoft, line 39: nit: brief is usually separated by a space line

src/rpc/handlers/AccountTx.cpp
```diff
@@ -146,6 +156,13 @@ AccountTxHandler::process(AccountTxHandler::Input const& input, Context const& c
             continue;
         }
 
+        std::optional<xrpl::AccountID> relevantAccount;
+        if (txFilter) {
+            relevantAccount = txFilter->check(txnPlusMeta);
```
godexsoft: nit: to avoid using `input.delegateFilter` again below this could return a "check result" type which will have the things needed below

src/rpc/handlers/AccountTx.cpp
```diff
@@ -204,6 +221,17 @@ AccountTxHandler::process(AccountTxHandler::Input const& input, Context const& c
                         obj[JS(close_time_iso)] = xrpl::toStringIso(ledgerHeader->closeTime);
                     }
                 }
+
+                if (relevantAccount) {
+                    // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
+                    if (input.delegateFilter->delegateType ==
```
godexsoft: I'm not a fan of how this depends on several levels of optionals..

tests/unit/rpc/RPCHelpersTests.cpp
```diff
@@ -616,6 +618,114 @@ TEST_F(RPCHelpersTest, FetchAndCheckAnyFlagExists_TrustLineIsFrozenAndCheckFreez
...
+        // NOLINTBEGIN(bugprone-unchecked-optional-access)
+        EXPECT_EQ(result->delegateType, DelegateFilter::Role::Authorizer);
+        EXPECT_FALSE(result->counterParty.has_value());
+        // NOLINTEND(bugprone-unchecked-optional-access)
+    }
+
+    // delegate agent + counter_party is valid
+    {
+        auto const json = boost::json::parse(
+                              fmt::format(
```
godexsoft: nit: Damn this is ugly formatting 😢

godexsoft (changes requested):
Let's fix a few small easy things.. i leave most of them up to you to decide how you prefer it.

godexsoft (approved):
👍

mathbunnyru (approved):

PR #2828, opened 2025-12-02
chore: Enable TSAN without ignoring errors

godexsoft:
Let's wait for last TSAN issues to be fixed in @kuznetsss 's ongoing PR before merging this.
Another thing is that we may want to set up so that any warning/error from sanitizers immediately halts clio_tests.


.github/workflows/reusable-test.yml
```diff
@@ -65,46 +61,9 @@ jobs:
...
-          GH_TOKEN: ${{ github.token }}
-        with:
-          labels: "bug"
-          title: "[${{ inputs.conan_profile }}] reported issues"
-          body: >
-            Clio tests failed one or more sanitizer checks when built with ${{ inputs.conan_profile }}`.
-
-            Workflow: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}/
-            Reports are available as artifacts.
-
```
kuznetsss, line 107: We probably should still create a ticket when it fails since it is a nightly workflow

src/feed/impl/TrackableSignal.hpp
```diff
@@ -73,10 +73,15 @@ class TrackableSignal {
...
+        // either. `track_foreign` is racey because the control block can be released from some thread without a mutex
+        // protecting it while another one is invoking the slot. Therefore we are storing a weak_ptr of the trackable
+        // and explicitly checking it in the slot invocation instead. The invocation is guaranteed to happen under an
+        // internal mutex so it's safe.
         connections->emplace(
-            trackable.get(), signal_.connect(typename SignalType::slot_type(slot).track_foreign(trackable))
+            trackable.get(), signal_.connect([slot, weakTrackable = std::weak_ptr(trackable)](Args&&... args) {
+                if (not weakTrackable.expired())
+                    std::invoke(slot, std::forward<Args...>(args)...);
+            })
```
kuznetsss: The fix is not very clear, but I'm ok if you think it fixes the issue

.github/workflows/reusable-test.yml
mathbunnyru: We will get a notification, but I think having an issue would also be good

mathbunnyru (commented):
I think it would be great if you commented on GitHub what was wrong with the initial code

godexsoft:
> I think it would be great if you commented on GitHub what was wrong with the initial code

It's in the comment above the code that had to change

tests/unit/feed/SubscriptionManagerTests.cpp
```diff
@@ -72,11 +72,11 @@ class SubscriptionManagerBaseTest : public util::prometheus::WithPrometheus, pub
     }
 
     StrictMockAmendmentCenterSharedPtr mockAmendmentCenterPtr_;
-    std::shared_ptr<SubscriptionManager> subscriptionManagerPtr_ =
-        std::make_shared<SubscriptionManager>(Execution(2), backend_, mockAmendmentCenterPtr_);
     web::SubscriptionContextPtr session_ = std::make_shared<MockSession>();
     MockSession* sessionPtr_ = dynamic_cast<MockSession*>(session_.get());
     uint32_t const networkID_ = 123;
+    std::shared_ptr<SubscriptionManager> subscriptionManagerPtr_ =
```
godexsoft, line 78: This was moved so that order of destruction is correct for all the tests (vs. adding a destructor that calls stop for the suite).

src/feed/impl/SingleFeedBase.cpp
```diff
@@ -64,7 +64,7 @@ SingleFeedBase::unsub(SubscriberSharedPtr const& subscriber)
 void
 SingleFeedBase::pub(std::string msg)
 {
-    [[maybe_unused]] auto task = strand_.execute([this, msg = std::move(msg)]() {
+    strand_.submit([this, msg = std::move(msg)] {
```
godexsoft, line 67: This was not a problem but 'submit' is a better API for this type of task.

mathbunnyru (approved):

src/feed/impl/TrackableSignal.hpp
```diff
@@ -73,10 +73,15 @@ class TrackableSignal {
...
-        // either. track_foreign will hold a weak_ptr to the connection, which makes sure the connection is valid when
-        // the slot is called.
+        // either. `track_foreign` is racey when one shared_ptr is tracked by multiple signals. Therefore we are storing
+        // a weak_ptr of the trackable and using weak_ptr::lock() to atomically check existence and acquire a shared_ptr
+        // during slot invocation. This guarantees to keep the trackable alive for the duration of the slot call and
+        // avoids potential race conditions.
         connections->emplace(
-            trackable.get(), signal_.connect(typename SignalType::slot_type(slot).track_foreign(trackable))
+            trackable.get(), signal_.connect([slot, weakTrackable = std::weak_ptr(trackable)](Args&&... args) {
+                if (auto lifeExtender = weakTrackable.lock(); lifeExtender)
```
godexsoft, line 82: While i believe it was safe simply checking the weakTrackable, this is even safer (just in case). The reason i believe it was safe without extending lifetime is that the race was between invocation of the slot AND track_foreign's internal mechanism which was accessing the shared control block without a mutex. With this new change it's an even stronger guarantee - no potential issue with any other external code that might do a similar thing to what track_foreign was doing.

kuznetsss (approved):

PR #2831, opened 2025-12-04
feat: Add observable value util

godexsoft:
This implements a simple observable value. Can be used for a more reactive approach. Will be used in ETL state and across the codebase with time.

### DIFF
--- a/src/util/ObservableValue.hpp
+++ b/src/util/ObservableValue.hpp
@@ -64,6 +64,8 @@ protected:
     boost::signals2::signal<void(T const&)> onUpdate_;
 
 public:
+    virtual ~ObservableValueBase() = default;
+
     /**
      * @brief Registers an observer callback for value changes.
      * @param fn Callback function/lambda that accepts T const&
@@ -84,6 +86,15 @@ public:
     {
         return not onUpdate_.empty();
     }
+
+    /**
+     * @brief Forces notification of all observers with the current value.
+     *
+     * This method will notify all observers with the current value regardless
+     * of whether the value has changed since the last notification.
+     */
+    virtual void
+    forceNotify() = 0;
 
 protected:
     /**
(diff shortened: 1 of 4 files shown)


tests/unit/util/ObservableValueTest.cpp
```diff
@@ -0,0 +1,854 @@
...
+    int count = 0;
+    std::vector<T> values;
+
+    void
+    operator()(T const& value)
+    {
+        ++count;
+        values.push_back(value);
+    }
+};
```
kuznetsss: This could be done with MockFunction

src/util/ObservableValue.hpp
```diff
@@ -0,0 +1,520 @@
...
+
+    /**
+     * @brief Provides direct access to the underlying atomic.
+     *
+     * Use with caution - direct atomic operations bypass observation.
+     * Prefer using set() or the guard mechanism for observable changes.
+     *
+     * @return Reference to the underlying atomic
+     */
+    [[nodiscard]] std::atomic<T>&
```
godexsoft: This should probably just be removed unless it is really used and needed somewhere.

src/util/ObservableValue.hpp
```diff
@@ -0,0 +1,520 @@
...
+namespace util {
+
+template <typename T>
+struct IsAtomic : std::false_type {};
+
+template <typename T>
+struct IsAtomic<std::atomic<T>> : std::true_type {};
+
+template <typename T>
+constexpr bool kIS_ATOMIC_V = IsAtomic<T>::value;
```
kuznetsss: Couldn't this be done using `std::same_as`?

src/util/ObservableValue.hpp
```diff
@@ -0,0 +1,520 @@
...
+ *     boost::signals2::scoped_connection scoped = counter.observe([](int const& value) {
+ *         std::cout << "Scoped observer: " << value << std::endl;
+ *     });
+ *     counter = 200;  // Prints "Scoped observer: 200"
+ * } // Automatically disconnects here
+ * counter = 300;  // No output - observer was disconnected
+ * @endcode
+ */
+template <Observable T>
+    requires(not kIS_ATOMIC_V<T>)
```
kuznetsss: Maybe it will work without this concept? If I remember correctly, compiler chooses the least generic overload.

src/util/ObservableValue.hpp
godexsoft: It's tricky but possible 👍

src/util/ObservableValue.hpp
godexsoft: It would work but it's dangerous - someone may change something in the future and the main template might be used instead of the partial specialization without us noticing. So let's leave it just for our own safety

src/util/ObservableValue.hpp
```diff
@@ -0,0 +1,428 @@
...
+ *
+ * - Must be equality comparable to detect changes
+ * - Must be copy constructible for capturing old values in guards
+ * - Must be move constructible for efficient value updates
+ *
+ * @note Copy assignment is intentionally not required since we use move semantics
+ *       for value updates and only need copy construction for change detection.
+ */
+template <typename T>
+concept Observable = std::equality_comparable<T> && std::copy_constructible<T> && std::move_constructible<T>;
```
kuznetsss, line 49: Maybe also trivially copyable at least for now, wdyt?

src/util/ObservableValue.hpp
```diff
@@ -0,0 +1,428 @@
...
+    /**
+     * @brief Checks if there are any active observers.
+     * @return true if there are observers, false otherwise
+     */
+    [[nodiscard]] bool
+    hasObservers() const
+    {
+        return not onUpdate_.empty();
+    }
+
```
kuznetsss, line 89: Should we also add `forceNotify()` to notify all subscribers with the current value?

src/util/ObservableValue.hpp
godexsoft: It would break a lot of tests atm. I think we can postpone until we find we need a more optimized specialization for vector etc.

kuznetsss (approved):
👍

PR #2851, opened 2025-12-12
WIP: fix race condition in ETL

kuznetsss:
This is a WIP fix for potential race condition introduced here: #2842 

- `taskManager_` should be protected by a strand
- Async framework changed to post where possible since spawn may be inlined and it may cause a deadlock in ETL

We should check that it is not possible to update cache twice or skip a ledger somewhere.


kuznetsss:
Closing this in favour of #2894

PR #2859, opened 2025-12-15
feat: Basic support for channels

godexsoft:
This PR implements go-like channels wrapper (on top of asio experimental channels). 
In the future this will be integrated into the AsyncFramework.


src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+namespace util {
+
+template <typename T>
+class Channel {
```
mathbunnyru, line 53: Right now it doesn't look like a class, it has 0 logic or members, let's make it a namespace

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+    class Shared {
+        boost::asio::any_io_executor executor_;
+        boost::asio::experimental::concurrent_channel<void(boost::system::error_code, T)> ch_;
+
+    public:
+        Shared(auto&& context, std::size_t capacity) : executor_(context.get_executor()), ch_(context, capacity)
+        {
+        }
+
+        [[nodiscard]] auto&
```
mathbunnyru: Let's add a type here

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+namespace util {
+
+template <typename T>
+class Channel {
+private:
+    class Shared {
```
mathbunnyru: Maybe call it ChannelImpl or ControlBlock?

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+        [[nodiscard]] bool
+        isClosed() const
+        {
+            return not ch_.is_open();
+        }
+    };
+
+    class Sender {
+        std::shared_ptr<Shared> shared_;
```
mathbunnyru: Do I understand correctly, that technically, you don't need this one (since you can access from inner)?
But for perf reasons you save it as well?

maybe let's add a comment?

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+        Sender&
+        operator=(Sender const&) = default;
+
+        template <typename D>
+        bool
+        asyncSend(D&& data, boost::asio::yield_context yield)
+            requires(std::same_as<std::remove_cvref_t<T>, std::remove_cvref_t<D>>)
+        {
+            boost::system::error_code ec;
+            shared_->channel().async_send(ec, std::forward<decltype(data)>(data), yield[ec]);
```
mathbunnyru:
```suggestion
            shared_->channel().async_send(ec, std::forward<D>(data), yield[ec]);
```

mathbunnyru: and in places below as well

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+        template <typename D>
+        void
+        asyncSend(D&& data, std::invocable<bool> auto&& fn)
+            requires(std::same_as<std::remove_cvref_t<T>, std::remove_cvref_t<D>>)
+        {
+            boost::system::error_code ec;
+            shared_->channel().async_send(
+                ec,
+                std::forward<decltype(data)>(data),
+                [fn = std::forward<decltype(fn)>(fn)](boost::system::error_code ec) mutable { fn(not ec); }
```
mathbunnyru: maybe let's rename `ec` to sth else here?
because otherwise we use the same name for 2 kind-of-different ec's

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+        std::optional<T>
+        tryReceive()
+        {
+            std::optional<T> result;
+            auto const received = shared_->channel().try_receive([&result](boost::system::error_code ec, auto&& value) {
+                if (not ec)
+                    result = std::forward<decltype(value)>(value);
+            });
+
+            if (received)
```
mathbunnyru: Is it possible for some reason to have error in `ec`, but received True?
And why is there 2 ways to communicate the error?

src/util/Channel.hpp
```diff
@@ -0,0 +1,195 @@
...
+        {
+            return shared_->isClosed();
+        }
+    };
+
+public:
+    static std::pair<Sender, Receiver>
+    create(auto&& context, std::size_t capacity)
+    {
+        auto shared = std::make_shared<Shared>(context, capacity);
```
mathbunnyru: forward context?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,756 @@
...
+        std::string context = (contextType == ContextType::IOContext) ? "IOContext" : "ThreadPool";
+        std::string approach = (approachType == ApproachType::Spawn) ? "Spawn" : "Callback";
+        return context + "_" + approach;
+    }
+};
+
+class ContextWrapper {
+public:
+    using ContextVariant =
+        std::variant<std::unique_ptr<boost::asio::io_context>, std::unique_ptr<boost::asio::thread_pool>>;
```
mathbunnyru: Why can't we store them by values here?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,756 @@
...
+
+    explicit ContextWrapper(ContextType type)
+    {
+        if (type == ContextType::IOContext) {
+            context_ = std::make_unique<boost::asio::io_context>();
+        } else {
+            context_ = std::make_unique<boost::asio::thread_pool>(4);
```
mathbunnyru: let's move 4 to a constant?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,756 @@
...
+    {
+        return *std::get<std::unique_ptr<T>>(context_);
+    }
+
+    void
+    run()
+    {
+        std::visit(
+            [](auto& contextPtr) {
+                using ContextType = std::decay_t<decltype(*contextPtr)>;
```
mathbunnyru: do we have `overloaded` pattern?
let's use it here (or add it)?

mathbunnyru (changes requested):
I haven't read test file yet, will review a bit later

src/util/Channel.hpp
godexsoft: Inner is only for tracking the last instance, not for any other use

src/util/Channel.hpp
godexsoft: I'm not too sure we actually need the `ec` at all as we don't actually use it when we send. when we receive i imagine there should be an internal `ec` somewhere. I will check if we can get rid of it

tests/unit/util/ChannelTests.cpp
godexsoft: Should be possible - this is a consequence of a different initial storage strategy

src/util/Channel.hpp
godexsoft: Hm, i could not remove error_code from the channel signature. I expected that would be easy but looks like you need that error_code in the signature

src/util/Channel.hpp
godexsoft: I agree it does not look like a class but a namespace would have to be called `channel` and will not be able to have a template parameter `T` (would be on the `create` function instead). So i'm going to leave it as is as it reads better

mathbunnyru: Ok, let's leave it 👍

src/util/Channel.hpp
mathbunnyru: 👌

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,755 @@
...
+private:
+    ContextVariant context_;
+};
+
+}  // namespace
+
+class ChannelParameterizedTest : public TestWithParam<ChannelTestParams> {
+protected:
+    void
+    SetUp() override
```
mathbunnyru: If I remember correctly, we removed usage of `SetUp`

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,755 @@
...
+    {
+        params_ = GetParam();
+        context_ = std::make_unique<ContextWrapper>(params_.contextType);
+    }
+
+    ChannelTestParams params_{};
+    std::unique_ptr<ContextWrapper> context_;
```
mathbunnyru: do you need unique_ptr here?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,755 @@
...
+protected:
+    template <typename Sender, typename Receiver>
+    void
+    runIOContextTest(Sender sender, Receiver receiver, boost::asio::io_context& executor)
+    {
+        std::vector<int> receivedValues;
+
+        if (params_.approachType == ApproachType::Spawn) {
+            util::spawn(executor, [&receiver, &receivedValues](boost::asio::yield_context yield) mutable {
+                while (receivedValues.size() < kTOTAL_EXPECTED) {
```
mathbunnyru: Maybe we should check that we have exactly kTOTAL_EXPECTED values?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,755 @@
...
+                    }
+                }
+            });
+
+            for (auto senderId = 0uz; senderId < kNUM_SENDERS; ++senderId) {
+                util::spawn(executor, [sender, senderId](boost::asio::yield_context yield) mutable {
+                    for (auto i = 0uz; i < kVALUES_PER_SENDER; ++i) {
+                        int value = (senderId * 100) + i;
+                        bool success = sender.asyncSend(value, yield);
+                        if (!success)
```
mathbunnyru: nit: can merge these lines
and below as well

mathbunnyru (dismissed):

tests/unit/util/ChannelTests.cpp
godexsoft: We check it below already, is that not what you meant?

mathbunnyru: Sorry, my wording is bad.

Right now you're pushing to vector only if `while (receivedValues.size() < kTOTAL_EXPECTED)`, so it's impossible you're going to have more than `kTOTAL_EXPECTED` values pushed to vector

This doesn't check that receiver is not able to receive a few more values though.
Imagine, there is a bug, and we do not account for some values in queue when receiving.

I suggest after this `while` to write `tryReceive` and check it fails

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+namespace util {
+
+/**
+ * @brief Represents a go-like channel, a single-producer (Sender) multi-consumer (Receiver) thread-safe data pipe.
```
kuznetsss: Looks like it is multi-producer single-consumer because only Sender is copyable

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+        std::shared_ptr<ControlBlock> shared_;
+        struct Inner {
+            std::shared_ptr<ControlBlock> shared;
+
+            ~Inner()
+            {
+                shared->close();
+            }
+        };
+        std::shared_ptr<Inner> inner_;
```
kuznetsss: Maybe would be good to leave a comment about the purpose of `Inner`

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+        Sender(Sender const&) = default;
+        Sender&
+        operator=(Sender&&) = default;
+        Sender&
+        operator=(Sender const&) = default;
+
+        template <typename D>
+        bool
+        asyncSend(D&& data, boost::asio::yield_context yield)
+            requires(std::same_as<std::remove_cvref_t<T>, std::remove_cvref_t<D>>)
```
kuznetsss: Maybe convertible instead of same?

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+            std::optional<T> result;
+            auto const received = shared_->channel().try_receive([&result](boost::system::error_code ec, auto&& value) {
+                if (not ec)
+                    result = std::forward<decltype(value)>(value);
+            });
+
+            if (received)
+                return result;
+
+            return std::nullopt;
```
kuznetsss: Why not always return `result`?

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+        [[nodiscard]] InternalChannelType&
+        channel()
+        {
+            return ch_;
+        }
+
+        void
+        close()
+        {
+            ch_.close();
```
kuznetsss: I randomly found this issue, probably worth checking and add a workaround if it is not fixed yet: https://github.com/chriskohlhoff/asio/issues/1575

src/util/Channel.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+public:
+    /**
+     * @brief Factory function to create channel components.
+     * @param context A supported context type (either io_context or thread_pool)
+     * @param capacity Size of the internal buffer on the channel
+     * @return A pair of Sender and Receiver
+     */
+    static std::pair<Sender, Receiver>
+    create(auto&& context, std::size_t capacity)
```
kuznetsss, line 307: Capacity could be optional to have unlimited buffer size

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,752 @@
...
+                });
+            });
+        }
+
+        context_.run();
+        EXPECT_TRUE(testCompleted);
+    } else {
+        auto& executor = context_.get<boost::asio::thread_pool>();
+
+        util::Mutex<bool> testCompleted{false};
```
kuznetsss: Why not atomic?

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,752 @@
...
+    Values(
+        ChannelTestParams{ContextType::IOContext, ApproachType::Spawn},
+        ChannelTestParams{ContextType::IOContext, ApproachType::Callback},
+        ChannelTestParams{ContextType::ThreadPool, ApproachType::Spawn},
+        ChannelTestParams{ContextType::ThreadPool, ApproachType::Callback}
+    ),
+    [](TestParamInfo<ChannelTestParams> const& info) { return info.param.toString(); }
+);
+
+TEST_P(ChannelParameterizedTest, ChannelClosureScenarios)
```
kuznetsss: I think all the tests will look better without parametrisation because they are already separate if branches

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,752 @@
...
+                                    boost::asio::post(executor, [self = std::move(self), i]() mutable { self(i + 1); });
+                            }
+                        );
+                    };
+                    sendNext(0);
+                });
+            }
+        }
+
+        context_.run();
```
kuznetsss, line 179: Maybe better to use `run_for()` to handle hanging in case of an issue

src/util/Channel.hpp
godexsoft: It's now a multi-producer multi-consumer 👍

src/util/Channel.hpp
godexsoft: Hm, maybe we address this if we see it happen some day. The tests we have show that things work as expected for our use cases.

src/util/Channel.hpp
godexsoft: IIRC @mathbunnyru had a strong opinion on this. Unbound queues are a bad idea normally so perhaps it's better to always have a capacity, wdyt?

tests/unit/util/ChannelTests.cpp
godexsoft: Imo it's good as is.

src/util/Channel.hpp
kuznetsss: I'm fine to leave it as it is. We can change later if needed

src/util/Channel.hpp
```diff
@@ -0,0 +1,207 @@
...
+        trySend(D&& data)
+            requires(std::convertible_to<std::remove_cvref_t<D>, std::remove_cvref_t<T>>)
+        {
+            boost::system::error_code ec;
+            return shared_->channel().try_send(ec, std::forward<D>(data));
+        }
+    };
+
+    class Receiver {
+        std::shared_ptr<ControlBlock> shared_;
```
kuznetsss, line 208: The last `Receiver` should also close the channel, so there should be `Inner` here as well

godexsoft: Now that it's mcmp i think you are right, let's change that as well.

src/util/Channel.hpp
```diff
@@ -0,0 +1,305 @@
...
+
+        void
+        close()
+        {
+            if (not isClosed()) {
+                ch_.close();
+                // Workaround for Boost bug: close() alone doesn't cancel pending async operations.
+                // We must call cancel() to unblock them. The bug also causes cancel() to return
+                // error_code 0 instead of channel_cancelled, so async operations must check
+                // isClosed() to detect this case.
```
kuznetsss, line 79: Maybe let's add a link to the bug?

src/util/Channel.hpp
```diff
@@ -0,0 +1,305 @@
...
+     * @param context A supported context type (either io_context or thread_pool)
+     * @param capacity Size of the internal buffer on the channel
+     * @return A pair of Sender and Receiver
+     */
+    static std::pair<Sender, Receiver>
+    create(auto&& context, std::size_t capacity)
+    {
+        auto shared = std::make_shared<ControlBlock>(std::forward<decltype(context)>(context), capacity);
+        auto sender = Sender{shared};
+        auto receiver = Receiver{shared};
```
kuznetsss: nit:

```suggestion
        auto receiver = Receiver{std::move(shared)};
```

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,691 @@
...
+
+        EXPECT_EQ(receivedValues.lock()->size(), kTOTAL_EXPECTED);
+        std::ranges::sort(receivedValues.lock().get());
+
+        std::vector<int> expectedValues;
+        for (auto senderId = 0uz; senderId < kNUM_SENDERS; ++senderId) {
+            for (auto i = 0uz; i < kVALUES_PER_SENDER; ++i) {
+                expectedValues.push_back(static_cast<int>((senderId * 100) + i));
+            }
+        }
```
kuznetsss: nit: You could hardcode values to send into test fixture field and then you will not have to generate them twice in each test.

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,691 @@
...
+    // Spawn multiple senders that will all block (no receiver is consuming)
+    for (auto i = 0uz; i < kPENDING_NUM_SENDERS; ++i) {
+        util::spawn(pool, [senderCopy = sender, i, &completedSends](boost::asio::yield_context yield) mutable {
+            senderCopy.asyncSend(static_cast<int>(i), yield);
+            ++completedSends;
+        });
+    }
+
+    // Give senders time to start and block
+    std::this_thread::sleep_for(std::chrono::milliseconds{100});
```
kuznetsss: Better to use semaphore here.

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,691 @@
...
+        [[maybe_unused]] auto temp = std::move(receiverCopy);
+    }
+    EXPECT_FALSE(sender.trySend(44));  // all receivers destroyed, channel closed
+}
+
+// This test verifies the workaround for a bug in boost::asio::experimental::concurrent_channel where close() does not
+// cancel pending async operations. Our Channel wrapper calls cancel() after close() to ensure pending operations are
+// unblocked.
+// See: https://github.com/chriskohlhoff/asio/issues/1575
+TEST(ChannelTest, PendingAsyncSendsNotCancelledOnClose)
```
kuznetsss: We are checking here that they are actually cancelled even though there is a bug in asio.
```suggestion
TEST(ChannelTest, PendingAsyncSendsAreCancelledOnClose)
```

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,691 @@
...
+
+    // Channel with capacity 0 - all sends will block waiting for a receiver
+    auto [sender, receiver] = util::Channel<int>::create(pool, 0);
+
+    std::atomic<std::size_t> completedSends{0};
+
+    // Spawn multiple senders that will all block (no receiver is consuming)
+    for (auto i = 0uz; i < kPENDING_NUM_SENDERS; ++i) {
+        util::spawn(pool, [senderCopy = sender, i, &completedSends](boost::asio::yield_context yield) mutable {
+            senderCopy.asyncSend(static_cast<int>(i), yield);
```
kuznetsss: We should check that cancellation returns an error:
```suggestion
            EXPECT_FALSE(senderCopy.asyncSend(static_cast<int>(i), yield));
```

tests/unit/util/ChannelTests.cpp
```diff
@@ -0,0 +1,691 @@
...
+class ChannelCallbackTest : public TestWithParam<ContextType> {
+protected:
+    ChannelCallbackTest() : context_(GetParam())
+    {
+    }
+
+    ContextWrapper context_;
+};
+
+TEST_P(ChannelSpawnTest, MultipleSendersOneReceiver)
```
kuznetsss, line 152: Maybe would also be good to add a test to verify the order when we have only one sender and one receiver.

kuznetsss (commented):
Looks good, a few comments, feel free to ignore nits.

kuznetsss (approved):

PR #2860, opened 2025-12-15
revert: "refactor: Add writing command to etl::SystemState"

mathbunnyru:
Reverts XRPLF/clio#2842

@kuznetsss told there is a bug
And tests fail a lot recently, so let's revert for now


godexsoft (approved):
:rage1:

PR #2866, opened 2025-12-18
fix: WorkQueue contention

godexsoft:
This PR fixes some performance issues in the recently changed WorkQueue with priority support.
The way this was fixed is by creating a micro benchmark and copying the old code into the benchmark to compare against. I'm leaving the benchmark in so that we can always quickly measure again if we need to. 

Result of these changes is ~2x performance on 8 threads with 5000 queue limit (similar to s1/s2 setup).


benchmarks/rpc/WorkQueueBenchmarks.cpp
```diff
@@ -0,0 +1,123 @@
...
+}
+
+}  // namespace
+
+static void
+benchmarkWorkQueue(benchmark::State& state)
+{
+    init();
+
+    auto const kTOTAL = static_cast<size_t>(state.range(0));
```
mathbunnyru:
```suggestion
    auto const total = static_cast<size_t>(state.range(0));
```

nit: It's a bit strange we're using different styling here

benchmarks/rpc/WorkQueueBenchmarks.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+                    boost::asio::steady_timer timer(yield.get_executor(), std::chrono::milliseconds{delayMs});
+                    timer.async_wait(yield);
+                },
+                /* isWhiteListed = */ false
+            ));
+        }
+
+        queue.stop();
+        ASSERT(totalQueued <= kTOTAL && totalExecuted == totalQueued, "Totals don't match");
```
mathbunnyru: We should also assert that at least `maxSize` have been queued and executed

benchmarks/rpc/WorkQueueBenchmarks.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+// Usage example:
+/*
+  clio_benchmark
+    --benchmark_repetitions=10
+    --benchmark_display_aggregates_only=true
+    --benchmark_min_time=1x
+    --benchmark_filter="WorkQueue"
```
mathbunnyru:
```suggestion
  ./clio_benchmark \
    --benchmark_repetitions=10 \
    --benchmark_display_aggregates_only=true \
    --benchmark_min_time=1x \
    --benchmark_filter="WorkQueue"
```

benchmarks/rpc/WorkQueueBenchmarks.cpp
```diff
@@ -0,0 +1,123 @@
...
+*/
+BENCHMARK(benchmarkWorkQueue)
+    ->ArgsProduct({{1'000, 10'000, 100'000}, {2, 4, 8}, {0, 5'000}, {10, 100, 250}})
```
mathbunnyru, line 126: What happens on 1 thread?
If it's slow, let's write why

mathbunnyru (approved):

benchmarks/rpc/WorkQueueBenchmarks.cpp
mathbunnyru: Right now left a TODO

PR #2867, opened 2025-12-18
ci: Remove unnecessary creation of build directory

bthomee:
The `conan install` command creates the directory if it doesn't exist.


.github/actions/conan/action.yml
```diff
@@ -23,7 +23,9 @@ runs:
   steps:
     - name: Create build directory
       shell: bash
-      run: mkdir -p "${{ inputs.build_dir }}"
+      env:
+        BUILD_DIR: "${{ inputs.build_dir }}"
+      run: mkdir -p "${BUILD_DIR}"
```
mathbunnyru: Let's remove these lines completely. `conan install` creates a dir if it doesn't exist (I checked locally)

mathbunnyru (changes requested):

.github/actions/conan/action.yml
mathbunnyru: Less code - less semgrep warnings 😂

mathbunnyru (approved):

PR #2877, opened 2026-01-02
chore: Pin ruamel.yaml<0.19 in pre-commit-hooks

mathbunnyru:
Please, see https://github.com/pre-commit/pre-commit-hooks/issues/1229


mathbunnyru:
The upstream issue has been fixed, so this is no longer required.

PR #2883, opened 2026-01-06
feat: Option to save cache asyncronously

kuznetsss:
This PR adds an option to save cache to file asynchronously in parallel with shutting down the rest of Clio services.

### DIFF
--- a/tests/unit/data/LedgerCacheSaverTests.cpp
+++ b/tests/unit/data/LedgerCacheSaverTests.cpp
@@ -47,17 +47,23 @@ struct LedgerCacheSaverTest : virtual testing::Test {
     constexpr static auto kFILE_PATH = "./cache.bin";
 
     static ClioConfigDefinition
-    generateConfig(bool cacheFilePathHasValue)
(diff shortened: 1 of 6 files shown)


tests/unit/data/LedgerCacheSaverTests.cpp
```diff
@@ -144,13 +150,61 @@ struct LedgerCacheSaverAssertTest : LedgerCacheSaverTest, common::util::WithMock
...
     saver.save();
     EXPECT_CLIO_ASSERT_FAIL({ saver.save(); });
+    semaphore.release();
+
+    saver.waitToFinish();
+}
+
+TEST_F(LedgerCacheSaverTest, SyncSaveWaitsForCompletion)
+{
+    auto const config = generateConfig(true, false);  // async_save = false
```
godexsoft: Let's use `/* async_save= */ false` style instead like we do in some other tests already. Also would be nice to include this above in each `generateConfig` call too 👍

godexsoft (changes requested):
Looks alright, just one nit to improve readability 👍

godexsoft (approved):
:rage1:

PR #2887, opened 2026-01-07
fix: Faster implementation of work queue

godexsoft:
The new work queue implementation seems to be too slow because of its dispatcher component.
According to micro benchmark, this implementation appears to be nearly as fast as the original that did not support priorities.

### DIFF
--- a/benchmarks/rpc/WorkQueueBenchmarks.cpp
+++ b/benchmarks/rpc/WorkQueueBenchmarks.cpp
@@ -115,10 +115,8 @@ benchmarkWorkQueue(benchmark::State& state)
             });
         }
 
-        for (auto& t : threads) {
-            if (t.joinable())
-                t.join();
-        }
+        for (auto& t : threads)
+            t.join();
 
         queue.stop();
 
--- a/benchmarks/rpc/WorkQueueBenchmarks.cpp
+++ b/benchmarks/rpc/WorkQueueBenchmarks.cpp
@@ -29,13 +29,18 @@
 
 #include <benchmark/benchmark.h>
 #include <boost/asio/steady_timer.hpp>
+#include <boost/asio/thread_pool.hpp>
+#include <boost/json/object.hpp>
 
+#include <algorithm>
 #include <atomic>
 #include <cassert>
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <mutex>
+#include <thread>
+#include <vector>
 
 using namespace rpc;
 using namespace util::config;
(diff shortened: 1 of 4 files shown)


src/rpc/WorkQueue.cpp
```diff
@@ -227,4 +187,24 @@ WorkQueue::size() const
     return curSize_.get().value();
 }
 
+void
+WorkQueue::executeTask(std::chrono::system_clock::time_point opTime, boost::asio::yield_context yield)
+{
+    std::optional<TaskType> task;
+    {
+        auto state = queueState_.lock();
+        task = state->popNext();
```
kuznetsss: We should either remove optional or check that task is not empty before executing it

kuznetsss (changes requested):

benchmarks/rpc/WorkQueueBenchmarks.cpp
```diff
@@ -75,36 +80,54 @@ benchmarkWorkQueue(benchmark::State& state)
...
 
-    auto const total = static_cast<size_t>(state.range(0));
-    auto const numThreads = static_cast<uint32_t>(state.range(1));
-    auto const maxSize = static_cast<uint32_t>(state.range(2));
-    auto const delayMs = static_cast<uint32_t>(state.range(3));
+    auto const queueThreads = static_cast<uint32_t>(state.range(0));
+    auto const clientThreads = static_cast<uint32_t>(state.range(1));
+    auto const itemsPerThread = static_cast<uint32_t>(state.range(2));
+    auto const maxSize = static_cast<uint32_t>(state.range(3));
+    auto const delayMs = static_cast<uint32_t>(state.range(4));
```
mathbunnyru:
```suggestion
    auto const wqThreads = static_cast<uint32_t>(state.range(0));
    auto const maxQueueSize = static_cast<uint32_t>(state.range(1));
    auto const clientThreads = static_cast<uint32_t>(state.range(2));
    auto const itemsPerClient = static_cast<uint32_t>(state.range(3));
    auto const clientProcessingMs = static_cast<uint32_t>(state.range(4));
```

src/rpc/WorkQueue.cpp
```diff
@@ -77,12 +94,14 @@ WorkQueue::~WorkQueue()
 void
 WorkQueue::startProcessing()
 {
-    util::spawn(strand_, [this](auto yield) {
-        ASSERT(not hasDispatcher_, "Dispatcher already running");
+    ASSERT(not processingStarted_, "Attempt to start processing work queue more than once");
+    processingStarted_ = true;
```
mathbunnyru, line 98: Strictly speaking, this doesn't prevent from starting multiple times, if all threads run ASSERT first

src/rpc/WorkQueue.cpp
```diff
@@ -77,12 +94,14 @@ WorkQueue::~WorkQueue()
...
 
-        hasDispatcher_ = true;
-        dispatcherLoop(yield);
-    });
+    // Spawn workers for all tasks that were queued before processing started
+    auto const numTasks = size();
+    for (auto i = 0uz; i < numTasks; ++i) {
+        util::spawn(ioc_, [this](auto yield) { executeTask(std::chrono::system_clock::now(), yield); });
```
mathbunnyru: I think we should save the time when the task is being queued. Otherwise we will give incorrect results, if for some reason `startProcessing` is delayed

src/rpc/WorkQueue.cpp
```diff
@@ -194,6 +150,10 @@ WorkQueue::stop()
         requestStop();
 
     ioc_.join();
+
+    auto onTasksComplete = onQueueEmpty_.lock();
+    ASSERT(onTasksComplete->operator bool(), "onTasksComplete must be set when stopping is true.");
+    onTasksComplete->operator()();
```
mathbunnyru: Maybe let's put this to `{}` section, to show that it's a lock here?
I know it's in the end, but I think it would make it more clear

src/rpc/WorkQueue.cpp
mathbunnyru: And probably add a test which fails the current implementation

mathbunnyru (changes requested):

src/rpc/WorkQueue.cpp
godexsoft: This startProcessing function is not called by any "threads". it's either called by the constructor or by user in case the user specified a special tagged constructor. This is only really used for tests too - no one should use it in actual Clio code. I don't think this needs more protection than it has

mathbunnyru: Got it, let's keep it as is

tests/unit/rpc/WorkQueueTests.cpp
```diff
@@ -54,7 +58,7 @@ struct RPCWorkQueueTestBase : public virtual ::testing::Test {
     }
 };
 
-struct WorkQueueTest : WithPrometheus, RPCWorkQueueTestBase {
+struct WorkQueueTest : WithPrometheus, RPCWorkQueueTestBase, LoggerFixture {
```
mathbunnyru:
```suggestion
struct WorkQueueTest : WithPrometheus, RPCWorkQueueTestBase {
```

mathbunnyru: We don't read logs, we tried it for assert, but it didn't work :(

mathbunnyru (dismissed):

kuznetsss:
```
[ RUN      ] WorkQueueDeathTest.ExecuteTaskAssertsWhenQueueIsEmpty
[       OK ] WorkQueueDeathTest.ExecuteTaskAssertsWhenQueueIsEmpty (124036 ms)
```
124 seconds for 1 test with codecov :rage1:

kuznetsss (dismissed):

godexsoft:
> ```
> [ RUN      ] WorkQueueDeathTest.ExecuteTaskAssertsWhenQueueIsEmpty
> [       OK ] WorkQueueDeathTest.ExecuteTaskAssertsWhenQueueIsEmpty (124036 ms)
> ```
> 
> 124 seconds for 1 test with codecov :rage1:

That's so painful.. but i don't see a better way to implement this test since AssertMock failed us 😢 

Should we consider to disable the test until better times? perhaps it's not important enough to waste 124 seconds every CI run.. we know the ASSERT will work.

kuznetsss:
>  AssertMock failed us 

It's WorkQueue's design, not AssertMock. 

We could disable this test and create an issue to fix it later.

mathbunnyru (approved):

mathbunnyru (approved):

kuznetsss (approved):

PR #2889, opened 2026-01-08
ci: Use updated prepare-runner in actions and worfklows

bthomee:


mathbunnyru (approved):
I changed slightly the PR title to be more specific, thanks for the change 👍

Issue #2895, opened 2026-01-09
AmendmentCenter throws on very old ledgers


godexsoft:
## Issue Description

Related to #2881 

The underlying issue is that AmendmentCenter throws a `std::runtime_error` when the queried ledger does not contain an `Amendments` ledger object. 
Investigation is needed as to why a ledger object that is ALWAYS expected to be there is not available. It could very well be that such object was not part of the protocol back in the day, hence we could see this with some old ledgers but not with newer ones.

PR #2897, opened 2026-01-09
fix: Workaround an edge case exception in `AmendmentCenter`

### LINKED ISSUE #2881: `account_info` fails with internal error (73) for specific historical ledger on Clio
## Environment

Clio version: 2.6.0
Network: XRPL Mainnet
Clio RPC: 51233
Local rippled RPC: 5005 (non-archive; recent history only)

ETL sources:
* Local rippled (127.0.0.1) – recent validated range
* Remote archive rippled – historical validated range

## Issue Description

`account_info` returns internal error (error_code: 73) for ledger_index = 20,880,000, while:
* the ledger itself exists and is validated
* ledger state is readable via ledger_data
* the same account_info call succeeds at nearby ledgers (e.g. 20,980,000)
* the request is not forwarded to rippled (served entirely by Clio)

This suggests a Clio-side bug or edge case in historical `account_info` handling for that ledger.

## Steps to Reproduce

repro call 

```
{
  "method": "account_info",
  "params": [
    {
      "account": "r437okHk47PYQrFvkstnYwKwHoxzPcW36b",
      "ledger_index": 20880000
    }
  ]
}
```

## Steps to conclusion

If you parse `20890000` which returns valid data. 

Here are list of checks 

#### 1) Ledger exists and is validated 

(ledger RPC succeeds)

```
ledger_index: 20880000
ledger_hash: 787BFB87C46DD23F...
validated: true
```

#### 2) State tree is readable at this ledger

Returns valid state entries (AccountRoot, Offer, RippleState, etc.)

```
ledger_data (ledger_hash=787BFB87C46D...)
```

#### 3) `account_info` fails at ledger 20,880,000

```
account_info
ledger_index: 20880000
```

Response:
Error occurs for multiple accounts, including long-lived account:
`rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh`

```
{
  "error": "internal",
  "error_code": 73
}
```

#### 4) Same call succeeds at nearby ledger

Returns valid AccountRoot.

```
account_info
ledger_index: 20980000
```

#### 5) Not forwarded to rippled (Clio-only failure)

Using server_info counters before/after each call:

* Ledger 20,880,000

```
DELTA:
{ "started": 0, "finished": 0, "errored": 0, "forwarded": 0 }
```

* Ledger 20,980,000

```
DELTA:
{ "started": 1, "finished": 1, "errored": 0, "forwarded": 0 }
```

Confirms both requests were handled by Clio locally.
Failure path at 20,880,000 does not even increment `account_info` metrics.

#### 6) Local rippled cannot serve this ledger (expected)

```
account_info via localhost:5005
ledger_index: 20880000
```

Returns: 

```
"error": "lgrNotFound"
```

Confirms historical data is not coming from local rippled.

### Conclusion

Clio can serve:
* ledger
* ledger_data

But fails:
* `account_info` for ledger 20,880,000 only.

This is not:
* missing historical data
* rippled limitation
* ETL forwarding issue

It appears to be a Clio internal lookup/indexing bug or edge case affecting `account_info` at that specific historical ledger.

godexsoft:
Fixes #2881

From my tests the output of `account_info` is now identical with `xrpld`.
I'd like to investigate this further in the future and possibly find a better fix if there is one in #2895

### DIFF
--- a/src/data/AmendmentCenter.cpp
+++ b/src/data/AmendmentCenter.cpp
@@ -146,9 +146,12 @@ AmendmentCenter::isEnabled(AmendmentKey const& key, uint32_t seq) const
 bool
 AmendmentCenter::isEnabled(boost::asio::yield_context yield, AmendmentKey const& key, uint32_t seq) const
 {
-    if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments)
-        return lookupAmendment(all_, *listAmendments, key);
-
+    try {
+        if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments)
+            return lookupAmendment(all_, *listAmendments, key);
+    } catch (std::runtime_error const&) {
+        return false;  // Some old ledger does not contain Amendments ledger object so do best we can for now
+    }
     return false;
 }
 
@@ -157,13 +160,19 @@ AmendmentCenter::isEnabled(boost::asio::yield_context yield, std::vector<Amendme
 {
     namespace rg = std::ranges;
 
-    if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments) {
-        std::vector<bool> out;
-        rg::transform(keys, std::back_inserter(out), [this, &listAmendments](auto const& key) {
-            return lookupAmendment(all_, *listAmendments, key);
-        });
+    try {
+        if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments) {
+            std::vector<bool> out;
+            rg::transform(keys, std::back_inserter(out), [this, &listAmendments](auto const& key) {
+                return lookupAmendment(all_, *listAmendments, key);
+            });
 
-        return out;
+            return out;
+        }
+    } catch (std::runtime_error const&) {
+        return std::vector<bool>(
+            keys.size(), false
+        );  // Some old ledger does not contain Amendments ledger object so do best we can for now
     }
 
     return std::vector<bool>(keys.size(), false);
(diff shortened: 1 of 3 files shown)


src/data/AmendmentCenter.cpp
```diff
@@ -146,9 +146,12 @@ AmendmentCenter::isEnabled(AmendmentKey const& key, uint32_t seq) const
 bool
 AmendmentCenter::isEnabled(boost::asio::yield_context yield, AmendmentKey const& key, uint32_t seq) const
 {
-    if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments)
-        return lookupAmendment(all_, *listAmendments, key);
-
+    try {
+        if (auto const listAmendments = fetchAmendmentsList(yield, seq); listAmendments)
+            return lookupAmendment(all_, *listAmendments, key);
+    } catch (std::runtime_error const&) {
```
mathbunnyru, line 152: Using exceptions for control flow is not great.
Can we change `fetchAmendmentsList` and/or `lookupAmendment` implementations, so they give better result, rather than relying on exceptions?

Catching general runtime_error exception can potentially also hide lots of other issues

mathbunnyru (changes requested):

src/data/AmendmentCenter.cpp
godexsoft: Indeed, we can do all that in #2895 as mentioned in the description.

mathbunnyru: Ok, I think it's a bit dangerous to make last-minute changes like this, but let's try it out

mathbunnyru (approved):

kuznetsss (approved):

PR #2904, opened 2026-01-12
ci: Pass version explicitly and don't rely on tags

mathbunnyru:
Relying on tags doesn't work when making nightly releases.
If we make a nightly release first time, version will be sth like `develop-<date>-<short-rev>`.
But if we don't commit for a while and nightly runs again, we will see previously created tag, and will use it.

This PR makes our build process not rely on tags, and purely rely on the type of event being triggered and passes this information to CMake explicitly.

### DIFF
--- a/.github/actions/cmake/action.yml
+++ b/.github/actions/cmake/action.yml
@@ -38,7 +38,7 @@ inputs:
     required: true
     default: "false"
   version:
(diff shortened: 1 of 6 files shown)


godexsoft (approved):
:rage1:

PR #2905, opened 2026-01-12
fix: No output from failed asserts in tests

kuznetsss:


mathbunnyru (dismissed):

src/util/Assert.cpp
```diff
@@ -54,7 +54,7 @@ OnAssert::resetAction()
 void
 OnAssert::defaultAction(std::string_view message)
 {
-    if (LogServiceState::initialized()) {
+    if (LogServiceState::initialized() and LogServiceState::hasSinks()) {
         LOG(LogService::fatal()) << message;
     } else {
         std::cerr << message;
```
mathbunnyru:
```suggestion
        std::cerr << message << std::endl;
```

Let's make it symmetric with logger, to always have newline

mathbunnyru (approved):

PR #2923, opened 2026-01-15
ci: Restart colima on macOS

mathbunnyru:


godexsoft (approved):
This fixes the build and seems like the only option to produce a release. Note we are not changing any code so this is safe.

PR #2947, opened 2026-02-13
feat: Metrics for requested ledger age

kuznetsss:
Adding metrics to be able to analyse requested ledger age distribution.


src/rpc/Counters.cpp
```diff
@@ -217,6 +260,67 @@ Counters::onInternalError()
...
 
+void
+Counters::recordLedgerRequest(boost::json::object const& params, std::uint32_t currentLedgerSequence)
+{
+    // Determine the requested ledger type
+    std::optional<std::uint32_t> requestedLedgerSeq;
+    bool isCurrent = false;
+    bool isValidated = false;
+
+    if (params.contains("ledger_index")) {
```
godexsoft: This looks like some generic helper we might already have somewhere. and if we don't then maybe this is the time to extract it

src/rpc/Counters.cpp
```diff
@@ -217,6 +260,67 @@ Counters::onInternalError()
...
+    } else if (requestedLedgerSeq.has_value()) {
+        ++ledgerSpecificCounter_.get();
+
+        // Calculate age and update histograms
+        if (*requestedLedgerSeq <= currentLedgerSequence) {
+            auto const ageLedgers = static_cast<std::int64_t>(currentLedgerSequence - *requestedLedgerSeq);
+            ledgerAgeLedgersHistogram_.get().observe(ageLedgers);
+
+            // Estimate age in seconds (assuming ~4 seconds per ledger)
+            static constexpr std::int64_t kSECONDS_PER_LEDGER = 4;
```
godexsoft: Not sure this is a good approach.. it's cheap one but not accurate i think

godexsoft (commented):
Looks pretty useful overall!

src/web/ng/RPCServerHandler.hpp
```diff
@@ -274,6 +274,7 @@ class RPCServerHandler {
             } else {
                 // This can still technically be an error. Clio counts forwarded requests as successful.
                 rpcEngine_->notifyComplete(context->method, us);
+                rpcEngine_->recordLedgerMetrics(context->params, context->range.maxSequence);
```
kuznetsss: This should be called from `notifyComplete()`

src/rpc/Counters.hpp
```diff
@@ -66,6 +68,13 @@ class Counters {
     CounterType unknownCommandCounter_;
     CounterType internalErrorCounter_;
 
+    // ledger request tracking metrics
+    CounterType ledgerCurrentCounter_;
+    CounterType ledgerValidatedCounter_;
+    CounterType ledgerSpecificCounter_;
+    std::reference_wrapper<util::prometheus::HistogramInt> ledgerAgeSecondsHistogram_;
```
kuznetsss: We don't need this metric

src/rpc/Counters.cpp
```diff
@@ -217,6 +260,67 @@ Counters::onInternalError()
...
+            auto const parsed = util::getLedgerIndex(indexValue);
+            if (parsed.has_value()) {
+                requestedLedgerSeq = *parsed;
+            }
+        }
+    } else if (params.contains("ledger_hash")) {
+        // For hash-based requests, we can't easily determine age without additional lookup
+        // Count it as "specific" but don't add to histogram
+        ++ledgerSpecificCounter_.get();
+        return;
```
kuznetsss: I think we shouldn't do it because we will not be able to estimate how many items are on the histogram

godexsoft (dismissed):
👍 :rage1:

src/rpc/Counters.cpp
```diff
@@ -217,6 +231,27 @@ Counters::onInternalError()
     ++internalErrorCounter_.get();
 }
 
+void
+Counters::recordLedgerRequest(boost::json::object const& params, std::uint32_t currentLedgerSequence)
+{
+    if (not params.contains("ledger_index")) {
```
godexsoft: There is also "ledger_hash" that can be used to refer to the same ledger.. not sure how to deal with it though.

godexsoft (commented):
I found one more potential thing to fix but i don't have a solution for it.

src/rpc/Counters.cpp
kuznetsss: I don't think there is an easy way to get ledger index by hash and querying DB or rippled for that is probably too heavy operation.

kuznetsss: Added a counter for requests with `ledger_hash`.

godexsoft (dismissed):
:rage1:

godexsoft (approved):
👍

PR #2963, opened 2026-02-23
style: Fix lint comments due to style changes

mathbunnyru:


src/etl/NFTHelpers.cpp
```diff
@@ -137,7 +137,6 @@ getNFTokenMintData(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
 
     // Find the first NFT ID that doesn't match.  We're looking for an
     // added NFT, so the one we want will be the mismatch in finalIDs.
-    // NOLINTNEXTLINE(modernize-use-ranges)
```
mathbunnyru, line 140: This was the reason it didn't trigger before

kuznetsss (approved):

PR #2972, opened 2026-03-03
feat: Implement VaultList for Lending Protocol

mathbunnyru:


src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    auto const addToResponse = [&](ripple::SLE&& sle) {
+        if (sle.getType() != ripple::ltVAULT)
+            return true;
+
+        boost::json::object summary;
+
+        auto vaultSle = std::move(sle);
+
+        summary["vault_id"] = ripple::strHex(vaultSle.key());
```
mathbunnyru: Magic constants - will get rid of these

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,218 @@
...
+        *sharedPtrBackend_, ctx.yield, input.ledgerHash, input.ledgerIndex, range->maxSequence
+    );
+
+    if (!expectedLgrInfo.has_value())
+        return Error{expectedLgrInfo.error()};
+
+    auto const& lgrInfo = *expectedLgrInfo;
+
+    // Parse the MPT Issuance ID from input
+    auto const mptID = ripple::uint192{input.tokenID.c_str()};
```
mathbunnyru, line 57: Will check if this is correct way to name/get the id

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <xrpl/protocol/Serializer.h>
+#include <xrpl/protocol/jss.h>
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+namespace rpc {
+
+VaultListHandler::VaultListHandler(std::shared_ptr<BackendInterface> const& sharedPtrBackend)
```
godexsoft: nit: Prefer pass by value and move

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+VaultListHandler::process(VaultListHandler::Input const& input, Context const& ctx) const
+{
+    auto const range = sharedPtrBackend_->fetchLedgerRange();
+    ASSERT(range.has_value(), "VaultList's ledger range must be available");
+
+    auto const expectedLgrInfo = getLedgerHeaderFromHashOrSeq(
+        *sharedPtrBackend_, ctx.yield, input.ledgerHash, input.ledgerIndex, range->maxSequence
+    );
+
+    if (!expectedLgrInfo.has_value())
```
godexsoft: nit: We like to use `not` when it reads more like plain english: `if (not expectedLgrInfo.has_value())` for example

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+        return Error{expectedLgrInfo.error()};
+
+    auto const& lgrInfo = *expectedLgrInfo;
+
+    // Parse the MPT Issuance ID from input
+    auto const mptID = ripple::uint192{input.tokenID.c_str()};
```
godexsoft, line 57: I wonder if there is a better constructor we can use (one that is not taking char*)

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    auto const issuerAccountID = issuanceSle.getAccountID(ripple::sfIssuer);
+
+    // Verify the issuer account exists
+    auto const accountKeylet = ripple::keylet::account(issuerAccountID);
+    auto const accountLedgerObject =
+        sharedPtrBackend_->fetchLedgerObject(accountKeylet.key, lgrInfo.seq, ctx.yield);
+
+    if (!accountLedgerObject)
+        return Error{Status{RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
```
godexsoft: Do we need to add the custom `accountNotFound` message? is it not the default already?

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+        auto vaultSle = std::move(sle);
+
+        summary[JS(vault_id)] = ripple::strHex(vaultSle.key());
+        summary[JS(account)] = ripple::toBase58(vaultSle.getAccountID(ripple::sfAccount));
+        summary[JS(owner)] = ripple::toBase58(vaultSle.getAccountID(ripple::sfOwner));
+
+        // AssetsTotal is an STNumber field; extract via JSON serialization
+        auto const vaultJson = toBoostJson(vaultSle.getJson(ripple::JsonOptions::none));
+        if (vaultJson.as_object().contains("AssetsTotal")) {
+            summary["total_assets"] = vaultJson.as_object().at("AssetsTotal");
```
godexsoft, line 103: nit: I assume there is no JS for these?

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+            };
+            summary["total_shares"] = shareSle.getFieldU64(ripple::sfOutstandingAmount);
+        } else {
+            summary["total_shares"] = std::uint64_t{0};
+        }
+
+        // Status derived from flags
+        auto const flags = vaultSle.getFlags();
+        // A vault with no special flags set is considered active
+        summary[JS(status)] = (flags == 0) ? "active" : "modified";
```
godexsoft, line 126: nit: No JS for active/modified?

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+        issuerAccountID,
+        lgrInfo.seq,
+        input.limit,
+        input.marker,
+        ctx.yield,
+        addToResponse,
+        false
+    );
+
+    if (!expectedNext.has_value())
```
godexsoft: nit: `not` would read better

src/rpc/handlers/VaultList.cpp
```diff
@@ -0,0 +1,219 @@
...
+    VaultListHandler::Output const& output
+)
+{
+    jv = boost::json::object{
+        {"token_id", output.tokenID},
+        {JS(ledger_hash), output.ledgerHash},
+        {JS(ledger_index), output.ledgerIndex},
+        {JS(validated), output.validated},
+        {JS(limit), output.limit},
+        {"vaults", output.vaults},
```
godexsoft, line 167: nit: No JS string for this and token_id above?

tests/unit/rpc/handlers/VaultListTests.cpp
```diff
@@ -0,0 +1,797 @@
...
+constexpr auto kASSET_CURRENCY = "USD";
+constexpr auto kASSET_ISSUER = "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW";
+constexpr uint32_t kSEQ = 30;
+
+}  // namespace
+
+struct RPCVaultListHandlerTest : HandlerBaseTest {
+    RPCVaultListHandlerTest()
+    {
+        backend_->setRange(10, kSEQ);
```
godexsoft: nit: Why 10 here is not a kSEQ_START or similar?

tests/unit/rpc/handlers/VaultListTests.cpp
```diff
@@ -0,0 +1,797 @@
...
+        auto const output = handler.process(req, Context{.yield = yield, .apiVersion = 2});
+        ASSERT_FALSE(output);
+
+        auto const err = rpc::makeError(output.result.error());
+        EXPECT_EQ(err.at("error").as_string(), testBundle.expectedError);
+        EXPECT_EQ(err.at("error_message").as_string(), testBundle.expectedErrorMessage);
+    });
+}
+
+// -- Ledger not found --
```
godexsoft: Not sure these kind of comments are too useful. I'll leave it up to you to decide if you want to keep them.

tests/unit/rpc/handlers/VaultListTests.cpp
```diff
@@ -0,0 +1,797 @@
...
+            R"JSON({{
+                "token_id": "{}"
+            }})JSON",
+            kMPT_ID
+        )
+    );
+
+    auto const handler = AnyHandler{VaultListHandler{backend_}};
+    runSpawn([&](auto yield) {
+        auto const output = handler.process(kINPUT, Context{.yield = yield, .apiVersion = 2});
```
godexsoft: Seems like .apiVersion is always 2 - let's extract as constant?

godexsoft (changes requested):
Overall this is looking good. Leaving a few questions and mostly nits.

src/rpc/handlers/VaultList.cpp
mathbunnyru: Let's merge this first then, and then I'll fix here: https://github.com/XRPLF/clio/pull/2974

src/rpc/handlers/VaultList.cpp
mathbunnyru: https://github.com/XRPLF/clio/pull/2977

src/rpc/handlers/VaultList.cpp
mathbunnyru: Remember the problem in base_uint? There will be a better ctor once we fix the ctor upstream

src/rpc/handlers/VaultList.cpp
mathbunnyru: The default message is `Account not found.`, which is as reasonable, so no need to provide custom string here, fixed, thanks

mathbunnyru: https://github.com/XRPLF/clio/pull/2978

src/rpc/handlers/VaultList.cpp
mathbunnyru: unfortunately, no :(

src/rpc/handlers/VaultList.cpp
mathbunnyru: I will check how to correctly name them, there is already `closed` with no JS for the status in the repo

src/rpc/handlers/VaultList.cpp
mathbunnyru: https://github.com/XRPLF/clio/pull/2979

src/rpc/handlers/VaultList.cpp
mathbunnyru: Unfortunately, not :(

src/rpc/handlers/VaultList.cpp
mathbunnyru: @Tapanito what's the best way to call this status?
The 0 is the default value from the XLS: https://github.com/XRPLF/XRPL-Standards/blob/master/XLS-0066-lending-protocol/README.md#322-fields

But I'm not sure what naming we should choose here

src/rpc/handlers/VaultList.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+
+namespace rpc {
+
+/**
+ * @brief The vault_list command retrieves all Vaults created by a particular Token.
```
Tapanito: Why is this RPC only for MPTs? WHy not IOUs and XRP?

Edit: a Token cannot create a Vault. I assume you meant to say that `vault_list` retrieves all vaults for a given token?

mathbunnyru: I fixed the comment, thanks. I'll take a look how to implement it for all token types in a similar fashion

mathbunnyru: I was looking at https://github.com/XRPLF/clio/blob/develop/src/rpc/handlers/VaultInfo.cpp#L138, which seems to deal with MPTs only

tests/unit/rpc/handlers/VaultListTests.cpp
mathbunnyru: https://github.com/XRPLF/clio/pull/3017

PR #2975, opened 2026-03-05
chore: Remove copyright from all source files

godexsoft:
XRPLF/rippled recently did the same so that we don't duplicate what is already in `LICENSE.md` file and more importantly don't pollute AI context when reviewing code and using AI to create new code.


mathbunnyru (commented):
GitHub says `+1 -15,078`, where does this `+1` comes from

emreariyurek:
> GitHub says `+1 -15,078`, where does this `+1` comes from

https://github.com/XRPLF/clio/pull/2975/changes#diff-4673a3aba01813b595de187a7a6e9e63a3491d55821606fecd9f13a10c188a1dR3 - present or forever :)

mathbunnyru:
> > GitHub says `+1 -15,078`, where does this `+1` comes from
> 
> [#2975 (changes)](https://github.com/XRPLF/clio/pull/2975/changes#diff-4673a3aba01813b595de187a7a6e9e63a3491d55821606fecd9f13a10c188a1dR3) - present or forever :)

Thanks! 👍

mathbunnyru (changes requested):
I think we should do it differently and add:
`// SPDX-License-Identifier: ISC` to every file

This is:
- machine-readable
- and easily implemented as a pre-commit hook
- recognised by scanners
- minimal file clutter
- widely used in large projects (e.g., Linux kernel, https://github.com/torvalds/linux/blob/master/io_uring/advise.c)
- looks like a better approach than removing the header completely
- also has almost zero effect on AI tokens

mathbunnyru (approved):

PR #2978, opened 2026-03-05
chore: Remove explicit accountNotFound message

mathbunnyru:
I searched rippled code, and there is no `accountNotFound` anywhere.
So, we should probably stick to the default, it looks reasonable


godexsoft (commented):
I hope this does not break any of parity checks @mounikakun has. Please confirm that the same RPCs in xrpld produce the same string.
Technically this would not be an api break as this is an error message, not the error code

src/rpc/handlers/AccountChannels.cpp
```diff
@@ -107,7 +107,7 @@ AccountChannelsHandler::process(
     );
 
     if (!accountLedgerObject)
-        return Error{Status{RippledError::rpcACT_NOT_FOUND, "accountNotFound"}};
+        return Error{Status{RippledError::rpcACT_NOT_FOUND}};
```
mathbunnyru, line 110: @godexsoft I verified that:
- `accountNotFound` string is not present at rippled at all
- The rippled code doesn't provide custom strings for this handler: https://github.com/XRPLF/rippled/blob/develop/src/xrpld/rpc/handlers/AccountChannels.cpp#L72
- And I tried:

https://xrpl.org/resources/dev-tools/websocket-api-tool#account_channels on devnet
request:
```
{
  "command": "account_channels",
  "account": "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun",
  "destination_account": "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun",
  "ledger_index": "validated"
}
```

s.devnet.rippletest.net (rippled): `"error_message": "Account not found."`
clio.devnet.rippletest.net (clio): `"error_message": "accountNotFound",`

So, after my change the string will match

godexsoft (approved):
As it matches `xrpld` output this is a good change :rage3:

PR #2985, opened 2026-03-06
feat: Limit cache loading in cluster

### LINKED ISSUE #2707: Limit of simultaneous cache loaders in cluster
## Summary

Clio cluster could automatically limit the number of simultaneous cache loaders using cluster communication.

kuznetsss:
This PR adds an option to limit simultaneous cache loading in a cluster to one node at a time.
Fixes #2707


src/cluster/CacheLoaderDecider.cpp
```diff
@@ -0,0 +1,92 @@
...
+{
+}
+
+void
+CacheLoaderDecider::onNewState(
+    ClioNode::CUuid selfId,
+    std::shared_ptr<Backend::ClusterData const> clusterData
+)
+{
+    if (not clusterData->has_value())
```
godexsoft, line 52: clusterData can't be nullptr here?

src/cluster/CacheLoaderDecider.cpp
```diff
@@ -0,0 +1,92 @@
...
+                return;
+
+            std::vector<ClioNode> notFullNodes;
+            std::ranges::copy_if(
+                clusterData, std::back_inserter(notFullNodes), [](ClioNode const& node) {
+                    return not node.cacheIsFull;
+                }
+            );
+
+            if (std::ranges::any_of(notFullNodes, [](ClioNode const& node) {
```
godexsoft: I feel like this could better be stored in a constant before the `if` - it's a bit messy as it is, wdyt?

src/cluster/CacheLoaderDecider.cpp
```diff
@@ -0,0 +1,92 @@
...
+                    return node.cacheIsCurrentlyLoading;
+                })) {
+                return;
+            }
+
+            std::ranges::sort(notFullNodes, [](ClioNode const& lhs, ClioNode const& rhs) {
+                return *lhs.uuid < *rhs.uuid;
+            });
+
+            if (*notFullNodes.front().uuid == *selfId) {
```
godexsoft, line 86: I assume clang-tidy would do this but it looks like curlies are redundant for a one liner `if` here

src/cluster/CacheLoaderDecider.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * 1. Doing nothing if this node's cache is already full
+ * 2. Doing nothing if any node in the cluster is currently loading the cache
+ * 3. Sorting all nodes whose cache is not yet full by UUID for deterministic ordering
+ * 4. Permitting loading on this node if it is the first in the sorted list
+ *
+ * This ensures at most one node in the cluster loads the cache at a time.
+ */
+class CacheLoaderDecider {
+    /** @brief Thread pool for spawning asynchronous tasks */
+    boost::asio::thread_pool& ctx_;
```
godexsoft, line 46: You know how i feel about reference members, but see for yourself if in this case it's ok or not.

src/cluster/ClioNode.cpp
```diff
@@ -77,35 +85,49 @@ tag_invoke(boost::json::value_from_tag, boost::json::value& jv, ClioNode const&
...
     }
 
-    auto const dbRoleValue = jv.as_object().at(JsonFields::kDB_ROLE).as_int64();
-    if (dbRoleValue > static_cast<int64_t>(ClioNode::DbRole::MAX))
-        throw std::runtime_error("Invalid db_role value");
+    // Each field has a default value for backward compatibility
+    auto dbRole = ClioNode::DbRole::Fallback;
+    if (auto const* v = obj.if_contains(JsonFields::kDB_ROLE)) {
+        auto const dbRoleValue = v->as_int64();
+        if (dbRoleValue > static_cast<int64_t>(ClioNode::DbRole::MAX))
```
godexsoft, line 108: Shouldn't this include MAX?

godexsoft (changes requested):
Just a few nits and questions 👍

src/cluster/CacheLoaderDecider.cpp
kuznetsss: No, we always put non-empty shared_ptr here.
https://github.com/XRPLF/clio/blob/f82196d9a2c8b92ae3e4a03cbc18d81b846f2350/src/cluster/Backend.cpp#L64

src/cluster/CacheLoaderDecider.cpp
kuznetsss: I think our rule is that `{}` is mandatory for multiline but optional for a single line

src/cluster/ClioNode.cpp
kuznetsss: No, `MAX` has the same value as `Fallback`

src/cluster/CacheLoaderDecider.cpp
kuznetsss: Fixed

src/cluster/CacheLoaderDecider.cpp
godexsoft: Multiline wont do the right thing without curlies, yes. I think you meant `if/else` vs just a single line `if`. Either way it's up to you 👍

godexsoft (approved):
:rage1:

PR #3006, opened 2026-03-19
fix: Resolve proxy ip before processing any request

kuznetsss:


src/web/impl/HttpBase.hpp
```diff
@@ -241,6 +241,16 @@ class HttpBase : public ConnectionBase {
         if (ec)
             return httpFail(ec, "read");
 
+        if (auto resolvedIp = proxyIpResolver_->resolveClientIp(clientIp_, req_);
```
godexsoft, line 244: nit: `auto const`?

godexsoft (approved):
One nit, otherwise good to go 👍

PR #3010, opened 2026-03-23
fix: Forwarded stream ignores specified `api_version`

### LINKED ISSUE #3007: Inconsistent format in `transactions_proposed` stream
## Issue Description

There is a report on X (https://x.com/i/status/2035005550174306789) about inconsistent subscription behaviour with specified api_version:

We may have found an unexpected behaviour in XRPL transaction streams when using `transactions_proposed` with `api_version=2`.

According to the docs, proposed transactions should only differ by:

`validated = false`
no `meta`
`ledger_current_index` instead of `ledger_index`
Doc: [https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/subscription-methods/subscribe#transaction-streams](https://t.co/9j83tgWVF8)

However, when subscribing with API v2, we receive mixed formats:

some messages use `tx_json`
some use `transaction`

This makes parsing inconsistent.

Interesting detail:
This behavior seems to happen on public servers running Clio (s1 / s2),
while the public xrplcluster node returns consistent format.
Not sure if this is expected behaviour, backward compatibility, or a Clio-specific issue.

---
Needs investigation.

godexsoft:
Fixes #3007 

The core issue is that `api_version` was ignored when forwarding xrpld's stream. 
This PR adds "repackaging" into api v2 format when needed.


src/feed/impl/ProposedTransactionFeed.hpp
```diff
@@ -46,9 +47,27 @@ namespace feed::impl {
...
  * @brief Feed that publishes the Proposed Transactions.
- * @note Be aware that the Clio only forwards this stream, not respect api_version.
  */
 class ProposedTransactionFeed {
+    // Hold two versions of transaction messages: [0] = v1, [1] = v2
+    using AllVersionMsgsType = std::array<std::shared_ptr<std::string>, 2>;
+
+    struct ProposedTransactionSlot {
+        std::reference_wrapper<ProposedTransactionFeed> feed;
+        std::weak_ptr<Subscriber> subscriptionContextWeakPtr;
```
kuznetsss: Shouldn't the fields be private, or do we need an access to them outside of the class?

src/feed/impl/ProposedTransactionFeed.hpp
```diff
@@ -46,9 +47,27 @@ namespace feed::impl {
 
 /**
  * @brief Feed that publishes the Proposed Transactions.
- * @note Be aware that the Clio only forwards this stream, not respect api_version.
  */
 class ProposedTransactionFeed {
+    // Hold two versions of transaction messages: [0] = v1, [1] = v2
+    using AllVersionMsgsType = std::array<std::shared_ptr<std::string>, 2>;
```
kuznetsss: It seems we could use `std::shared_ptr<AllVersionsMsgsType>` instead and it will save one dynamic allocation. Where
```cpp
struct AllVersionsMsgsType {
  std::string v1;
  std::string v2;
};
```

src/feed/impl/ProposedTransactionFeed.hpp
```diff
@@ -57,10 +76,16 @@ class ProposedTransactionFeed {
     std::reference_wrapper<util::prometheus::GaugeInt> subAllCount_;
     std::reference_wrapper<util::prometheus::GaugeInt> subAccountCount_;
 
-    TrackableSignalMap<ripple::AccountID, Subscriber, std::shared_ptr<std::string>> accountSignal_;
-    TrackableSignal<Subscriber, std::shared_ptr<std::string>> signal_;
+    TrackableSignalMap<ripple::AccountID, Subscriber, AllVersionMsgsType const&> accountSignal_;
+    TrackableSignal<Subscriber, AllVersionMsgsType const&> signal_;
```
kuznetsss: Why is it const ref now?

kuznetsss (commented):
Looks good, a few questions.

src/feed/impl/ProposedTransactionFeed.hpp
godexsoft: It's a copy of what we already did in `TransactionFeed`. Let me try to refactor both i guess.

kuznetsss (approved):
👍

PR #3011, opened 2026-03-24
fix: Allow deleted libXRPL amendments to still be queried via `feature`

godexsoft:
Recently some amendments in libXRPL were completely deleted. This made it impossible to query them in Clio, even for historical ledger indices where this amendment was still enabled. While we can't know whether libXRPL "supported" some amendment at any given ledger index, we sure can find out whether it was enabled at any point. It makes sense to allow Clio to keep this knowledge for historical queries.


kuznetsss (dismissed):
👍 
Can we add tests for this change?

kuznetsss (approved):

PR #3016, opened 2026-03-25
feat: Optional log rotation

kuznetsss:
This PR adds an option to disable log rotation.


src/util/log/Logger.cpp
```diff
@@ -184,12 +185,20 @@ spdlog::sink_ptr
 LogService::createFileSink(FileLoggingParams const& params, std::string const& format)
 {
     std::filesystem::path const dirPath(params.logDir);
-    // the below are taken from user in MB, but spdlog needs it to be in bytes
-    auto const rotationSize = mbToBytes(params.rotationSizeMB);
+    std::shared_ptr<spdlog::sinks::sink> fileSink;
```
godexsoft: I'd prefer this to be initialized by immediately invoked lambda, wdyt? the path can also be computed once as a capture ```[path = (dirPath / "clio.log").string()] { ... }()```

tests/unit/util/log/LoggerTests.cpp
```diff
@@ -91,3 +103,138 @@ TEST_F(LoggerTest, ManyDynamicLoggers)
...
+/**
+ * @brief Fixture for testing real log-file rotation behaviour.
+ *
+ * Unlike LoggerTest (which uses LoggerFixture's in-memory buffer), this fixture
+ * initialises the LogService with a real file sink and redirects all spdlog
+ * loggers to that sink so that written messages actually land on disk.
+ */
+struct LogFileRotationTests : ::testing::Test {
+    std::filesystem::path const tmpDir =
+        std::filesystem::temp_directory_path() / "clio_log_rotation_tests";
```
godexsoft: Let's make sure temp_directory_path is actually unique every run

godexsoft (changes requested):
Some small suggestions 👍

godexsoft (dismissed):
Cool 👍

godexsoft (dismissed):

PR #3044, opened 2026-04-24
refactor: Use error code in `make_address()` calls

kuznetsss:
Function `ip::make_address()` throws an exception on invalid IP. Refactor to a better error handling without exceptions.

### DIFF
--- a/src/app/ClioApplication.cpp
+++ b/src/app/ClioApplication.cpp
@@ -222,10 +222,13 @@ ClioApplication::run(bool const useNgWebServer)
         config_, backend, rpcEngine, etl, dosGuard
     );
 
-    auto const httpServer = web::makeHttpServer(config_, ioc, dosGuard, handler, cache);
+    auto const expectedHttpServer = web::makeHttpServer(config_, ioc, dosGuard, handler, cache);
+    if (not expectedHttpServer.has_value())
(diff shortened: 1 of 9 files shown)


src/app/ClioApplication.cpp
```diff
@@ -222,10 +227,15 @@ ClioApplication::run(bool const useNgWebServer)
...
+    auto const expectedHttpServer = web::makeHttpServer(config_, ioc, dosGuard, handler, cache);
+    if (not expectedHttpServer.has_value()) {
+        LOG(util::LogService::fatal()) << expectedHttpServer.error();
+        return EXIT_FAILURE;
+    }
+
     appStopper_.setOnStop(
         Stopper::makeOnStopCallback(
-            *httpServer,
+            **expectedHttpServer,
```
godexsoft, line 238: this looks odd. probably technically correct though :/

godexsoft (approved):
Looks good! one off looking line spotted but not sure if we can improve it somehow

src/app/ClioApplication.cpp
kuznetsss: One star for the `expected` where the errors lie,
One star for the `shared_ptr` before its ref count dies,
One Segfault to rule them all, One Core Dump to find them,
One Debugger to bring them all, and in the backtrace bind them,
In the Land of Runtime where the Shadows lie.

PR #3054, opened 2026-04-30
chore: Enable more clang-tidy checks

godexsoft:
This PR enables more checks to align with XRPLF/xrpld.
For optional access check, in most places a NOLINT comment has been added. The access itself was safe already. `.value()` was mostly updated to `*opt` squeezing out a little extra performance.


godexsoft:
A bunch of nolint comments are misaligned atm after clang-format ran. I'll fix that.

src/data/BackendInterface.cpp
```diff
@@ -188,15 +188,16 @@ BackendInterface::fetchBookOffers(
         auto mid2 = std::chrono::system_clock::now();
         numSucc++;
         succMillis += getMillis(mid2 - mid1);
-        if (!offerDir || offerDir->key >= bookEnd) {
+        if (!offerDir || (*offerDir).key >= bookEnd) {
```
kuznetsss: `*` and `->` are exactly the same, so there is no point of replacing one to another. `(*offerDir).key` is less readable comparing to `offerDir->key`

.clang-tidy
```diff
@@ -64,16 +66,17 @@ Checks: "-*,
...
   bugprone-unhandled-exception-at-new,
   bugprone-unhandled-self-assignment,
   bugprone-unique-ptr-array-mismatch,
   bugprone-unsafe-functions,
-  bugprone-unused-local-non-trivial-variable,
+  bugprone-use-after-move,
   bugprone-unused-raii,
   bugprone-unused-return-value,
-  bugprone-use-after-move,
+  bugprone-unused-local-non-trivial-variable,
```
mathbunnyru: This order was correct, you probably need to change it in rippled instead

.clang-tidy
```diff
@@ -105,11 +109,15 @@ Checks: "-*,
...
   modernize-use-equals-default,
   modernize-use-equals-delete,
+  modernize-use-nodiscard,
   modernize-use-override,
   modernize-use-ranges,
+  modernize-use-scoped-lock,
   modernize-use-starts-ends-with,
   modernize-use-std-numbers,
   modernize-use-using,
+  modernize-deprecated-headers,
```
mathbunnyru: not sorted order

.clang-tidy
```diff
@@ -105,11 +109,15 @@ Checks: "-*,
...
+  llvm-namespace-comment,
```
mathbunnyru: this duplicates the flag above

.clang-tidy
```diff
@@ -5,6 +5,7 @@ Checks: "-*,
   bugprone-bad-signal-to-kill-thread,
   bugprone-bool-pointer-implicit-conversion,
   bugprone-casting-through-void,
+  bugprone-capturing-this-in-member-variable,
```
mathbunnyru, line 7: not sorted

.clang-tidy
```diff
@@ -26,6 +27,7 @@ Checks: "-*,
   bugprone-misplaced-operator-in-strlen-in-alloc,
   bugprone-misplaced-pointer-arithmetic-in-alloc,
   bugprone-misplaced-widening-cast,
+  bugprone-misleading-setter-of-reference,
```
mathbunnyru: not sorted

src/data/BackendInterface.cpp
godexsoft: Yeah, i'm not sure why Claude thought this was a good idea. I'd expect it to only change the `.value()`. I'll bring those more readable `->` back.

godexsoft: I think i understand now. There is a check that reports suspicious use of 'reset()' through '->' on types like optional/shared_ptr etc. and for that check, clang-tidy will automatically change it to `(*obj).reset();` to make it clear we are not trying to call optional's member function `reset` but rather the member function of our type stored in the optional. So i guess Claude thought we should apply this pattern everywhere.. which is nonsense.

src/data/BackendInterface.cpp
```diff
@@ -208,6 +208,8 @@ BackendInterface::fetchBookOffers(
             auto nextKey = ripple::keylet::page(uTipIndex, next);
             auto nextDir = fetchLedgerObject(nextKey.key, ledgerSequence, yield);
             ASSERT(nextDir.has_value(), "Next dir must exist");
+            if (!nextDir)
+                break;
```
mathbunnyru: Don't our asserts throw?
Maybe it would be better to have a NOLINT comment instead of `if`

mathbunnyru: If you could somehow check for other places like that this PR changes, that would be great

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1002,7 +1003,7 @@ class CassandraBackendFamily : public BackendInterface {
             return false;
         }
 
-        if (not maybeSuccess.value()) {
+        if (not *maybeSuccess) {
```
kuznetsss, line 1006: I'm thinking now, is it a useful check? The performance benefit will not be noticeable but if something changes in the code and the optional will become empty `.value()` will throw so it will be visible at least in some way while `*` on empty optional is UB.

src/data/BackendInterface.cpp
kuznetsss: ASSERT() calls std::abort. Looks like we don't need to add this if.

godexsoft: This if is what makes it possible to not have to nolint, but i agree that it's better to nolint in this case too 👍

src/data/cassandra/CassandraBackendFamily.hpp
godexsoft: I'm not sure i understand. it can't be UB because we have this just above: 
```cpp
    if (not maybeSuccess) {
        LOG(log_.error()) << "executeSyncUpdate - error getting result - no row";
        return false;
    }

    if (not *maybeSuccess) {
```

mathbunnyru (approved):

src/data/cassandra/CassandraBackendFamily.hpp
kuznetsss: I'm saying imagine we add some code resetting it or somehow add a case so it could be `nullopt` and forget to update `*maybeSuccess` (not exactly in this place, but in general). Having `.value()` before would throw in such case, so it is easier to notice. What is the benefit we get of having `*` instead of `.value()`?

godexsoft: It's a tiny bit faster as we don't need to check internally and throw. Idk., i'd just leave it as is now.

kuznetsss (approved):

PR #3058, opened 2026-05-01
fix: Don't cache responses to admin's requests

kuznetsss:


src/etl/LoadBalancer.cpp
```diff
@@ -275,7 +275,7 @@ LoadBalancer::forwardToRippled(
         return std::unexpected{rpc::ClioError::RpcCommandIsMissing};
 
     auto const cmd = boost::json::value_to<std::string>(request.at("command"));
-    if (forwardingCache_) {
+    if (forwardingCache_ and not isAdmin) {
```
mathbunnyru: You should probably use new method `shouldUseCache` here and in the next diff :)

mathbunnyru (changes requested):

src/etl/LoadBalancer.cpp
kuznetsss: Fixed

mathbunnyru (dismissed):
LGTM, let's wait for clang-tidy though, please

mathbunnyru (approved):

PR #3061, opened 2026-05-02
chore: Fix clang-tidy

godexsoft:
This PR fixes the remaining issues clang-tidy found in CI. 
According to the docs ASSERT_TRUE should work (in v23 onwards). ASSERT_TRUE(has_value()) should be marked as 'safe' for optional access check. For now though we still need the NOLINTS as well.


tests/unit/rpc/common/CheckersTests.cpp
```diff
@@ -22,7 +22,6 @@ TEST_F(DeprecatedTests, Field)
 {
     auto warning = Deprecated<>::check(json, "some_string");
     ASSERT_TRUE(warning);
```
mathbunnyru: Let's replace this with `.has_value()` in this file

tests/unit/util/config/ClioConfigDefinitionTests.cpp
```diff
@@ -482,11 +482,10 @@ TEST(ClioConfigDefinitionParse, unexpectedFields)
...
-        // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
-        *result,
-        [](auto const& lhs, auto const& rhs) { return lhs.error < rhs.error; }
-    );
+    ASSERT_TRUE(result.has_value());
+    std::ranges::sort(*result, [](auto const& lhs, auto const& rhs) {
+        return lhs.error < rhs.error;
+    });
     ASSERT_TRUE(result.has_value());
     // NOLINTBEGIN(bugprone-unchecked-optional-access)
```
mathbunnyru, line 492: This looks like it could also be deleted.

I think it's worth searching for remaining `bugprone-unchecked-optional-access` and checking that they are needed

mathbunnyru (changes requested):

tests/unit/util/config/ClioConfigDefinitionTests.cpp
godexsoft: So turns out none of my ASSERT_TRUE actually works without NOLINT in clang-tidy 21. We will need to wait for v23 for this 😢 
I added back all the NOLINTs but also kept the ASSERT_TRUE so we are a bit more prepared for the future.

tests/unit/migration/MigratorRegisterTests.cpp
```diff
@@ -250,19 +257,20 @@ TEST_F(MultipleMigratorRegisterTests, MigrateNormalMigrator)
 
 TEST_F(MultipleMigratorRegisterTests, canBlock)
 {
+    ASSERT_TRUE(migratorRegister.has_value());
     // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
     auto canBlock = migratorRegister->canMigratorBlockClio("SimpleTestMigrator");
-    EXPECT_TRUE(canBlock);
+    ASSERT_TRUE(canBlock);
```
mathbunnyru:
```suggestion
    ASSERT_TRUE(canBlock.has_value());
```

And in the places down below

mathbunnyru (approved):

PR #3072, opened 2026-05-11
test: Fix integration tests

kuznetsss:
Fix 2 issues in integration tests:
- `127.0.0.2` is forwarded to localhost on macos
- Keyspace is always dropped in test's destructor but it may not exist

### DIFF
--- a/tests/integration/data/BackendFactoryTests.cpp
+++ b/tests/integration/data/BackendFactoryTests.cpp
@@ -81,7 +81,7 @@ public:
         // drop the keyspace for next test
         data::cassandra::Handle const handle{TestGlobals::instance().backendHost};
         EXPECT_TRUE(handle.connect());
-        EXPECT_TRUE(handle.execute("DROP KEYSPACE " + std::string{kKEYSPACE}));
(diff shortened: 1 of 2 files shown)


tests/integration/data/BackendFactoryTests.cpp
```diff
@@ -98,7 +98,7 @@ TEST_F(BackendCassandraFactoryTest, CreateCassandraBackendDBDisconnect)
         {"database": {
             "type": "cassandra",
             "cassandra": {
-                "contact_points": "127.0.0.2"
+                "contact_points": "128.0.2.1"
```
mathbunnyru, line 101: Is there any meaning behind this value, or is it just some IP address?
Why is it not also `192.0.2.1`?

kuznetsss: No meaning, let me make `128.0.2.1` everywhere, becuse `192.0.x.x` may appear on some machine

mathbunnyru: Yep, that's why I was curios, what's going on 😂 
Thanks!

mathbunnyru (approved):

mathbunnyru:
I would probably suggest being a bit more specific in the PR title, but other than that good to go

PR #3090, opened 2026-06-04
ci: Ubuntu 20.04 -> 22.04

kuznetsss:


mathbunnyru (approved):
I'm ok with this if you made sure we don't have ubuntu:20.04 at all

PR #3091, opened 2026-06-04
feat: Add storage and backend for mpt_issuance_history

BryanJ1ang:
Why
This is the first PR in the broader `mptoken_issuance_history` plan. It lands the additive schema/backend layer first so later PRs can safely add live ETL indexing, historical backfill, and finally the RPC handler gated on backfill completion. Keeping this step dark avoids exposing partial history while still making the stack reviewable in small pieces.

Summary
Adds the storage and backend primitives for MPT transaction-history filtering.

- Adds Cassandra tables:

  - mpt_transactions

  - account_mpt_transactions
- Adds MPTTransactionsData and backend interface methods for writing/fetching both MPT index shapes.
Implements Cassandra write/fetch paths with forward/reverse pagination, raw-page cursor behavior, and case-insensitive tx_type filtering before transaction hydration.
- Adds MockBackend coverage for the new pure-virtual methods.
- Adds Cassandra integration tests for round-trip behavior, ordering, pagination, missing transaction blobs, sparse tx_type filtering, account fanout, and account-side filtering.


PeterChen13579:
@BryanJ1ang  Can you please fix the precommit hook and clang tidy issue? 
Please refer to https://github.com/XRPLF/clio/blob/develop/docs/build-clio.md for clang tidy
For pre-commit: do: `brew install pre-commit` in terminal and then `pre-commit run --all-files` for the fix

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1103,100 @@ class CassandraBackendFamily : public BackendInterface {
...
+        std::vector<ripple::uint256> hashes = {};
+        // The marker rides the raw index page boundary, so the full page must be counted even
+        // when the tx_type filter drops some rows from the hydrated result.
+        auto const rawRowCount = results.numRows();
+        auto remaining = rawRowCount;
+        LOG(log_.info()) << "num_rows = " << rawRowCount;
+
+        for (auto const& [hash, data, rowTxType] :
+             extract<ripple::uint256, std::tuple<uint32_t, uint32_t>, std::string>(results)) {
+            if (not txType.has_value() || boost::iequals(rowTxType, *txType))
```
Kassaking7: Do we need to store tx_type in the database? The advantage I agree is here we query less txns from the transaction table & store less data into memory. But the consequence is we do store many `text` in cassandra. We need to consider whether we want to keep it or not. One example is "account_tx" which you can find in src/rpc/handler/AccountTx.cpp. It fetches tx_type in handler by using `toExpandJson(txnPlusMeta)`. I mentioned this concern in the design doc as well :D

src/data/cassandra/Schema.hpp
```diff
@@ -273,6 +273,41 @@ class Schema {
...
+                  (
+                     mpt_id blob,
+                    seq_idx tuple<bigint, bigint>,
+                       hash blob,
+                    tx_type text,
+                     PRIMARY KEY (mpt_id, seq_idx)
+                  )
+             WITH CLUSTERING ORDER BY (seq_idx DESC)
+            )",
+                qualifiedTableName(settingsProvider_.get(), "mpt_transactions")
```
yinyiqian1: Since mpt_issuance and mptoken are distinct concepts, let’s rename it to mptoken_issuance_transactions to avoid any confusion.

src/data/cassandra/Schema.hpp
```diff
@@ -273,6 +273,41 @@ class Schema {
...
+                     mpt_id blob,
+                    account blob,
+                    seq_idx tuple<bigint, bigint>,
+                       hash blob,
+                    tx_type text,
+                     PRIMARY KEY ((mpt_id, account), seq_idx)
+                  )
+             WITH CLUSTERING ORDER BY (seq_idx DESC)
+            )",
+                qualifiedTableName(settingsProvider_.get(), "account_mpt_transactions")
```
yinyiqian1: account_mptoken_issuance_transactions

src/data/cassandra/Schema.hpp
```diff
@@ -273,6 +273,41 @@ class Schema {
             )
         );
 
+        statements.emplace_back(
+            fmt::format(
+                R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  (
+                     mpt_id blob,
```
yinyiqian1: mptoken_issuance_id

src/data/cassandra/Schema.hpp
```diff
@@ -273,6 +273,41 @@ class Schema {
...
+                qualifiedTableName(settingsProvider_.get(), "mpt_transactions")
+            )
+        );
+
+        statements.emplace_back(
+            fmt::format(
+                R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  (
+                     mpt_id blob,
```
yinyiqian1: mptoken_issuance_id

src/data/cassandra/Schema.hpp
```diff
@@ -474,6 +509,32 @@ class Schema {
             );
         }();
 
+        PreparedStatement insertMPTTx = [this]() {
```
yinyiqian1: Let's go with insertMPTokenIssuanceTx here, and apply similar naming updates across the rest of this lambda function for consistency.

src/data/cassandra/Schema.hpp
```diff
@@ -474,6 +509,32 @@ class Schema {
...
+                INSERT INTO {}
+                       (mpt_id, seq_idx, hash, tx_type)
+                VALUES (?, ?, ?, ?)
+                )",
+                    qualifiedTableName(settingsProvider_.get(), "mpt_transactions")
+                )
+            );
+        }();
+
+        PreparedStatement insertAccountMPTTx = [this]() {
```
yinyiqian1: let's use insertAccountMPTokenIssuanceTx and apply similar naming updates across the rest of this lambda

src/data/cassandra/Schema.hpp
```diff
@@ -740,6 +801,72 @@ class Schema {
             );
         }();
 
+        PreparedStatement selectMPTTx = [this]() {
```
yinyiqian1: Let's also apply the naming to this section from line 804 to 869

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1103,100 @@ class CassandraBackendFamily : public BackendInterface {
...
+     * `cursorIdx + 1`)
+     * @param txType Optional `TxFormats` transaction type name to filter on (case-insensitive)
+     * @param limit The maximum number of transactions per result page
+     * @param forward Whether the page is fetched forwards or backwards
+     * @param cursorIn The cursor to resume fetching from
+     * @param yield The coroutine context
+     * @return Results and a cursor to resume from
+     */
+    TransactionsAndCursor
+    fetchMPTTransactionsImpl(
```
yinyiqian1: let's make it consistent with the existing patterns.
We can call this helper function `doFetch....`

src/data/DBHelpers.hpp
```diff
@@ -199,6 +199,33 @@ struct MPTHolderData {
...
+ * A single instance drives writes to both MPT transaction index tables
+ * (`mpt_transactions` and `account_mpt_transactions`). The @ref accounts set fans out the
+ * per-account table, and @ref txType is the canonical mixed-case `TxFormats` name persisted on
+ * every row (filtered case-insensitively at read time for the `tx_type` query shapes).
+ *
+ * @note This type is an aggregate populated from already-extracted transaction details. It
+ * intentionally has no `ripple::TxMeta`-based constructor so the MPT extraction logic can live in
+ * one ETL/backfill helper rather than in the storage data model.
+ */
+struct MPTTransactionsData {
```
yinyiqian1: The doxygen style seems inconsistent with the current file. Let's keep it consistent with the rest of file.

And also let's modify the naming as it is mptokenissuance

yinyiqian1 (commented):
Let's standardize on MPTokenIssuance everywhere. Since MPTokenIssuance and MPToken represent distinct concepts, keeping them separate will prevent confusion and naming collisions when we add new MPToken-related APIs in the future.

src/data/cassandra/CassandraBackendFamily.hpp
yinyiqian1: Agree, let's filter at the handler layer.

src/data/BackendInterface.hpp
```diff
@@ -408,6 +408,54 @@ class BackendInterface {
         boost::asio::yield_context yield
     ) const = 0;
 
+    /**
+     * @brief Fetches transactions that touched a given MPT issuance.
```
PeterChen13579: I would name this `Fetches transactions for a particular MPT issuance ID.` "touched" is a bit weird here

src/data/BackendInterface.hpp
```diff
@@ -408,6 +408,54 @@ class BackendInterface {
...
+     *
+     * Returns one page of the issuance-wide index, newest-first (or oldest-first when @p forward).
+     * Type filtering, when requested, is applied post-hydration in the handler off the hydrated
```
PeterChen13579: I don't see "hydration" in Clio codebase before... Also not sure what "index" means here. I think a more understandable comment would be something like:
* Returns one page of transactions for this issuance, newest-first (or oldest-first when
* @p forward is set). Each row of the `mpt_transactions` table holds only a transaction hash
* and its ledger position, not the transaction itself or its type. So the query cannot filter
* by type, but when a type filter is requested, the handler looks up each full transaction by hash
* and drops the ones whose type does not match, the same way `account_tx` does.

src/data/BackendInterface.hpp
```diff
@@ -408,6 +408,54 @@ class BackendInterface {
...
+     * transaction blob, exactly as `account_tx` (the index stores no transaction type).
+     *
+     * @param mptID The 24-byte MPT issuance ID
```
PeterChen13579: nit: add a `.` to end of each sentence here

src/data/BackendInterface.hpp
```diff
@@ -408,6 +408,54 @@ class BackendInterface {
...
+    fetchMPTTransactions(
+        ripple::uint192 const& mptID,
+        std::uint32_t limit,
+        bool forward,
+        std::optional<TransactionsCursor> const& cursorIn,
+        boost::asio::yield_context yield
+    ) const = 0;
+
+    /**
+     * @brief Fetches transactions that touched a given MPT issuance and involved a given account.
```
PeterChen13579: Please apply the same comments above in `fetchMPTTransactions` to here

src/data/BackendInterface.hpp
```diff
@@ -725,6 +773,24 @@ class BackendInterface {
     virtual void
     writeNFTTransactions(std::vector<NFTTransactionsData> const& data) = 0;
 
+    /**
+     * @brief Write MPT transaction index rows to the `mpt_transactions` table.
+     *
+     * @param data A vector of MPTTransactionsData objects
```
PeterChen13579: nit: add "`.`" to the end of the sentence

src/data/BackendInterface.hpp
```diff
@@ -725,6 +774,28 @@ class BackendInterface {
...
+    virtual void
+    writeMPTokenIssuanceTransactions(std::vector<MPTokenIssuanceTransactionsData> const& data) = 0;
+
+    /**
+     * @brief Write MPTokenIssuance transaction index rows to the
+     * `account_mptoken_issuance_transactions` table.
+     *
+     * One row is written per affected account in each record.
+     *
+     * @param data A vector of MPTokenIssuanceTransactionsData objects
```
PeterChen13579: same as above

src/data/DBHelpers.hpp
```diff
@@ -199,6 +199,26 @@ struct MPTHolderData {
     ripple::AccountID holder;
 };
 
+/**
+ * @brief Represents a link from a transaction to an MPTokenIssuance that it touched.
+ *
+ * A single instance drives writes to both MPTokenIssuance transaction index tables
```
PeterChen13579: IMO, I feel like the comments can be cleaner here. I understood what this line meant after a while, but it could just be rewritten as `Writing one of these records inserts into two tables: mptoken_issuance_transactions and account_mptoken_issuance_transactions`. This makes it much easier to read. I think the same can be applied to other comments.

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1102,87 @@ class CassandraBackendFamily : public BackendInterface {
 
         return true;
     }
+
+    /**
+     * @brief Shared implementation of the two MPTokenIssuance transaction-index fetchers.
+     *
+     * Mirrors `fetchNFTTransactions`: binds the cursor/limit onto an already partition-bound
```
PeterChen13579: I think we can get rid of this paragraph, no need to mention it mirrors `fetchNFTTransactions`. Add something like:

* @note The forward path queries with an inclusive `seq_idx >=`, so the returned cursor's transaction index is advanced by one to avoid re-reading the last row on the next page.

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1102,87 @@ class CassandraBackendFamily : public BackendInterface {
...
+     * statement, reads `(hash, seq_idx)` index rows, then hydrates the blobs via
+     * @ref fetchTransactions. The forward path uses an inclusive `seq_idx >=`, so the returned
+     * cursor's transaction index is advanced by one (matching the NFT history convention).
+     *
+     * @param statement The statement already bound with the partition-key columns
```
PeterChen13579: nit: add `.` end of each sentence.

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1102,87 @@ class CassandraBackendFamily : public BackendInterface {
...
+        std::optional<TransactionsCursor> const& cursorIn,
+        boost::asio::yield_context yield
+    ) const
+    {
+        auto rng = fetchLedgerRange();
+        if (!rng)
+            return {.txns = {}, .cursor = {}};
+
+        auto cursor = cursorIn;
+        if (cursor) {
```
PeterChen13579: I know other places didn't do this, but adding `.hasvalue()` here is good practice. Feel free to change it for other places too

src/data/cassandra/CassandraBackendFamily.hpp
PeterChen13579: @yinyiqian1 , our doFetch is only if the functions are virtual and exists in `BackendInterface.hpp`. Since this is a private helper, I think it should be without the "do", same like `fetchAccountMPTokenIssuanceTransactions`, `fetchNFTTransactions` etc
I may be wrong, would like confirmation: CC @kuznetsss

src/data/cassandra/Schema.hpp
```diff
@@ -740,6 +801,76 @@ class Schema {
...
+                   AND seq_idx >= ?
+              ORDER BY seq_idx ASC
+                 LIMIT ?
+                )",
+                    qualifiedTableName(
+                        settingsProvider_.get(), "account_mptoken_issuance_transactions"
+                    )
+                )
+            );
+        }();
```
PeterChen13579, line 872: Good job on these tables! Looks good

PeterChen13579 (changes requested):

src/data/DBHelpers.hpp
```diff
@@ -199,6 +199,20 @@ struct MPTHolderData {
     ripple::AccountID holder;
 };
 
+/**
+ * @brief Represents a transaction link for an MPTokenIssuance.
+ *
+ * Writing one of these records inserts into two tables:
```
PeterChen13579: either add @ note here or keep it within the brief section

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1461,8 +1462,378 @@ TEST_F(CacheBackendCassandraTest, CacheFetchLedgerBySeq)
     });
 }
 
+struct BackendCassandraMPTokenIssuanceTest : BackendCassandraTest {
+    static ripple::uint192
+    makeMptIssuanceId()
```
PeterChen13579, line 1468: There already exists helper functions that do the same thing as makeMptissuanceID (same with makeAccount and makeHash) I suggest you can keep these functions, but in the bodies call the helpers instead of writing the logic. ie: 

```
 ripple::makeMptID(1, makeAccount(0x01));   
and use ripple::AccountID and  ripple:: uint256 for other 2
```

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1461,8 +1462,378 @@ TEST_F(CacheBackendCassandraTest, CacheFetchLedgerBySeq)
...
+            uint256ToString(hash),
+            seq,
+            0,
+            "tx_" + ripple::strHex(hash),
+            "meta_" + ripple::strHex(hash)
+        );
+    }
+};
+
+TEST_F(BackendCassandraMPTokenIssuanceTest, RoundTripBothShapes)
```
PeterChen13579: Please think of a better name instead of `RoundTripBothShapes`

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1461,8 +1462,378 @@ TEST_F(CacheBackendCassandraTest, CacheFetchLedgerBySeq)
...
+            expectFetchedSingle(txns);
+            EXPECT_FALSE(cursor);
+        }
+        // Both affected accounts are indexed: one row was written per account.
+        {
+            auto [txns, cursor] = backend_->fetchAccountMPTokenIssuanceTransactions(
+                mptIssuanceId, secondAccount, 100, false, {}, yield
+            );
+            expectFetchedSingle(txns);
+            EXPECT_FALSE(cursor);
```
PeterChen13579, line 1571: I would add another test at the end of this func, where you insert another row, fetchtable with limit 1, and expect_true cursor

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1461,8 +1462,378 @@ TEST_F(CacheBackendCassandraTest, CacheFetchLedgerBySeq)
...
+                    for (auto const& tx : txns)
+                        seen.push_back(txBlobToString(tx));
+                    cursor = retCursor;
+                } while (cursor);
+
+                // No duplicates.
+                std::set<std::string> const seenSet(seen.begin(), seen.end());
+                EXPECT_EQ(seen.size(), seenSet.size()) << "pagination returned duplicate rows";
+                // No gaps: union equals the full expected set.
+                EXPECT_EQ(seenSet, expected) << "pagination dropped or repeated rows";
```
PeterChen13579: This checks we got all the rows, but not that they came back in order. Pagination could return pages out of order and still pass. Since seen keeps page order, can we also assert it's sorted (descending for forward=false, ascending for forward=true)?

tests/integration/data/cassandra/BackendTests.cpp
```diff
@@ -1461,8 +1462,378 @@ TEST_F(CacheBackendCassandraTest, CacheFetchLedgerBySeq)
...
+                    .ledgerSequence = seq,
+                    .transactionIndex = i,
+                    .txHash = hash
+                };
+                backend_->writeMPTokenIssuanceTransactions({record});
+                expectedB.insert(expectedBlob(i));
+            }
+            backend_->waitForWritesToFinish();
+
+            auto pageThroughB = [&](bool forward) {
```
PeterChen13579: I think pageThroughB is nearly identical to pageThrough above... the only difference is it expects a trailing empty page. Can we merge them into one helper with a flag for that case?

PeterChen13579 (changes requested):

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -485,6 +485,43 @@ class CassandraBackendFamily : public BackendInterface {
...
+        boost::asio::yield_context yield
+    ) const override
+    {
+        auto const statement = [this, forward, &mptIssuanceID, &account]() {
+            if (forward)
+                return schema_->selectAccountMPTokenIssuanceTxForward.bind(mptIssuanceID, account);
+
+            return schema_->selectAccountMPTokenIssuanceTx.bind(mptIssuanceID, account);
+        }();
+        return fetchMPTokenIssuanceTransactionsImpl(statement, 2, limit, forward, cursorIn, yield);
```
kuznetsss: `2` here is a magic number. Please move it into a named constant. Probably the same for this function call above

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -1016,6 +1102,86 @@ class CassandraBackendFamily : public BackendInterface {
...
+        auto const res = executor_.read(yield, statement);
+        auto const& results = res.value();
+        if (not results.hasRows()) {
+            LOG(log_.debug()) << "No rows returned";
+            return {};
+        }
+
+        std::vector<ripple::uint256> hashes = {};
+        auto numRows = results.numRows();
+        LOG(log_.info()) << "num_rows = " << numRows;
```
kuznetsss: Is it left from debug?

src/data/cassandra/Schema.hpp
```diff
@@ -273,6 +273,39 @@ class Schema {
             )
         );
 
+        statements.emplace_back(
+            fmt::format(
+                R"(
+           CREATE TABLE IF NOT EXISTS {}
+                  (
+        mptoken_issuance_id blob,
```
kuznetsss: Please format queries in the same way as in the other code

src/data/cassandra/CassandraBackendFamily.hpp
kuznetsss: @PeterChen13579, I guess you know this part of the code better than me. Maybe @godexsoft knows the convention here

src/data/cassandra/CassandraBackendFamily.hpp
bryanjiang1: Yes removed, thanks!

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -77,6 +77,9 @@ class CassandraBackendFamily : public BackendInterface {
     // TODO: move to interface level
     mutable FetchLedgerCacheType ledgerCache_{};
 
+    static constexpr std::size_t kMPTokenIssuanceTxCursorBindIndex = 1;
```
PeterChen13579, line 82: Really like this change; I suggest in doing so, there are other parts of the file that uses magic number `1 `and `2`,   where you can replace for these constants as well. search for `statement.bindAt ` places

PeterChen13579 (changes requested):
Last thing I found, will approve after fixing

Kassaking7 (approved):
LGTM

bryanjiang1:
Thank you guys!

PeterChen13579 (approved):
👍

src/data/cassandra/CassandraBackendFamily.hpp
PeterChen13579: @BryanJ1ang  I suggest keeping it without the `do` here.

bryanjiang1:
Force pushed to sign commits

kuznetsss (approved):

PR #3094, opened 2026-06-05
ci: Remove arm

kuznetsss:


.github/workflows/update-docker-ci.yml
```diff
@@ -77,106 +77,17 @@ jobs:
...
+            type=raw,value=${{ env.GCC_VERSION }}
+            type=raw,value=${{ github.sha }}
           platforms: linux/amd64
           build_args: |
             GCC_MAJOR_VERSION=${{ env.GCC_MAJOR_VERSION }}
             GCC_VERSION=${{ env.GCC_VERSION }}
           dockerhub_repo: ${{ github.repository_owner == 'XRPLF' && 'rippleci/clio_gcc' || '' }}
           dockerhub_description: GCC compiler for XRPLF/clio.
 
-  gcc-arm64:
```
mathbunnyru, line 91: Maybe we would be better writing `if: false`, so the diff is not that huge in case we need to change something in the future?

mathbunnyru (changes requested):

mathbunnyru (approved):

PR #3098, opened 2026-06-08
ci: Use Nix-based images for all workflows except pre-commit

### LINKED ISSUE #2863: Add all possible UndefinedBehaviorSanitizer flags in Clang
There are additional flags for Clang when running with UndefinedBehaviorSanitizer

https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

> -fsanitize=undefined: All of the checks listed above other than float-divide-by-zero, unsigned-integer-overflow, implicit-conversion, local-bounds, vptr and the nullability-* group of checks.

We should:
- either enable all additional flags, all comment why we didn't add them- 
- add a link to docs

mathbunnyru:
This:

- uses rippled's Nix images, latest version, with 0 additional changes to them
- removes old code to build old Docker images
- doesn't change the configuration matrix scripts (it will be done separately)
- uses rippled's conan configurations
- so, it allows multiple sanitizers
- and removes duplication of sanitizers settings in cmake
- simplifies dependabot
- adds a useful PatchBinary utility to patch the binary right after it's being linked. Since we use cpack (as of now), we have to do it this way, so the binary in debian package ends up being correct.
- updates `benchmark`, so it builds fine with modern compilers (https://github.com/google/benchmark/pull/2108)
- removes `static` option and always builds a static binary on gcc and clang on Linux
- respects `SSL_CERT_FILE` env variable - this is how it should be done before checking system paths, and is needed as we now use Nix certificate in our Linux CI images
- adds a wrapper for `std::expected<std::any, ...>` (ideologically reverts https://github.com/XRPLF/clio/pull/1336), to make it work with `clang` + `libstdc++`
- force removes `scylladb` container, because in some cases even after colima restart it seems to be present
- fixes link bug - clio_rpc depends on clio_data, but it was not declared so in cmake
- adds build-identifier action, and reuse it everywhere, to ensure that all artifacts are named the same way (on both upload and download), and for ccache, keeping backwards compatibility for everything but sanitized executables

The result is much less code, but with many more features, and closer to rippled.
Built with old libc on a modern toolchain.

**Note 1**: `cppstd` for dependencies is kept `20` for now, though, since we still use old xrpl and old dependencies, which don't work with `23`.

**Note 2**: clang built before used `libc++` (LLVM's implementation of standard library), clang built in nix uses `libstdc++` (GCC's).
This is not ideal, because changing the standard library can produce significantly different results, which is great for testing (especially with sanitizers) and might impact performance.
I think that's a reasonable downside for now, which will be later (sometime in the future) addressed in rippled's nix images, so here we will easily get these improvements too.

**Note 3**: This PR does change a clang-tidy version, but the workflow is only launched on changed files. This is on purpose, because otherwise it will introduce too much noise. It will be addressed after this PR is merged.

**Note 4**: `set-compiler-env` will be moved to actions after this PR has been merged (to allow a bit easier experiments if needed)

Fix: https://github.com/XRPLF/clio/issues/2863


kuznetsss (dismissed):

godexsoft (dismissed):
:rage1:

godexsoft (approved):

PR #3099, opened 2026-06-08
ci: Use glibc version in conan profile

kuznetsss:


docker/ci/conan/clang.profile
```diff
@@ -10,3 +10,7 @@ os=Linux
 [conf]
 tools.build:compiler_executables={"c": "/usr/bin/clang-19", "cpp": "/usr/bin/clang++-19"}
 grpc/1.50.1:tools.build:cxxflags+=["-Wno-missing-template-arg-list-after-template-kw"]
+{% if os == "Linux" %}
```
mathbunnyru: Not needed here, because there are purely Linux profiles

mathbunnyru (changes requested):
This will be redundant in nix-based images though

docker/ci/conan/clang.profile
kuznetsss: Fixed

docker/ci/conan/sanitizer_template.profile
```diff
@@ -35,3 +35,6 @@ tools.build:sharedlinkflags+={{ sanitizer_link_flags }}
 {% endif %}
 
 tools.info.package_id:confs+=["tools.build:cflags", "tools.build:cxxflags", "tools.build:exelinkflags", "tools.build:sharedlinkflags", "tools.build:defines"]
+
+user.package:libc_version=2.32
+tools.info.package_id:confs+=["user.package:libc_version"]
```
mathbunnyru: Here `if Linux` is needed though, because this profile is also used locally with apple-clang 🙂

mathbunnyru: Actually, this section is not needed at all

mathbunnyru: Because adding sanitizer happens on top of existing profiles

kuznetsss: Then maybe lets put `if Linux` everywhere for consistency?

mathbunnyru: I would only add libc version to gcc and clang profiles without Linux if, because it doesn't contradict with anything and is quite consistent.

kuznetsss: Updated

mathbunnyru (approved):

PR #3102, opened 2026-06-08
feat: Update MPToken ETL to support all transactions types

### LINKED ISSUE #2695: Update getMPTHolderFromTx function for rippled 3.0.0 release
## Summary

Features released in rippled 3.0.0 like lending and future features (MPT DEX) introduce ability to create MPToken for existing transactions. Currently, the ETL only filters for MPTokenAuthorize, which is no longer adequate. We'd need to remove the transactions-specific filter and traverse metadata for all transaction types.

shawnxie999:
## Summary

Fixes incomplete MPT holder indexing in ETL.

Previously, Clio only indexed MPT holders when an `ltMPTOKEN` ledger node was created by an `MPTokenAuthorize` transaction. However, with future amendments, XRPL can also create `ltMPTOKEN` nodes as part of other successful transactions, such as `Payment`, when a holder receives an MPT. Those holders were silently skipped and therefore missing from the `mpt_holders` RPC results.

This change makes MPT holder indexing metadata-driven instead of transaction-type-driven. ETL now scans successful transaction metadata for newly created `ltMPTOKEN` nodes and writes every holder it finds.

## What Changed

- MPT holder extraction now works for any successful transaction that creates an `ltMPTOKEN`.
- Multiple MPT holders created in a single transaction are all indexed.
- The `mpt_holders` table is kept consistent with ledger metadata instead of relying on assumptions about which transaction types can create holder nodes.
- Added regression coverage for MPT holders created by a `Payment` transaction.

## Testing

- `build/clio_tests '--gtest_filter=MPTExtTests.*'`
- `clang-tidy -p build src/etl/MPTHelpers.cpp src/etl/MPTHelpers.hpp src/etl/impl/ext/MPT.cpp tests/unit/etl/ext/MPTTests.cpp`

### DIFF
--- a/src/etl/MPTHelpers.cpp
+++ b/src/etl/MPTHelpers.cpp
@@ -15,41 +15,34 @@
 
 #include <optional>
 #include <string>
+#include <vector>
 
 namespace etl {
 
-/**
- * @brief Get the MPToken created from a transaction
- *
- * @param txMeta Transaction metadata
- * @return MPT and holder account pair
- */
-std::optional<MPTHolderData>
-getMPTokenAuthorize(ripple::TxMeta const& txMeta)
+std::vector<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const&)
 {
+    if (txMeta.getResultTER() != ripple::tesSUCCESS)
+        return {};
+
+    std::vector<MPTHolderData> holders;
+
     for (ripple::STObject const& node : txMeta.getNodes()) {
         if (node.getFieldU16(ripple::sfLedgerEntryType) != ripple::ltMPTOKEN)
             continue;
 
         if (node.getFName() == ripple::sfCreatedNode) {
             auto const& newMPT = node.peekAtField(ripple::sfNewFields).downcast<ripple::STObject>();
-            return MPTHolderData{
-                .mptID = newMPT[ripple::sfMPTokenIssuanceID],
-                .holder = newMPT.getAccountID(ripple::sfAccount)
-            };
+            holders.push_back(
+                MPTHolderData{
+                    .mptID = newMPT[ripple::sfMPTokenIssuanceID],
+                    .holder = newMPT.getAccountID(ripple::sfAccount)
+                }
+            );
         }
     }
-    return {};
-}
 
-std::optional<MPTHolderData>
-getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const& sttx)
-{
-    if (txMeta.getResultTER() != ripple::tesSUCCESS ||
-        sttx.getTxnType() != ripple::TxType::ttMPTOKEN_AUTHORIZE)
-        return {};
-
-    return getMPTokenAuthorize(txMeta);
+    return holders;
 }
 
 std::optional<MPTHolderData>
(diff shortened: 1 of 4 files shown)


src/etl/impl/ext/MPT.cpp
```diff
@@ -49,8 +49,8 @@ MPTExt::writeMPTHoldersFromTransactions(model::LedgerData const& data)
     std::vector<MPTHolderData> holders;
 
     for (auto const& tx : data.transactions) {
-        if (auto const mptHolder = getMPTHolderFromTx(tx.meta, tx.sttx); mptHolder.has_value())
-            holders.push_back(*mptHolder);
+        auto const mptHolders = getMPTHolderFromTx(tx.meta, tx.sttx);
+        holders.insert(holders.end(), mptHolders.begin(), mptHolders.end());
```
PeterChen13579: `holders.append_range(mptHolders);` is probs cleaner

src/etl/MPTHelpers.hpp
```diff
@@ -6,16 +6,20 @@
...
+
 namespace etl {
 
 /**
  * @brief Pull MPT data from TX via ETLService.
  *
  * @param txMeta Transaction metadata
  * @param sttx The transaction
- * @return The MPTIssuanceID and holder pair as a optional
+ * @return The MPTIssuanceID and holder pairs created by the transaction
```
PeterChen13579: Maybe add "empty if the transaction failed or created no MPToken"

src/etl/MPTHelpers.cpp
```diff
@@ -11,45 +11,37 @@
...
- *
- * @param txMeta Transaction metadata
- * @return MPT and holder account pair
- */
-std::optional<MPTHolderData>
-getMPTokenAuthorize(ripple::TxMeta const& txMeta)
+std::vector<MPTHolderData>
+getMPTHolderFromTx(ripple::TxMeta const& txMeta, ripple::STTx const&)
 {
+    if (txMeta.getResultTER() != ripple::tesSUCCESS)
```
PeterChen13579, line 24: I think there isn't a test which tests for a failed tx here. Might be worth adding

PeterChen13579 (changes requested):

PeterChen13579 (approved):
LGTM

src/etl/MPTHelpers.hpp
```diff
@@ -6,16 +6,21 @@
 #include <ripple/protocol/STTx.h>
 #include <ripple/protocol/TxMeta.h>
 
+#include <optional>
```
Kassaking7, line 9: I think `MPTHelpers.cpp` should include `etl/MPTHelpers.hpp` directly, ideally first, instead of re-declaring all dependencies through separate includes. This helps catch declaration/definition drift.

kuznetsss: We are using enforced [Include-What-You-Use policy](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md) so everything used in the file should have an include.

kuznetsss (approved):

Kassaking7 (approved):

shawnxie999:
@kuznetsss would you be able to merge this? thanks!

PR #3104, opened 2026-06-15
feat: Add ETL Indexing for mpt_issuance_history

BryanJ1ang:
## Summary

Adds live ETL indexing for mptoken_issuance_history by extracting MPT issuance references from transaction metadata and transaction fields, then writing both MPT transaction index shapes during the existing MPT ETL pass. Successful transactions are indexed from affected MPT ledger objects; failed transactions are indexed only when they carry an attached MPT issuance reference.

## Changes

- Adds `getMPTokenIssuanceTxsFromTx` to extract issuance transaction index records from `tesSUCCESS` transactions.
- Scans `AffectedNodes` generically for `MPToken` and `MPTokenIssuance` ledger objects instead of relying on a transaction-type allowlist.
- Reconstructs issuance IDs for `MPTokenIssuance` objects with `makeMptID(sequence, issuer)`, avoiding the hashed ledger-key trap.
- Deduplicates distinct issuances per transaction and attaches the transaction’s affected accounts for account-level fanout.
- Extends `MPTExt` so live and initial ETL data writes:
  - existing MPT holder index data
  - `mptoken_issuance_transactions`
  - `account_mptoken_issuance_transactions`
- Adds a Prometheus counter for MPT issuance transaction index rows written by ETL.
- Logs unexpectedly large per-transaction fanout.

## Tests

Adds unit coverage for:

- failed transactions producing no records
- `MPTokenIssuanceCreate`, `Destroy`, `Set`, and `MPTokenAuthorize`
- generic transaction-type coverage through `Payment`, `Clawback`, `OfferCreate`, and `AMMDeposit`
- ID reconstruction from issuance nodes
- multi-issuance fanout and deduplication
- affected-account propagation
- ETL extension writes to both backend index paths while preserving existing holder indexing


src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+
+}  // namespace
+
+std::vector<MPTokenIssuanceTransactionsData>
+getMPTokenIssuanceTxsFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const& sttx)
+{
+    // Collect each distinct issuance only once per transaction; the same set of affected accounts
+    // is attached to every record produced below.
+    MPTokenIssuanceIDs issuanceIDs;
+    for (xrpl::STObject const& node : txMeta.getNodes()) {
```
PeterChen13579: prefer use `auto` here for for loops

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+    if (not fields.isFieldPresent(xrpl::sfSequence) || not fields.isFieldPresent(xrpl::sfIssuer))
+        return {};
+
+    return xrpl::makeMptID(
+        fields.getFieldU32(xrpl::sfSequence), fields.getAccountID(xrpl::sfIssuer)
+    );
+}
+
+void
+addMPTokenIssuanceIDFromAmount(MPTokenIssuanceIDs& issuanceIDs, xrpl::STAmount const& amount)
```
PeterChen13579: This function is only used once and 2 lines long, I would remove it and put the code body to where it is

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+{
+    if (amount.holds<xrpl::MPTIssue>())
+        issuanceIDs.insert(amount.get<xrpl::MPTIssue>().getMptID());
+}
+
+void
+addMPTokenIssuanceIDFromIssue(MPTokenIssuanceIDs& issuanceIDs, xrpl::STIssue const& issue)
```
PeterChen13579: ditto above

src/etl/impl/ext/MPT.cpp
```diff
@@ -40,21 +42,50 @@ void
...
     std::vector<MPTHolderData> holders;
+    std::vector<MPTokenIssuanceTransactionsData> issuanceTxs;
+    std::size_t indexRowsWritten = 0;
+    static constexpr std::size_t kIndexRowsPerTxWarningThreshold = 1000;
 
     for (auto const& tx : data.transactions) {
         auto const mptHolders = getMPTHolderFromTx(tx.meta, tx.sttx);
         holders.append_range(mptHolders);
+
+        auto txIndexData = getMPTokenIssuanceTxsFromTx(tx.meta, tx.sttx);
```
PeterChen13579, line 60: If I understood correctly, line 57 will scan through every txn on every incoming ledger number, if the txn's related `ledger_entry` is not `mpt`, it will return `{}`. But line 60, you scan the same data again for mpt fields, which won't exist if it already returned `{}` prior. Therefore, I would have a way to exit early instead of doing another scan

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
     return holders;
 }
 
+namespace {
+
+using MPTokenIssuanceIDs = boost::container::flat_set<xrpl::uint192>;
+
+/**
+ * @brief Derive the MPTokenIssuanceID from an affected node in transaction metadata
```
PeterChen13579: please add `.` to end of each sentence

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+ */
+std::optional<xrpl::uint192>
+getMPTokenIssuanceIDFromNode(xrpl::STObject const& node)
+{
+    auto const entryType = node.getFieldU16(xrpl::sfLedgerEntryType);
+    if (entryType != xrpl::ltMPTOKEN && entryType != xrpl::ltMPTOKEN_ISSUANCE)
+        return {};
+
+    auto const& fieldsName =
+        node.getFName() == xrpl::sfCreatedNode ? xrpl::sfNewFields : xrpl::sfFinalFields;
```
PeterChen13579, line 70: Can we confirm the affected-node field selection here is always valid?
CreatedNode uses `sfNewFields`, while ModifiedNode/DeletedNode use `sfFinalFields`. (I'm pretty sure this is right checking rippled code, but what is the mechanism that allows it to change? 🤔 ).  (cc @yinyiqian1  @shawnxie999 )

tests/unit/etl/ext/MPTTests.cpp
```diff
@@ -68,6 +75,100 @@ constinit auto const kHash = "6005B465CBBF7FA8E41AC0C0CD38491026D9411FCB7BA46E2A
...
+        EXPECT_EQ(lhs[i].mptIssuanceID, rhs[i].mptIssuanceID);
+        EXPECT_EQ(lhs[i].accounts, rhs[i].accounts);
+        EXPECT_EQ(lhs[i].ledgerSequence, rhs[i].ledgerSequence);
+        EXPECT_EQ(lhs[i].transactionIndex, rhs[i].transactionIndex);
+        EXPECT_EQ(lhs[i].txHash, rhs[i].txHash);
+    }
+}
+
+xrpl::STObject
+createMPTokenNode(xrpl::uint192 const& issuanceID, std::string_view holder)
```
PeterChen13579: I think these create objects helpers for test already exist. I would use the ones in `MPTHelpersTests.cpp`. As a bonus, I would create a new file named MPTokenTestObjects.cpp  (don't move it to `testObject.cpp` as it is a mess there) and move the shared util there.

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,117 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+{
+    if (amount.holds<xrpl::MPTIssue>())
+        issuanceIDs.insert(amount.get<xrpl::MPTIssue>().getMptID());
+}
+
+void
+addMPTokenIssuanceIDFromIssue(MPTokenIssuanceIDs& issuanceIDs, xrpl::STIssue const& issue)
+{
+    if (issue.holds<xrpl::MPTIssue>())
+        issuanceIDs.insert(issue.value().get<xrpl::MPTIssue>().getMptID());
```
PeterChen13579: I know this isn't followed strictly, but use .`has_value() `for checking optional values and use `->` to access

PeterChen13579 (changes requested):
There is also Codecov that shows you some lines from source file that are missing tests. Please cover them.

src/etl/impl/ext/MPT.cpp
bryanjiang1: Not quite. Line 57 only extracts newly created `ltMPTOKEN` holder records from successful metadata. I believe returning empty does not mean the transaction is unrelated to MPT.

 Line 60 builds the MPT issuance transaction-history lookup, so it also needs to catch `ltMPTOKEN_ISSUANCE` metadata and MPT references in the submitted transaction body, such as `sfMPTokenIssuanceID`, `MPT Amount,` or `MPT Issue`. Early exit after line 57 would miss these valid MPT-related transactions.

src/etl/MPTHelpers.cpp
bryanjiang1: I believe the pinned STIssue doesn't define operator `->`, so `issue->get<xrpl::MPTIssue>() `does not compile.

PeterChen13579 (dismissed):

godexsoft (dismissed):
Looks fine to me from Clio/code perspective 👍 
Would be nice to get answer for @PeterChen13579 's question just to make sure.

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,109 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
     return holders;
 }
 
+namespace {
+
+using MPTokenIssuanceIDs = boost::container::flat_set<xrpl::uint192>;
```
mathbunnyru, line 54: Can we use `std::flat_set`?

src/etl/MPTHelpers.cpp
```diff
@@ -44,6 +49,109 @@ getMPTHolderFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const&)
...
+ *
+ * @param node An entry of the metadata's AffectedNodes array.
+ * @return The 192-bit issuance ID if the node is an MPTokenIssuance or MPToken object.
+ */
+std::optional<xrpl::uint192>
+getMPTokenIssuanceIDFromNode(xrpl::STObject const& node)
+{
+    auto const entryType = node.getFieldU16(xrpl::sfLedgerEntryType);
+    if (entryType != xrpl::ltMPTOKEN && entryType != xrpl::ltMPTOKEN_ISSUANCE)
+        return {};
```
mathbunnyru: I would prefer being explicit and returning `std::nullopt` in places like this, otherwise it's slightly unclear

tests/unit/etl/ext/MPTTests.cpp
```diff
@@ -68,6 +77,141 @@ constinit auto const kHash = "6005B465CBBF7FA8E41AC0C0CD38491026D9411FCB7BA46E2A
...
 
+// The issuance ID carried by the ltMPTOKEN CreatedNode in kTxnMeta.
+constinit auto const kIssuanceID = "002DBD1817E0AF9FDE4F9978B8FCD8A5063630B5737DA605";
+
+constinit auto const kAccount = "rM2AGCCCRb373FRuD8wHyUwUsh2dV4BW5Q";
+constinit auto const kAccount2 = "rnd1nHuzceyQDqnLH8urWNr4QBKt4v7WVk";
+constexpr auto kHighFanoutAccountCount = 1001u;
+
+void
+expectSameRecords(
```
mathbunnyru: This feels like we should add a comparison operator for `MPTokenIssuanceTransactionsData` and this function will just go away

src/etl/MPTHelpers.cpp
bryanjiang1: I believe our documented min compiler Apple clang 17 doesn't support`std::flat_set` so sticking with boost would be better for consistency.

src/etl/MPTHelpers.cpp
bryanjiang1: Yes, it should be guaranteed by rippled's metadata construction in `ApplyStateTable::apply.` Each affected node is wrapped in one of `sfCreatedNode/sfModifiedNode/sfDeletedNode` based on the action, and each has its own field: CreatedNode → sfNewFields, ModifiedNode/DeletedNode → sfFinalFields.

src/etl/MPTHelpers.cpp
mathbunnyru: We already updated to Apple Clang 21, maybe forgot to update the compiler list (but in rippled it's already 21), so should be fine?
We can do it after this gets merged though, no worries

mathbunnyru (approved):
LGTM, if we could switch to std::flat_set in a separate PR when we merge this, it would be great 👍

tests/unit/etl/ext/MPTTests.cpp
```diff
@@ -191,13 +308,33 @@ createTestData()
...
+    auto const metaBlob = xrpl::strUnHex(kTxnMeta);
+    EXPECT_TRUE(metaBlob.has_value());
+    // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
+    xrpl::SerialIter sitMeta{xrpl::makeSlice(*metaBlob)};
+    xrpl::STObject metaObj{sitMeta, xrpl::sfMetadata};
+    metaObj.setFieldU32(xrpl::sfTransactionIndex, txIndex);
+
+    xrpl::uint256 hash;
+    EXPECT_TRUE(hash.parseHex(hashStr));
+    tx.meta = xrpl::TxMeta{hash, kSeq, metaObj.getSerializer().peekData()};
```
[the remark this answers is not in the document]
bryanjiang1: `assert_true` doesn't compile here. The deref is safe since `kTxnMeta` is a compile time constant.

yinyiqian1 (approved):

PR #3109, opened 2026-06-17
ci: Fix docs workflow && build on PRs

mathbunnyru:
I update all workflows to use new image (which now has git-lfs)
Also, to check it, we should build docs in PRs - it's fast, so shouldn't affect anyone.


kuznetsss (dismissed):

godexsoft (approved):
:rage1:

PR #3113, opened 2026-06-23
fix: Don't cache requests with params

kuznetsss:
`ResponseExpirationCache` is using only `method` (or `command`) as a key, but responses might be different with different params. This may lead to a client getting incorrect response because of caching. This PR fixes it by caching only requests without any additional params.


src/util/ResponseExpirationCache.cpp
```diff
@@ -44,9 +47,36 @@ ResponseExpirationCache::shouldCache(std::string const& cmd)
...
+    // "command" and "method" merely name the RPC command — they do not affect response content.
+    //
+    // "id" is safe to ignore because the web layer re-applies the current request's id to the
+    // response AFTER the cache lookup (see src/web/RPCServerHandler.hpp), so a cached body never
+    // leaks a stale id back to the client.  Ignoring "id" is also necessary for WebSocket
+    // requests, which almost always carry an id and would otherwise never benefit from caching.
+    //
+    // "api_version" is intentionally NOT in this ignore-set: it can change the response body
+    // content (not just an echoed field), so requests carrying it must bypass the cache entirely.
+    static constexpr std::array<std::string_view, 3> kIgnoredKeys{"command", "method", "id"};
```
godexsoft: probably can use std::to_array

src/util/ResponseExpirationCache.cpp
```diff
@@ -44,9 +47,36 @@ ResponseExpirationCache::shouldCache(std::string const& cmd)
...
+
+    for (auto const& kv : request) {
+        std::string_view const key{kv.key()};
+        if (std::ranges::find(kIgnoredKeys, key) == kIgnoredKeys.end())
```
godexsoft: not contains()?

tests/unit/util/ResponseExpirationCacheTests.cpp
```diff
@@ -11,40 +11,105 @@ using namespace util;
 struct ResponseExpirationCacheTests : public ::testing::Test {
 protected:
     ResponseExpirationCache cache_{std::chrono::seconds{100}, {"key"}};
+    // A bare request: only the "command" key, which is in the ignore-set.
```
godexsoft: Probably not very useful comment

godexsoft (changes requested):
Few nits 👍

godexsoft (approved):
🥳

PR #3118, opened 2026-06-26
build: Switch to new XRPLF Conan remote

mathbunnyru:
This switched the remote, updates lockfile, so it's consistent with new remote.
I uploaded `xrpl/3.2.0` manually to the new remote.

Also updated openssl to 3.6.3, to ensure we're using our custom openssl recipe.


conanfile.py
```diff
@@ -47,7 +47,7 @@ def requirements(self):
         # So they must be direct requirements of clio to get their include dirs;
         # the version pins match xrpl's, so this does not change any package_id.
         self.requires("grpc/1.78.1")
-        self.requires("openssl/3.6.2")
+        self.requires("openssl/3.6.3", force=True)
```
godexsoft, line 50: This is something more than switching remote. Is this to match xrpld, right ❓

godexsoft (approved):
In theory this is fine but i see you also updated openssl. It is matching rippled though so probably this is ok. :rage1:  :rage2:

conanfile.py
mathbunnyru: Not exactly, I wrote in PR description, but let me give more info.

We upload custom openssl recipe, but only for the latest minor version present currently at official CCI (they delete the previous one).
So, in new Conan remote, there is no 3.6.2, so we have to update.

I don't think this would cause any issues

PR #3123, opened 2026-06-29
build: Use apple-clang 21; remove SourceLocation stub

mathbunnyru:


godexsoft (approved):
Yay. hopefully it really does work on all platforms now :rage2:

PR #3126, opened 2026-06-30
ci: Pin scylladb docker to fix integration tests

mathbunnyru:
It seems that new scylladb versions have changed `SimpleStrategy` that we use and now code fails with:
```
C++ exception with description "Could not create keyspace: future::get(): SimpleStrategy doesn't support tablet replication" thrown in the test fixture's constructor.
```

This should fix it for now in integration tests, but we should take a closer look if we need to migrate the database

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -104,7 +104,7 @@ jobs:
 
     services:
       scylladb:
-        image: ${{ inputs.container != '' && 'scylladb/scylla' || '' }}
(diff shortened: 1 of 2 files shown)


godexsoft (approved):
We should be careful with scylla and their driver.. we have seen that updating the driver also broke things last time we tried. My understanding is that this change is pinning the older version, right?

mathbunnyru:
> We should be careful with scylla and their driver.. we have seen that updating the driver also broke things last time we tried. My understanding is that this change is pinning the older version, right?

Yes, it's pinning the version before the latest release, which broke something.
And this change is tests-only, no effect on production.

PR #3128, opened 2026-06-30
feat: Add migration backfill for mptoken_issuance_history

bryanjiang1:
Note: this PR depends on previous ETL indexing PR [here](https://github.com/XRPLF/clio/pull/3104)
## Summary

Adds `MPTTransactionHistoryMigrator`, a Cassandra migration that backfills historical MPT issuance transaction history into the new MPT index tables:

- `mptoken_issuance_transactions`
- `account_mptoken_issuance_transactions`

The migrator full-scans the `transactions` table, reuses the live ETL MPT extractor, writes the same deterministic index rows produced by live indexing, and waits for queued async writes before marking the migration as complete.

## Details

- Registers `MPTTransactionHistoryMigrator` with the Cassandra migration manager.
- Uses migration scan settings for full-table scanning:
  - `migration.full_scan_threads`
  - `migration.full_scan_jobs`
  - `migration.cursors_per_job`
- Adds Cassandra driver paging support so token-range scans read all result pages.
- Makes migration scans select explicit columns instead of relying on table column order.
- Propagates full-table scanner worker failures so failed scans do not get marked migrated.
- Keeps the migration non-blocking for Clio startup.

## Testing

Added coverage for:

- Migrator status before and after running.
- Backfilling MPT issuance create paths and multi-issuance transaction fan-out.
- Idempotent reruns.
- Empty transaction tables.
- Parity between migration backfill rows and live ETL rows.
- Failed transactions with explicit MPT issuance references.
- Cassandra paging across more than one driver page.
- Full-table scanner validation and worker error propagation.

### DIFF
--- a/src/data/cassandra/impl/Collection.hpp
+++ b/src/data/cassandra/impl/Collection.hpp
@@ -1,18 +1,29 @@
 #pragma once
 
-#include "data/cassandra/Error.hpp"
 #include "data/cassandra/impl/ManagedObject.hpp"
 
 #include <cassandra.h>
 #include <xrpl/basics/base_uint.h>
 
 #include <cstdint>
+#include <stdexcept>
+#include <string>
+#include <string_view>
 #include <vector>
 
 namespace data::cassandra::impl {
 
 class Collection : public ManagedObject<CassCollection> {
     static constexpr auto kDeleter = [](CassCollection* ptr) { cass_collection_free(ptr); };
+
+    static void
+    throwErrorIfNeeded(CassError const rc, std::string_view const label)
+    {
+        if (rc == CASS_OK)
+            return;
+        auto const tag = '[' + std::string{label} + ']';
+        throw std::logic_error(tag + ": " + cass_error_desc(rc));
+    }
 
 public:
     /* implicit */ Collection(CassCollection* ptr);
--- a/src/data/cassandra/impl/Collection.hpp
+++ b/src/data/cassandra/impl/Collection.hpp
@@ -1,29 +1,18 @@
 #pragma once
 
+#include "data/cassandra/Error.hpp"
 #include "data/cassandra/impl/ManagedObject.hpp"
 
 #include <cassandra.h>
 #include <xrpl/basics/base_uint.h>
 
 #include <cstdint>
-#include <stdexcept>
-#include <string>
-#include <string_view>
 #include <vector>
 
 namespace data::cassandra::impl {
 
 class Collection : public ManagedObject<CassCollection> {
     static constexpr auto kDeleter = [](CassCollection* ptr) { cass_collection_free(ptr); };
-
-    static void
-    throwErrorIfNeeded(CassError const rc, std::string_view const label)
-    {
-        if (rc == CASS_OK)
-            return;
-        auto const tag = '[' + std::string{label} + ']';
-        throw std::logic_error(tag + ": " + cass_error_desc(rc));
-    }
 
 public:
     /* implicit */ Collection(CassCollection* ptr);
--- a/src/data/cassandra/impl/Statement.hpp
+++ b/src/data/cassandra/impl/Statement.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "data/cassandra/Error.hpp"
 #include "data/cassandra/Types.hpp"
 #include "data/cassandra/impl/Collection.hpp"
 #include "data/cassandra/impl/ManagedObject.hpp"
@@ -17,7 +18,6 @@
 
 #include <cstddef>
 #include <cstdint>
-#include <stdexcept>
 #include <string>
 #include <string_view>
 #include <tuple>
@@ -170,14 +170,6 @@ public:
             // type not supported for binding
             static_assert(util::Unsupported<DecayedType>);
         }
-    }
-
-private:
-    static void
-    throwErrorIfNeeded(CassError const rc, std::string_view const label)
-    {
-        if (rc != CASS_OK)
-            throw std::logic_error(fmt::format("[{}]: {}", label, cass_error_desc(rc)));
     }
 };
 
--- a/src/data/cassandra/impl/Statement.hpp
+++ b/src/data/cassandra/impl/Statement.hpp
@@ -1,6 +1,5 @@
 #pragma once
 
-#include "data/cassandra/Error.hpp"
 #include "data/cassandra/Types.hpp"
 #include "data/cassandra/impl/Collection.hpp"
 #include "data/cassandra/impl/ManagedObject.hpp"
@@ -18,6 +17,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <stdexcept>
 #include <string>
 #include <string_view>
 #include <tuple>
@@ -75,7 +75,8 @@ public:
     setPagingSize(std::int32_t const pageSize) const
     {
         auto const rc = cass_statement_set_paging_size(*this, pageSize);
-        throwErrorIfNeeded(rc, "Set paging size");
+        if (rc != CASS_OK)
+            throw std::logic_error(fmt::format("[Set paging size]: {}", cass_error_desc(rc)));
     }
 
     /**
@@ -87,7 +88,8 @@ public:
     setPagingState(Result const& result) const
     {
         auto const rc = cass_statement_set_paging_state(*this, result);
-        throwErrorIfNeeded(rc, "Set paging state");
+        if (rc != CASS_OK)
+            throw std::logic_error(fmt::format("[Set paging state]: {}", cass_error_desc(rc)));
     }
 
     /**
@@ -101,9 +103,12 @@ public:
     bindAt(std::size_t const idx, Type&& value) const
     {
         using std::to_string;
-        auto throwBindingErrorIfNeeded = [idx](CassError rc, std::string_view label) {
-            if (rc != CASS_OK)
-                throwErrorIfNeeded(rc, fmt::format("{} at idx {}", label, idx));
+        auto throwErrorIfNeeded = [idx](CassError rc, std::string_view label) {
+            if (rc != CASS_OK) {
+                throw std::logic_error(
+                    fmt::format("[{}] at idx {}: {}", label, idx, cass_error_desc(rc))
+                );
+            }
         };
 
         auto bindBytes = [this, idx](auto const* data, size_t size) {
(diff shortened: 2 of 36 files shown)


src/data/cassandra/impl/Statement.hpp
```diff
@@ -76,12 +101,8 @@ class Statement : public ManagedObject<CassStatement> {
...
     {
         using std::to_string;
-        auto throwErrorIfNeeded = [idx](CassError rc, std::string_view label) {
-            if (rc != CASS_OK) {
-                throw std::logic_error(
-                    fmt::format("[{}] at idx {}: {}", label, idx, cass_error_desc(rc))
-                );
-            }
+        auto throwBindingErrorIfNeeded = [idx](CassError rc, std::string_view label) {
+            throwErrorIfNeeded(rc, fmt::format("{} at idx {}", label, idx));
```
Kassaking7: the fmt::format is a function argument, so it's evaluated on every call — including every successful bind — whereas the old lambda only formatted inside the rc != CASS_OK branch. bindAt is the per-parameter hot path for every prepared statement in clio (every ETL write and RPC read goes through PreparedStatement::bind's fold expression), and typical labels like "Bind xrpl::base_uint at idx 0" may exceed [SSO](https://stackoverflow.com/questions/10315041/meaning-of-acronym-sso-in-the-context-of-stdstring), so this will be a string construction + likely heap allocation per bound parameter, server-wide, to build an error message that's immediately discarded on success.

I would suggest this:
```
  auto throwBindingErrorIfNeeded = [idx](CassError rc, std::string_view label) {
      if (rc != CASS_OK)
          throwErrorIfNeeded(rc, fmt::format("{} at idx {}", label, idx));
  };
```

Kassaking7 (commented):
You may also want to add missing unit tests based on the codecov report

src/migration/cassandra/impl/TransactionsAdapter.cpp
```diff
@@ -1,19 +1,38 @@
...
+    // migration would hit it again forever. Log and skip the row instead of wedging the whole
+    // scan; genuine infrastructure failures (read errors) still fail closed in
+    // migrateInTokenRange.
+    std::optional<xrpl::STTx> sttx;
+    std::optional<xrpl::TxMeta> txMeta;
+    try {
+        xrpl::SerialIter it{txBlob.data(), txBlob.size()};
+        sttx.emplace(it);
+        txMeta.emplace(sttx->getTransactionID(), ledgerSeq, metaBlob);
+    } catch (std::exception const& e) {
```
Kassaking7: This catch is fail-open: skipping doesn't affect control flow, so MigratorsRegister still writes Migrated unconditionally, a rerun becomes a no-op, and the missing index rows are unrecoverable without manually editing migrator_status. It also catches more than the comment claims — std::bad_alloc from the allocation-heavy STTx/TxMeta parsing takes the same path, silently dropping a valid transaction. And with no counter or summary, even a 100% decode failure (e.g. a future same-typed column reorder in kSelectColumns, which Spec.hpp documents as undetectable) completes cleanly as Migrated with an empty index.

Since the transactions table is written by clio's own ETL from validated ledgers, an undecodable row means storage corruption or a code bug — both should abort the migration, not be skipped silently. Something like:
```
// TransactionsAdapter: narrow the catch, report failures to the owner instead of
// deciding policy here
} catch (std::runtime_error const& e) {  // parse errors only; bad_alloc etc. still fail the scan
    LOG(log_.error()) << "Failed to deserialize transaction: hash " << xrpl::strHex(txHash)
                      << ", ledger " << ledgerSeq << ": " << e.what();
    onDecodeFailure_();  // new callback, alongside onTransactionRead_
    return;
}

// MPTTransactionHistoryMigrator::runMigration: the migrator owns the policy
std::atomic<std::uint64_t> undecodableRows{0};
impl::TransactionsScanner scanner(
    {...},
    impl::TransactionsAdapter(backend, onTransaction, /* onDecodeFailure= */ [&] { ++undecodableRows; })
);
scanner.waitForAllAndThrowOnError();

if (auto const skipped = undecodableRows.load(); skipped > 0) {
    throw std::runtime_error(
        fmt::format("MPT backfill: {} transactions failed to deserialize; aborting so status stays NotMigrated", skipped)
    );
}
```
(Atomic because onRowRead runs concurrently across scan workers.) If you want tolerance for genuine bit-rot, make the threshold a config knob defaulting to 0 — but the count should be visible and a systematic failure must abort.

bryanjiang1:
> You may also want to add missing unit tests based on the codecov report

Added some unit tests but for `CassandraMigrationBackend.hpp` `MPTTransactionHistoryMigrator.cpp` and `MigrationApplication.cpp` these can't be unit tested since they need a Cassandra db. They are covered by integration tests.

bryanjiang1:
@PeterChen13579 @yinyiqian1 I can't add you guys as reviewers but please take a look when you guys can!

Kassaking7 (dismissed):
LGTM

src/data/cassandra/impl/Statement.hpp
```diff
@@ -76,12 +101,9 @@ class Statement : public ManagedObject<CassStatement> {
...
     {
         using std::to_string;
-        auto throwErrorIfNeeded = [idx](CassError rc, std::string_view label) {
-            if (rc != CASS_OK) {
-                throw std::logic_error(
-                    fmt::format("[{}] at idx {}: {}", label, idx, cass_error_desc(rc))
-                );
-            }
+        auto throwBindingErrorIfNeeded = [idx](CassError rc, std::string_view label) {
+            if (rc != CASS_OK)
```
PeterChen13579: You're checking `if (rc != CASS_OK)` twice, because `throwErrorIfNeeded` also does `if (rc != CASS_OK)` check

src/data/cassandra/impl/Collection.hpp
```diff
@@ -1,30 +1,19 @@
...
-#include <string_view>
 #include <vector>
 
 namespace data::cassandra::impl {
 
 class Collection : public ManagedObject<CassCollection> {
     static constexpr auto kDeleter = [](CassCollection* ptr) { cass_collection_free(ptr); };
 
-    static void
-    throwErrorIfNeeded(CassError const rc, std::string_view const label)
```
yinyiqian1, line 20: currently collection and statement is using the refactored function. But there are other places still using the local lambda function. So this refactor is not complete.
I suggest we revert these changes and leave the refactor in a separate PR.

PeterChen13579 (changes requested):
Just realized you have to merge with latest develop which updates to the newest libxrpl. Make sure to do a clean build. I will continue to review after this as merging with latest develop will change a lot

src/data/cassandra/impl/Statement.hpp
```diff
@@ -65,6 +66,30 @@ class Statement : public ManagedObject<CassStatement> {
...
+    /**
+     * @brief Set the Cassandra driver page size for this statement.
+     *
+     * @param pageSize Number of rows per driver page
+     */
+    void
+    setPagingSize(std::int32_t const pageSize) const
+    {
+        auto const rc = cass_statement_set_paging_size(*this, pageSize);
+        throwErrorIfNeeded(rc, "Set paging size");
```
yinyiqian1: the refactor of throwErrorIfNeeded is mainly for page.
We can keep the logic within Statement.hpp, instead of refactoring in error.hpp, collection.hpp, etc.

src/data/cassandra/impl/Statement.hpp
```diff
@@ -76,12 +101,9 @@ class Statement : public ManagedObject<CassStatement> {
     bindAt(std::size_t const idx, Type&& value) const
     {
         using std::to_string;
-        auto throwErrorIfNeeded = [idx](CassError rc, std::string_view label) {
-            if (rc != CASS_OK) {
-                throw std::logic_error(
-                    fmt::format("[{}] at idx {}: {}", label, idx, cass_error_desc(rc))
-                );
-            }
+        auto throwBindingErrorIfNeeded = [idx](CassError rc, std::string_view label) {
```
yinyiqian1: The change in `bindAt` does not change anything. The error message is almost the same after this change, just the `]`'s location is moved.

before change: [...] at idx ..
after change: [... at idx ..]

src/etl/impl/ext/MPT.cpp
```diff
@@ -51,26 +51,14 @@ MPTExt::writeMPTDataFromTransactions(model::LedgerData const& data)
...
-
-        if (txIndexRows > kIndexRowsPerTxWarningThreshold) {
-            LOG(log_.warn()) << "MPT issuance tx index fanout of " << txIndexRows
-                             << " rows exceeds the expected bound of "
-                             << kIndexRowsPerTxWarningThreshold << " for tx "
-                             << xrpl::strHex(tx.id);
-        }
-
-        indexRowsWritten += txIndexRows;
+            indexRowsWritten += 1 + record.accounts.size();
```
yinyiqian1: I don't think the change in MPT.cpp is needed

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -52,6 +55,46 @@ concept CanReadByTokenRange =
  */
 template <CanReadByTokenRange TableAdapter>
 class FullTableScanner {
+public:
+    /**
+     * @brief The full table scanner settings.
+     */
+    struct FullTableScannerSettings {
```
yinyiqian1: this struct does not need to be moved

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -52,6 +55,46 @@ concept CanReadByTokenRange =
...
+        std::uint32_t ctxThreadsNum;  ///< number of threads used in the execution context
+        std::uint32_t jobsNum;  ///< number of coroutines to run, it is the number of concurrent
+                                ///< database reads
+        std::uint32_t cursorsPerJob;  ///< number of cursors per coroutine
+    };
+
+private:
+    [[nodiscard]] static std::uint32_t
+    validatedCtxThreadsNum(FullTableScannerSettings const& settings)
```
yinyiqian1: this is not needed

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -52,6 +55,46 @@ concept CanReadByTokenRange =
...
+    {
+        ASSERT(
+            settings.ctxThreadsNum > 0,
+            "ctxThreadsNum for full table scanner must be greater than 0"
+        );
+        return settings.ctxThreadsNum;
+    }
+
+    [[nodiscard]] static std::size_t
+    computeCursorsNum(FullTableScannerSettings const& settings)
```
yinyiqian1: This is not needed

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -139,30 +172,45 @@ class FullTableScanner {
      */
     template <typename ExecutionContextType = util::async::CoroExecutionContext>
     FullTableScanner(FullTableScannerSettings settings, TableAdapter&& reader)
-        : ctx_(ExecutionContextType(settings.ctxThreadsNum))
-        , cursorsNum_(settings.jobsNum * settings.cursorsPerJob)
+        : ctx_(ExecutionContextType(validatedCtxThreadsNum(settings)))
```
yinyiqian1: we don't need to refactor to pass in validatedCtxThreadsNum

src/migration/cassandra/impl/FullTableScanner.hpp
```diff
@@ -139,30 +172,45 @@ class FullTableScanner {
...
         , reader_{std::move(reader)}
     {
-        ASSERT(settings.jobsNum > 0, "jobsNum for full table scanner must be greater than 0");
-        ASSERT(
-            settings.cursorsPerJob > 0,
-            "cursorsPerJob for full table scanner must be greater than 0"
-        );
-
-        auto const cursors = TokenRangesProvider{cursorsNum_}.getRanges();
+        auto const cursors =
```
yinyiqian1: We don't need the refactor.

PeterChen13579:
Continuing work from https://github.com/XRPLF/clio/pull/3211

PR #3131, opened 2026-07-01
build: Don't reuse binaries between different C++ versions

mathbunnyru:
By default, Conan tries to use binaries built with other cppstd version.
It makes code less reproducible.
Let's force all binaries to have an exact C++ version.

Small tweaks:
- Style in templates is unified
- Everything was pre-built here: https://github.com/XRPLF/clio/actions/runs/28514120255?pr=3131


godexsoft (approved):
:rage2:

PR #3137, opened 2026-07-07
fix: Shared SSL context

godexsoft:
Recreating SSL context unnecessarily is expensive. This PR makes it happen once and share the context instead.


kuznetsss (dismissed):

src/util/requests/impl/SslContext.cpp
```diff
@@ -79,23 +80,34 @@ getRootCertificate()
...
 
+}  // namespace
+
+std::expected<std::reference_wrapper<ssl::context>, RequestError>
+getClientSslContext()
+{
+    static std::expected<ssl::context, RequestError> context = makeClientSslContext();
+
+    if (not context.has_value())
+        return std::unexpected{context.error()};
```
mathbunnyru: This doesn't change on a client, maybe we should crash on app startup if we have an error in our configuration?

It would be better than not responding to all the clients due to misconfiguration.

Also, the on-client code can be simplified to only handle `std::reference_wrapper<ssl::context>` when getting ssl context

mathbunnyru (changes requested):

src/util/requests/impl/SslContext.cpp
godexsoft: I did not want to change behaviour with my changes. But you are not wrong.. it may be better to actually do that. I'm not 100% sure still. @kuznetsss wdyt?

kuznetsss: It will be a bit tricky because we would need to add a config option to allow no ssl forwarding and check that option to decide whether we should crash on assert

mathbunnyru: Do we really need to add a config option?
We don't require anything unusual in `makeClientSslContext`, so it should always work. Machine should be really misconfigured for it not to work, tbh.

godexsoft: I implemented the suggestion in the end. This does feel more correct indeed

godexsoft: I don't think we need config option either.. it's a bit simpler than that, albeit more complex than i wanted it to be

src/util/requests/impl/SslContext.cpp
```diff
@@ -47,55 +47,80 @@ constexpr std::array kCertFilePaths{
...
 {
     // Honor the OpenSSL-standard SSL_CERT_FILE environment variable first. Some
     // environments (e.g. the Nix-based CI/runtime image) point it at their CA
     // bundle instead of installing certificates at the well-known system paths.
     if (char const* const certFile = std::getenv("SSL_CERT_FILE"); certFile != nullptr) {
-        if (auto contents = readCertificateFile(certFile); contents.has_value()) {
-            return *std::move(contents);
-        }
+        if (auto contents = readCertificateFile(certFile); contents.has_value())
+            return contents;
```
mathbunnyru, line 71: I think you could've kept `*std::move` here and below - this will probably be marked by clang-tidy as missed optimization opportunity at some point

src/util/requests/impl/SslContext.hpp
```diff
@@ -11,8 +11,39 @@
 
 namespace util::requests::impl {
 
+/**
+ * @brief Create a client SSL context from the given root certificate.
+ *
+ * @param rootCertificate The PEM-encoded root certificate bundle to trust.
+ * @return The client SSL context, or a RequestError if @p rootCertificate is empty.
+ */
```
mathbunnyru: It's in `impl`, so this wasn't strictly needed, but I don't mind having doxygen here too

mathbunnyru (dismissed):
Overall, LGTM, but I wouldn't remove `*std::move`

kuznetsss (dismissed):

src/util/requests/impl/SslContext.cpp
godexsoft: i think it should be the opposite, should be marked as redundant move by clang-tidy as the compiler is supposed to implicitly treat this as rvalue in this context

mathbunnyru: Ah, you changed the type, so now it's the same as in the readCertificateFile, so yeah.

mathbunnyru (approved):

PR #3141, opened 2026-07-08
feat: Add mptoken_issuance_history RPC

bryanjiang1:
feat: add mptoken_issuance_history RPC

Summary

Adds mptoken_issuance_history, a Clio-only method that returns the transaction history for a given MPT issuance — the MPT-scoped sibling of nft_history. You can optionally filter by account and/or tx_type.

The index tables, backend fetch methods, and live ETL indexing landed earlier; this wires up the handler on top of them.

The handler

- Routes to fetchMPTokenIssuanceTransactions or fetchAccountMPTokenIssuanceTransactions depending on whether account is set. When tx_type is given, it filters on TransactionType post-fetch, exactly like account_tx.
- Follows the nft_history/account_tx conventions for ledger ranges, markers, binary, forward, and limit ([1,100], default 50), including api-version response branching.
- Gated on backfill completion so it never returns partial history: until this node's MPTTransactionHistoryMigrator reports Migrated, requests get notReady with a message pointing at the --migrate command. Once migrated, the result is cached and the check is skipped — the method turns on automatically, no restart.


yinyiqian1:
The handler PR should not contain the migration code

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,309 @@
...
+    auto const [blobs, retCursor] = txnsAndCursor;
+
+    if (retCursor)
+        response.marker = {.ledger = retCursor->ledgerSequence, .seq = retCursor->transactionIndex};
+
+    for (auto const& txnPlusMeta : blobs) {
+        // A hash with no matching Transactions row yields a default-constructed record in-position.
+        // Skip it before the range check so it neither shortens the page nor disturbs the marker.
+        if (txnPlusMeta.transaction.empty() || txnPlusMeta.metadata.empty())
+            continue;
```
yinyiqian1, line 203: can we add a log here?

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+MPTokenIssuanceHistoryHandler::Result
+MPTokenIssuanceHistoryHandler::process(
+    MPTokenIssuanceHistoryHandler::Input const& input,
+    Context const& ctx
+) const
+{
+    // Fail closed unless the backfill is done: partial history must never be served.
+    if (not migrated_->load(std::memory_order_relaxed)) {
+        auto const status = migrationInspector_->getMigratorStatusByName(kMigratorName);
```
yinyiqian1: we don't need to make migrationInspector_ a member to call this.
We can use sharedPtrBackend_.

yinyiqian1 (dismissed):

PeterChen13579:
@maria-robobug Please note there is an extra RPC handler we are adding to clio

mathbunnyru (changes requested):
Please, merge latest changes - CI doesn't seem to have run, and you also didn't allow maintainers to update your branch, so I can't do it myself

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,310 @@
...
+                migration::MigratorStatus::Status::Migrated) {
+            migrated_->store(true, std::memory_order_relaxed);
+        } else {
+            return Error{Status{
+                RippledError::RpcNotReady,
+                "mptoken_issuance_history is unavailable until the MPT transaction-history "
+                "backfill "
+                "completes on this node. Run: ./clio_server --migrate "
+                "MPTTransactionHistoryMigrator "
+                "CONFIG"
```
mathbunnyru: Formatting of comment doesn't look nice

src/rpc/handlers/MPTokenIssuanceHistory.hpp
```diff
@@ -0,0 +1,193 @@
...
+ * MPTokenIssuance, optionally filtered by an affected account and/or transaction type.
+ *
+ * @note This is a Clio-only method. Requests fail with `notReady` until the MPT
+ * transaction-history backfill reports `Migrated`, so partial history is never served.
+ */
+class MPTokenIssuanceHistoryHandler {
+    util::Logger log_{"RPC"};
+    std::shared_ptr<BackendInterface> sharedPtrBackend_;
+    // Status is monotonic, so the terminal Migrated result is cached across handler copies.
+    std::shared_ptr<std::atomic_bool> migrated_ = std::make_shared<std::atomic_bool>(false);
```
mathbunnyru, line 49: Do we actually copy our handlers?
If not, we should probably get rid of shared_ptr and make this class non-copyable

yinyiqian1:
@BryanJ1ang Could you add more tests regarding the coverage report?

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,310 @@
...
+            }};
+        }
+    }
+
+    auto const range = sharedPtrBackend_->fetchLedgerRange();
+    ASSERT(range.has_value(), "MPTokenIssuanceHistory's ledger range must be available");
+
+    auto [minIndex, maxIndex] = *range;  // NOLINT(bugprone-unchecked-optional-access)
+
+    if (input.ledgerIndexMin) {
```
mathbunnyru:
```suggestion
    if (input.ledgerIndexMin.has_value()) {
```

and in similar places too

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,310 @@
...
+#include <cstdint>
+#include <limits>
+#include <optional>
+#include <string>
+#include <utility>
+
+namespace rpc {
+
+MPTokenIssuanceHistoryHandler::Result
+MPTokenIssuanceHistoryHandler::process(
```
mathbunnyru, line 47: This function is very long, could you please refactor it in smaller, but meaningful private methods?

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,310 @@
...
+    MPTokenIssuanceHistoryHandler::Input const& input,
+    Context const& ctx
+) const
+{
+    // Fail closed unless the backfill is done: partial history must never be served.
+    if (not migrated_->load(std::memory_order_relaxed)) {
+        auto const statusString = sharedPtrBackend_->fetchMigratorStatus(kMigratorName, ctx.yield);
```
mathbunnyru: I'm not sure if it's a good idea, but maybe we should have a small refactoring PR, so fetchMigratorStatus returns MigratorStatus, or optional<MigratorStatus> or sth like that?

src/rpc/handlers/MPTokenIssuanceHistory.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+    // if marker exists
+    if (input.marker) {
+        cursor = {input.marker->ledger, input.marker->seq};
+    } else {
+        if (input.forward) {
+            cursor = {minIndex, 0};
+        } else {
+            cursor = {maxIndex, std::numeric_limits<int32_t>::max()};
+        }
```
mathbunnyru: Can we use designated initializers here to make it more clear what's going on?

mathbunnyru (changes requested):
I think process function is quite difficult to read, because it's too long, so this should be refactored.

General comments:
1. Please, fix clang-tidy
2. And PR title
3. And add more test coverage (I agree with @PeterChen13579)

src/rpc/handlers/MPTokenIssuanceHistory.hpp
bryanjiang1: Yes, handlers are copied per request. `getHandler()` returns `AnyHandler` by value, and `clone() `copies the concrete handler. 

The shared atomic keeps the cached `Migrated` state across those copies. otherwise each request would query the status again.

src/rpc/handlers/MPTokenIssuanceHistory.cpp
bryanjiang1: TransactionsCursor has user declared constructors so we can't use designated initializers here. Added some comments to make it more clear

src/rpc/handlers/MPTokenIssuanceHistory.cpp
bryanjiang1: Yeah I think this would make sense for a separate PR

src/rpc/CMakeLists.txt
```diff
@@ -58,4 +59,4 @@ target_sources(
         handlers/VaultInfo.cpp
 )
 
-target_link_libraries(clio_rpc PUBLIC clio_util clio_data)
+target_link_libraries(clio_rpc PUBLIC clio_util clio_data clio_migration)
```
yinyiqian1: This line should be reverted as well

src/rpc/handlers/MPTokenIssuanceHistory.hpp
mathbunnyru: > otherwise each request would query the status again.

Are you sure that we actually copy the handler on each request?

src/rpc/handlers/MPTokenIssuanceHistory.cpp
mathbunnyru: May I please ask you to do that?

src/rpc/handlers/MPTokenIssuanceHistory.hpp
bryanjiang1: Yeah. The trace is: RPCServerHandler → RPCEngine::buildResponse() → getHandler() → copies the map-held AnyHandler → clone() → copies the concrete handler.

So each locally handled request gets a handler copy

mathbunnyru: That's unfortunate, thanks for the explanation

mathbunnyru (approved):

src/rpc/handlers/MPTokenIssuanceHistory.cpp
PeterChen13579: @mathbunnyru Bryan intern finished. I'll create a task for this and pick it up when I have time.

PR #3142, opened 2026-07-09
style: Unify style for all Doxygen comments

mathbunnyru:


src/etl/CacheLoaderSettings.hpp
```diff
@@ -13,43 +13,52 @@ namespace etl {
...
 
-    /** @brief Settings for cache file operations */
+    /**
+     * @brief Settings for cache file operations
+     */
     struct CacheFileSettings {
         std::string
-            path; /**< path to the file to load cache from on start and save cache to on shutdown */
-        uint32_t maxAge = 5000; /**< max difference between latest sequence in cache file and DB */
+            path;  ///< path to the file to load cache from on start and save cache to on shutdown
```
godexsoft, line 26: nit: Didn't we prefer `/**<` for these?

src/rpc/handlers/GetAggregatePrice.hpp
```diff
@@ -181,8 +181,8 @@ class GetAggregatePriceHandler {
 private:
     /**
      * @brief Calls callback on the oracle ledger entry
-     If the oracle entry does not contains the price pair, search up to three previous metadata
-     objects. Stops early if the callback returns true.
+     * If the oracle entry does not contains the price pair, search up to three previous metadata
```
godexsoft, line 185: nit: We usually try to separate brief from detail by a space line

src/util/Channel.hpp
```diff
@@ -29,20 +29,19 @@ struct ChannelInstantiated;
  * @brief Specifies the producer concurrency model for a Channel.
  */
 enum class ProducerType {
-    Single, /**< Only one Sender can exist (non-copyable). Uses direct Guard ownership for zero
-               overhead. */
-    Multi   /**< Multiple Senders can exist (copyable). Uses shared_ptr<Guard> for shared ownership.
-             */
+    Single,  ///< Only one Sender can exist (non-copyable). Uses direct Guard ownership for zero
+             ///< overhead.
```
godexsoft: Is it because of this? Alignment looks off now

godexsoft (changes requested):

src/etl/CacheLoaderSettings.hpp
mathbunnyru: In rippled, there are actually more places with `///<`.
In clio, there are already many places with `///<` too.

So, overall - incosistent.

Both these styles are in their own way good:
- `/**<` uses stars more similar to regular doxygen comments we have
- `///<` doesn't require you to clutter the code on the right side, just as our regular doxygen comments

Ans since inline doxygen comments are a bit different, I think it's ok to use `/` instead of `*`.
In the end, it's all getting more consistent, and I like not having to worry to close your comment, so I suggest we do it this way.

src/util/Channel.hpp
mathbunnyru: Fixed, and I think it reads better than before

src/rpc/handlers/GetAggregatePrice.hpp
mathbunnyru: Fixed

src/etl/CacheLoaderSettings.hpp
godexsoft: Meh, i'm not gonna fight over this, especially since we already had inconsistent comments before. It is getting better overall so i'm happy.

godexsoft (approved):
:rage1:

PR #3144, opened 2026-07-20
chore: Use libxrpl 3.3.0-rc1

mathbunnyru:
There is a non-trivial fix because of sponsored feature: https://github.com/XRPLF/clio/pull/3144/changes/f5b0802a88f05a0318f47abfb5373bda0c99e6f7


tests/unit/rpc/handlers/AccountObjectsTests.cpp
```diff
@@ -1418,15 +1424,18 @@ TEST_F(RPCAccountObjectsHandlerTest, NFTLimitAdjust)
         .WillOnce(Return(Blob{'f', 'a', 'k', 'e'}));
 
     std::string first{kIndex1};
-    auto current =
-        xrpl::keylet::nftpage(xrpl::keylet::nftpageMin(account), xrpl::uint256{first.c_str()}).key;
+    auto current = xrpl::keylet::nftokenPage(
```
godexsoft, line 1427: Yikes formatting here... oneway to do better is to manually add `//` at end of some line so .key is not in middle of nowhere like this :( clang-format letting us down a bit.

godexsoft (approved):
Formatting could be nicer but it's not crucial

tests/unit/rpc/handlers/AccountObjectsTests.cpp
mathbunnyru: As agreed, we can merge without this for now

PR #3149, opened 2026-07-21
fix: Amendment block initial ledger load and cache load

godexsoft:
Fix: don't crash / silently stall when connecting to a rippled newer than the bundled libxrpl**

When Clio connects to a rippled whose network has an amendment this build's libxrpl doesn't understand, it hit two bad failure modes. This PR makes both fail safely.

**What changed**
- **ETL extraction → clean amendment-block.** Ledger deserialization failures (e.g. `Invalid transaction type`, `Unknown field` from a new amendment) were thrown inside the ETL extractor and silently swallowed by the coroutine's error handler — ETL just died with no log. The `Extractor` now takes the shared `AmendmentBlockHandler`; a deserialization failure during extraction transitions to the amendment-blocked state (like the loader already does) and is logged, instead of vanishing. Extraction completion logs now also reflect the real outcome instead of always printing "Extracted and transformed…".
- **Cache loader → graceful degrade instead of process abort.** The cache-loader worker only guarded against `DatabaseTimeout` (via `retryOnTimeout`); any other failure (non-timeout DB error, etc.) escaped the worker coroutine and took the whole process down via `std::terminate`. The worker now catches everything, disables the cache, and continues serving from the database.

**Edge cases fixed**
- Backbone rippled ahead of Clio's libxrpl (unknown tx/object types): ETL now amendment-blocks with a clear log, rather than silently stalling or busy-looping.
- Non-timeout backend error during cache loading: cache is disabled and Clio keeps running, instead of aborting.

**Tests**
- `ExtractionTests`: extraction deserialization failure triggers `notifyAmendmentBlocked` and returns no data (both `extractLedgerOnly` and `extractLedgerWithDiff`).
- `CacheLoaderTests`: a non-timeout backend error disables the cache and the loader winds down cleanly (no escape/terminate).


src/etl/impl/CacheLoader.hpp
```diff
@@ -98,46 +99,62 @@ class CacheLoaderImpl {
...
-                auto cursor = queue_.tryPop();
-                if (not cursor.has_value()) {
-                    return;  // queue is empty
-                }
-
-                auto [start, end] = *cursor;
-                LOG(log_.debug()) << "Starting a cursor: " << xrpl::strHex(start);
-
+            try {
                 while (not token.isStopRequested() and not cache_.get().isDisabled()) {
```
mathbunnyru: Can we put this whole block to a separate method/function and then add a test which checks which exception gets thrown?

src/etl/impl/Extraction.cpp
```diff
@@ -180,19 +182,36 @@ Extractor::unpack()
     };
 }
 
+std::optional<model::LedgerData>
+Extractor::guardedUnpack(std::optional<PBLedgerResponseType> response, uint32_t seq)
```
mathbunnyru:
```suggestion
Extractor::guardedUnpack(std::optional<PBLedgerResponseType>&& response, uint32_t seq)
```

This is only called with rvalue, and we can avoid a copy (probably expensive), on each call

src/etl/impl/Extraction.cpp
```diff
@@ -180,19 +182,36 @@ Extractor::unpack()
...
+{
+    try {
+        return std::move(response).and_then(unpack());
+    } catch (std::runtime_error const& e) {
+        LOG(log_.fatal()) << "Failed to extract/deserialize ledger " << seq
+                          << " - the network likely has an amendment this Clio does not support: "
+                          << e.what();
+        amendmentBlockHandler_->notifyAmendmentBlocked();
```
mathbunnyru, line 192: Just in case - this is safe to run from other threads, right?

tests/unit/etl/ExtractionTests.cpp
```diff
@@ -370,9 +371,23 @@ struct MockFetcher : etl::LedgerFetcherInterface {
...
+    etl::impl::Extractor extractor{fetcher, amendmentBlockHandler};
 };
 
+inline etl::impl::PBLedgerResponseType
+makeResponseWithUndeserializableTx()
+{
+    auto response = util::createData();
+    auto& txs = *response.mutable_transactions_list();
+    std::string const badTxBlob{
+        static_cast<char>(0x12), static_cast<char>(0xff), static_cast<char>(0xff)
```
mathbunnyru: I hope this is not the one which we be added in 3.3.0? 😂 
Otherwise we'll have to change the test to make it bad again 🙂

mathbunnyru (commented):
How do you think - maybe we need integration test too?

src/etl/impl/Extraction.cpp
godexsoft: Turns out it was not anymore, although most internals are thread safe, the optional holding the operation is not.
A better refactoring would be to simplify all this so that it's just an atomic bool flag but that's out of scope for this bugfix so let's just add a mutex for now to make it 100% safe.

tests/unit/etl/ExtractionTests.cpp
godexsoft: No but also turns out the way it was written it triggered a different error path, not exactly the one we were trying to fix for. So i'm updating this to exercise the actual error we wanted to fix.

mathbunnyru (approved):

PR #3152, opened 2026-07-23
feat: Support Confidential Transfer Fields

PeterChen13579:
Add Confidential Transfer Fields to MPToken, MPToken Issuance, and mpt_holders API


yinyiqian1 (approved):
LGTM

PR #3154, opened 2026-07-24
feat: Add accounts filter to mpt_holders

bryanjiang1:
## Add an accounts filter to mpt_holders
Adds an optional accounts parameter. When present, the handler skips the holder-index scan entirely and looks the requested accounts up directly by key (keylet::mptoken). Response shape remains the same as before, including all the confidential-transfer fields.

Note:
- accounts is a list of base58 addresses, non-empty, capped at 100. 
- Accounts that don't hold the token are left out of the response.
- This mode isn't paginated, so accounts + marker is rejected.


PeterChen13579:
Continuing work from https://github.com/XRPLF/clio/pull/3180 as we do not have permission to push to Bryan's personal fork

PR #3166, opened 2026-08-04
fix: Remove clio_etl from rpc cmake

PeterChen13579:


src/etl/MPTHelpers.hpp
```diff
@@ -64,11 +152,23 @@ getMPTokenIssuanceTxsFromTx(xrpl::TxMeta const& txMeta, xrpl::STTx const& sttx);
  * @param mptIssuanceID The MPT issuance to check for.
  * @return true if the transaction references mptIssuanceID.
  */
-bool
+inline bool
```
mathbunnyru: This is hiding the problem, not fixing it.

It doesn't remove dependency between components, just doesn't force you to declare it and makes it implicit.

Can we move the code which needs to be shared to some shared component?
Maybe somewhere to util?

mathbunnyru (changes requested):

src/etl/MPTHelpers.hpp
PeterChen13579: Thanks, I never knew this. Yea, I think then moving to util is the way to go

src/util/MPTIssuanceUtils.hpp
```diff
@@ -0,0 +1,58 @@
+/** @file */
```
mathbunnyru: This look strange

mathbunnyru (changes requested):
There is one suspicious place, otherwise looks like moving the code to the appropriate places (which is exactly what was needed 👍 )

mathbunnyru (approved):

PR #3167, opened 2026-08-05
fix: Introduce retry for transient errors from DB

godexsoft:


src/data/cassandra/SettingsProvider.cpp
```diff
@@ -25,6 +26,12 @@ SettingsProvider::SettingsProvider(util::config::ObjectView const& cfg)
     , keyspace_{cfg.get<std::string>("keyspace")}
     , tablePrefix_{cfg.maybeValue<std::string>("table_prefix")}
     , replicationFactor_{cfg.get<uint16_t>("replication_factor")}
+    , initialRetryDelay_{util::config::ClioConfigDefinition::toMilliseconds(
+          cfg.get<float>("initial_request_retry_delay")
```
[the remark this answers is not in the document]
godexsoft: This is not an issue.

docs/config-description.md
```diff
@@ -132,10 +132,26 @@ This document provides a list of all available Clio configuration properties in
 ### database.cassandra.request_timeout
 
 - **Required**: False
-- **Type**: int
+- **Type**: double
 - **Default value**: None
-- **Constraints**: The minimum value is `1`. The maximum value is `4294967295`.
-- **Description**: The maximum amount of time in seconds that the system waits for a request to be fetched from the database.
+- **Constraints**: The value must be a positive double number.
+- **Description**: The maximum amount of time in seconds that the system waits for a request to be fetched from the database. Should be set higher than the server side read timeout. If omitted, no request timeout is applied.
```
mathbunnyru, line 138: `Should be set higher than the server side read timeout.` - do we check this on startup?

src/data/cassandra/CassandraBackendFamily.hpp
```diff
@@ -139,6 +139,24 @@ class CassandraBackendFamily : public BackendInterface {
      */
     CassandraBackendFamily(CassandraBackendFamily&&) = delete;
 
+    /**
+     * @return The configured delay before the first retry
+     */
+    [[nodiscard]] std::chrono::milliseconds
+    retryInitialDelay() const override
```
mathbunnyru: Naming is a bit incosistent, should probably be `initialRetryDelay` here (and rename similarly below)

src/data/BackendInterface.hpp
```diff
@@ -35,41 +36,117 @@
...
+     */
+    explicit DatabaseError(std::string message) : message_{std::move(message)}
+    {
+    }
+
     /**
      * @return The error message as a C string
      */
     [[nodiscard]] char const*
-    what() const throw() override
```
mathbunnyru, line 46: If you could search the codebase and if it has something like that anywhere else, would be nice to fix

src/data/cassandra/impl/ExecutionStrategy.hpp
```diff
@@ -551,11 +552,13 @@ class DefaultExecutionStrategy {
...
-        if (err.isTimeout())
-            throw DatabaseTimeout();
-
+        // NOTE: etl::impl::Loader and etl::impl::Extractor treat std::runtime_error as
+        // "amendment blocked", so only genuinely permanent failures may be thrown as one.
         if (err.isInvalidQuery())
             throw std::runtime_error("Invalid query");
+
+        // anything else, including unclassified codes, is transient and gets retried
+        throw DatabaseError{"Database error: " + err.message()};
```
mathbunnyru: Maybe it would be worth saving err.code here, and then logging it?

src/data/BackendInterface.hpp
```diff
@@ -35,41 +36,117 @@
...
+    util::ExponentialBackoffStrategy backoff{initialDelay, maxDelay};
+    std::size_t attempt = 1;
+
     while (true) {
         try {
             return func();
-        } catch (DatabaseTimeout const&) {
-            LOG(log.error()) << "Database request timed out. Sleeping and retrying ... ";
-            std::this_thread::sleep_for(std::chrono::milliseconds(waitMs));
+        } catch (DatabaseError const& e) {
```
mathbunnyru, line 141: Are all errors we can get are safe to retry?

src/etl/impl/LedgerPublisher.hpp
```diff
@@ -130,9 +130,12 @@ class LedgerPublisher : public LedgerPublisherInterface {
...
             }
 
-            auto lgr = data::synchronousAndRetryOnTimeout([&](auto yield) {
-                return backend_->fetchLedgerBySequence(ledgerSequence, yield);
-            });
+            // flat delay: this blocks a thread of the shared pool
+            auto lgr = data::synchronousAndRetryOnTimeout(
+                [&](auto yield) { return backend_->fetchLedgerBySequence(ledgerSequence, yield); },
+                data::kDefaultWaitBetweenRetry,
+                data::kDefaultWaitBetweenRetry
```
mathbunnyru: I think it's worth creating a struct, holding 2 delays, and instead of 2 params pass this struct

In LedgerPublisher you can even have a field with 2 `data::kDefaultWaitBetweenRetry`, and pass it

docs/config-description.md
godexsoft: No because it's "should be" not "must be". It's just a recommendation by ScyllaDB folks

src/data/BackendInterface.hpp
godexsoft: Should be yeah.

src/data/BackendInterface.hpp
godexsoft: There is only 2 more so let's do it in this PR

src/data/cassandra/impl/ExecutionStrategy.hpp
godexsoft: DatabaseError is part of the global BackendInterface so it should not have to know about details of specific DB errors (like codes). We could include that in the string that we hold in DatabaseError though if that helps, although i think message alone is probably enough

src/data/cassandra/CassandraBackendFamily.hpp
godexsoft: Inconsistent with what exactly? This was chosen because the two functions are talking about "retry". So the logical order of the name here is "Component(retry)->Setting(initialDelay)" not "first half of setting(initial)->component(retry)->second part of setting(delay)". I don't mind changing it, just don't understand what you compared it with so it's inconsistent in your opinion.

mathbunnyru: I see you changed it, there was incosistency in naming:
`retryInitialDelay` sounded like a verb, like it does something (while it doesn't)
`initialRetryDelay` is correct

src/data/BackendInterface.hpp
mathbunnyru: Thanks

src/data/cassandra/impl/ExecutionStrategy.hpp
mathbunnyru: Yes, including in the message is a good idea - keeps it more generic, while not loosing information

src/data/BackendInterface.hpp
mathbunnyru: Let's hope 🙂

src/data/BackendInterface.hpp
```diff
@@ -35,41 +36,119 @@
...
+ * @return The same as the return type of func
+ */
+template <typename FnType>
+auto
+retryOnTimeout(FnType func, boost::asio::yield_context yield, RetryDelays delays = {})
+{
+    static util::Logger const log{"Backend"};  // NOLINT(readability-identifier-naming)
+
+    auto retry = util::makeRetryExponentialBackoff(
+        delays.initialDelay, delays.maxDelay, yield.get_executor()
```
mathbunnyru: Let's move RetryDelays definition to utils, and simplify interfaces here as well?

src/data/cassandra/SettingsProvider.hpp
```diff
@@ -62,6 +65,24 @@ class SettingsProvider {
...
+    }
+
+    /**
+     * @return Upper bound for the retry backoff
+     */
+    [[nodiscard]] std::chrono::milliseconds
+    getMaxRetryDelay() const
+    {
+        return maxRetryDelay_;
+    }
```
mathbunnyru, line 84: Can be stored as a single RetryDelays field, and a single getter

src/util/Retry.cpp
```diff
@@ -98,4 +118,16 @@ makeRetryExponentialBackoff(
     return Retry(std::make_unique<ExponentialBackoffStrategy>(delay, maxDelay), std::move(strand));
 }
 
+Retry
+makeRetryExponentialBackoff(
+    std::chrono::steady_clock::duration delay,
+    std::chrono::steady_clock::duration maxDelay,
```
mathbunnyru: Should be RetryDelays here too

src/data/cassandra/SettingsProvider.hpp
godexsoft: Can be but should not be imo - these represent independent settings on cassandra/scylla

src/util/Retry.cpp
godexsoft: I did not want to change the Retry interface too much but i guess this is better long term so let's do it.

mathbunnyru (dismissed):

src/data/cassandra/SettingsProvider.hpp
mathbunnyru: Well, you can actually make it a struct - it would look great on a config side too

mathbunnyru (approved):

PR #3169, opened 2026-08-11
refactor: Separate handler registry

godexsoft:
This is RPC spec migration PR 1 out of many.

In this PR the handlers are routed through the new registry which is a list of handlers and their properties but without creating the actual handlers yet - instantiation still happens where it used to happen before but we gain ability to peek into the list of supported handlers before we instantiate them (useful for spec dump later on).


bthomee (dismissed):

src/rpc/common/impl/HandlerRegistry.cpp
```diff
@@ -0,0 +1,319 @@
+#include "rpc/common/impl/HandlerRegistry.hpp"
+
+#include "rpc/Counters.hpp"  // NOLINT(misc-include-cleaner)
```
mathbunnyru: I think we write iwyu pragma differently

godexsoft: Good find 👍

mathbunnyru (dismissed):

mathbunnyru (approved):

PR #3171, opened 2026-08-11
chore: Introduce `xrpl-rpc-spec` 0.1.6 dependency

godexsoft:
This is RPC spec migration PR 2 out of many.

In this PR we add a dependency on the xrpl-rpc-spec that is now available to us through [conan-center-index](https://github.com/XRPLF/conan-center-index) and remove `JsonBool` which is now an alias to one in the spec repo.


src/rpc/common/JsonBool.hpp
```diff
@@ -1,58 +1,13 @@
...
-
-#include <stdexcept>
+#include <rpcspec/JsonBool.hpp>
 
 namespace rpc {
 
 /**
- * @brief A wrapper around bool that allows to convert from any JSON value
- */
-struct JsonBool {
```
godexsoft, line 15: This stuff moved to the spec repo, that's why here it's removed and aliased instead.

godexsoft: fyi: Eventually this entire file will be gone.

bthomee (dismissed):

cmake/deps/rpcspec.cmake
```diff
@@ -0,0 +1,5 @@
+find_package(xrpl-rpc-spec REQUIRED CONFIG)
+
+if(NOT DEFINED RPCSPEC_IS_CLIO)
+    add_compile_definitions(RPCSPEC_IS_CLIO=1)
```
mathbunnyru: This sets a global compile definition, even for targets which don't know anything about rpc/rpc-spec.

To do it properly, it should be a param of xrpl-rpc-spec, and depending on that the conan package itself for xrpl-rpc-spec will set it when being used as a dependency.
This way, only cmake targets which depend on xrpl-rpc-spec will get this define

mathbunnyru (changes requested):

cmake/deps/rpcspec.cmake
godexsoft: This was applied in spec 0.1.6 package and in this PR

conanfile.py
```diff
@@ -1,3 +1,4 @@
+from conan.errors import ConanInvalidConfiguration
```
mathbunnyru: You don't need this here

CMakeLists.txt
```diff
@@ -84,6 +84,8 @@ include(deps/cassandra)
 include(deps/libbacktrace)
 include(deps/spdlog)
 
+include(deps/rpcspec)
```
mathbunnyru, line 84: How about we sort these?

mathbunnyru (changes requested):

mathbunnyru (approved):

PR #3173, opened 2026-08-12
refactor: Use error types from rpc-spec

godexsoft:
This is PR 3 out of many. It depends on merging #3171 first.

Moves the RPC error model into the shared spec library. `src/rpc/Errors.hpp` becomes a shim over `<rpcspec/Errors.hpp>`, keeping only the Clio-side rendering: `makeError`, `getErrorInfo`, `getEtlErrorInfo`.

`ClioError::Etl*` is carved out into a Clio-side `etl::EtlError` (new `src/etl/Errors.hpp`), since ETL codes aren't an RPC-spec concern. `Source::forwardToRippled` now returns `EtlError` instead of `ClioError`, which is what most of the diff is.

**Wire-visible change:** the "clio only serves validated data" warning now says *"if you want to talk to **xrpld**"* instead of *"rippled"* — the string moved into rpcspec, which renamed it. 10 expectations updated in `RPCServerHandlerTests.cpp`.

**Drive-by cleanups**:

- Deleted `InvalidParamsError` and `AccountNotFoundError` - unreferenced anywhere in `src/` or `tests/`.
- Both error-info tables become `std::to_array` + `ranges::find` with a projection.


src/etl/Errors.hpp
```diff
@@ -0,0 +1,11 @@
+/** @file */
+#pragma once
+
+#include <rpcspec/Errors.hpp>
+
+namespace etl {
+
+/** @brief Error codes produced by ETL source operations. Aliased from rpc::EtlError. */
+using EtlError = rpc::EtlError;
```
mathbunnyru: This makes the whole etl depend on rpcspec library, which doesn't make much sense.
I think you should make 2 cmake components in your rpc-spec, if you want to have EtlError as part of rpc-spec.
Clio's rpc layer will depend on rpc component, etl one will depend on etl

mathbunnyru (changes requested):

src/etl/CMakeLists.txt
```diff
@@ -29,4 +29,4 @@ target_sources(
         impl/ext/Successor.cpp
 )
 
-target_link_libraries(clio_etl PUBLIC clio_data clio_util)
+target_link_libraries(clio_etl PUBLIC clio_data clio_util rpcspec::rpcspec)
```
mathbunnyru, line 32: This shouldn't be like so

src/etl/Errors.hpp
godexsoft: The whole idea here was to factor ETL errors out of the RPC layer.. but turns out they are not actual ETL errors, they are RPC errors from ETL source/forwarding code - customer facing. So the cleanest fix is to actually bring them back as they were and keep them in rpc-spec without trying to call them etl::EtlError.

src/etl/CMakeLists.txt
godexsoft: Yeah i agree, ETL should not depend on RPC at all

src/etl/Errors.hpp
godexsoft: Ok so after more research it looks like we can't really do this easily. ETL today is not just the part writing data but also the forwarding component that uses the "etl" error codes. I suggest we move them back into rpc-spec but call them "RpcForwarding[...]" instead. Then at a later stage we factor all forwarding code out of ETL module entirely. This will allow us to cleanly and honestly remove the dependency. Today though ETL will have to depend on rpc-spec simply because ETL is not just the writer part :/

src/etl/CMakeLists.txt
godexsoft: Having that said, for now we need to keep this because ETL is not only the writer but also the forwarding code, which uses the rpc-spec errors for forwarding failures.

src/app/WebHandlers.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <boost/asio/spawn.hpp>
 #include <boost/beast/http/status.hpp>
+#include <rpcspec/Errors.hpp>
```
mathbunnyru, line 17: nit: Slightly unexpected to see 0 code changes here, but include being added

src/etl/impl/ForwardingSource.cpp
```diff
@@ -58,25 +58,25 @@ ForwardingSource::forwardToRippled(
     auto expectedConnection = connectionBuilder.connect(yield);
     if (not expectedConnection) {
         LOG(log_.debug()) << "Couldn't connect to rippled to forward request.";
-        return std::unexpected{rpc::ClioError::EtlConnectionError};
+        return std::unexpected{rpc::ClioError::RpcForwardingConnectionError};
```
mathbunnyru, line 61: Since you have ClioError, I think it would be great to make rpc-spec not have access to it, when server is rippled.

src/rpc/common/AnyHandler.hpp
```diff
@@ -60,7 +60,7 @@ class AnyHandler final {
      *
      * @param value The JSON to process
      * @param ctx Request context
-     * @return JSON result or @ref Status on error
```
mathbunnyru, line 63: Doesn't this `@ref` work or what this change is about?

src/rpc/Errors.cpp
```diff
@@ -57,100 +69,100 @@ operator<<(std::ostream& stream, Status const& status)
...
+        {
+            .code = ClioError::RpcForwardingInvalidResponse,
+            .error = "invalidResponse",
+            .message = "Rippled returned an invalid response.",
+        },
+    });
 
-    auto matchByCode = [code](auto const& info) { return info.code == code; };
-    if (auto it = ranges::find_if(kInfos, matchByCode); it != end(kInfos))
+    if (auto it = ranges::find(kInfos, code, &ClioErrorInfo::code); it != end(kInfos))
```
mathbunnyru, line 165: Not your code exactly, but how where we able to use `ranges` without `std`?

src/rpc/common/AnyHandler.hpp
godexsoft: Unfortunately yes, since Status is now in rpc-spec it stopped finding it and i could not find an easy way to fix it decided to just drop the ref

src/etl/impl/ForwardingSource.cpp
godexsoft: Maybe as a final cleanup step when everything is working everywhere :)

src/rpc/Errors.cpp
godexsoft: `using namespace std;` at line 21 or around there. it's ok in .cpp files, although we usually don't use this

src/etl/impl/ForwardingSource.cpp
mathbunnyru: That's easier & better to do now, before someone (probably you 😆) accidentally start using clio codes in rpc-spec for rippled or rippled itself

src/app/WebHandlers.cpp
godexsoft: I think the file must have used something RippledError local errors but they included transitively before. Now the RippledError moved to rpc-spec and so we need this include here. Looks strange i agree :)

src/etl/impl/ForwardingSource.cpp
godexsoft: I'll make a pr for the rpc-spec repo, let's not enforce it in this PR yet to avoid having to make a new version and publish through conan-center

mathbunnyru (approved):

PR #3174, opened 2026-08-12
refactor: Use shared ledger and tx type tables from rpcspec

godexsoft:
PR 4 out of many. Based on #3171 .

`util::LedgerTypes` becomes a thin facade over `rpcspec`'s ledger-types table, and `util/TxUtils.{hpp,cpp}` is deleted in favour of the shared tx-type set. 

Two notes:

- `AccountTx.hpp` is repointed at the shared tx-type function, since the legacy spec still needs it. Same signature, one-line swap.
- `TxUtilTests.cpp` is kept and retargeted rather than deleted — it validates the set against the real `xrpl::TxFormats`, which the spec repo can't do (its tests run against a mocked xrpl).


src/etl/LoadBalancer.cpp
```diff
@@ -292,7 +292,7 @@ LoadBalancer::forwardToRippled(
     auto xUserValue = isAdmin ? kAdminForwardingXUserValue : kUserForwardingXUserValue;
 
     std::optional<boost::json::object> response;
-    rpc::ClioError error = rpc::ClioError::EtlConnectionError;
+    rpc::ClioError error = rpc::ClioError::RpcForwardingConnectionError;
     while (numAttempts < sources_.size()) {
         auto [res, duration] = util::timed([&]() {
```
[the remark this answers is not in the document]
godexsoft: I think this is fine, we can leave as is for the time being

src/util/LedgerUtils.hpp
```diff
@@ -175,7 +34,7 @@ class LedgerTypes {
     {
         std::array<char const*, std::size(kLedgerTypes)> res{};
         std::ranges::transform(kLedgerTypes, std::begin(res), [](auto const& item) {
-            return item.rpcName_;
+            return item.rpcName.data();
         });
```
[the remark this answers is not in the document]
godexsoft: Fixed this one.

mathbunnyru (changes requested):
This has lots of unrelated changes, so I think you need to do something with your branches

src/rpc/handlers/AccountTx.hpp
```diff
@@ -114,7 +114,9 @@ class AccountTxHandler {
     static RpcSpecConstRef
     spec([[maybe_unused]] uint32_t apiVersion)
     {
-        auto const& typesKeysInLowercase = util::getTxTypesInLowercase();
+        // TODO: goes away when account_tx moves to the shared spec, where the tx_type
```
bthomee, line 117: I see this `auto const& typesKeysInLowercase = spec::txTypesInLowercase()` occur more than once in this PR, but the comment is only added here.

Should the comment appear in both places, or would it be better to simply create a GitHub issue rather than adding a TODO that's easy to forget about?

bthomee (dismissed):

src/rpc/handlers/AccountTx.hpp
godexsoft: I'll add a TODO line to other places. Remember that these smaller PRs come from a huge branch that already contains all the changes needed, including these. So eventually we will arrive to the complete picture and these TODOs will be addressed.

src/rpc/common/Validators.cpp
```diff
@@ -116,8 +116,7 @@ CustomValidator CustomValidators::ledgerTypeValidator =
             }};
         }
 
-        auto const type =
-            util::LedgerTypes::getLedgerEntryTypeFromStr(boost::json::value_to<std::string>(value));
+        auto const type = spec::ledgerEntryTypeFromStr(boost::json::value_to<std::string>(value));
```
mathbunnyru: The word `spec` is too general, let's use `rpc::spec` explicitly, to make it consistent and clear that this comes from another library.

mathbunnyru (dismissed):

mathbunnyru (approved):

PR #3175, opened 2026-08-13
refactor: Follow up to use public util from rpc-spec

godexsoft:
This is the follow up to #3174.


godexsoft:
This is no longer needed as the changes are directly in #3174 now.

PR #3176, opened 2026-08-17
fix: Rare deadlock in subscriptions

godexsoft:
Clio would silently stop delivering all subscription messages while HTTP RPC kept working. The feed's strand thread was permanently deadlocked on a mutex it already held.

`TrackableSignalMap::emit` held the map mutex while running slots. If the last reference to a subscriber dropped during delivery, the session was destroyed *inside* `emit` → its destructor fired `onDisconnect` → `disconnect()` → tried to lock the same non-recursive mutex on the same thread. The narrow race window (session dying exactly during delivery) is why it was rare.

Fix: `emit` copies the signal out under the lock, releases it, then emits. The map value became `shared_ptr<TrackableSignal>` because once the lock is dropped, the re-entrant `disconnect` can erase and destroy the signal mid-emit — an owning pointer keeps it alive.

Scope: legacy web stack only (ng calls `disconnect()` explicitly instead of from the destructor), which is the deployed default. `TrackableSignal` itself was never affected — it takes no lock during emit.

Test `MapDisconnectFromTrackableDestructorWhileEmitting` hangs forever on the old code.

### DIFF
--- a/src/feed/impl/TrackableSignalMap.hpp
+++ b/src/feed/impl/TrackableSignalMap.hpp
@@ -30,8 +30,10 @@ template <Hashable Key, typename Session, typename... Args>
 class TrackableSignalMap {
     using ConnectionPtr = Session*;
     using ConnectionSharedPtr = std::shared_ptr<Session>;
+    using SignalType = TrackableSignal<Session, Args...>;
+    using SignalPtr = std::shared_ptr<SignalType>;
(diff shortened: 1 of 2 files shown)


mathbunnyru (commented):
Could you please describe in the PR description what the situation is when the problem happens and why?

mathbunnyru (approved):

PR #3177, opened 2026-08-17
fix: Internal error on warning injection

godexsoft:
This PR fixes a bug which produced internal error output instead of the correct plain-text error output in cases where it overlapped with maxing out the DOSGuard session allowance.

### DIFF
--- a/src/web/impl/WsBase.hpp
+++ b/src/web/impl/WsBase.hpp
@@ -4,6 +4,7 @@
 #include "rpc/common/Types.hpp"
 #include "util/Taggable.hpp"
 #include "util/log/Logger.hpp"
+#include "web/LoadWarning.hpp"
 #include "web/SubscriptionContext.hpp"
 #include "web/SubscriptionContextInterface.hpp"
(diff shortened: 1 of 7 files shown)


src/web/impl/WsBase.hpp
```diff
@@ -197,28 +198,21 @@ class WsBase : public ConnectionBase,
 
     /**
      * @brief Send a message to the client
+     *
+     * The message length is added to the DOSGuard. If that puts the client over its limit and the
+     * message parses as a JSON object, a "load" warning is attached to it; messages that are not
+     * JSON objects are sent unchanged.
+     *
      * @param msg The message to send
```
mathbunnyru: This almost duplicates the doxygen comment of src/web/impl/HttpBase.hpp.
If I remember correctly, there is a way to refer to your base class documentation

godexsoft: Perhaps this should not be here in the first place as we don't normally doxygen impl

mathbunnyru (approved):

PR #3180, opened 2026-08-18
feat: Add accounts filter to mpt_holders

PeterChen13579:
Continuing work from https://github.com/XRPLF/clio/pull/3154 as we do not have permission to push to their personal fork


src/rpc/handlers/MPTHolders.cpp
```diff
@@ -56,72 +129,43 @@ MPTHoldersHandler::process(MPTHoldersHandler::Input const& input, Context const&
...
-                )
-            );
+    // Account-list filter: bounded lookup by key, so non-holders are dropped and no
+    // paging marker is produced.
+    if (input.accounts) {
+        std::vector<xrpl::uint256> keys;
+        keys.reserve(input.accounts->size());
+        for (auto const& account : *input.accounts) {
+            auto const accountID = util::parseBase58Wrapper<xrpl::AccountID>(account);
+            if (!accountID) {
```
godexsoft: nit: Curlies look unnecessary here

godexsoft: Also probably `if (not accountID.has_value())`?

godexsoft (dismissed):
Small nits that can be ignored 👍

src/rpc/handlers/MPTHolders.cpp
```diff
@@ -134,18 +135,15 @@ MPTHoldersHandler::process(MPTHoldersHandler::Input const& input, Context const&
...
-        for (auto const& account : *input.accounts) {
-            auto const accountID = util::parseBase58Wrapper<xrpl::AccountID>(account);
-            if (!accountID) {
-                return Error{Status{RippledError::RpcInvalidParams, "accountsMalformed"}};
-            }
-            // NOLINTNEXTLINE(bugprone-unchecked-optional-access)
-            keys.push_back(xrpl::keylet::mptoken(mptID, *accountID).key);
+        for (auto const& accountID : *input.accounts) {
+            auto const key = xrpl::keylet::mptoken(mptID, accountID).key;
+            if (std::ranges::find(keys, key) == keys.end())
```
godexsoft: `not std::ranges::contains(keys, key)`?

tests/unit/rpc/handlers/MPTHoldersTests.cpp
```diff
@@ -791,7 +791,7 @@ TEST_F(RPCMPTHoldersHandlerTest, AccountsMalformedEntry)
         ASSERT_FALSE(output);
         auto const err = rpc::makeError(output.result.error());
         EXPECT_EQ(err.at("error").as_string(), "invalidParams");
-        EXPECT_EQ(err.at("error_message").as_string(), "accountsMalformed");
+        EXPECT_EQ(err.at("error_message").as_string(), "accounts'sItemMalformed");
```
godexsoft: Strange error message.. this would have been `accounts' item malformed` but in any case it reads strange with the apostrophe. Consider changing

tests/unit/rpc/handlers/MPTHoldersTests.cpp
```diff
@@ -1244,23 +1264,26 @@ TEST_F(RPCMPTHoldersHandlerTest, AccountsFilterDuplicates)
         .WillByDefault(Return(Blob{'f', 'a', 'k', 'e'}));
 
     auto const mptoken1 = createMpTokenObject(kHoldeR1Account, xrpl::uint192(kMptId));
+    auto const mptoken2 = createMpTokenObject(kHoldeR2Account, xrpl::uint192(kMptId));
     std::vector<Blob> const objs = {
-        mptoken1.getSerializer().peekData(), mptoken1.getSerializer().peekData()
+        mptoken1.getSerializer().peekData(), mptoken2.getSerializer().peekData()
+    };
+    std::vector<xrpl::uint256> const expectedKeys = {
+        xrpl::keylet::mptoken(xrpl::uint192(kMptId), getAccountIdWithString(kHoldeR1Account)).key,
```
godexsoft: Let's fix `kHoldeR` -> `kHolder` :)

godexsoft (changes requested):
Leaving few nits to address 👍

tests/unit/rpc/handlers/MPTHoldersTests.cpp
PeterChen13579: I changed it to without the apostrophe, but kept the format the same because all of our error codes follow camel style. but yea definitely looked weird with it.

src/rpc/handlers/MPTHolders.cpp
godexsoft: This is about to disappear in favor of rpc-spec version of the same thing :rage3:

godexsoft, line 36: This part should now go through input.ledger

src/rpc/handlers/MPTHolders.hpp
godexsoft: We will have to adopt the new framework here now that this part has been merged

godexsoft (changes requested):
I think my recent changes broke a bunch here.. let's fix what i noted and see if you can get it to work with current rpc-spec instead of old spec system 👍

src/rpc/handlers/MPTHolders.cpp
```diff
@@ -21,13 +27,146 @@
...
+}
+
+}  // namespace
+
+std::expected<MPTHoldersHandler::Input, Status>
+MPTHoldersHandler::parseInput(boost::json::value jv, uint32_t apiVersion)
+{
+    // The shared rpc-spec fields (mpt_issuance_id, marker, limit, ledger) are validated and
+    // parsed the same way for xrpld and Clio; only `accounts` is Clio-only and has no home in
+    // the shared spec, so it is validated by hand below.
```
godexsoft: This is not quite as it should be i think. The spec for Clio should have validated the field already and ignored it for xrpld. I'll fix this myself in later PRs

src/rpc/handlers/MPTHolders.hpp
```diff
@@ -1,33 +1,45 @@
...
- * @brief The mpt_holders command asks the Clio server for all holders of a particular
+ * @brief The mpt_holders command asks the Clio server for holders of a particular
  * MPTokenIssuance.
+ *
+ * When `accounts` is provided, those accounts are looked up directly by
+ * `keylet::mptoken` instead of scanning the holder index. Duplicate accounts are
+ * collapsed in first-seen order and non-holders are omitted. This filtered mode is
+ * not paginated, so `marker` and `limit` are rejected.
  */
-class MPTHoldersHandler : public rpc::spec::HandlerFor<rpc::spec::handlers::mpt_holders::Input> {
```
godexsoft, line 24: Why is this not going through rpc::spec?

src/rpc/handlers/MPTHolders.hpp
```diff
@@ -41,6 +53,15 @@ class MPTHoldersHandler : public rpc::spec::HandlerFor<rpc::spec::handlers::mpt_
         std::optional<std::string> marker;
     };
 
+    /**
+     * @brief Input for mpt_holders: the shared rpc-spec fields (`mpt_issuance_id`, `marker`,
+     * `limit`, `ledger`), plus Clio's own `accounts` filter, which has no equivalent in the
+     * shared spec and is validated/parsed separately in @ref parseInput.
+     */
+    struct Input : rpc::spec::handlers::mpt_holders::Input {
```
godexsoft: I don't think this is what we will end up with, we can have rpc::spec Input type conditionally inject members (via preprocessor). So this local type should not be a thing

godexsoft (dismissed):
I think this looks about right but it currently contains a lot of changes that are done in #3214 so i suggest we merge it before this one - the diff should collapse to just your handler changes 🔢 👍

godexsoft (approved):

PR #3181, opened 2026-08-19
fix: WebServer NG send queue

godexsoft:
`ws_max_sending_queue_size` was not being applied correctly on the ng web server path.
The check counted in-flight coroutines rather than queued messages, which does
not track queue depth because non-draining senders return immediately.

Moves the limit into `SendingQueue`, which owns the queue and can enforce it
directly, and wires the configured value through on connection upgrade. Matches
the existing `WsBase` behaviour of failing the connection with `timed_out`.

Adds unit tests for the queue bound and for the config value reaching the
connection.

### DIFF
--- a/src/web/ng/impl/ConnectionHandler.cpp
+++ b/src/web/ng/impl/ConnectionHandler.cpp
@@ -24,6 +24,7 @@
 
 #include <chrono>
 #include <cstddef>
+#include <functional>
 #include <memory>
 #include <optional>
 #include <string>
@@ -135,12 +136,9 @@ ConnectionHandler::processConnection(ConnectionPtr connectionPtr, boost::asio::y
     if (connectionRef.wasUpgraded()) {
         auto* ptr = dynamic_cast<impl::WsConnectionBase*>(connectionPtr.get());
         ASSERT(ptr != nullptr, "Casted not websocket connection");
+        ptr->setMaxSendingQueueSize(maxSubscriptionSendQueueSize_);
         subscriptionContext = std::make_shared<SubscriptionContext>(
-            tagFactory_,
-            *ptr,
-            maxSubscriptionSendQueueSize_,
-            yield,
-            [this](Error const& e, Connection const& c) { return handleError(e, c); }
+            tagFactory_, *ptr, yield, std::bind_front(&ConnectionHandler::handleError, this)
         );
         LOG(log_.trace()) << connectionRef.tag()
                           << "Created SubscriptionContext for the connection";
--- a/src/web/ng/impl/ConnectionHandler.cpp
+++ b/src/web/ng/impl/ConnectionHandler.cpp
@@ -76,7 +76,6 @@ ConnectionHandler::ConnectionHandler(
     ProcessingPolicy processingPolicy,
     std::optional<size_t> maxParallelRequests,
     util::TagDecoratorFactory& tagFactory,
-    std::optional<size_t> maxSubscriptionSendQueueSize,
     ProxyIpResolver proxyIpResolver,
     OnDisconnectHook onDisconnectHook,
     OnIpChangeHook onIpChangeHook
@@ -84,7 +83,6 @@ ConnectionHandler::ConnectionHandler(
     : processingPolicy_{processingPolicy}
     , maxParallelRequests_{maxParallelRequests}
     , tagFactory_{tagFactory}
-    , maxSubscriptionSendQueueSize_{maxSubscriptionSendQueueSize}
     , proxyIpResolver_(std::move(proxyIpResolver))
     , onDisconnectHook_{std::move(onDisconnectHook)}
     , onIpChangeHook_(std::move(onIpChangeHook))
@@ -136,7 +134,6 @@ ConnectionHandler::processConnection(ConnectionPtr connectionPtr, boost::asio::y
     if (connectionRef.wasUpgraded()) {
         auto* ptr = dynamic_cast<impl::WsConnectionBase*>(connectionPtr.get());
         ASSERT(ptr != nullptr, "Casted not websocket connection");
-        ptr->setMaxSendingQueueSize(maxSubscriptionSendQueueSize_);
         subscriptionContext = std::make_shared<SubscriptionContext>(
             tagFactory_, *ptr, yield, std::bind_front(&ConnectionHandler::handleError, this)
         );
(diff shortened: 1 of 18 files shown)


src/web/ng/impl/ConnectionHandler.cpp
```diff
@@ -135,12 +136,9 @@ ConnectionHandler::processConnection(ConnectionPtr connectionPtr, boost::asio::y
...
         auto* ptr = dynamic_cast<impl::WsConnectionBase*>(connectionPtr.get());
         ASSERT(ptr != nullptr, "Casted not websocket connection");
+        ptr->setMaxSendingQueueSize(maxSubscriptionSendQueueSize_);
         subscriptionContext = std::make_shared<SubscriptionContext>(
-            tagFactory_,
-            *ptr,
-            maxSubscriptionSendQueueSize_,
-            yield,
-            [this](Error const& e, Connection const& c) { return handleError(e, c); }
+            tagFactory_, *ptr, yield, std::bind_front(&ConnectionHandler::handleError, this)
```
mathbunnyru, line 138: To me, `std::bind_front` + member function pointer is more difficult to read than a lambda

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -23,18 +27,32 @@ class SendingQueue {
...
+    }
+
     std::expected<void, Error>
     send(T message, boost::asio::yield_context yield)
     {
         if (error_)
             return std::unexpected{error_};
 
+        if (maxSize_.has_value() and queue_.size() >= *maxSize_) {
+            error_ = boost::asio::error::timed_out;
```
mathbunnyru, line 43: Is there maybe a better error type?

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -23,18 +27,32 @@ class SendingQueue {
...
 
 public:
-    SendingQueue(Sender sender) : sender_{std::move(sender)}
+    SendingQueue(Sender sender, std::optional<size_t> maxSize = std::nullopt)
+        : sender_{std::move(sender)}, maxSize_{maxSize}
     {
     }
 
+    void
+    setMaxSize(std::optional<size_t> maxSize)
```
mathbunnyru: Do we need to make it optional? I think it should always be bounded

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -23,18 +27,32 @@ class SendingQueue {
...
+
     std::expected<void, Error>
     send(T message, boost::asio::yield_context yield)
     {
         if (error_)
             return std::unexpected{error_};
 
+        if (maxSize_.has_value() and queue_.size() >= *maxSize_) {
+            error_ = boost::asio::error::timed_out;
+            queue_ = {};
```
mathbunnyru: Why do we empty the queue?

src/web/ng/impl/ConnectionHandler.cpp
```diff
@@ -135,12 +136,9 @@ ConnectionHandler::processConnection(ConnectionPtr connectionPtr, boost::asio::y
     if (connectionRef.wasUpgraded()) {
         auto* ptr = dynamic_cast<impl::WsConnectionBase*>(connectionPtr.get());
         ASSERT(ptr != nullptr, "Casted not websocket connection");
+        ptr->setMaxSendingQueueSize(maxSubscriptionSendQueueSize_);
```
mathbunnyru: Do we need the runtime manipulation of the queue size? Shouldn't it be in ctor?

mathbunnyru (changes requested):

src/web/ng/impl/ConnectionHandler.cpp
godexsoft: It's strictly better - no need to know what parameters we need to pass nor remember to type them out. Also we used exactly this pattern elsewhere  (Monitor i think)

src/web/ng/impl/SendingQueue.hpp
godexsoft: This is the correct error type, the one used by the library itself for timeouts

src/web/ng/impl/SendingQueue.hpp
godexsoft: Actually we don't have to. This is just deallocating sooner (otherwise the queue stays in memory until connection is destroyed, which happens later). We can simplify

src/web/ng/impl/ConnectionHandler.cpp
godexsoft: I agree, we should not need it

src/web/ng/impl/ConnectionHandler.cpp
mathbunnyru: It is strictly worse: it's not obvious how these params will be passed.
Also, refactoring of unrelated code in a "fix" PR makes it more difficult to review and increases the scope.

src/web/ng/impl/SendingQueue.hpp
mathbunnyru: Why is it correct? We're not timing out; we're rejecting the message completely

godexsoft: The reason for rejection is "client is too slow" - timeout.

src/web/ng/impl/SendingQueue.hpp
```diff
@@ -23,18 +27,32 @@ class SendingQueue {
...
+            return std::unexpected{error_};
```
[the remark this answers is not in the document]
mathbunnyru: @godexsoft please take a look here

src/web/ng/impl/SendingQueue.hpp
mathbunnyru: That makes sense, thanks

src/web/ng/impl/SendingQueue.hpp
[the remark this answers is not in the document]
godexsoft: Yea this is done

mathbunnyru (changes requested):
Please, don't make my comments "resolved" if you simply disagree with them

godexsoft:
> Please, don't make my comments "resolved" if you simply disagree with them

I can keep it for you to resolve if you prefer - since i'm not going to apply your suggestion i did not feel it was worth keeping unresolved 🤷

mathbunnyru (approved):
Approving this to move forward. Please, don't make refactoring changes in unrelated code in the future, especially if we don't agree whether the refactoring makes the code better.

src/web/ng/impl/ConnectionHandler.cpp
mathbunnyru: Resolving, because was asked to resolve

PR #3187, opened 2026-08-21
fix: Fiber suspension during exception

godexsoft:
This PR fixes a segfault that can be caused by Retry logic when used via coroutines during cache load. 
The regression was introduced in #3167.

### DIFF
--- a/src/data/BackendInterface.hpp
+++ b/src/data/BackendInterface.hpp
@@ -103,16 +103,21 @@ retryOnTimeout(
     auto retry = util::makeRetryExponentialBackoff(delays, yield.get_executor());
 
     while (true) {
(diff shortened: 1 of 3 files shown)


kuznetsss (dismissed):

mathbunnyru (dismissed):
This looks good. Could you ask AI to check if we don't have other places in code like this?

godexsoft:
> This looks good. Could you ask AI to check if we don't have other places in code like this?

There is no more places like this in the codebase. 
However, to avoid re-adding this i'm adding an assert inside the wait function of Retry. This will help us catch it sooner while developing Clio if we ever re-introduce this bug in Retry specifically.

mathbunnyru (approved):

PR #3198, opened 2026-09-01
refactor: Dual path for rpc handlers

godexsoft:
Started from #3173 

Adds a second dispatch path so handlers can use the shared rpc-spec (`HandlerFor<Input>`), plus the `LedgerSpecifier` → ledger-header helper they'll need — no handler migrates, nothing changes on the wire.


mathbunnyru (changes requested):
I see lots of unrelated changes here too

godexsoft:
> I see lots of unrelated changes here too

Yeah, idk how stacking works here.. seems like it does not. I would not look at this before we merge the previous PRs in the series.

src/rpc/common/impl/Processors.hpp
```diff
@@ -19,8 +22,29 @@ struct DefaultProcessor final {
...
+
+        static_assert(
+            kIsSingleInputPath<HandlerType>,
+            "handler satisfies both the legacy and the typed input path; dispatch would be "
+            "decided by the order of the branches below rather than by the handler"
+        );
+
+        // New `rpc-spec`-based handler
+        if constexpr (SomeHandlerWithTypedInput<HandlerType>) {
+            auto warnings = rpc::spec::toJsonArray(HandlerType::spec(ctx.apiVersion).check(value));
```
bthomee, line 39: I would move this lower, to just above the `return ReturnType{value_from(std::move(ret).value()), std::move(warnings)};`.

Namely, if either the `input` or `ret` are invalid, then the `warnings` statement is run unnecessarily.

src/rpc/common/impl/Processors.hpp
```diff
@@ -19,8 +22,29 @@ struct DefaultProcessor final {
...
+            auto input = HandlerType::parseInput(value, ctx.apiVersion);
+            if (not input)
+                return ReturnType{Error{std::move(input).error()}, std::move(warnings)};
+
+            auto ret = handler.process(*input, ctx);
+            if (not ret)
+                return ReturnType{Error{std::move(ret).error()}, std::move(warnings)};
+
+            return ReturnType{value_from(std::move(ret).value()), std::move(warnings)};
+        } else if constexpr (SomeHandlerWithInput<HandlerType>) {
```
bthomee: Shouldn't this just be a regular `if` on a separate line instead of an `else if`, given the `return` statements in the `if` above? (I would've expected clang-tidy to complain.)

src/rpc/RPCHelpers.cpp
```diff
@@ -547,6 +548,47 @@ getLedgerHeaderFromHashOrSeq(
...
+        auto const lgrInfo =
+            backend.fetchLedgerByHash(std::get<xrpl::uint256>(resolved.value), yield);
+        if (!lgrInfo || lgrInfo->seq > maxSeq)
+            return err;
+
+        return *lgrInfo;
+    }
+
+    if (resolved.isShortcut()) {
+        auto const shortcut = std::get<rpc::spec::LedgerShortcut>(resolved.value);
```
bthomee, line 572: Don't you need a [[maybe_unused]] here, assuming the ASSERT is compiled away in release mode? In that case, returning an error in case `shortcut != rpc::spec::LedgerShortcut::Validated` seems better than falling through...

src/rpc/common/impl/Processors.hpp
godexsoft: Not sure, maybe it's because there are more than two branches there in the if. I'll see if the other option is better

src/rpc/RPCHelpers.cpp
godexsoft: This is Clio, asserts are in all configurations, no falling through or ignoring the assert is possible here

src/rpc/common/impl/Processors.hpp
godexsoft: Are you suggesting we should compute it 3 times in 3 separate places instead? It looks like it's used regardless of `input` and `ret`.

src/rpc/common/impl/Processors.hpp
godexsoft: Done some refactoring, it does seem a bit better 👍 The `static_assert` had to change now that we can't use `static_assert(false)` as we are outside of the `else` branch and it would unconditionally fail.

src/rpc/common/impl/Processors.hpp
bthomee: No, my bad - the lines got wrapped in my browser and I missed that it was used by all three returns and not just the last.

bthomee (dismissed):

src/rpc/common/impl/Processors.hpp
```diff
@@ -19,9 +21,37 @@ struct DefaultProcessor final {
...
+                SomeHandlerWithoutInput<HandlerType>,
+            "handler matches none of the branches below"
+        );
+
+        // New `rpc-spec`-based handler
+        if constexpr (SomeHandlerWithTypedInput<HandlerType>) {
+            auto input = HandlerType::parseInput(value, ctx.apiVersion);
+            auto warnings = rpc::spec::toJsonArray(HandlerType::spec(ctx.apiVersion).check(value));
+
+            if (not input)
```
mathbunnyru:
```suggestion
            if (not input.has_value())
```

src/rpc/common/impl/Processors.hpp
```diff
@@ -19,9 +21,37 @@ struct DefaultProcessor final {
...
+                return ReturnType{Error{std::move(input).error()}, std::move(warnings)};
+
+            auto ret = handler.process(*input, ctx);
+
+            if (not ret)
```
mathbunnyru:
```suggestion
            if (not ret.has_value())
```

src/rpc/common/impl/Processors.hpp
```diff
@@ -19,9 +21,37 @@ struct DefaultProcessor final {
...
+                return ReturnType{Error{std::move(ret).error()}, std::move(warnings)};
```
mathbunnyru: Should we use `and_then` or sth like this for expected chaining?

src/rpc/common/impl/Processors.hpp
```diff
@@ -33,23 +63,19 @@ struct DefaultProcessor final {
...
+
             return ReturnType{value_from(std::move(ret).value()), std::move(warnings)};
-        } else if constexpr (SomeHandlerWithoutInput<HandlerType>) {
+        }
+
+        if constexpr (SomeHandlerWithoutInput<HandlerType>) {
             // no input to pass, ignore the value
             auto const ret = handler.process(ctx);
-            if (not ret) {
+            if (not ret)
```
mathbunnyru:
```suggestion
            if (not ret.has_value())
```

src/rpc/RPCHelpers.cpp
```diff
@@ -547,6 +548,47 @@ getLedgerHeaderFromHashOrSeq(
...
+    auto const err = std::unexpected{Status{RippledError::RpcLgrNotFound, "ledgerNotFound"}};
+    auto const resolved = ledger.resolved();
+
+    if (resolved.isHash()) {
+        auto const lgrInfo =
+            backend.fetchLedgerByHash(std::get<xrpl::uint256>(resolved.value), yield);
+        if (!lgrInfo || lgrInfo->seq > maxSeq)
+            return err;
+
+        return *lgrInfo;
```
mathbunnyru: We sometimes call it expectedLgrInfo or maybeLgrInfo, and use `.has_value()` as well

mathbunnyru (dismissed):

src/rpc/common/impl/Processors.hpp
godexsoft: Probably not

src/rpc/RPCHelpers.cpp
godexsoft: There is more of the same few lines down.. i'll refactor both of these here.

mathbunnyru (approved):

PR #3199, opened 2026-09-03
refactor: Migrate first handlers to rpc-spec

godexsoft:
This PR migrates `account_info` and `account_currencies` as the first two handlers to use rpc-spec system.
It also migrates Clio to rpc-spec 0.1.8 and fixes a discrepancy in how we handle malformed ledger hash/index.


src/rpc/handlers/AccountCurrencies.cpp
```diff
@@ -33,25 +37,21 @@ AccountCurrenciesHandler::process(
...
+        input.ledger,
         range->maxSequence  // NOLINT(bugprone-unchecked-optional-access)
     );
 
     if (not expectedLgrInfo.has_value())
         return Error{expectedLgrInfo.error()};
 
     auto const& lgrInfo = *expectedLgrInfo;
-    auto const accountID = accountFromStringStrict(input.account);
+    auto const& accountID = input.account;
```
mathbunnyru, line 45: Is it numeric? If not, shouldn't be ID

src/rpc/handlers/AccountCurrencies.hpp
```diff
@@ -25,7 +21,8 @@ namespace rpc {
  *
  * For more details see: https://xrpl.org/account_currencies.html
  */
-class AccountCurrenciesHandler {
+class AccountCurrenciesHandler
+    : public spec::HandlerFor<spec::handlers::account_currencies::Input> {
```
mathbunnyru: Please, use the fully qualified `rpc::spec` namespace everywhere

src/rpc/handlers/AccountInfo.cpp
```diff
@@ -3,17 +3,18 @@
...
 
 #include <boost/json/array.hpp>
 #include <boost/json/conversion.hpp>
 #include <boost/json/object.hpp>
 #include <boost/json/value.hpp>
-#include <boost/json/value_to.hpp>
 #include <rpcspec/Errors.hpp>
+#include <rpcspec/HandlerFor.hpp>
+#include <rpcspec/HandlerForDefs.hpp>              // IWYU pragma: keep
+#include <rpcspec/handlers/account_info/Spec.hpp>  // IWYU pragma: keep
```
mathbunnyru: Why do you need IWYU here?
It won't look nice, if we'll have this everywhere, better to fix the reason why this is needed

src/rpc/handlers/AccountInfo.cpp
```diff
@@ -32,6 +33,8 @@
 #include <utility>
 #include <vector>
 
+template struct rpc::spec::HandlerFor<rpc::spec::handlers::account_info::Input>;
```
mathbunnyru: Please, explain, why this is needed

src/rpc/handlers/AccountInfo.hpp
```diff
@@ -28,7 +24,7 @@ namespace rpc {
  *
  * For more details see: https://xrpl.org/account_info.html
  */
-class AccountInfoHandler {
+class AccountInfoHandler : public spec::HandlerFor<spec::handlers::account_info::Input> {
```
mathbunnyru: Also `rpc::spec`

tests/unit/rpc/handlers/AccountInfoTests.cpp
```diff
@@ -106,21 +107,21 @@ generateTestValuesForParametersTest()
...
-            .expectedErrorMessage = "ledger_hashMalformed"
+            .expectedErrorMessage = "Invalid field 'ledger_hash', not hex string."
         },
         AccountInfoParamTestCaseBundle{
             .testName = "LedgerHashNotString",
             .testJson =
                 R"JSON({"ident": "rLEsXccBGNR3UPuPu2hUXPjziKC3qKSBun", "ledger_hash": 1})JSON",
             .expectedError = "invalidParams",
-            .expectedErrorMessage = "ledger_hashNotString"
+            .expectedErrorMessage = "Invalid field 'ledger_hash', not hex string."
```
mathbunnyru, line 117: The error message now is the same for this one and the previous one. Maybe we should keep them separate?

mathbunnyru (changes requested):

src/rpc/handlers/AccountCurrencies.cpp
godexsoft: It's a `xrpl::AccountID`, do you want to rename it?

src/rpc/handlers/AccountInfo.cpp
godexsoft: Unfortunately these are needed with the current design. I agree it's not the nicest and we may comeup with some way to do slightly better. But overall this is why (In HandlerForDefs.hpp): 
```
// Out-of-line definitions for HandlerFor<Input>. Include ONLY from a handler's .cpp (which also
// includes that handler's <rpcspec/handlers/<name>/Spec.hpp>, so `specFor` is visible to ADL), and
// emit the single explicit instantiation:
//
//     template struct rpc::spec::HandlerFor<rpc::spec::handlers::<name>::Input>;
//
// That confines the handler's consteval spec instantiation to its own translation unit.
```

All this is to compile instantiations together with the handler rather than during HandlerRegistry.cpp. This actually saves a lot of compile time. Without this all these would instantiate and build in one TU and that takes a long while.

src/rpc/handlers/AccountInfo.cpp
godexsoft: See explanation above.
We can do this in rpc-spec (TU per handler) but that means it will no longer be a header-only library.
What do you think is a good alternative/workaround?

tests/unit/rpc/handlers/AccountInfoTests.cpp
godexsoft: This makes it match rippled. Since we are now based on a "spec" it will produce the same output on Clio and xrpld. And so i rather change Clio's output than xrpld's.

src/rpc/handlers/AccountInfo.cpp
godexsoft: Ok i think i have a decent workaround. Remove all the explicit instantiations and includes from Clio/xrpld handler impls and add two lines to rpc cmake: 
```
rpcspec_generate_instantiations(OUT_VAR rpcspec_instantiations)
target_sources(clio_rpc PRIVATE ${rpcspec_instantiations})
```

I'm making a conan package for the rpc-spec 0.1.9 and will update this PR with the changes soon.

src/rpc/handlers/AccountCurrencies.cpp
mathbunnyru: If the type is called AccountID, then the variable name makes sense, thanks

src/rpc/handlers/AccountInfo.cpp
mathbunnyru: Not nice, but we agreed this should be fixed after all rpcs are switched in rippled and clio

tests/unit/rpc/handlers/AccountInfoTests.cpp
mathbunnyru: I think we agreed to allow the change of error messages? (without changing error codes)

`not hex string` sounds like `it's a string, but not hex`, while in fact it's not even a string at all.

I think we should give easy-to-understand errors to people when something goes wrong

godexsoft: When Clio and xrpld had their own ways to do this it makes sense to allow different error message.. but this comes from a shared spec so it's expected to be the same, don't you think? technically we can probably do it but i'd prefer to have one spec that is as shared as possible. There may also be a discussion later to change it uniformly in both but use a more reasonable error message, like we had in Clio

mathbunnyru (approved):

tests/unit/rpc/handlers/AccountInfoTests.cpp
godexsoft: fyi @mathbunnyru: https://github.com/XRPLF/clio/issues/3205

PR #3206, opened 2026-09-11
refactor: Migrate handlers to rpc-spec (A)

godexsoft:
This batch of handlers don't need further changes in `rpc-spec` to align error messages so it is going first.


tests/unit/rpc/handlers/BookChangesTests.cpp
```diff
@@ -62,19 +62,19 @@ generateTestValuesForParametersTest()
             .testName = "LedgerHashInvalid",
             .testJson = R"JSON({"ledger_hash": "1"})JSON",
             .expectedError = "invalidParams",
-            .expectedErrorMessage = "ledger_hashMalformed"
+            .expectedErrorMessage = "Invalid field 'ledger_hash', not hex string."
```
bthomee, line 65: Is changing the error message (but keeping the error type/code/name = invalidParams the same) is an API-breaking change or is that fine?

src/rpc/handlers/GatewayBalances.cpp
```diff
@@ -169,7 +161,7 @@ GatewayBalancesHandler::process(
     if (!ret.has_value())
         return Error{ret.error()};
 
-    output.accountID = input.account;
+    output.accountID = xrpl::to_string(accountID);
```
bthomee, line 164: Before this just copied the input value straight to the output, and now it re-serializes the parsed input value. In which scenarios would the new output representation of the account be different from the out output representation?

bthomee (approved):

tests/unit/rpc/handlers/BookChangesTests.cpp
godexsoft: This is all temporary to some extent - we want to unify the messages using the spec and so currently i'm trying to copy what rippled does as much as possible. Changing these strings in Clio is acceptable (as long as the code don't change)

src/rpc/handlers/GatewayBalances.cpp
godexsoft: It used to be a string that we converted back to accountID but now that part is hidden inside of the spec so we directly get xrpl::accountID type here, hence the to_string for output.

Issue #3205, opened 2026-09-11
Unify error messages through rpc-spec


godexsoft:
## Summary
In #3199 there was a request from @mathbunnyru to change the error message to a more informative one.
This is a good idea but it's best done in one PR fixing/unifying all error messages across the spec - both xrpld and Clio should have the same messages across the board. Today some are better in xrpld, some in Clio.

## Motivation

Better error messages and unification.

## Solution

Unify in rpc-spec.

## Paths Not Taken

Change on demand - would lead to worse history/visibility of the changes and require agreements on case by case basis instead of one shot.

PR #3207, opened 2026-09-14
refactor: Migrate handlers to rpc-spec (B)

godexsoft:
More handlers migrating with rpc-spec 0.1.11 aligning the errors closer to xrpld's style. 
This changes some wire format in Clio but since these are only error messages (not codes) it's not a big issue.
We will also align both xrpld and Clio fully in a separate PR towards the end of this migration.


tests/unit/rpc/handlers/NFTBuyOffersTests.cpp
```diff
@@ -183,7 +186,7 @@ TEST_F(RPCNFTBuyOffersHandlerTest, NFTIDInvalidFormat)
         ASSERT_FALSE(output);
         auto const err = rpc::makeError(output.result.error());
         EXPECT_EQ(err.at("error").as_string(), "invalidParams");
-        EXPECT_EQ(err.at("error_message").as_string(), "nft_idMalformed");
+        EXPECT_EQ(err.at("error_message").as_string(), "Invalid field 'nft_id'.");
```
bthomee, line 189: Previously the error messages provided more detail on what was wrong with the field.

Before you'd be able to tell whether the field was malformed or not a string, while now you just see `"Invalid field 'nft_id'."` and you'd have to guess what you did wrong. Same for the field `marker` below.

For the `ledger_hash` and `ledger_index` I see some more detail in some (but not all) cases - it's a bit inconsistent. I'd suggest if possible to always provide a rationale if the field is invalid.

* `"Invalid field 'ledger_hash'."` => no rationale
* `"Invalid field 'ledger_hash', not string."` => rationale
* `"Invalid field 'ledger_index', not string or number."` => rationale

godexsoft: I agree, but this is in tune with today's `xrpld` output so i try to copy that for now. I'm saving all these for unification at a later stage - we will need to add some code in `xrpld` to support the effort too 👍

godexsoft: Fyi there is https://github.com/XRPLF/clio/issues/3205 that was created so we don't forget :)

bthomee (approved):

PR #3209, opened 2026-09-15
refactor: Migrate handlers to rpc-spec (D)

godexsoft:
Slightly more complicated handlers batch. After this lands we will have about 3 handlers left to migrate - the hardest ones.
This PR also bumps used rpc-spec to 0.1.12.


src/rpc/filters/TransactionFilter.hpp
```diff
@@ -3,15 +3,22 @@
...
 
+/**
+ * @brief The delegate filter used to narrow account_tx by delegation.
+ */
+using DelegateFilter = rpc::spec::handlers::account_tx::DelegateFilter;
+
 /**
  * @brief Interface for filtering transactions.
  */
+// TODO: consider moving to rpc-spec later on
```
bthomee, line 21: Instead of a TODO I'd recommend creating a GitHub Issue instead.

src/rpc/handlers/AccountTx.hpp
```diff
@@ -35,23 +27,17 @@ namespace rpc {
...
     std::shared_ptr<BackendInterface> sharedPtrBackend_;
     std::shared_ptr<etl::ETLServiceInterface const> etl_;
 
 public:
-    static constexpr auto kLimitMin = 1;
-    static constexpr auto kLimitMax = 1000;
-    static constexpr auto kLimitDefault = 200;
+    static constexpr auto kLimitMin = rpc::spec::handlers::account_tx::kLimitMin;
+    static constexpr auto kLimitMax = rpc::spec::handlers::account_tx::kLimitMax;
+    static constexpr auto kLimitDefault = rpc::spec::handlers::account_tx::kLimitDefault;
```
bthomee, line 38: Can't these limits be expressed in the spec itself?

src/rpc/handlers/DepositAuthorized.cpp
```diff
@@ -176,8 +143,11 @@ tag_invoke(
         {JS(ledger_index), output.ledgerIndex},
         {JS(validated), output.validated}
     };
-    if (output.credentials)
-        jv.as_object()[JS(credentials)] = *output.credentials;
+    if (output.credentials.has_value()) {
+        jv.as_object()[JS(credentials)] = *output.credentials                             //
+            | std::views::transform([](auto const& cred) { return xrpl::strHex(cred); })  //
+            | std::ranges::to<boost::json::array>();
+    }
```
bthomee, line 150: I guess this is the one thing that's a bit worse-looking with the new spec...

bthomee (approved):

src/rpc/handlers/AccountTx.hpp
godexsoft: Well, they are, just not in the DSL part of the spec as it's easier to just have a kConstant for that. I'm not sure putting them in the DSL is better or not

src/rpc/handlers/DepositAuthorized.cpp
godexsoft: I mean.. i could keep the old code here basically but this ranges code looks better to me :) In any case we now do need to wrap it as json::array because we need to cross the model->json boundary somewhere

src/rpc/filters/TransactionFilter.hpp
godexsoft: I don't have an issue but i have a local plan where this is mentioned for followups. I'm not actually decided yet whether it's better or not (hence the 'consider' wording)

src/rpc/handlers/AccountTx.hpp
bthomee: If they'd be part of the DSL part of the spec it seems like it'd be more self-contained.

PR #3210, opened 2026-09-15
refactor: Migrate handlers to rpc-spec (E)

godexsoft:
This PR migrates Subscribe and Unsubscribe handlers to rpc-spec.


src/rpc/common/Validators.cpp
```diff
@@ -242,37 +217,6 @@ CustomValidator CustomValidators::bookTakerValidator =
...
-            if (not kValidStreams.contains(boost::json::value_to<std::string>(v)))
-                return Error{Status{RippledError::RpcStreamMalformed}};
-        }
-
-        return MaybeError{};
-    }};
-
 CustomValidator CustomValidators::subscribeAccountsValidator =
     CustomValidator{[](boost::json::value const& value, std::string_view key) -> MaybeError {
         if (!value.is_array())
```
bthomee: A scan by Claude showed that this one is now no longer in use. If so, it can be deleted (plus the test).

bthomee (dismissed):

src/rpc/common/Validators.cpp
godexsoft: Ok removed now. The other report by copilot was incorrect so ignoring it.

bthomee (approved):

PR #3212, opened 2026-09-15
refactor: Migrate handlers to rpc-spec (F)

godexsoft:
This PR migrates `ledger_entry` to rpc-spec.


src/rpc/handlers/LedgerEntry.cpp
```diff
@@ -1,266 +1,314 @@
...
+    };
+
     xrpl::uint256 key;
+    // For locators supplied as a raw ledger-entry hex key, the type is implied and
+    // enforced below; a precisely-computed keylet leaves this as ltANY (no check).
+    xrpl::LedgerEntryType expectedType = xrpl::ltANY;
 
-    if (input.index) {
-        key = xrpl::uint256{std::string_view(*(input.index))};
+    if (input.index.has_value()) {
```
bthomee, line 221: The massive amount of if-else statements makes my head spin. I would suggest figuring out a better way to do this, e.g. with helper functions, and perhaps some restructuring so you don't need if/else if/else if/..../else but perhaps just if+return / if+return etc.

tests/unit/rpc/handlers/LedgerEntryTests.cpp
```diff
@@ -3730,6 +3731,56 @@ TEST_F(RPCLedgerEntryTest, UnexpectedLedgerType)
     });
 }
 
+// Each variant field's hex-key arm sets its own expectedType, so a wrong constant there would
+// silently accept the wrong object type.
+struct LedgerEntryHexKeyTypeTestBundle {
```
bthomee, line 3736: Given the complex logic in src/rpc/handlers/LedgerEntry.cpp I find the number of test cases here on the light side - they only cover `credential`, `xchain_owned_claim_id`, `xchain_owned_create_account_claim_id` right now.

src/rpc/handlers/LedgerEntry.cpp
godexsoft: Agree, refactored a bit, i think it's more readable now

tests/unit/rpc/handlers/LedgerEntryTests.cpp
godexsoft: Added all missing tests

src/rpc/handlers/LedgerEntry.cpp
bthomee: Nice stuff 😄

bthomee (dismissed):

bthomee (approved):

PR #3214, opened 2026-09-17
refactor: Migrate handlers to rpc-spec (H)

godexsoft:
This PR brings back Clio wire compatibility with 2.8.0 where applicable. Previously we tried to unify towards xrpld error messages but as we are only migrating Clio to rpc-spec atm., it's best to leave the errors be and have no wire format change at all for first release with rpc-spec. Eventually we will unify error messages but that requires xrpld to move to rpc-spec as well.


bthomee (approved):
I'm looking forward to seeing rpc-spec in use in xrpld too, it will make things much more consistent.

